PR #33, opened 2021-12-18
Fixing CI to use branch caching

jonaprieto:
This PR will solve some issues related to our previous workflow. We have divided our build into three main steps, build executables, type check and website.  We will efficiently cache the executables for type checking the Agda files. The Agda cache will only store files from the same branch. Lastly, we will automatically cancel any previous run pushed in the same branch. I believe this should be enough to get CI working correctly for a while. :)

 CI workflow summary:

 - **Build**: install Haskell and Agda, and make them available for other jobs/builds.
   1. Cancel Previous Runs
   1. Branch caching
   2. Setup Haskell
   3. Add .cabal/bin into PATH
   4. Checkout Agda repository
   5. Install Agda
   
 - **Check**: depends on Build and the branch.
   1. Use cache to get an Agda executable.
   6. Checkout the main repository
   7. Verify the whole formalisation

 - **HTML**: only depends on Check.
   1. Setup Graphviz to generate the Agda dependency graph
   9. Setup Pandoc using setup-pandoc 
   10. Generate HTML
   11. Deploy HTML to Github pages

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,9 +16,9 @@
 name: build
 on:
   push:
-    branches: [ master ]
-  pull_request:
     branches: [ master , develop ]
+  pull_request:
+   
 
 jobs:
   build:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # CI workflow summary:
 # - build
-#   1. Caching
+#   1. Branch caching
 #   2. Setup Haskell
 #   3. Add .cabal/bin into PATH
 #   4. Checkout Agda repository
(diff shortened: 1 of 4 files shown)


favonia:
I have been playing with GitHub CI extensively, and here's one suggestion: we could use Docker images for Agda to have a more reliable experience. The reason is that GitHub cache does not last very long and you might end up with rebuilding Agda more than necessary. Using Docker images as starting points could avoid the problem. One could even use Docker images to permanently cache intermediate results (once the Agda development becomes large enough), but there are trade-offs between GitHub cache, transferring speed of Docker images, etc., which I don't want to go into now.  (PS: I am happy to tweak the CI once my teaching for this semester is done, but perhaps you can figure it out way before I become available.)

jonaprieto:
> I have been playing with GitHub CI extensively, and here's one suggestion: we could use Docker images for Agda to have a more reliable experience. The reason is that GitHub cache does not last very long and you might end up with rebuilding Agda more than necessary. Using Docker images as starting points could avoid the problem. One could even use Docker images to permanently cache intermediate results (once the Agda development becomes large enough), but there are trade-offs between GitHub cache, transferring speed of Docker images, etc., which I don't want to go into now. (PS: I am happy to tweak the CI once my teaching for this semester is done, but perhaps others can figure it out way before I become available.)

You're right, and it seems that the cache only lasts seven days on Github. One thing we can do is schedule a night build before it expires. Please, open a PR when you get some time for this. I'll be interested in testing a better approach  ;) I'm not familiar with Docker, but thanks for letting me know.

jonaprieto:
I'll merge this PR and open it again if I forget to solve anything.

PR #39, opened 2021-12-21
Instance args categories

elisabethbonnevier:
* Adds equivalences of large precategories.
* Adds notation using instance arguments.


EgbertRijke:
It causes an error in the category of groups. That kind of confirms my worry about instance arguments. Maybe it works well in an abstract setting, eg categories for their own sake, but we want the categories to be as functional as possible for our applications. So if the instance arguments don't work with the applications, I'd definitely say it's not worth the trouble to have them.

elisabethbonnevier:
I'll see if I can make the instance arguments work for the category of groups. Otherwise I'll just remove all uses of instance arguments.

elisabethbonnevier:
The error was just that I had renamed some arguments and I forgot to change them in the category of groups.

EgbertRijke:
Ok, glad it's working now. It seems that you are convinced that instance arguments are a good thing:) 

Can we rename the weird composition symbol to something more conventional though? We shouldn't try to come up with the weirdest symbols. Perhaps `_∘Precat_` could do, which also leaves the door open for `_∘Cat_`.

elisabethbonnevier:
Yes, I think instance arguments are a good thing :) I can change the notation. The ¤ symbol was the closest thing I could come up with that wasn't the usual composition symbol.

EgbertRijke:
I would also like to suggest that we should try to stick to one kind of symbol for evaluation of functors. It's a bit of a stretch to use `⟨_⟩` for evaluation, because it is very unconventional, and I think it is a bit too much if we use a whole range of variations of angled brackets. So I suggest `_obj-⟨_⟩` for the action on objects of functors and `_hom-⟨_⟩` for the action on morphisms of functors. But it will still look weird to me. 

When we write `obj-functor-Precat F x`, then we actually have `F x` in the notation if you ignore the word `obj-functor-Precat`, so to me that's closer to the usual way of writing `F` applied to an object `x`.

EgbertRijke:
We'll go with instance arguments then:)

Martin Escardo told me a horror story though, where his whole library once broke because the Agda implementers changed how instance arguments work. So this feature is still a bit experimental, it seems.

EgbertRijke:
By the way, I saw that you also removed the category from the type of objects. Are you sure that won't give trouble?

In an adjunction we say that for every `x : obj C` and every `y : obj D` there is an equivalence `Hom(x,G(y)) ≃ Hom (F(x),y)`. With the new notation, we will have `x : obj` and `y : obj`. Will Agda be able to figure out that `x` is in `C` and `y` is in `D`?

I'm doing adjunctions now in the other pull request, that's why I ask.

elisabethbonnevier:
Where in the code do you mean?

EgbertRijke:
In the Ljubljana branch in a new file called adjunctions.lagda.md.

But I'll wait with merging the pull request, because I don't want to give you the headache of these merge conflicts. I'll deal with them. I'm afraid they will be quite severe, because we've both been working on large-categories.lagda.md

EgbertRijke:
Anyway, for objects it seems not unreasonable to mention the category, even if Agda can be set up so that this is not technically necessary.

elisabethbonnevier:
Oh, I meant, where in the code do you mean I have removed the category for objects? In the large or small categories?

EgbertRijke:
Ah, it looks like that only happened in `categories.lagda.md`

elisabethbonnevier:
Do you mean obj-Precat? This takes the category as an explicit argument

elisabethbonnevier:
So does obj-Cat

EgbertRijke:
Ay sorry, my bad. It is in a module. I only saw from the git diff that the category was removed, but I missed the module. I said nothing:)

elisabethbonnevier:
I have removed the mention of C in the Precat functions below associative-composition-Precat because I put them into a module with the category as a parameter. But this parameter is explicit :)

elisabethbonnevier:
Ah, I see how you mistook that :) no worries!

EgbertRijke:
Sorry for giving you a hard time. Let me know when the PR is ready to be merged.

elisabethbonnevier:
Don't worry, I gladly take input on my code :) I don't know how much time I'll have to work on the code over the holidays so you can merge your work on adjunctions to master and don't have to wait for me. Then I can sort out any merge conflicts later when I'm done :)

EgbertRijke:
Cool! I'll merge it tomorrow, and try to set up the instance arguments for the adjunctions too. Happy holidays Elisabeth!

elisabethbonnevier:
Happy holidays!

PR #56, opened 2022-01-19
minor revisions. using builtins for some types

jonaprieto:


EgbertRijke:
Can you please put function arrows before the line breaks and not as the first character of a line?

PR #106, opened 2022-03-16
square free natural numbers: correct variable name in idea

Taneb:


EgbertRijke:
Many thanks for the fix!

PR #183, opened 2022-05-18
Editing CSB and split-surjective files & Adding set theory folder

ElifUskuplu:
I've made changes as @jonaprieto suggested.


EgbertRijke:
The preperation could be replaced by an import from `foundation/iterating-functions`.

EgbertRijke:
In general, names should not refer to variables

EgbertRijke:
Can you change the file name to make it all lower case? This is to avoid messing up the ordering in how the files are listed by `everything`

ElifUskuplu:
> The preperation could be replaced by an import from `foundation/iterating-functions`.

I've made this.

PR #196, opened 2022-05-31
Add precedences to equational/equivalence reasoning.

elisabethbonnevier:


tomdjong:
@elisabethbonnevier Thanks for noticing that the precedences were missing and fixing it!
Out of curiosity (I don't really mind): why did you change from the infix notation and move to `syntax`?

elisabethbonnevier:
Short answer: because that's how the Agda standard library does it.
Long answer: when writing the precedence for the transitivity syntax I started to wonder what 'infixr' means for a mixfix operator with three arguments. The examples in the [documentation](https://agda.readthedocs.io/en/v2.6.2.1/language/mixfix-operators.html) only talk about binary operators. I did a quick internet search to try to figure it out but did not find anything and I wanted to do this fix quickly because I want to use it for myself. So I looked at the standard library and noticed that they use syntax, instead of defining it as '_=⟨_⟩_', like you did. Why? You only need to use syntax when you want to bind a variable. But then I noticed that they flip the second and third argument from 'step-=' to the syntax. Assuming that they have a good reason for doing so, my hypothesis is that 'infixr' applies to the first two arguments, and thus we have to write the fixity on something where the arguments are flipped. Then we just add the syntax we actually want with 'syntax'.

PR #199, opened 2022-06-02
Change file names + move comp-emb out of abstract block

elisabethbonnevier:


EgbertRijke:
Can you add the unique-existence imports to README and foundation (in the correct spot with alphabetic ordering)

elisabethbonnevier:
Sure :) (forgot to do make check before commiting)

elisabethbonnevier:
I put it in foundation.lagda.md, but where do you want it in README? I don't see any imports there

EgbertRijke:
It's a bit confusing that this file is also called readme.

EgbertRijke:
Can you also put the imports that you changed in alphabetic order?

elisabethbonnevier:
Ah, yes of course :) Done!

PR #217, opened 2022-06-24
Additions from development on set theory in HoTT

elisabethbonnevier:


elisabethbonnevier:
@EgbertRijke please have an extra look at `foundation.set-truncations`. I might have messed something up because I did it quite quickly yesterday.

src/foundation/surjective-maps.lagda.md
```diff
@@ -46,9 +46,25 @@ A map `f : A → B` is surjective if all of its fibers are inhabited.
...
+
+module _
+  {l1 l2 : Level} (A : UU l1) (B : UU l2) (f : A ↠ B)
+  where
+
+  map-surjection : A → B
+  map-surjection = pr1 f
+
+  is-surjective-map-surj : is-surjective map-surjection
+  is-surjective-map-surj = pr2 f
```
EgbertRijke, line 67: For more consistent naming we should either use `map-surj` or `map-surjection` in both names, but not mix.

EgbertRijke (changes requested):
Just a few minor comments

PR #229, opened 2022-07-05
Alcohols

EgbertRijke:


plt-amy:
I'll admit that on my phone I can't exactly read the Agda, but does this definition ensure that the hydroxyl group is bound to a saturated carbon?

PR #231, opened 2022-07-05
Definition of alkanes, alkenes and alkynes

plt-amy:
I added some prose to the definitions section of the file explaining how I arrived at the definitions, thinking about the chemistry. I hope that's okay 😄


EgbertRijke:
Excellent! Thank you so much for this.

If I may suggest some changes, I think I'd put saturatedness, alkanes, alkenes, and alkynes in four separate files, named accordingly. Then the file names can be more descriptive than the generic "classes". 

Once you do that, can you also add the files to `src/Readme.lagda.md` so that they will be listed on the main indexing page of the webpage?

plt-amy:
Sure! I'm also pretty sure that, chemically speaking, the carbons to which hydroxyls are attached (in the definition of `alcohol`) must be saturated..

EgbertRijke:
Good point about alcohols

PR #233, opened 2022-07-05
Meet and join for the standard finite types

plt-amy:
Needed these for my idea about how to prove the standard polygon graphs are connected: Order the vertices and then compare them — the resulting difference will tell you exactly what path to build. Also to match the natural numbers.

I added a small note on the definition of `Fin` because `inr` being the top element caught me off-guard: hopefully nobody else trips on this!

To be honest I can't ever see a need for `min-Fin-Fin` and `max-Fin-Fin` but `min-Fin-Fin` has such great prosody that I just _had_ to include it.

### DIFF
--- a/src/elementary-number-theory/inequality-standard-finite-types.lagda.md
+++ b/src/elementary-number-theory/inequality-standard-finite-types.lagda.md
@@ -14,6 +14,7 @@ open import elementary-number-theory.natural-numbers using (ℕ; zero-ℕ; succ-
 open import foundation.coproduct-types using (inl; inr)
 open import foundation.dependent-pair-types using (Σ; pair; pr1; pr2)
 open import foundation.empty-types using (empty; ex-falso; is-prop-empty)
+open import foundation.decidable-types
 open import foundation.identity-types using (_＝_; refl; ap)
 open import foundation.propositions using (is-prop; UU-Prop)
 open import foundation.unit-type using (unit; star; is-prop-unit)
@@ -113,4 +114,14 @@ pr1 (pr2 (fin-Poset k)) = leq-fin-Prop
 pr1 (pr1 (pr2 (pr2 (fin-Poset k)))) = refl-leq-Fin
 pr2 (pr1 (pr2 (pr2 (fin-Poset k)))) = transitive-leq-Fin
 pr2 (pr2 (pr2 (fin-Poset k))) = antisymmetric-leq-Fin
+```
+
+### Ordering on the standard finite types is decidable
+
+```agda
+decide-leq-Fin : {k : ℕ} (x y : Fin k) → is-decidable (leq-Fin x y)
+decide-leq-Fin {k = succ-ℕ k} (inl x) (inl y) = decide-leq-Fin x y
+decide-leq-Fin {k = succ-ℕ k} (inl x) (inr y) = inl star
+decide-leq-Fin {k = succ-ℕ k} (inr x) (inl y) = inr (λ x → x)
+decide-leq-Fin {k = succ-ℕ k} (inr x) (inr y) = inl star
 ```
--- a/src/elementary-number-theory/inequality-standard-finite-types.lagda.md
+++ b/src/elementary-number-theory/inequality-standard-finite-types.lagda.md
@@ -119,9 +119,9 @@ pr2 (pr2 (pr2 (fin-Poset k))) = antisymmetric-leq-Fin
 ### Ordering on the standard finite types is decidable
 
 ```agda
-decide-leq-Fin : {k : ℕ} (x y : Fin k) → is-decidable (leq-Fin x y)
-decide-leq-Fin {k = succ-ℕ k} (inl x) (inl y) = decide-leq-Fin x y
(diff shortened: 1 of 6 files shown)


src/elementary-number-theory/inequality-standard-finite-types.lagda.md
```diff
@@ -114,3 +115,13 @@ pr1 (pr1 (pr2 (pr2 (fin-Poset k)))) = refl-leq-Fin
 pr2 (pr1 (pr2 (pr2 (fin-Poset k)))) = transitive-leq-Fin
 pr2 (pr2 (pr2 (fin-Poset k))) = antisymmetric-leq-Fin
 ```
+
+### Ordering on the standard finite types is decidable
+
+```agda
+decide-leq-Fin : {k : ℕ} (x y : Fin k) → is-decidable (leq-Fin x y)
```
EgbertRijke, line 122: To be consistent with the naming scheme, we'd call this `is-decidable-leq-Fin`

src/elementary-number-theory/maximum-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,57 @@
...
+open import elementary-number-theory.inequality-standard-finite-types using (leq-Fin)
+open import elementary-number-theory.natural-numbers using (ℕ; zero-ℕ; succ-ℕ)
+
+open import foundation.coproduct-types using (inl; inr)
+open import foundation.unit-type using (star)
+
+open import univalent-combinatorics.standard-finite-types using (Fin)
+```
+
+# Maximum on the standard finite types
```
EgbertRijke, line 19: We usually don't have markdown headers of level `#` in our files, although some old files that I haven't cleaned up yet might still have it. But it's better to have a header `## Idea` here, which explains what the max is supposed to do, and after that a header `## Definition` where you start your formal definition.

src/elementary-number-theory/maximum-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,57 @@
...
+
+```agda
+max-Fin : ∀ {k} → Fin k → Fin k → Fin k
+max-Fin {k = succ-ℕ k} (inl x) (inl y) = inl (max-Fin x y)
+max-Fin {k = succ-ℕ k} (inl x) (inr _) = inr star
+max-Fin {k = succ-ℕ k} (inr _) (inl x) = inr star
+max-Fin {k = succ-ℕ k} (inr _) (inr _) = inr star
+
```
EgbertRijke, line 27: Now the definition is complete. What follows are properties, so we usually close the code block here and have a header `## Properties`. After that you can have level `###` headers that say briefly what the next property is.

src/elementary-number-theory/maximum-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,57 @@
...
+leq-right-leq-max-Fin {succ-ℕ k} (inr x) (inl y) (inl z) p = star
+leq-right-leq-max-Fin {succ-ℕ k} (inr x) (inl y) (inr z) p = star
+leq-right-leq-max-Fin {succ-ℕ k} (inr x) (inr y) (inl z) p = star
+leq-right-leq-max-Fin {succ-ℕ k} (inr x) (inr y) (inr z) p = star
+leq-right-leq-max-Fin {succ-ℕ k} (inl x) (inl y) (inr z) p = p
+
+max-Fin-Fin : {k : ℕ} (n : ℕ) → (Fin (succ-ℕ n) → Fin k) → Fin k
+max-Fin-Fin zero-ℕ f     = f (inr star)
+max-Fin-Fin (succ-ℕ n) f = max-Fin (f (inr star)) (max-Fin-Fin n (λ k → f (inl k)))
+```
```
EgbertRijke, line 57: It would probably be good if this file imports `order-theory.least-upper-bounds` and proves that the max is an instance of a least upper bound

src/elementary-number-theory/minimum-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,59 @@
+---
```
EgbertRijke, line 1: The comments for max also apply here for min

EgbertRijke (changes requested):
Thanks so much for this PR! It looks good. I made some minor comments.

Furthermore, please add `maximum-standard-finite-types` and `minimum-standard-finite-types` to `src/README.md` and to `elementary-number-theory.lagda.md`.

plt-amy:
@EgbertRijke Should I go ahead and make the same changes to `elementary-number-theory.minimum-natural-numbers` and `elementary-number-theory.maximum-natural-numbers` while I'm at it? Not a big deal but I understand if it'd be better in another PR 🙂

EgbertRijke:
Sure! It would be great if you could do that!

src/elementary-number-theory/inequality-standard-finite-types.lagda.md
plt-amy: Fixed

EgbertRijke (approved):

PR #251, opened 2022-07-09
wip: standard linear alkanes

plt-amy:
Step 1 is defining the "generic finite path" with n vertices. Ran into some trouble with the definition of `path-Undirected-Graph` when proving it's connected, though


EgbertRijke:
I moved `is-set-ℕ` recently to `elementary-number-theory.natural-numbers`

EgbertRijke:
I think `standard-finite-path` would be a better name for what you call `generic-finite-path`. "Generic finite path" sounds more like the most general possible finite path, which would be a variable of type `finite-path`. What you're defining, however, is not generic, but it is a special finite path.

plt-amy:
I called it "generic" since it (co)represents finite paths of a given length (so it's "the generic situation of having a finite path in a graph"), but I suppose that's a universal property; and this is a _standard_ construction of an object having that universal property :slightly_smiling_face:

src/graph-theory/standard-finite-paths.lagda.md
```diff
@@ -44,58 +44,58 @@ open import univalent-combinatorics.finite-types
...
+  standard-finite-path-vertex : UU
+  standard-finite-path-vertex = Fin length
 ```
 
 Our encoding of the edges is a bit obfuscated by the use of `fib`, but it boils down to: For an unordered pair of vertices `p`, there is an edge between the two elements of `p` iff there are points `x, y : type-unordered-pair(p)` such that `p(x) = suc(p(y))`.
 
 ```agda
-  generic-finite-path-edge : unordered-pair generic-finite-path-vertex → UU
-  generic-finite-path-edge vertices =
+  standard-finite-path-edge : unordered-pair standard-finite-path-vertex → UU
```
EgbertRijke, line 66: We usually put what is defined first in the name, so that would be `edge-standard-finite-path` in this case

src/graph-theory/standard-finite-paths.lagda.md
```diff
@@ -44,58 +44,58 @@ open import univalent-combinatorics.finite-types
...
-  generic-finite-path-edge-is-finite :
-    (vertices : unordered-pair generic-finite-path-vertex) →
-    is-finite (generic-finite-path-edge vertices)
-  generic-finite-path-edge-is-finite vertices =
-    is-finite-Σ (is-finite-mere-equiv (has-two-elements-type-unordered-pair vertices) is-finite-Fin)
-      ( λ x → is-finite-Σ (is-finite-mere-equiv (has-two-elements-type-unordered-pair vertices) is-finite-Fin)
+      fib (nat-Fin length ∘ element-unordered-pair vertices)
+        (succ-ℕ (nat-Fin length (element-unordered-pair vertices x)))
+
+  standard-finite-path-edge-is-finite :
```
EgbertRijke, line 72: In this case, you're showing `is-finite` so that has to go first in the name. It would be `is-finite-edge-standard-finite-path`. Notice how it matches the codomain of the construction:)

src/graph-theory/standard-finite-paths.lagda.md
```diff
@@ -44,58 +44,58 @@ open import univalent-combinatorics.finite-types
...
 
-We construct the definition of the generic finite path in stages: The type of vertices is the standard finite type on $n$ elements, but the type of edges needs more care to define.
+We construct the standard finite path in stages: The type of vertices is the standard finite type on $n$ elements, but the type of edges needs more care to define.
 
 ```agda
 module _ (length : ℕ) where
-  generic-finite-path-vertex : UU
-  generic-finite-path-vertex = Fin length
+  standard-finite-path-vertex : UU
+  standard-finite-path-vertex = Fin length
```
EgbertRijke, line 60: The name here would be `vertex-standard-finite-path` (see my comment about edges).

src/graph-theory/standard-finite-paths.lagda.md
```diff
@@ -106,33 +106,33 @@ Our encoding of the edges is a bit obfuscated by the use of `fib`, but it boils
...
 ```
 
 ### The generic finite path is connected
 
 ```agda
 module _ (len : ℕ) where
-  generic-finite-path-is-connected-Undirected-Graph
-    : is-connected-Undirected-Graph (generic-finite-path-Undirected-Graph len)
-  generic-finite-path-is-connected-Undirected-Graph x y = unit-trunc-Prop {!   !}
+  standard-finite-path-is-connected-Undirected-Graph
```
EgbertRijke, line 121: The name here would be `is-connected-standard-finite-path-Undirected-Graph`. Again it will match the codomain of the construction.

src/graph-theory/standard-finite-paths.lagda.md
```diff
@@ -142,23 +142,23 @@ module _ (len : ℕ) where
           (raise-path _ _ (find-path-to-top x))
       find-path-to-top (inr star) = refl-path-Undirected-Graph
 
-      absurd : ∀ {l} {P : UU l} {k : ℕ} (x : Fin k) → le-ℕ k (nat-Fin x) → P
```
EgbertRijke: This looks like a statement about `nat-Fin` so it shouldn't be in a file about the standard finite path in undirected graphs. We won't be able to find it here if we need it for another purpose.

EgbertRijke (commented):
Just some minor comments. Thanks so much for this PR!!

EgbertRijke:
The error message says that the `--without-K` flag is not loaded. We also usually load the `--exact-split` flag

EgbertRijke:
One interesting way to get the standard linear graphs is from the polygons:

The type of linear undirected graphs with n vertices is equivalent to the type of (n+1)-gons with a marked vertex. To pass back and forth: From a linear undirected graph to a polygon add one extra vertex and connect it with the ends; from a polygon to a linear graph take the complement of the marked vertex.

EgbertRijke:
I will close this pull request, since it is failing and no new commits have happened in a while. Feel free to make a PR again when it is finished.

PR #257, opened 2022-07-15
Kernels of group homomorphisms and alternating groups A_n

plt-amy:
Piecing together existing definitions: The kernel of `f` is defined as the subgroup generated by a particular subset, and the alternating group is defined as the kernel of the sign homomorphism


plt-amy:
You can add a `flags:` section to the `agda-lib` file. So e.g. the 1Lab has:
```
flags:
  --cubical
  --no-import-sorts
  --postfix-projections
  --rewriting
  --guardedness
  --two-level
  -W noNoEquivWhenSplitting
```
For agda-unimath it'd probably be just:
```
flags:
  --without-K
  --exact-split
```
unless you can think of other flags that should be on in every file :slightly_smiling_face:

PR #297, opened 2022-08-18
distance-between-multiples is decidable

blu-bird:
in modular-arithmetic:  int-ℤ-Mod-bounded : (k : ℕ) → (x : ℤ-Mod (succ-ℕ k)) → leq-ℤ (int-ℤ-Mod (succ-ℕ  k) x) (int-ℕ (succ-ℕ k))
in integers: nonneg-nonpos-eq-zero : (x : ℤ) → (is-nonnegative-ℤ x) → (is-nonnegative-ℤ (neg-ℤ     x)) → is-zero-ℤ x
in absolute-value-integers: nonneg-int-eq-int-abs-ℤ : (x : ℤ) → is-nonnegative-ℤ x → int-abs-ℤ x ＝ x

these names could probably be improved, but maybe wait to change them until after the main proof is merged in?


blu-bird:
other lemmas added: 
neg-div-ℤ: (x y : ℤ) → div-ℤ x y → div-ℤ (neg-ℤ x) y
is-decidable-div-ℤ 
is-decidable-div-ℤ-Mod

used to prove decidability of distance-between-multiples in `bezouts-lemma` file

src/elementary-number-theory/absolute-value-integers.lagda.md
```diff
@@ -46,6 +46,10 @@ abs-neg-ℤ (inl x) = refl
 abs-neg-ℤ (inr (inl star)) = refl
 abs-neg-ℤ (inr (inr x)) = refl
 
+nonneg-int-eq-int-abs-ℤ : (x : ℤ) → is-nonnegative-ℤ x → int-abs-ℤ x ＝ x
```
EgbertRijke, line 49: I would name this `int-abs-is-nonnegative-ℤ`

src/elementary-number-theory/integers.lagda.md
```diff
@@ -392,6 +392,10 @@ decide-is-nonnegative-ℤ :
   {x : ℤ} → (is-nonnegative-ℤ x) + (is-nonnegative-ℤ (neg-ℤ x))
 decide-is-nonnegative-ℤ {inl x} = inr star
 decide-is-nonnegative-ℤ {inr x} = inl star
+
+nonneg-nonpos-eq-zero :
```
EgbertRijke, line 396: I would name this `is-zero-is-nonnegative-neg-is-nonnegative-ℤ`

src/elementary-number-theory/modular-arithmetic.lagda.md
```diff
@@ -760,3 +775,42 @@ has-no-fixed-points-succ-Fin :
 has-no-fixed-points-succ-Fin {succ-ℕ k} x =
   has-no-fixed-points-succ-ℤ-Mod (succ-ℕ k) x
 ```
+
+# Corollary: Divisibility is decidable
```
EgbertRijke, line 779: No need to say corollary. Simply say `### The divisibility relation is decidable`.

We generally don't use single `#`. This is only appropriate for the title of the file, but we have a different format for that. I see that I have used a lot of double `##` in this file. I will change those all to triple `###` to be consistent with the rest of the library. 

Some of the older files don't have the correct markup. That's my fault, so you don't need to worry about that.

src/elementary-number-theory/modular-arithmetic.lagda.md
```diff
@@ -760,3 +775,42 @@ has-no-fixed-points-succ-Fin :
...
+
+```agda 
+is-decidable-div-ℤ : (d x : ℤ) → is-decidable (div-ℤ d x)
```
EgbertRijke, line 782: This proof can be shortened a lot by using stuff that is already there:

```
is-decidable-div-ℤ : (d x : ℤ) → is-decidable (div-ℤ d x)
is-decidable-div-ℤ d x =
  is-decidable-iff
    ( div-div-int-abs-ℤ ∘ div-is-zero-mod-ℤ (abs-ℤ d) x)
    ( is-zero-mod-div-ℤ (abs-ℤ d) x ∘ div-int-abs-div-ℤ)
    ( has-decidable-equality-ℤ-Mod
      ( abs-ℤ d)
      ( mod-ℤ (abs-ℤ d) x)
      ( zero-ℤ-Mod (abs-ℤ d)))
```
  
You need to add

```
open import foundation.functions using (_∘_)
```

at the top of the file, and import `div-div-int-abs-ℤ` and `div-int-abs-div-ℤ` from the file `divisibility-integers`

src/elementary-number-theory/modular-arithmetic.lagda.md
```diff
@@ -760,3 +775,42 @@ has-no-fixed-points-succ-Fin :
...
+  path : int-ℕ (abs-ℤ d) ＝ neg-ℤ d
+  path = ap (int-ℕ) (inv (abs-neg-ℤ d)) 
+    ∙ (nonneg-int-eq-int-abs-ℤ (neg-ℤ d) nonpos)
+
+  neg-div-neg-d-x : div-ℤ (neg-ℤ d) x → empty
+  neg-div-neg-d-x = tr (λ p → (div-ℤ p x → empty)) path 
+    (λ div → neg-mod (is-zero-mod-div-ℤ (abs-ℤ d) x div))
+   
+```
+
```
EgbertRijke, line 816: Just a remark about the philosophy of this library: In general, we avoid proofs with long `where` blocks. This one isn't so bad at all, but in general `where` blocks are a sign that the code should be refactored into reusable bits. Note that definitions in `where` blocks cannot be reused outside the current definition. We like it if everything is made reusable as much as possible.

EgbertRijke (commented):
Thank you very much for this pull request. It's very nice work! I left a few remarks in your code. I have the feeling that the lemmas in `bezouts-lemma` can also refactored quite a bit, but I'd need to study it some more to see what is going on, and how we can factor out reusable facts from your proof. In the other comments I made some suggestions for smaller changes.

PR #311, opened 2022-08-31
Add a solver for group equations

anka-213:
This allows writing things like
```agda
example1 : (x y : type-Group G) →
  Id
    (inv-Group G
      (mul-Group G (mul-Group G (mul-Group G x y) (inv-Group G x)) (inv-Group G y)))
      (mul-Group G (mul-Group G (mul-Group G y x) (inv-Group G y)) (inv-Group G x))
example1 x y = simplifyExpr G (x ∷ y ∷ empty-vec) (λ x y → gInv (x *' y *' gInv x *' gInv y))
```
or
```agda
example2 : (x y : type-Group G) →
      Id (mul-Group G (mul-Group G (mul-Group G y x) (inv-Group G x)) (inv-Group G y))
      (unit-Group G)
example2 x y = simplifyExpr G (x ∷ y ∷ empty-vec) (λ x y → (y *' x *' gInv x *' gInv y))
```
in order to automatically reduce an expression according to the group laws.

I've tried to make it so that the proof terms are simple and doesn't contain redundant `refl`s and such.

I don't know if this is something that belongs in this library, but I made it and thought that I might as well show it by making a PR.


EgbertRijke:
Very cool! Thanks so much for doing this!

PR #328, opened 2022-09-15
Towards reals

Ryunaq:
The first commit is my suggestion for separating the sub modules in `elementary-number-theory`, other than the last 7 submodules, I think every submodule can be neatly categorized by the main structure being studied (finite types, N, Z or Q). I got very lost when searching for stuff as a beginner. If you do not like it, I understand.

I'm new to contributing in general, and I wish to start contributing a lot from now, on. So I have a couple of questions before I continue:
- You mentioned on twitch (and in the readme) that `pr1, pr2` should be avoided if possible, but here `presim-unit-ℤ x y` does not have named projections, so it was unavoidable. Should I have created their names?
- Is it fine to split the equational reasoning like I did? i.e.
```
         ＝ mul-ℤ (int-unit-ℤ (pr1 (presim-x-y p))) x by tr (λ - → neg-ℤ x ＝ mul-ℤ - x)
                                                           (inv q)
                                                           (left-neg-unit-law-mul-ℤ x)
         ＝ y                                         by pr2 (presim-x-y p)))
```
- Is it better to do PRs with few theorems (~5), or wait until I have more? Since I'm new, I think I will need more feedback at the beginning.

Best regards,
Fernando

### DIFF
--- a/src/elementary-number-theory.lagda.md
+++ b/src/elementary-number-theory.lagda.md
@@ -7,8 +7,21 @@ title: Elementary Number Theory
 
 module elementary-number-theory where
 
+-- Integers
 open import elementary-number-theory.absolute-value-integers public
 open import elementary-number-theory.addition-integers public
+open import elementary-number-theory.congruence-integers public
+open import elementary-number-theory.difference-integers public
+open import elementary-number-theory.distance-integers public
+open import elementary-number-theory.divisibility-integers public
+open import elementary-number-theory.equality-integers public
+open import elementary-number-theory.greatest-common-divisor-integers public
+open import elementary-number-theory.group-of-integers public
+open import elementary-number-theory.integers public
+open import elementary-number-theory.multiplication-integers public
+open import elementary-number-theory.relatively-prime-integers public
+
+-- Naturals
 open import elementary-number-theory.addition-natural-numbers public
 open import elementary-number-theory.arithmetic-functions public
 open import elementary-number-theory.bezouts-lemma public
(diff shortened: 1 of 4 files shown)


src/elementary-number-theory.lagda.md
```diff
@@ -7,8 +7,21 @@ title: Elementary Number Theory
 
 module elementary-number-theory where
```
EgbertRijke, line 9: I'm not entirely sold on the changes in this file. When settling on a way to organize files, it should not be the case that some files don't have a clear place anymore. This is a significant flaw of this proposal. Furthermore, we should encourage the interplay between integers, natural numbers, rational numbers and itegers modulo k in elementary number theory, not try to separate them. So I propose to undo this change.

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -125,7 +126,12 @@ pr2 (div-neg-ℤ x y (pair d p)) = left-negative-law-mul-ℤ d x ∙ ap neg-ℤ
 ```agda
 neg-div-ℤ : (x y : ℤ) → div-ℤ x y → div-ℤ (neg-ℤ x) y
 pr1 (neg-div-ℤ x y (pair d p)) = neg-ℤ d
-pr2 (neg-div-ℤ x y (pair d p)) = (left-negative-law-mul-ℤ d (neg-ℤ x) ∙ (ap neg-ℤ (right-negative-law-mul-ℤ d x) ∙ (neg-neg-ℤ (mul-ℤ d x) ∙ p)))
+pr2 (neg-div-ℤ x y (pair d p)) = equational-reasoning
```
EgbertRijke, line 129: This is nice! Thank you for incorporating equational reasoning here. Can you start `equational-reasoning` on a new line? Usually, when not everything fits on one line, we start the entire construction on a new line.

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -178,9 +184,11 @@ div-is-unit-ℤ :
 pr1 (div-is-unit-ℤ x y (pair d p)) = mul-ℤ y d
 pr2 (div-is-unit-ℤ x y (pair d p)) =
   associative-mul-ℤ y d x ∙ (ap (mul-ℤ y) p ∙ right-unit-law-mul-ℤ y)
+```
 
--- An integer is a unit if and only if it is 1 or -1.
+### An integer is a unit if and only if it is `1` or `-1`.
```
EgbertRijke, line 189: Thank you so much for making the comments into headers. Perfect!

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -473,3 +499,41 @@ div-div-int-abs-ℤ :
 div-div-int-abs-ℤ {x} {y} =
   div-sim-unit-ℤ (sim-unit-abs-ℤ x) (refl-sim-unit-ℤ y)
 ```
+
+### If we have that `sim-unit-ℤ x y`, then they must differ only by sign
+
+```agda
+sim-unit-ℤ-sign : {x y : ℤ} → sim-unit-ℤ x y → (x ＝ y) + (mul-ℤ neg-one-ℤ x ＝ y)
```
EgbertRijke, line 506: Probably it is better to make a new predicate for `x = ± y`. Probably `is-plus-or-minus-ℤ` is a good name for such a predicate. When we do this, it is also clear what the name of your construction should be: `is-plus-or-minus-sim-unit-ℤ`.

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -473,3 +499,41 @@ div-div-int-abs-ℤ :
...
+  where
+    presim-x-y : (p : is-nonzero-ℤ x) → presim-unit-ℤ x y
+    presim-x-y p = H (λ - → p (pr1 -))
+```
+
+### If we have that `sim-unit-ℤ x y` and both `x` and `y` have the same sign,
+then they are the same
+
+```agda
+-- is-nonnegative-sim-unit-id :
```
EgbertRijke, line 536: A better name for this would be `eq-sim-unit-is-nonnegative-ℤ`. This construction constructs an equality, so that should be the first part of the name. Then it uses `sim-unit` and `is-nonnegative`.

EgbertRijke (commented):
Thank you so much for your PR! Most of your proposals are very good. I would just like to ask you to undo the reorganization of file names in `elementary-number-theory.lagda.md`

It is indeed much better to make smaller pull requests, especially since you're a new contributor. Smaller pull requests are just easier to review.

Many thanks again!

Ryunaq:
Thanks for your comments, I've done all the suggested changes. Thanks for explaining your reasons for rejecting the `elementary-number-theory.lagda.md` changes, I've been convinced. 
I'll continue working towards the reals :)

EgbertRijke:
Thank you so much for your good work Fernando!

PR #329, opened 2022-09-16
reduce-fraction gives a reduced fraction

shiranaiyo:
Added `is-reduced-reduce-fraction-ℤ` and the definitions I needed for it. I have some doubts

- Is the proof of `is-reduced-reduce-fraction-ℤ` fine as it is? All the "mini-lemmas" are very situation specific and I'm unsure if it's better to make them their own definition, though the proof is certainly a bit long.
- Are comments on the proof structure fine?
- Should the definitions involving reduced fractions stay in the Rational file? I'll probably have to add some more definitions that I'll need for plus, minus, times, etc; so maybe they deserve their own file?

Thanks again for your feedback,
Fernando

### DIFF
--- a/src/elementary-number-theory/divisibility-integers.lagda.md
+++ b/src/elementary-number-theory/divisibility-integers.lagda.md
@@ -17,7 +17,7 @@ open import elementary-number-theory.equality-integers using
 open import elementary-number-theory.integers using
   ( ℤ; zero-ℤ; one-ℤ; is-zero-ℤ; neg-ℤ; int-ℕ; is-nonnegative-ℤ; is-one-ℤ;
     is-injective-int-ℕ; is-nonnegative-eq-ℤ; is-nonnegative-int-ℕ; is-neg-one-ℤ;
-    neg-one-ℤ; is-nonzero-ℤ; neg-neg-ℤ; decide-is-zero-ℤ;
+    neg-one-ℤ; is-nonzero-ℤ; neg-neg-ℤ;
     is-zero-is-nonnegative-neg-is-nonnegative-ℤ )
 open import elementary-number-theory.multiplication-integers using
   ( mul-ℤ; mul-ℤ'; left-unit-law-mul-ℤ; associative-mul-ℤ; right-unit-law-mul-ℤ;
@@ -526,7 +526,7 @@ is-plus-or-minus-sim-unit-ℤ {x} {y} H =
       ( is-one-or-neg-one-is-unit-ℤ
         (int-unit-ℤ (pr1 (presim-x-y p)))
         (is-unit-int-unit-ℤ (pr1 (presim-x-y p))) ))
-    ( decide-is-zero-ℤ x)
+    ( is-decidable-is-zero-ℤ x)
   where
     presim-x-y : (p : is-nonzero-ℤ x) → presim-unit-ℤ x y
     presim-x-y p = H (λ - → p (pr1 -))
@@ -555,7 +555,7 @@ eq-sim-unit-is-nonnegative-ℤ {a} {b} H H' K =
             a
             H
             (tr is-nonnegative-ℤ (inv p) H'))))
-      ( decide-is-zero-ℤ a )
+      ( is-decidable-is-zero-ℤ a )
       )
     ( is-plus-or-minus-sim-unit-ℤ K )
 ```
--- a/src/elementary-number-theory/divisibility-integers.lagda.md
+++ b/src/elementary-number-theory/divisibility-integers.lagda.md
@@ -54,6 +54,18 @@ An integer `d` divides an integer `x` if there is an integer `k` such that `mul-
 ```agda
 div-ℤ : ℤ → ℤ → UU lzero
 div-ℤ d x = Σ ℤ (λ k → mul-ℤ k d ＝ x)
+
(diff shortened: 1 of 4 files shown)


src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -53,6 +54,18 @@ An integer `d` divides an integer `x` if there is an integer `k` such that `mul-
 ```agda
 div-ℤ : ℤ → ℤ → UU lzero
 div-ℤ d x = Σ ℤ (λ k → mul-ℤ k d ＝ x)
+
```
EgbertRijke, line 57: excellent!

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -114,6 +127,15 @@ pr2 (div-add-ℤ x y z (pair d p) (pair e q)) =
   ( ap-add-ℤ p q)
 ```
 
+### If `x` divides `y` then `x` divides any multiple of `y`
```
EgbertRijke, line 130: perfect!

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -535,7 +605,26 @@ is-plus-or-minus-sim-unit-ℤ {x} {y} H =
 then they are the same
 
 ```agda
--- eq-sim-unit-is-nonnegative-ℤ :
---   {a b : ℤ} → is-nonnegative-ℤ a → is-nonnegative-ℤ b → sim-unit-ℤ a b → a ＝ b
--- eq-sim-unit-is-nonnegative-ℤ {a} {b} H H' K =
+eq-sim-unit-is-nonnegative-ℤ :
+  {a b : ℤ} → is-nonnegative-ℤ a → is-nonnegative-ℤ b → sim-unit-ℤ a b → a ＝ b
+eq-sim-unit-is-nonnegative-ℤ {a} {b} H H' K =
```
EgbertRijke, line 610: isn't this possible with pattern matching instead of calling ind-coprod?

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -115,9 +126,91 @@ reduce-fraction-ℤ x =
       ( int-reduce-denominator-fraction-ℤ x)
       ( is-positive-int-reduce-denominator-fraction-ℤ x))
 
-{-
 is-reduced-reduce-fraction-ℤ :
   (x : fraction-ℤ) → is-reduced-fraction-ℤ (reduce-fraction-ℤ x)
-is-reduced-reduce-fraction-ℤ x = {!!}
--}
+is-reduced-reduce-fraction-ℤ x =
+  -- Induction on whether alpha is 0
```
EgbertRijke, line 132: again, it is probably better to do pattern matching here:)

EgbertRijke (commented):
excellent PR! Just some minor remarks about using pattern matching instead of ind-coprod.

src/elementary-number-theory/divisibility-integers.lagda.md
shiranaiyo: Do you mean by using a `with ...` abstraction? If yes, I haven't been able to do so with `where` blocks, I'm sorry but I haven't been able to figure that out, not so sure if it's even possible. Same for other ind-coprod.

shiranaiyo:
Hello, I'm asking again here in case you did not see it
> isn't this possible with pattern matching instead of calling ind-coprod?

Do you mean by using a `with ...` abstraction? If yes, I haven't been able to do so with where blocks, I'm sorry but I haven't been able to figure that out, not so sure if it's even possible. Same for other ind-coprod.

EgbertRijke:
Not using `with` blocks. Pattern matching is done by typing a variable in the goal, and then press `C-c C-c` to case split the goal. This allows you to bypass explicit `ind-coprod`. If you're not familiar with this, watch Martin Escardo's first Agda lecture of the HoTTEST summer school. 

https://youtu.be/fJxWLQaaCX4

shiranaiyo:
I think coproducts can't be pattern matched inside a lambda. Trying to split them gives `Cannot split on local variable -`. This is for the inner `ind-coprod`; for the outer one, the element of the coproduct comes from another type, so I think that something like a `with...` abstraction must be done.

In any case, I'd like to ask you to refactor my code, if I see you once I can probably learn how to tackle this sort of stuff better in the future, as I plan to contribute a lot. I'd really appreciate it if you do so, maybe during a stream, if you can/want.

Taneb:
> I think coproducts can't be pattern matched inside a lambda. Trying to split them gives `Cannot split on local variable -`. This is for the inner `ind-coprod`; for the outer one, the element of the coproduct comes from another type, so I think that something like a `with...` abstraction must be done.
> 
> In any case, I'd like to ask you to refactor my code, if I see you once I can probably learn how to tackle this sort of stuff better in the future, as I plan to contribute a lot. I'd really appreciate it if you do so, maybe during a stream, if you can/want.

Agda can pattern-match inside a lambda, but it needs curly bracket. I'm not sure if C-c C-c works for this but you can do it manually. See https://agda.readthedocs.io/en/v2.6.2.2/language/lambda-abstraction.html

PR #333, opened 2022-09-23
refactoring @shiranaiyo's PR (#329)

blu-bird:
successfully refactored the proofs of the lemmas in #329 so that they do not use `ind-coprod` and instead use with-abstraction.


EgbertRijke:
I tried to fix the merge conflicts but I'm afraid I messed something up. Sorry!!

PR #335, opened 2022-09-25
refactor isretr-function-functional-correspondence

malarbol:
Use `contraction` instead of `eq-is-contr`

### DIFF
--- a/src/foundation/functional-correspondences.lagda.md
+++ b/src/foundation/functional-correspondences.lagda.md
@@ -170,7 +170,7 @@ module _
             ( is-functional-functional-correspondence
               ( functional-correspondence-function f)
               ( x))
-          ( f x , refl)))
+            ( f x , refl)))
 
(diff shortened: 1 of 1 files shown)


src/foundation/functional-correspondences.lagda.md
```diff
@@ -166,16 +166,11 @@ module _
...
             ( is-functional-functional-correspondence
               ( functional-correspondence-function f)
               ( x))
-            { ( pair
-                ( function-functional-correspondence
-                  ( functional-correspondence-function f)
-                  ( x))
-                ( pr2 (center (is-contr-total-path (f x)))))}
-            { ( f x , refl)}))
+          ( f x , refl)))
```
EgbertRijke, line 173: The proper indentation level would be with two further spaces. Since it is an argument of `contraction`, it should not start further left than `contraction`.

PR #342, opened 2022-10-05
Minor addition to cardinalities

ElifUskuplu:
I added 
*cardinal inequality is transitive
*(assuming LEM) cardinal inequality is anti-symmetric

To obtain the second, I also added
*(assuming LEM) If there are embeddings between  X and Y in both directions, then there is a mere equivalence between them.

### DIFF
--- a/src/set-theory/cardinalities.lagda.md
+++ b/src/set-theory/cardinalities.lagda.md
@@ -108,7 +108,7 @@ refl-≤-cardinality {l} =
 
 transitive-≤-cardinality :
   {l1 l2 l3 : Level} (X : cardinal l1) (Y : cardinal l2) (Z : cardinal l3) →
-  (X ≤-cardinality Y → Y ≤-cardinality Z → X ≤-cardinality Z)
+  X ≤-cardinality Y → Y ≤-cardinality Z → X ≤-cardinality Z
 transitive-≤-cardinality {l1} {l2} {l3} X Y Z =
   apply-dependent-universal-property-trunc-Set'
   (λ u →
@@ -158,13 +158,13 @@ is-effective-cardinality X Y =
   ( is-effective-unit-trunc-Set (Set _) X Y)
 ```
 
-### Using this and assuming excluded middle , we can show `≤-cardinality` is a partial order by showing that it is anti-symmetric.
+### Using this and assuming excluded middle , we can show `≤-cardinality` is a partial order by showing that it is antisymmetric.
 
 ```agda
-anti-symmetric-≤-cardinality :
+antisymmetric-≤-cardinality :
   {l1 : Level} (X Y : cardinal l1) → (LEM l1) → 
   X ≤-cardinality Y → Y ≤-cardinality X → X ＝ Y
-anti-symmetric-≤-cardinality {l1} X Y lem =
+antisymmetric-≤-cardinality {l1} X Y lem =
   apply-dependent-universal-property-trunc-Set'
   (λ u →
     set-Prop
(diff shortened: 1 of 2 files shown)


src/set-theory/cardinalities.lagda.md
```diff
@@ -93,11 +95,54 @@ unit-leq-cardinality :
...
+
 refl-≤-cardinality : {l : Level} (X : cardinal l) → X ≤-cardinality X
 refl-≤-cardinality {l} =
   apply-dependent-universal-property-trunc-Set'
     ( λ X → set-Prop (leq-cardinality-Prop X X))
     ( λ A → unit-leq-cardinality A A (refl-mere-emb (type-Set A)))
+
+transitive-≤-cardinality :
+  {l1 l2 l3 : Level} (X : cardinal l1) (Y : cardinal l2) (Z : cardinal l3) →
+  (X ≤-cardinality Y → Y ≤-cardinality Z → X ≤-cardinality Z)
```
EgbertRijke, line 111: The outer parentheses are not necessary here

src/set-theory/cardinalities.lagda.md
```diff
@@ -93,11 +95,54 @@ unit-leq-cardinality :
...
+transitive-≤-cardinality {l1} {l2} {l3} X Y Z =
```
EgbertRijke, line 112: Wouldn't this proof simplify if you assume the extra two hypotheses? I think it will

src/set-theory/cardinalities.lagda.md
```diff
@@ -112,3 +157,36 @@ is-effective-cardinality X Y =
   ( equiv-trunc-Prop (extensionality-Set X Y)) ∘e
   ( is-effective-unit-trunc-Set (Set _) X Y)
 ```
+
+### Using this and assuming excluded middle , we can show `≤-cardinality` is a partial order by showing that it is anti-symmetric.
+
+```agda
+anti-symmetric-≤-cardinality :
```
EgbertRijke, line 164: In other places in the library we use "antisymmetric" as one word without a hyphen. Probably better to use it here too.

src/set-theory/cardinalities.lagda.md
```diff
@@ -112,3 +157,36 @@ is-effective-cardinality X Y =
...
+  {l1 : Level} (X Y : cardinal l1) → (LEM l1) → 
+  X ≤-cardinality Y → Y ≤-cardinality X → X ＝ Y
+anti-symmetric-≤-cardinality {l1} X Y lem =
```
EgbertRijke, line 167: I think the proof will be cleaner if you assume H : X \leq Y before starting the proof

EgbertRijke: But let me know if that isn't the case

EgbertRijke (commented):
Thanks so much for these additions! Very good!

src/set-theory/cardinalities.lagda.md
ElifUskuplu: Fixed!

src/set-theory/cardinalities.lagda.md
ElifUskuplu: I've tried and checked the proof in HoTT book again. We should apply induction on Set truncation in X, Y, Z respectively. So I don't have any simplification here.

src/set-theory/cardinalities.lagda.md
ElifUskuplu: Fixed!

src/set-theory/cardinalities.lagda.md
ElifUskuplu: Same like the transitivity.

PR #343, opened 2022-10-05
Defined an infinite distributive law for a meet semilattice and a join complete semilattice.

IanRay11:
Included is the statement of the infinite distributive law in the context of a poset that has binary meets and arbitrary joins. I didn't have a structured guide when creating this file but did my best to stay in line with other developments. I've made notes where I plan to add more of the standard stuff: projecting out the order relation and its properties as well as the fact that the underlying poset is a set (this to stay in line with the other order structure files in the library).

### DIFF
--- a/src/order-theory/infinite-distributive-law.lagda.md
+++ b/src/order-theory/infinite-distributive-law.lagda.md
@@ -1,0 +1,118 @@
+# Title: Infinite-distributive-law 
+
+```agda
+{-# OPTIONS --without-K --exact-split #-}
+
+module order-theory.infinite-distributive-law where
+
+open import foundation.cartesian-product-types using (_×_)
+open import foundation.dependent-pair-types using (Σ; pair; pr1; pr2)
+open import foundation.propositions using
+  ( Prop; is-prop; type-Prop; is-prop-type-Prop; all-elements-equal;
+    is-prop-all-elements-equal; prod-Prop; Π-Prop; function-Prop)
+open import foundation.subtypes using (eq-type-subtype)
+open import foundation.universe-levels
+open import foundation.identity-types
+open import foundation.sets
+
+open import order-theory.posets
+open import order-theory.least-upper-bounds-posets
+open import order-theory.join-complete-semilattice
+open import order-theory.meet-semilattices
+
+```
+
+## Idea
+Let A be a poset that has all binary meets and arbitrary joins. The infinite distributive law states
+for all a : A and for all families b : I → A the following identity holds
+                  a ∧ (‌‌‌⋁ᵢ bᵢ) ＝ ⋁ᵢ (a ∧ bᵢ)
+Note: One could inquire about the dual infinite distributive law but it is not needed at this time.
+
+
+```agda
+
+module _
+  {l1 l2 : Level} (l3 : Level) (P : Poset l1 l2)
+  where
+
+  is-meet-semilattice-and-join-complete-semilattice-poset-Prop : Prop (l1 ⊔ l2 ⊔ lsuc l3)
+  is-meet-semilattice-and-join-complete-semilattice-poset-Prop =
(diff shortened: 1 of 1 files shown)


src/order-theory/infinite-distributive-law.lagda.md
```diff
@@ -0,0 +1,118 @@
+# Title: Infinite-distributive-law
```
EgbertRijke, line 1: Can you write the title in the following format:

```
---
title: Infinite distributive law
---
```

Note also that we use ordinary English for titles of files

src/order-theory/infinite-distributive-law.lagda.md
```diff
@@ -0,0 +1,118 @@
+# Title: Infinite-distributive-law 
+
+```agda
+{-# OPTIONS --without-K --exact-split #-}
```
EgbertRijke, line 4: These options are now loaded globally in agda-unimath.agda-lib. So it is not necessary to include them anymore

src/order-theory/infinite-distributive-law.lagda.md
```diff
@@ -0,0 +1,118 @@
...
+
+module order-theory.infinite-distributive-law where
+
+open import foundation.cartesian-product-types using (_×_)
```
EgbertRijke, line 8: We're moving away from explicit imports. So you can just open import your files without `using`.

src/order-theory/infinite-distributive-law.lagda.md
```diff
@@ -0,0 +1,118 @@
...
+Note: One could inquire about the dual infinite distributive law but it is not needed at this time.
+
+
+```agda
+
+module _
+  {l1 l2 : Level} (l3 : Level) (P : Poset l1 l2)
+  where
+
+  is-meet-semilattice-and-join-complete-semilattice-poset-Prop : Prop (l1 ⊔ l2 ⊔ lsuc l3)
```
EgbertRijke, line 38: Another suggestion: We could make this the condition on meet-semilattices, that it is join complete. Then we would have `is-join-complete-meet-semilattice-Prop`, `is-join-complete-Meet-Semilattice`, `is-prop-is-join-complete-Meet-Semilattice`, and at the end `Join-Complete-Meet-Semilattice`. Especially the latter reads better than the current proposal.

src/order-theory/infinite-distributive-law.lagda.md
```diff
@@ -0,0 +1,118 @@
...
+    element-Meet-Semilattice-and-Join-Complete-Semilattice
+  join-Meet-Semilattice-and-Join-Complete-Semilattice I f = pr1 (is-join-complete-semilattice-Meet-Semilattice-and-Join-Complete-Semilattice I f)
+
+```
+
+## We now state the infinite distributive law
+
+```agda
+
+Infinite-Distributive-Law : (l1 l2 l3 : Level) → UU (lsuc l1 ⊔ lsuc l2 ⊔ lsuc l3)
```
EgbertRijke, line 106: This can be lower case. Probably call it `infinite-distributive-law-Join-Complete-Meet-Semilattice` because there might be other infinite distributive laws, and this one is for join complete meet semilattices

EgbertRijke (commented):
Nice work. Thanks for working on this!

src/order-theory/infinite-distributive-law.lagda.md
IanRay11: It may read nicer but I was trying to give a defintion that captured the concept that a frame is a Poset with binary meets and arbitrary joins and satisfies the infinite distributive law. I didn't want to have any biases towards one structure or the other. Let me know if this is misguided.

IanRay11: I do agree with shortening the name by rearranging as you suggested.

PR #345, opened 2022-10-07
Addition on rationals and a few lemmas

shiranaiyo:
I now realize it's unfeasible to use rationals directly, what I need is to show that `sim-fraction-ℤ` preserves all relevant properties, I'll do that next time.


src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -56,16 +60,55 @@ is-reduced-fraction-ℤ x =
 ```agda
 ℚ : UU lzero
 ℚ = Σ fraction-ℤ is-reduced-fraction-ℤ
+
```
shiranaiyo, line 63: I'm unsure of this names

PR #346, opened 2022-10-13
Renaming...

IanRay11:
Renamed file and most terms


IanRay11:
Looks like the check whole formalization fails because infinite-distributive-law calls the old file name which can’t be found. I’ll update the infinite-distributive-law and make a PR later today.

PR #359, opened 2022-11-26
addition on rationals [DO NOT MERGE, VERY SLOW]

blu-bird:
proved enough lemmas about addition on rationals to show Q is a group under addition


blu-bird:
also testing whether if it is just my computer that compiles this slowly or if I screwed up

blu-bird:
> also testing whether if it is just my computer that compiles this slowly or if I screwed up
as you can see from the check, the proofs of all of these lemmas is really slow, and I'm not sure how to streamline this -- I wouldn't like to singlehandedly triple the checking time of the whole system so I wouldn't merge this until this gets fixed!

PR #361, opened 2022-12-06
UU-Contr to Contr

fredrik-bakke:
Looked like an oversight to me.


EgbertRijke:
Thanks! That was indeed an oversight. Previously, the type of propositions was called `UU-Prop`, and the type of sets was called `UU-Set`. Since those changed to `Prop` and `Set` respectively, it makes sense to change `UU-Contr` to `Contr` too. Many thanks!

PR #364, opened 2023-01-05
Transport lemmas

morphismz:
This is a collection of transport lemmas, essentially those from the IAS HoTT book. This is my first pull request (ever), so please let me know if something needs to be changed.


EgbertRijke:
Thank you so much. It is a great idea to collect the results about transport in one file!

Note that I have often managed to avoid computations of transport, by coming up with some other methods that don't require it. This was the reason why there aren't a lot of transport computations in the library. But it's great to have them!

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+open import foundation.function-extensionality
+open import foundation.identity-types
+open import foundation.universe-levels
+open import foundation.universal-property-dependent-pair-types
+```
+
+Transporting through along p through a family B ∘ f is transporting along ap f p through B. Useful when B≡id and f is a type family defined by univalence. 
+
+```agda
+tr-through-pre-comp :
```
EgbertRijke, line 26: This function already exists in the library and is called `tr-ap`. You can find it in foundation-core.identity-types among the laws for transport.

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+Transport through a family of cartesian products
+
+```agda
+tr-through-product :
+  {l1 l2 : Level} {A : UU l1} {a0 a1 : A} (B C : A → UU l2) (p : a0 ＝ a1) (u : B a0 × C a0) →
+  (tr (λ a → B a × C a) p u) ＝ (pair (tr B p (pr1 u)) (tr C p (pr2 u)))
+tr-through-product B C refl u = refl
+```
```
EgbertRijke, line 39: I would call this `tr-prod`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+
+Transport over a base space that is a cartesian product
+
+```agda
+tr-over-product :
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {a0 a1 : A} {b0 b1 : B} (C : A × B → UU l3) (p : a0 ＝ a1) (q : b0 ＝ b1) (u : C (a0 , b0)) →
+  (tr C (eq-pair p q) u) ＝ (tr (λ x → C (a1 , x)) q (tr (λ x → C (x , b0)) p u))
+tr-over-product C refl refl u = refl
```
EgbertRijke, line 47: I would call this `tr-eq-pair`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+When one of the paths is refl.
+
+```agda
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {a0 a1 : A} {b0 b1 : B}
+  where
+  tr-over-product-left-refl :
```
EgbertRijke, line 56: Hm... I'm wondering whether `left-unit-law-tr-eq-pair` would be a good name for this. And `right-unit-law-tr-eq-pair` for the other one. Not sure...

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+A coherence path for transporting over a cartesian product, in case it is useful
+
+```agda
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {a0 a1 : A} {b0 b1 : B} {p : a0 ＝ a1} {q : b0 ＝ b1}
+    where
+
+    expand-pair-outer : (eq-pair p q) ＝ ((eq-pair p refl) ∙ (eq-pair refl q))
```
EgbertRijke, line 74: This lemma doesn't really belong here. Instead it should go to `foundation/equality-cartesian-product-types.lagda.md`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+    expand-pair-outer = ap (λ x → eq-pair x q) (inv right-unit) ∙ (eq-pair-concat p refl refl q)
+
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {a0 a1 : A} {b0 b1 : B}
+  where
+{- the following code works but pops up an error due to the agda version. Keep commented out until agda version is changed.
+
+  tr-over-product-coh : (C : A × B → UU l3) (p : a0 ＝ a1) (q : b0 ＝ b1) (u : C (a0 , b0)) →
```
EgbertRijke, line 82: `coh-tr-eq-pair`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+    (tr-over-product C p q u) ＝ ((ap (λ x → tr C x u) expand-pair-outer)  ∙ (tr-concat (eq-pair p refl) (eq-pair refl q) u ∙ (
+    (ap (tr C (eq-pair refl q)) (tr-over-product-right-refl C p u)) ∙ (tr-over-product-left-refl C q (tr (λ x → C (x , b0)) p u) ))))
+  tr-over-product-coh C refl refl u = refl -}
+```
+
+Now for dependent pairs. First, transporting through a family of dependent pairs.
+
+```agda
+tr-through-dependent-pair :
```
EgbertRijke, line 91: `tr-Σ`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+  {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A}  {B : A → UU l2} (C : (x : A) → B x → UU l3) (p : a0 ＝ a1) (z : Σ (B a0) (λ x → C a0 x)) →
+  (tr (λ a → (Σ (B a) (λ x → C a x))) p z) ＝ pair (tr B p (pr1 z)) (tr (ind-Σ C) (eq-pair-Σ p refl) (pr2 z))
+tr-through-dependent-pair C refl z = refl
+```
+
+Transporting over a base space of dependent pairs. 
+
+```agda
+tr-over-dependent-pair : {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A}  {B : A → UU l2} {b0 : B a0} {b1 : B a1}
```
EgbertRijke, line 100: `tr-eq-pair-Σ`

I'm not sure if transport over a family of the form `ind-Σ C` is a wise thing to have in the signature. Perhaps better to make C a type family over the Σ-type.

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+  (C : (x : A) → B x → UU l3) (p : a0 ＝ a1) (q : path-over (B) p b0 b1) (u : C a0 b0) →
+  (tr (ind-Σ C) (eq-pair-Σ p q) u) ＝ tr (λ x → C a1 x) q (tr (ind-Σ C) (eq-pair-Σ p refl) u)
+tr-over-dependent-pair C refl refl u = refl
+```
+
+Transporting through a family of functions.
+
+```agda
+tr-through-function : {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A} (B : A → UU l2) (C : A → UU l3) (p : a0 ＝ a1) (f : B a0 → C a0) →
```
EgbertRijke, line 109: `tr-function-type`

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+Transporting through a family of identity types. Note that tr-id-right is already defined in foundation-core.identity-types
+
+```agda
+tr-id-left :
+  {l1 : Level} {A : UU l1} {a b c : A} (q : Id b c) (p : Id b a) →
+  Id (tr (λ y → Id y a) q p) ((inv q) ∙ p)
+tr-id-left refl refl = refl
+
```
EgbertRijke, line 121: `tr-id-right` seems missing.

Usually `id` in a name refers to the identity function. Perhaps better call them `tr-Id-left` and `tr-Id-right`.

EgbertRijke (commented):
Great work! I'm just suggesting some names that should fit better with the naming scheme of the library.

src/foundation/transport.lagda.md
morphismz: Ah, I misunderstood what the purpose of ```tr-ap``` was. I'll delete this.

src/foundation/transport.lagda.md
morphismz: that seems like good name!

src/foundation/transport.lagda.md
morphismz: should I edit that in with this revision, or with a new pull request?

src/foundation/transport.lagda.md
morphismz: I think tr-id-right is in foundation-core.identity-types . Should it be moved? I thought that might mess with preexisting code.

src/foundation/transport.lagda.md
```diff
@@ -0,0 +1,136 @@
...
+tr-id-two-sided :
+  {l1 : Level} {A : UU l1} {a0 a1 : A} (q : a0 ＝ a1) (p : a0 ＝ a0) →
+  (tr (λ y → y ＝ y) q p) ＝ (((inv q) ∙ p) ∙ q)
+tr-id-two-sided refl p = inv right-unit
+
+tr-through-fx＝gy :
```
morphismz, line 127: Are these latter two names good? Or should I change them to something like ```tr-id-fx\= fy```+

src/foundation/transport.lagda.md
morphismz: I agree that the ```tr-Id-____``` is a better convention. But, we already seem to have ```tr-id-right``` in foundation-core.identity-types``` using the lower case i. Would it be better to avoid this confusion and just use ```tr-=-left```, ect.?

morphismz:
it seems like the definition ```coh-tr-eq-pair``` causes an error. I was told this error is due to the agda version. So I simply commented that line of code out. Maybe there is a better way to deal with that?

PR #366, opened 2023-01-07
Lemmas Regarding Dependent Paths

morphismz:
we define the groupoidal operators on dependent paths, show they satisfy the expected identites, and then prove they are equivalences.


morphismz:
The file is loading very slowly and it seems to keeping popping a certain error. I was told this error is due to the agda version and only happens on 2.6.2 and 2.6.3; but the work flows check the file against 2.6.2, so I expect an error. I'll try to figure out how to fix this.

morphismz:
Fixed some of the issues with the definitions. I have not yet defined the equivalences. But the equivalences will take a while to define and it would be good to get feed back on what I already have.

morphismz:
Will open a different pr with just definitions of operators and the coherence paths.

PR #372, opened 2023-01-26
Suspension Lemmas

morphismz:
Lemmas showing that given a function defined in terms of a suspension cocone, we can recover the suspension cocone by applying the function to the suspension structure cocone


morphismz:
I need to rewrite the definition of `suspension-cocone-htpy-suspension-cocone-eq'`

morphismz:
I haven't defined the retraction yet, but the pr is already too long

PR #375, opened 2023-01-27
Change preorder reasoning syntax

fredrik-bakke:
Another simple PR to suggest changing the preorder reasoning syntax to match the other ones.

New usage example:

```agda
preorder X reasoning
  x ≤ a by ineq-1 inside X
    ≤ b by ineq-2 inside X
    ≤ c by ineq-3 inside X
```

as opposed to the old

```agda
x ≤ X by ineq-1
a ≤ X by ineq-2
b ≤ X by ineq-3
c ∎ X
```

Note in particular the insertion of the preorder in `preorder X reasoning_` and the use of the preposition `inside`.

Depending on preference this can be changed, of course. For instance, the preposition "within" can be used instead.

### DIFF
--- a/src/foundation/equational-reasoning.lagda.md
+++ b/src/foundation/equational-reasoning.lagda.md
@@ -171,35 +171,32 @@ logical-equivalence-reasoning
 
 ### Equational reasoning for preorders
 
-Note: In an equational reasoning argument, the preorder is always specified at the last step. So do we really need to specify it at each of the earlier steps?
-
 ```agda
-private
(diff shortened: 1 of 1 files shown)


EgbertRijke:
That's a good improvement, thanks!

I remembered I tried with "in" instead of "inside", which agda wouldn't let me do because "in" is taken up by the syntax of agda.

fredrik-bakke:
Yes, I noticed the same thing 😅

PR #389, opened 2023-01-29
reorganize `section-comp` and variants

fredrik-bakke:
### Summary
- Move `section-comp` and its variants from `foundation-core.equivalences` to their rightful home, `foundation-core.sections` (and similar for retractions).
- Rename `section-comp` to `section-left-factor-htpy` and add a non-homotopy variant (and similar for retractions).
- Rename `section-comp'` to `section-comp-htpy` and add a non-homotopy variant (and similar for retractions).
- Change the title of the `foundation.sections` files from "Sections of type families" to just "Sections".
- Rename `is-equiv-comp` to `is-equiv-comp-htpy` and change to using the non-homotopy variant where possible. In my opinion, it is clear from the usage of `is-equiv-comp` in the library that its name was misleading.

I hope my pull requests don't come across as pointlessly nitpicky.


EgbertRijke:
These are indeed good improvements over what we had previously. Thank you for those changes!

PR #390, opened 2023-01-29
Some type arithmetic with empty and unit types

fredrik-bakke:
Just some type arithmetic with the empty and unit types.

Am I breaking style with the `is-empty`-variants?

### DIFF
--- a/src/foundation-core/contractible-types.lagda.md
+++ b/src/foundation-core/contractible-types.lagda.md
@@ -251,6 +251,15 @@ abstract
       ( λ x → contraction (H x) (f x))
 ```
 
+### The type of functions into a contractible type is contractible
+
+```agda
+is-contr-function-types :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  is-contr B → is-contr (A → B)
+is-contr-function-types is-contr-B = is-contr-Π λ _ → is-contr-B
+```
+
 ### The type of equivalences between contractible types is contractible
 
 ```agda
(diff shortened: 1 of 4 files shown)


EgbertRijke:
The laws for the empty type were already present in `foundation.universal-property-empty-type`. The fact that a dependent product of contractible types is contractible is in `foundation-core.contractible-types` and it is called `is-contr-Π`

fredrik-bakke:
I see. I must've misunderstood the purpose of the type-arithmetic files. My bad!

EgbertRijke:
Since you were obviously looking for those laws when you were browsing the type-arithmetic files, perhaps the best thing to do is to mention that the universal property of the empty type is in its own file, and that products of contractible types are just under contractible-types.

fredrik-bakke:
Good idea. I'll change the pull request in the morning.

EgbertRijke:
Great, thanks so much!

Perhaps make a new section `## See also` at the bottom of the type-arithmetic files where you want to mention other files, and add html-links to the other files so that it renders nicely on the website.

fredrik-bakke:
Oh yeah, that sounds perfect!

fredrik-bakke:
Alright, I've committed the proposed changes. Did you want me to remove `is-contr-function-types` as well?

EgbertRijke:
Excellent! Can you just rename `is-contr-function-types` to singular `is-contr-function-type`? That's because when you use it, it makes more sense in singular form. We have some other names in the library that end in `-function-type`.

EgbertRijke:
Great! I'll merge this PR once the checks have all passed.

PR #393, opened 2023-01-30
Organizing homotopies

fredrik-bakke:
Some organization of `foundation.homotopies` and `foundation-core.homotopies`.

### Summary
- Add inverse-homotopy versions of most of the basic theorems and apply these where applicable.
  This is a rather pedantic addition, but it's in line with the practice of always recording the inverses of equivalences, and makes reading homotopy proofs marginally easier. For an example, see the changes to `foundation.commuting-cubes.coherence-htpy-parallel-cone-coherence-cube`.
- Add references to function extensionality and equational reasoning
- Fix some titles and names
- Miscellaneous formatting
- Move some parts from `foundation` to `foundation-core`


fredrik-bakke:
Oh, really? I hadn't noticed that. I will have a look at it.

EgbertRijke:
I only cleaned up `foundation.homotopies`. Not `foundation-core.homotopies. So I think your PR is still useful, if you can make it compatible with my work from earlier today.

fredrik-bakke:
I think that is a good choice. I've been trying to avoid issues with this mostly, but I'll try to follow your example from now on.

fredrik-bakke:
One thing. I see you've been moving `inv-con-htpy` and the likes around in `foundation.homotopies`. I was considering moving these to `foundation-core` as they do not depend on the dependencies of `foundation.homotopies`. Do you want to keep them there?

fredrik-bakke:
The equivalence theorems will of course have to stay.

EgbertRijke:
I would like the `## Properties` header, to keep the file organisation consistent. Things under `## Idea` should only be informal explanations of what is in that file. Things under `## Properties` are formalized properties of the things that are defined.

fredrik-bakke:
That was an oversight on my part. My merging tool didn't make things particularly obvious.

fredrik-bakke:
Alright, if the checks go through I think that's it.

src/foundation/universal-property-pullbacks.lagda.md
```diff
@@ -49,22 +49,22 @@ The universal property of pullbacks describes the optimal way to complete a diag
...
 
 to a square
 
 ```md
-  C -----> B
-  |        |
-  |        |
-  V        V
-  A -----> X
+C -----> B
```
EgbertRijke, line 59: It looks like this square collapsed

fredrik-bakke:
Nice catch! I was trying out the automatic document formatting function in vscode. I guess that's what did it.

PR #396, opened 2023-01-30
Unit and counit of suspension loop space adjunction

morphismz:
we define the unit and counit of the suspension loop space adjunction. We still need to define the equivalence


morphismz:
Added the equivalence. Remains to verify this equivalence is given by applying the loopspace functor and precomposing with th unit

PR #400, opened 2023-01-31
Remove `using`s

fredrik-bakke:
After Egbert's suggestion yesterday, I figured I could try out regular expressions to remove `using`s in bulk. So here's the result of that. This PR touches all the namespaces. I hope this helps and doesn't just create a bunch of headaches for people when merging.

I did not touch imports of files with axioms in them, as I personally like when a file tells me upfront if it uses any special axioms.

Also, I initially started some other work in this branch, so it includes minor changes to some equivalence proofs. In particular, I changed the uses of `eq-htpy-equiv refl-htpy` to a new `eq-eq-map-equiv refl` which may be a more direct proof, although it still relies on function extensionality.


fredrik-bakke:
I'm still weeding out some small mistakes.

src/foundation/equivalences.lagda.md
```diff
@@ -292,6 +293,17 @@ module _
   pr1 (is-equiv-Prop f) = is-equiv f
   pr2 (is-equiv-Prop f) = is-property-is-equiv f
 
+  eq-eq-map-equiv :
+    {e e' : A ≃ B} → (map-equiv e) ＝ (map-equiv e') → e ＝ e'
+  eq-eq-map-equiv {e} {e'} p =
+    eq-pair-Σ
```
EgbertRijke, line 299: Can we do this more directly with `eq-type-subtype`?

EgbertRijke: Perhaps a clearer name would be `eq-equiv-eq-map-equiv`

src/univalent-combinatorics/distributivity-of-set-truncation-over-finite-products.lagda.md
```diff
@@ -82,7 +82,7 @@ abstract
                 ( map-prod (map-Π (λ x → unit-trunc-Set)) unit-trunc-Set)
                 ( type-Set B))
               ( is-equiv-ev-pair)
-              ( is-equiv-htpy-equiv
+              ( pr2
```
EgbertRijke, line 85: Code is clearer if we use explicit names instead of projections. Can you revert this change?

src/univalent-combinatorics/distributivity-of-set-truncation-over-finite-products.lagda.md
```diff
@@ -100,14 +100,13 @@ abstract
                           ( pr1
                             ( center
                               ( distributive-trunc-Π-Fin-Set k (A ∘ inl)))))
-                        ( Π-Set' (A (inr star)) (λ a → B)))) ∘e
+                        ( Π-Set' (A (inr star)) (λ a → B))))∘e
```
EgbertRijke, line 103: Can you put this space back?

EgbertRijke (commented):
Thanks for doing this! I made a few minor comments

src/univalent-combinatorics/distributivity-of-set-truncation-over-finite-products.lagda.md
EgbertRijke: is it perhaps just `is-equiv-map-equiv`?

src/foundation/equivalences.lagda.md
fredrik-bakke: I agree that's a better name. The name `eq-eq-map-equiv` is derived from `eq-htpy-equiv` and made sense inside `equvialence-extensionality`, but since the lemma now lives in `equivalences` I'll rename it.

fredrik-bakke: Thanks for the suggestion, that does give a cleaner proof.

src/univalent-combinatorics/distributivity-of-set-truncation-over-finite-products.lagda.md
fredrik-bakke: Thanks. I didn't recall the name of the projection at the time.

fredrik-bakke:
Thanks for the feedback! I think I've addressed all of it now.

fredrik-bakke:
By the way, do you have any thoughts on using `using` with axiom files?

EgbertRijke:
Great! The PR looks good now. 

I wouldn't mind if we also removed `using` from the axiom files. Would you be willing to do that? And then I'll merge.

fredrik-bakke:
Sure! But in this case I would like to propose moving `htpy-eq` to `foundation-core.homotopies` so files that don't rely on function extensionality but use `htpy-eq` don't have to import `foundation.function-extensionality`.

EgbertRijke:
Hm... in that case I'll just merge the PR now, and we can take some time to think about it.

The worry I have is that the obvious place to look for `htpy-eq` is in `function-extensionality`, so if we change that then `htpy-eq` won't be anymore where people might be looking for it.

fredrik-bakke:
No wait, Ill remove the usings for the other axiom files first

EgbertRijke:
Ok:)

EgbertRijke:
One thing that has been very important for the design of the agda-unimath library, is that the user shouldn't have to know what depends on what in order to find something. If they're looking for a concept, then they should just be able to look for that concept in order to find it, and not have to think about what was used in order to define it in order to locate it in the library. That's why I'm hesitant about "moving `htpy-eq` because it doesn't depend on function extensionality".

fredrik-bakke:
I understand where you're coming from. What about moving the axiom-part to `foundation.function-extensionality` and leave the safe parts in core. Would that be satisfactory?
Also, I realized the same problem holds for univalence and `equiv-eq`. 😅

fredrik-bakke:
Either way that should be the subject for a different PR. This one is ready for merging now after the checks pass.

EgbertRijke:
The core is currently not safe at all. It contains files that import files from non-core foundation. It's only purpose is to avoid cyclic dependencies while still maintaining logical imports (eg. all basic facts about equivalences can be imported by `open import foundation.equivalences` even though some of them refer to pullbacks which refer to equivalences`. So I'm not worried about the core not being safe.

If there are other solutions than having a 2-step foundation we can definitely consider it, but that would take a major redesign of the current setup. In that case, we should definitely first think it through all the way before we start moving stuff around.

fredrik-bakke:
No, you misunderstood me. I meant the safe parts of `foundation-core.function-extensionality` as in what doesn't rely on the function extensionality axiom. E.g. `htpy-eq` and `FUNEXT`. I'm not making a judgment or suggestion about the core as a whole.

EgbertRijke:
Ah, thanks for clarifying. I think it can't be done. It will create cyclic dependencies.

fredrik-bakke:
I see... I'll have a look anyways 🙃

EgbertRijke:
By the way, I saw your branch is called local-types. Are you interested in localizations and orthogonal factorization systems?

fredrik-bakke:
It could be done! #401

fredrik-bakke:
And yes! 😄
I have some code in a different repository I was considering moving over. Where would be the best place for it do you think?

PR #401, opened 2023-02-01
Separate `postulate`s from statement

fredrik-bakke:
Moves function extensionality postulates from `foundation-core` to `foundation`. Builds on #400.


EgbertRijke:
Why does this pull request repeat many of the changes made in the previous pull request? I don't want to review 452 files again, if it isn't necessary.

fredrik-bakke:
Yes I will see if I can rebase

EgbertRijke:
I still view this as a very non-essential change. Perhaps it is better to focus on more meaningful improvements of the library.

PR #412, opened 2023-02-05
Remove redundant option pragmas

fredrik-bakke:
This pull request removes redundant option pragmas in the library.

The options `--without-K` and `--exact-split` are enabled in all files by default, so enabling these explicitly in a file is redundant.


fredrik-bakke:
I have a couple of questions before I'll mark this PR as ready for review.

In the Makefile, the following is defined as the options
```
checkOpts :=--without-K --exact-split --guardedness
```
while in the `agda-unimath.agda-lib` file, the flags that are set are
```
flags: --without-K --exact-split --no-import-sorts --auto-inline
```

Should these not agree? Are `--guardedness` flags and/or `--no-import-sorts` flags redundant as well?

fredrik-bakke:
The modules
- `foundation.maybe`
- `foundation.partitions`
have the flag `--allow-unsolved-metas` for seemingly no reason.
Should these be removed as well?

EgbertRijke:
The guardedness option should be loaded by the makefile, but it should not be enabled on the entire library. Some files in the library use it (the files about type-theories), but most files don't. 

The no-import-sorts option allows us to write `Set` for the type of sets and `Prop` for the type of props. This option is used everywhere and is definitely not redundant.

There should be no files with --allow-unsolved-metas. This was an oversight on my part, that they're still there. Thanks for spotting them!

fredrik-bakke:
Thank you for the response!
What I mean is that enabling the option inside a file is redundant, as it is already enabled by the `.agda-lib`-file. For instance, is it not redundant to have the`--no-import-sorts` option explicitly enabled in `foundation-core.universe-levels` (although I think in this particular case it is demonstrative to have the flag there), as it is already enabled globally?

Also, regarding `--allow-unsolved-metas`, in this case, I should point out that the modules
- `finite-group-theory.groups-of-order-2`
- `type-theories.simple-type-theories`
- `type-theories.unityped-type-theories`
also have this flag enabled due to "unplugged" holes.

PR #413, opened 2023-02-06
New `orthogonal-factorization-systems` module

fredrik-bakke:
Here is that promised pull request.

### Summary
Add modules
- `orthogonal-factorization-systems`
- `orthogonal-factorization-systems.extensions-of-maps`
- `orthogonal-factorization-systems.lifts-of-maps`
- `orthogonal-factorization-systems.local-types`
and rename `foundation-core.identity-types.lift` to `foundation-core.identity-types.lift-eq-Σ`.


EgbertRijke:
Could you add a new file `orthogonal-factorization-systems.lagda.md` in `/src` which `open imports` everything in the folder publicly?

fredrik-bakke:
I did

EgbertRijke:
Ah, right, I see it now. Github showed it under a folder so I didn't notice it. Thanks!

src/orthogonal-factorization-systems/local-types.lagda.md
```diff
@@ -0,0 +1,133 @@
...
+
+  is-local-type-Prop : {l : Level} → UU l → Prop (l1 ⊔ l2 ⊔ l)
+  pr1 (is-local-type-Prop A) = is-local-type A
+  pr2 (is-local-type-Prop A) = is-property-is-local-type A
+```
+
+### Locality distributes over Π-types
+
+```agda
+  distributes-over-Π-is-local-family :
```
EgbertRijke, line 71: The naming scheme for distributive laws is `distributive-X-Y` to assert that `X` distributes over `Y`. Distributive laws of types are expected to assert equivalences, so this would be its underlying map. That makes the name `map-distributive-Π-is-local-family`.

EgbertRijke (commented):
just a comment about naming distributivity

src/orthogonal-factorization-systems/local-types.lagda.md
fredrik-bakke: Great, thanks!

fredrik-bakke:
Do you have any thoughts on the naming by the way?
For instance, `is-local-type` could be shortened to `is-local`, and `extensions` could be made singular `extension`, but then `lift` would conflict with equality lifting.

EgbertRijke:
I thought about it, and I concluded that it would be quite reasonable if extensions and lifts are in the OFS folder.

Also the current path lifting definition `lift` is not important enough to have that name. We almost never use it. Perhaps a better name would be `lift-eq-Σ`. Then we can use `lift` for the version of maps. Indeed, singular would be better in definitions, because `f : lift` reads better than `f : lifts` and similar for extension.

fredrik-bakke:
Now that you say it, having them there does make more sense. And sure, I will rename `lift` then.

fredrik-bakke:
In addition, I was thinking about moving `is-equiv-is-equiv-precomp` and its variants from `foundation.equivalences` to `foundation.local-types` and change the wording to `is-equiv-is-local` instead. Do you have any thoughts on this change?

fredrik-bakke:
Turns out `lift` wasn't used anywhere even once.

EgbertRijke:
I see where the idea is coming from to rename `is-equiv-is-equiv-precomp` to `is-equiv-is-local`. It says that a map `f` is an equivalence if every type is `f`-local. On the other hand, this lemma is very basic and is used in situations that aren't really about reflective subuniverses. So I'm inclined to keep the name as it is, at least for now.

The PR is good as it is, so I'll merge it.

PR #414, opened 2023-02-06
Remerge `foundation.slice`

fredrik-bakke:
That's pretty much it.


EgbertRijke:
The `open import foundation.slice` now appears among the imports of `foundation-core` files. Can you list `foundation.slice` among the `foundation` files?

fredrik-bakke:
Oh, damn, right. Sorry about that!

PR #420, opened 2023-02-08
cleaning up lifts and extensions

EgbertRijke:


src/orthogonal-factorization-systems/lifts-of-maps.lagda.md
```diff
@@ -55,25 +76,25 @@ module _
...
-    {X : UU l3} (f : X → B) → is-trunc k (lift-of i f)
-  is-trunc-lift-of k is-trunc-A is-trunc-B f =
+    {X : UU l3} (f : X → B) → is-trunc k (lift i f)
+  is-trunc-lift k is-trunc-A is-trunc-B f =
     is-trunc-Σ
       ( is-trunc-function-type k is-trunc-A)
       ( is-trunc-is-lift k is-trunc-B f)
   
-  is-trunc-lift :
+  is-trunc-total-lift :
```
fredrik-bakke, line 91: Given `compute-total-lift`, isn't this lemma obsolete?

src/orthogonal-factorization-systems/extensions-of-maps.lagda.md
```diff
@@ -160,26 +181,26 @@ module _
...
-    is-trunc k (extension-of i P f)
-  is-trunc-extension-of k is-trunc-P f =
+    is-trunc k (extension i P f)
+  is-trunc-extension k is-trunc-P f =
     is-trunc-Σ
       ( is-trunc-Π k is-trunc-P)
       ( is-trunc-is-extension k (is-trunc-succ-is-trunc k ∘ (is-trunc-P ∘ i)) f)
 
-  is-trunc-extension :
+  is-trunc-total-extension :
```
fredrik-bakke, line 197: Isn't this obsolete due to `compute-total-extension`?

fredrik-bakke (changes requested):
Thank you for this! It helps to not just have to guess at how things should be when I make additions.
While you're cleaning up, could you fix the links in the "See also" section? I forgot to change these when moving the files.

src/orthogonal-factorization-systems/lifts-of-maps.lagda.md
fredrik-bakke: At the very least we can strengthen the statement.

src/orthogonal-factorization-systems/extensions-of-maps.lagda.md
fredrik-bakke: Or I guess at least the proof can be simplified.

EgbertRijke:
The lemmas about truncation levels aren't obsolete, but they can be simplified. Thanks for catching this!

fredrik-bakke:
Looks good now, thanks again!

PR #422, opened 2023-02-08
update ap-const

morphismz:
change notation used in definition

### DIFF
--- a/src/foundation-core/identity-types.lagda.md
+++ b/src/foundation-core/identity-types.lagda.md
@@ -173,7 +173,7 @@ ap-inv f refl = refl
 
 ap-const :
   {i j : Level} {A : UU i} {B : UU j} (b : B) {x y : A}
-  (p : x ＝ y) → (ap (const A B b) p) ＝ (const (x ＝ y) (b ＝ b) refl) p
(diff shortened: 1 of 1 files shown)


src/foundation-core/identity-types.lagda.md
```diff
@@ -172,7 +173,7 @@ ap-inv f refl = refl
 
 ap-const :
   {i j : Level} {A : UU i} {B : UU j} (b : B) {x y : A}
-  (p : x ＝ y) → (ap (λ a → b) p) ＝ refl
+  (p : x ＝ y) → (ap (const A B b) p) ＝ (const (x ＝ y) (b ＝ b) refl) p
```
EgbertRijke, line 176: Instead of `(const (x ＝ y) (b ＝ b) refl) p` you can just say `refl`

morphismz: I thought it would be beneficial to phrase it like this, so that its easier to see this as a homotopy (and then use funext when we need an equality of functions). But I can change it to just refl

EgbertRijke: Ah, I understand your idea. Still, let's keep it simple. Thanks again for doing this!

PR #432, opened 2023-02-09
Lifting squares and lifting properties

fredrik-bakke:
### Summary
- Define the pullback hom
- Define lifting squares
- Define lifting properties
- Define orthogonality of maps
- If the domain and codomain of a function are local at that function, then the function is an equivalence
- Refactor split surjections
- Files on commuting 3-simplices, commuting 3-simplices of homotopies, commuting triangles of homotopies


EgbertRijke:
There's a nice way to look at lifting problems, which we could try to use in the library. Note that if `f : A -> B` and `g : X -> Y` are maps, then we get a commuting square

```md
  X^B -----> X^A
   |          |
   |          |
   V          V
  Y^B -----> Y^A
```

Let `P` be the pullback and let `u : X^B -> P` be the canonical map (we call it the gap map in the library). A "lifting problem" in your terminology asks of an element in `P` whether it is in the image of `u`, and a lifting structure asks of an element in `P` whether there is something in the fiber of `u`. 

Then all diagonal lifts exists if `u` is surjective, and `f` is orthogonal to `g` if `u` is an equivalence.

fredrik-bakke:
Thanks for the comment!
I was planning on including this as a theorem. Would you rather have it be the definition?

By the way, when you say surjective do you mean in the proof-irrelevant sense? I was thinking of defining this as a _mere lifting property_ or something in that direction.

EgbertRijke:
Surjective means the fibers are merely inhabited. It is defined in `foundation.surjective-maps`

src/orthogonal-factorization-systems/lifting-problems.lagda.md
```diff
@@ -0,0 +1,255 @@
...
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {Y : UU l4}
+  (h : A → B) (f : A → X) (g : B → Y) (i : X → Y) (H : coherence-square h f g i)
+  where
+
+  is-lifting-square : (j : X → B) → UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  is-lifting-square j = Σ
+    ( is-extension f h j × is-lift g i j)
```
fredrik-bakke, line 72: Would you have this associated to the right instead?

EgbertRijke:
We could go either way, to have it as a theorem or as a definition. I imagine that some things will be nicer from this perspective, but perhaps not everything.

src/orthogonal-factorization-systems/lifting-problems.lagda.md
EgbertRijke: Could be, yes. It probably doesn't make a huge difference, but we have some infrastructure for Sigma types of the form `Sigma (a : A), Sigma (b : B), C a b` (eg to swap the order of the Sigma). If it is associated to the left then we can't use that directly.

fredrik-bakke:
Also, I guess I can drop the use of "lifting problems" and only refer to "lifting squares" if that makes more sense.

EgbertRijke:
How about `diagonal-lift` for `lifting-structure` and `mere-diagonal-lift` for `is-inhabited diagonal-lift`. The type `is-inhabited` is defined in `foundation.inhabited-types`.

fredrik-bakke:
Sure! And then `is-lifting-square-diagonal` instead of `is-lifting-square-spine`. I think it's best to do the mere version in a different file, but I'm not entirely sure what to call the files in this case. Would you drop the terminology "lifting structure" completely?

fredrik-bakke:
I think you're right that redefining the lifting problem in terms of the pullback-hom would get our boots strapped more quickly, and then the explicit characterization as I've done can be a theorem. This way, there will be less need to transport across this particular equivalence I'm guessing.

fredrik-bakke:
I will mark this PR as a draft, so I can make these changes before merging.

EgbertRijke:
Let's also make a file `pullback-hom` then, to set up some nice infrastructure.

src/foundation/split-surjective-maps.lagda.md
```diff
@@ -34,14 +34,6 @@ is-split-surjective :
 is-split-surjective {A = A} {B} f = (b : B) → Σ A (λ a → f a ＝ b)
 ```
 
-### Split epimorphisms
-
-```agda
-is-split-epimorphism :
-  {l1 l2 : Level} {A : UU l1} {B : UU l2} → (A → B) → UU (l1 ⊔ l2)
-is-split-epimorphism f = sec f
-```
```
fredrik-bakke, line 41: This definition seemed superfluous to me. It was not used anywhere.

src/foundation/split-surjective-maps.lagda.md
```diff
@@ -22,16 +22,22 @@ A map `f : A → B` is split surjective if we can construct for every `b : B` an
...
-is-split-surjective {A = A} {B} f = (b : B) → Σ A (λ a → f a ＝ b)
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2}
+  where
+
+  is-split-surjective : (A → B) → UU (l1 ⊔ l2)
+  is-split-surjective f = (b : B) → Σ A (λ a → f a ＝ b)
+
+  split-surjection : UU (l1 ⊔ l2)
+  split-surjection = Σ (A → B) is-split-surjective
```
EgbertRijke, line 40: You can add `map-split-surjection` and `is-split-surjective-split-surjection` for the underlying map (just pr1) and the fact that it is split surjective (just pr2). The reason to do this is that it is much better to see `map-split-surjection` in later code, than the rather non-descriptive `pr1`.

fredrik-bakke: Sure! Thanks for that. I've been meaning to do this for my definitions, but I keep forgetting.

src/foundation/split-surjective-maps.lagda.md
EgbertRijke: True. I think it was ok to remove this one.

But don't make a habit of removing code that isn't used anywhere. If you do that recursively on the whole library we'll end up with an empty library:)

fredrik-bakke: 😂

src/orthogonal-factorization-systems/lifting-problems.lagda.md
fredrik-bakke: Sorry, but I can't be bothered with this after all. The lemma `issec-map-lifting-square-lift` does not play well with reassociating the definition :(

fredrik-bakke:
Alright, at this point I will mark this PR as ready for review. Of course, pieces are still missing, but I'd like to get some feedback and perhaps merge this PR before continuing.

Some notes:
- I went with `lifting-operations` instead of `lifting-structures`.
- At the moment, `mere-diagonal-lift` is defined in terms of the inhabitedness of the type of sections. But, in your first comment, you alluded to defining it in terms of surjectivity. I find the latter to be a more elegant definition, so perhaps this can be changed.
- Perhaps `orthogonal-maps` should be called `orthogonality-of-maps` instead.

EgbertRijke:
Great!
- Pullback-hom can be in the ofs folder, that's fine.
- Defining lifting property in terms of surjectivity is totally legit. Let's go with that.
- For orthogonality, I think we should have two variants: 
  * `is-left-orthogonal g f` which asserts that `f` is left orthogonal to `g`. Think of `is-left-orthogonal g` as the predicate on maps that asserts left orthogonality. 
  * `is-right-orthogonal f g` which asserts that `g` is right orthogonal to `f`
  Of course these predicates can be defined in terms of your current `is-orthogonal`, which we should keep. It might look superfluous, but I think it is probably good to have both variants to guide our intuition in the same way as happens in a text on OFSs.

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Definition
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {Y : UU l4}
+  (f : A → X) (g : B → Y)
+  where
+
+  pullback-hom : UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
```
EgbertRijke, line 59: I would suggest that the `pullback-hom` is defined to be the map, and perhaps `type-pullback-hom` for the type it maps into. At least in my understanding of the pullback-hom, it is usually the map.

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### The diagonal pullback-hom map
+
+The pullback-hom type comes equipped with a canonical comparison map
+`(X → B) → pullback-hom` which we can interpret as the map that takes
+a diagonal map `j` from the codomain of `f` to the domain of `g` to
+the fibered map `((g ∘ j) , (j ∘ f) , refl-htpy)`.
+
+```agda
```
EgbertRijke, line 77: It looks a little strange to me that `gap-pullback-hom` takes pre- and postcomposition maps as arguments. I thought it should take `f` and `g` as arguments directly. Perhaps I'm misunderstanding the purpose of `diagonal-pullback-hom`

src/foundation-core/propositions.lagda.md
```diff
@@ -138,6 +139,25 @@ module _
...
+module _
+  {l1 l2 : Level} {A : UU l1} (B : UU l2)
+  where
+
+  is-proof-irrelevant-retract-of :
+    A retract-of B → is-proof-irrelevant B → is-proof-irrelevant A
+  is-proof-irrelevant-retract-of r is-proof-irrelevant-B a =
+    is-contr-retract-of B r (is-proof-irrelevant-B (pr1 r a))
+  
+  is-prop-retract-of : A retract-of B → is-prop B → is-prop A
```
EgbertRijke, line 154: This can be proven using `is-trunc-retract-of` at truncation level -1. Then we don't need `is-proof-irrelevant-retract-of` anymore.

src/orthogonal-factorization-systems/pullback-hom.lagda.md
fredrik-bakke: Ah yes, that does make sense. I'm just used to calling it the Leibniz cotensor, but I didn't want to introduce that terminology. I'll fix it!

src/orthogonal-factorization-systems/pullback-hom.lagda.md
fredrik-bakke: Not sure what the issue is here exactly, but perhaps this will make more sense once I've renamed things.

src/foundation-core/propositions.lagda.md
fredrik-bakke: Oh, nice, thanks!

src/foundation/split-surjective-maps.lagda.md
```diff
@@ -20,41 +23,95 @@ A map `f : A → B` is split surjective if we can construct for every `b : B` an
...
+  pr1 (pr1 is-equiv-is-split-surjective-sec) = sec-is-split-surjective
+  pr2 (pr1 is-equiv-is-split-surjective-sec) = refl-htpy
+  pr1 (pr2 is-equiv-is-split-surjective-sec) = sec-is-split-surjective
+  pr2 (pr2 is-equiv-is-split-surjective-sec) = refl-htpy
+
+  equiv-sec-is-split-surjective : is-split-surjective f ≃ sec f
+  pr1 equiv-sec-is-split-surjective = sec-is-split-surjective
+  pr2 equiv-sec-is-split-surjective = is-equiv-sec-is-split-surjective
+
+  equiv-is-split-surjective-sec : sec f ≃ is-split-surjective f
```
EgbertRijke, line 84: The `foundation.sections` also contains this equivalence. Let's keep yours, delete the other, and make sure all the other places that used the other definition now use yours.

EgbertRijke: Although the other was shorter...

EgbertRijke: Not sure what is best. Perhaps move the shorter version to this file?

src/foundation/split-surjective-maps.lagda.md
```diff
@@ -20,41 +23,95 @@ A map `f : A → B` is split surjective if we can construct for every `b : B` an
...
 ```agda
-is-split-surjective :
-  {l1 l2 : Level} {A : UU l1} {B : UU l2} → (A → B) → UU (l1 ⊔ l2)
-is-split-surjective {A = A} {B} f = (b : B) → Σ A (λ a → f a ＝ b)
+module _
+  where
+
+  is-split-surjective : (A → B) → UU (l1 ⊔ l2)
+  is-split-surjective f = (b : B) → Σ A (λ a → f a ＝ b)
```
EgbertRijke, line 38: It's probably better to use `fib f b` instead of the explicit Sigma.

src/orthogonal-factorization-systems/lifting-operations.lagda.md
```diff
@@ -0,0 +1,65 @@
...
+open import foundation.universe-levels
+
+open import orthogonal-factorization-systems.pullback-hom
+```
+
+## Idea
+
+Given two maps, `f : A → X` and `g : B → Y`, a _lifting operation between `f` and `g`_
+is a choice of lifting square for every commuting square
+
```
EgbertRijke, line 21: Maybe lifting structure is a good name for a lift of a particular fibered map, and lifting operation is the universally quantified variant that assigns to each fibered map from f to g a lifting structure.

src/orthogonal-factorization-systems/lifting-squares.lagda.md
```diff
@@ -0,0 +1,268 @@
...
+  pr2 (pr1 (pr2 (map-lifting-square-extension _))) = refl-htpy
+  pr2 (pr2 (map-lifting-square-extension _)) _ = eq-is-contr (is-prop-unit star star)
+
+  isretr-map-lifting-square-extension :
+    ((map-extension-lifting-square h f (λ _ → star) (λ _ → star) refl-htpy) ∘ map-lifting-square-extension) ~ id
+  isretr-map-lifting-square-extension = refl-htpy
+
+  issec-map-lifting-square-extension :
+    (map-lifting-square-extension ∘ (map-extension-lifting-square h f (λ _ → star) (λ _ → star) refl-htpy)) ~ id
+  issec-map-lifting-square-extension el =
```
EgbertRijke, line 169: The better way to do this is to first characterize the identity type of `lifting-square` to get a nice function that constructs an equality in `lifting-square` and then use that, instead of all the explicit `eq-pair-Σ`'s

fredrik-bakke: I will have to fix that tomorrow

fredrik-bakke: but thank you for the tip

src/orthogonal-factorization-systems/lifting-squares.lagda.md
```diff
@@ -0,0 +1,268 @@
...
+    (j : X → B) → is-lifting-square j → is-lift g i j
+  is-lift-is-lifting-square j = pr2 ∘ pr1
+
+  map-lift-lifting-square : lifting-square → lift g i
+  pr1 (map-lift-lifting-square L) = pr1 L
+  pr2 (map-lift-lifting-square L) = pr2 (pr1 (pr2 L))
+```
+
+## Properties
+
```
EgbertRijke, line 98: The section of Properties should begin with a characterization of `lifting-square`, which defines the following functions:
- `htpy-lifting-square` which asserts what a homotopy is between elements of type `lifting-square`.
- `refl-htpy-lifting-square` defines the reflexivity homotopy of an element of type `lifting-square`.
- `htpy-eq-lifting-square` which turns an equality of elements of type `lifting-square` into a homotopy (using path induction).
- `is-contr-total-htpy-lifting-square` which proves that at an element of type `lifting-square` the type of pairs `(x,y)` where `x` is an element of type `lifting-square` and `y` is an element of type `htpy-lifting-square x y` is contractible. This will make repeated use of `is-contr-total-Eq-structure` but the proof should be very straightforward.
- `is-equiv-htpy-eq-lifting-square` which shows that `htpy-eq-lifting-square` is an equivalence. This should use `fundamental-theorem-id` and the previous proof of contractibility.
- `extensionality-lifting-square` which is the equivalence with underlying map `htpy-eq-lifting-square`
- `eq-htpy-lifting-square` which is the `map-inv-equiv` of `extensionality-lifting-square`.

This is also how in general such characterizations are done. You  might notice that it was better to define `is-lifting-square` fully associated to the right. You have it now defined as a Sigma over a cartesian product.

EgbertRijke: This proof characterizing the identity type in the type of cones is very similar, you can use it as your guiding example. 

https://unimath.github.io/agda-unimath/foundation-core.cones-pullbacks.html#3257

EgbertRijke: One more tip: When you define `htpy-lifting-square` make sure that it is *always* a Sigma-type. I.e., don't pattern match on the variables of type `lifting-square` in order to define `htpy-lifting-square`. Otherwise `is-contr-total-Eq-structure` won't work.

src/foundation/split-surjective-maps.lagda.md
fredrik-bakke: I agree, `distributive-Π-Σ` is so nice.

EgbertRijke:
Should we call 3-simplices "tetrahedra" instead?

fredrik-bakke:
Hmmm, I like "3-simplices". It makes the pattern very obvious

fredrik-bakke:
although I guess the correct terminology is "standard 3-simplices"

src/orthogonal-factorization-systems/lifting-operations.lagda.md
fredrik-bakke: That makes sense, although I find the terminology `lifting-square` a little less cryptic. Maybe I've just become too accustomed to it already.

src/orthogonal-factorization-systems/lifting-squares.lagda.md
fredrik-bakke: Resolved by omission :)

src/orthogonal-factorization-systems/pullback-hom.lagda.md
fredrik-bakke: Please review the current state of `pullback-hom`

src/orthogonal-factorization-systems/lifting-operations.lagda.md
```diff
@@ -0,0 +1,63 @@
...
+```agda
+module _
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {Y : UU l4}
+  (f : A → X) (g : B → Y)
+  where
+
+  diagonal-lift : UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  diagonal-lift = sec (pullback-hom f g)
+
+  _⧄_ = diagonal-lift -- This symbol doesn't have an input sequence :(
```
fredrik-bakke, line 55: Any opinions on this symbol usage?
Note that it there is no input sequence for it in Agda.

src/orthogonal-factorization-systems/lifts-of-maps.lagda.md
```diff
@@ -63,11 +219,15 @@ module _
...
+
+  is-small-total-lift : is-small (l1 ⊔ l3) (total-lift i X)
+  pr1 (is-small-total-lift) = X → A
+  pr2 (is-small-total-lift) = inv-compute-total-lift
 ```
 
 ## Properties
 
-### If `P` is `k`-truncated then the type of lifts is `k`-truncated
+### If `P` is `k`-truncated then the type of lifts of maps is `k`-truncated
```
fredrik-bakke, line 230: Is it preferred to not use ` in headers? It does look a little off like this I think.

src/orthogonal-factorization-systems/orthogonal-maps.lagda.md
```diff
@@ -0,0 +1,65 @@
...
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {Y : UU l4}
+  where
+
+  is-orthogonal : (A → X) → (B → Y) → UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  is-orthogonal f g = is-equiv (pullback-hom f g)
+
+  _⊥_ = is-orthogonal
```
fredrik-bakke, line 34: This symbol is the same as what logicians usually use for the trivial falsity.
Should this definition be removed?

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -0,0 +1,118 @@
...
+commuting square
+
+```md
+          - ∘ f  
+    B → X ----> B → A
+      |            |
+g ∘ - |            | g ∘ -
+      V            V
+    Y → X ----> Y → A.
+          - ∘ f
```
fredrik-bakke, line 31: Is the notation `X^B` etc. preferred?

PR #433, opened 2023-02-10
suspension-structure lemmas

morphismz:
a few lemmas showing that projections play well with the characterization of equalities in suspension-structure

### DIFF
--- a/src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
+++ b/src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
@@ -487,14 +487,11 @@ module _
                   ( equiv-up-suspension
                     ( type-Pointed-Type X)
                     ( type-Pointed-Type Y)))))))))
-
-  pointed-equiv-susp-loop-adj :  [(suspension-Pointed-Type X) →* Y ]  ≃* [ X →* (Ω Y) ]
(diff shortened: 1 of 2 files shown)


src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
```diff
@@ -180,6 +181,52 @@ module _
...
+      (is-contr-equiv (Σ (suspension-structure X Z) (λ c' → c ＝ c'))
+        (inv-equiv (equiv-tot (λ c' → extensionality-suspension-structure c c')))
+        (is-contr-total-path c)) P)
+```
+
+#### The action of paths of the projections have the expected effect
+
+```agda
+module _
+  {l1 l2 : Level} {X : UU l1} {Z : UU l2} (c : suspension-structure X Z)
```
morphismz, line 212: Do you think the c and c' should be made implicit?

EgbertRijke: I think it's better if we leave them explicit. Agda isn't always able to infer elements of Sigma-types.

morphismz: Sounds good.

PR #434, opened 2023-02-10
transport lemmas

morphismz:
adds some lemmas characterizing dependent paths in a type family of functions and dependent paths in a type family of identities

### DIFF
--- a/src/foundation/transport.lagda.md
+++ b/src/foundation/transport.lagda.md
@@ -105,29 +105,28 @@ tr-function-type B C refl f = refl
 #### This implies implies a charaterization of dependent paths in a family of function types
 
 ```agda
-extensionality-path-over-function-type :
+module _
   {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A} (B : A → UU l2) (C : A → UU l3)
-  (p : a0 ＝ a1) {f : B a0 → C a0} {g : B a1 → C a1} →
-  (path-over (λ a → B a → C a) p f g) ≃ (((tr C p) ∘ f) ~ (g ∘ (tr B p)))
-extensionality-path-over-function-type {a1 = a1} B C p {f = f} {g = g} =
-  equiv-funext ∘e
-  ((inv-equiv (equiv-concat (eq-htpy (λ a → ap (λ t → tr C p (f t)) (isretr-inv-tr B p a))) (g ∘ (tr B p)))) ∘e
-  ((equiv-ap (equiv-precomp (equiv-tr B p ) (C a1)) (λ x → tr C p (f (tr B (inv p) x))) g) ∘e
-  (inv-equiv (equiv-concat (tr-function-type B C p f) g))))
+  (p : a0 ＝ a1) {f : B a0 → C a0} {g : B a1 → C a1}
+  where
+  
+  compute-path-over-function-type :
+    (((tr C p) ∘ f) ~ (g ∘ (tr B p))) ≃ (path-over (λ a → B a → C a) p f g)
+  compute-path-over-function-type =
+    equiv-concat (tr-function-type B C p f) g ∘e
+    ((equiv-concat' ((tr C p ∘ f) ∘ tr B (inv p)) (eq-htpy (λ a → (ap g (issec-inv-tr B p a))))) ∘e
+    ((equiv-ap (equiv-precomp (equiv-tr B (inv p) ) (C a1)) ((tr C p) ∘ f) (g ∘ (tr B p))) ∘e
+    equiv-eq-htpy))
 
-coh-sq-path-over-function-type :
-  {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A} (B : A → UU l2) (C : A → UU l3)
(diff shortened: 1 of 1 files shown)


morphismz:
I'm wondering if some more of the variables can be made implicit, but I don't yet have a sense of whether or not this will cause problems later. E.g. extensionality-path-over-fx＝gx has 5 variables, each of which will be a long string in most use cases.

EgbertRijke:
I'm still not convinced that proving transport lemmas is the way to go :)

This was how we did HoTT 10 years ago, but the library is designed to make cleaner proofs possible that avoid transport lemmas. At least for me, when I notice myself computing a transport, then that is usually a sign for me that the proof can be improved and then I'll be looking for a better proof.

EgbertRijke:
Having said that, the lemmas that you proved are nice. I think there are some occasions where we want to use them, so thank you for formalizing them, and my apologies for always saying that transport lemmas should be avoided.

src/foundation/transport.lagda.md
```diff
@@ -122,6 +153,27 @@ tr-fx＝gx :
 tr-fx＝gx f g p q = inv (tr-eq-pair-diagonal (λ z → f (pr1 z) ＝ g (pr2 z)) p q) ∙ (tr-fx＝gy f g p p q)
 ```
 
+#### This implies a characterization of dependent paths in a family of identity types
+
+```agda
+extensionality-path-over-fx＝gx :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f g : A → B)
+  {a0 a1 : A} (p : a0 ＝ a1) (q : f a0 ＝ g a0) (q' : f a1 ＝ g a1) →
+  ((tr (λ x → f x ＝ g x) p q) ＝ q') ≃ ((q ∙ (ap g p)) ＝ ((ap f p) ∙ q'))
```
EgbertRijke, line 162: The type family `λ x → f x ＝ g x` is called `eq-value f g` and is defined in `foundation-core.homotopies`. So, it would be better to write `(tr (eq-value f g) p q) ＝ q'` instead of `(tr (λ x → f x ＝ g x) p q) ＝ q'`. 

Once we do that, we can also see how to improve the name:
`extensionality-path-over-eq-value`. This name is better, because it doesn't refer to variable names `f`, `g`, and `x`. Referring to variable names in names of functions is usually not a great idea.

src/foundation/transport.lagda.md
```diff
@@ -122,6 +153,27 @@ tr-fx＝gx :
...
+extensionality-path-over-fx＝gx f g p q q' =
+  equiv-concat' (q ∙ ap g p)  (ap (λ t → t ∙ q') (inv-inv (ap f p))) ∘e
+  ((equiv-inv-con (inv (ap f p)) (q ∙ ap g p) q') ∘e
+  ((equiv-concat (inv (assoc  (inv (ap f p)) q (ap g p))) q') ∘e
+  (equiv-concat (inv (tr-fx＝gx f g p q)) q')))
+
+nat-sq-path-over-fx＝gx :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f g : A → B)
+  {a0 a1 : A} (p : a0 ＝ a1) (q : f a0 ＝ g a0) (q' : f a1 ＝ g a1) →
+  ((tr (λ x → f x ＝ g x) p q) ＝ q') → ((q ∙ (ap g p)) ＝ ((ap f p) ∙ q'))
```
EgbertRijke, line 172: Here also use `eq-value`

src/foundation/transport.lagda.md
```diff
@@ -99,6 +102,34 @@ tr-function-type :
 tr-function-type B C refl f = refl
 ```
 
+#### This implies implies a charaterization of dependent paths in a family of function types
+
+```agda
+extensionality-path-over-function-type :
```
EgbertRijke, line 108: In other places in the library I've been using `compute-path-over` instead of `extensionality-path-over`.

src/foundation/transport.lagda.md
```diff
@@ -99,6 +102,34 @@ tr-function-type :
...
+  {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A} (B : A → UU l2) (C : A → UU l3)
+  (p : a0 ＝ a1) {f : B a0 → C a0} {g : B a1 → C a1} →
+  (path-over (λ a → B a → C a) p f g) ≃ (((tr C p) ∘ f) ~ (g ∘ (tr B p)))
```
EgbertRijke, line 111: Most liketly it is best to reverse the direction of this equivalence. 

Let's say we're proving a theorem and we want to construct this path-over. Then we want to use this equivalence, so we want to go from the computed version to the path-over version. So then we'll need to invert the equivalence.

Given that our use-case is likely to go in the reverse direction, I think it's best to revert this equivalence.

morphismz:
I will add analgous lemmas for dependent function in a subsequent pr. It will take a bit thought how best to do it.

EgbertRijke:
I made those changes in PR #438

PR #441, opened 2023-02-11
ap-lemma

morphismz:
lemmas for binary action on paths. These are useful for working with horizontal composition of 2-paths, when the latter is defined in terms of the former


morphismz:
I'm not sure if you'll find these definitions useful for this library, but they were necessary for formalizing some features of horizontal composition of 2-paths in a side project of mine. Feel free to close the pr if they aren't needed

EgbertRijke:
Great, thank you!

PR #454, opened 2023-02-16
Cumulative Hierarchy first PR

shiranaiyo:
Definition of the induction principle of the Cumulative Hierarchy


fredrik-bakke:
Thank you for this excellent PR! I have a few comments about your contribution, but will also request a review from Egbert. :)

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,125 @@
...
+open import foundation.transport
+open import foundation.universe-levels
+open import foundation.propositional-truncations
+```
+
+## Idea
+
+The cumulative hierarchy is a model of set theory.
+Instead of introducing it as a HIT, as in the HoTT Book [1, §10.5], we introduce
+it's induction principle, following [2].
```
fredrik-bakke, line 19: It is not clear to me what the _idea_ of cumulative hierarchies is from this text, only what it is used for. Would you care to elaborate a little on what the precise idea behind cumulative hierarchies is?

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,125 @@
...
+  (a : A) → Σ B (λ b → g b ＝ f a)
+
+have-equal-images :
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
+  (A → X) → (B → X) → UU (l1 ⊔ l2 ⊔ l3)
+have-equal-images {l1} {l2} {l3} {A} {B} {X} f g =
+  has-smaller-image f g × has-smaller-image g f
+```
+
+### A type is a pseudo cumulative hierarchy if it has the structure of a cumulative hierarchy, but not necessarily its induction principle.
```
fredrik-bakke, line 39: This title is a bit long. I would instead write
```
### Pseudo cumulative hierarchy

A type is a pseudo cumulative hierarchy if it has the structure of a cumulative hierarchy, but not necessarily its induction principle.
```

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,125 @@
...
+        → ((b : B) → type-trunc-Prop ( Σ A (λ a → Σ (g b ＝ f a)
+                         (λ p → tr P p (IH₂ b) ＝ IH₁ a ))))
+        → tr P (set-ext-pseudo-cumulative-hierarchy V f g e) (ρ f IH₁) ＝ ρ g IH₂)
+    → {A : UU l1} (f : A → type-pseudo-cumulative-hierarchy V) (IH : (a : A) → P (f a))
+    → IP P σ ρ τ (set-pseudo-cumulative-hierarchy V f) ＝ ρ f IH
+```
+
+## Properties
+
+## Examples
```
fredrik-bakke, line 120: I would avoid having a header here when there is no content in the section. Instead, add the header whenever there is something to have underneath it.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,125 @@
...
+
+## References
+
+1. Institute for Advanced Study. Homotopy Type Theory: Univalent Foundations of Mathematics.
+2. Tom de Jong, in collaboration with Nicolai Kraus, Fredrik Nordvall Forsberg and Chuangjie Xu. https://www.cs.bham.ac.uk/~mhe/agda/UF.CumulativeHierarchy.html
```
fredrik-bakke, line 125: Currently, we don't have guidelines for how to add references, although I really think we should have support for this. So what I will just say currently is that you can make the link clickable by adding angle brackets around it. I.e.
`<https://www.cs.bham.ac.uk/~mhe/agda/UF.CumulativeHierarchy.html>`

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke, line 19: its*

shiranaiyo: I'm not sure what you mean by this, do you mean its constructors?

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: I agree that is better, I've changed it

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Those were meant to be placeholders, my bad

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Done! I tried to follow https://unimath.github.io/agda-unimath/foundation.equational-reasoning.html?highlight=references#references

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: corrected, thanks

shiranaiyo:
I've realized that other `induction-principles` have been named the other way I did, I've corrected that

src/set-theory/cumulative-hierarchy.lagda.md
EgbertRijke: See for example the brief descriptions of what a cumulative hierarchy is on

https://en.wikipedia.org/wiki/Cumulative_hierarchy
https://ncatlab.org/nlab/show/cumulative+hierarchy

In this case I like the wiki description more than nlab's, but that is only my opinion. 

The idea of the `## Idea` section is that you briefly describe what this page is about. So answer the question "What is a cumulative hierarchy". The question "How is a specific cumulative hierarchy constructed" can be deferred to later, when the construction is already happening. This will be after you have given the formal specification of a cumulative hierarchy.

src/SUMMARY.lagda.md
```diff
@@ -785,6 +785,7 @@
   - [Cantor space](set-theory.cantor-space.md)
   - [Cardinalities](set-theory.cardinalities.md)
   - [Countable sets](set-theory.countable-sets.md)
+  - [Cumulative Hierarchy](set-theory.cumulative-hierarchy.md)
```
EgbertRijke, line 788: The letter `H` in `Cumulative Hierarchy` can be a small letter `h`.

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke: Right. Apologies for not elaborating in my first comment!

src/SUMMARY.lagda.md
shiranaiyo: Done, also made it lowercase in the file

shiranaiyo:
I also added the variables for the module inside properties, the names got really long, I'd appreciate suggestions

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -117,6 +117,20 @@ module _
...
+
+```agda
+module _
+  {l1 : Level} (l2 : Level) (cumulative-hierarchy : pseudo-cumulative-hierarchy l1)
+  (cumulative-hierarchy-has-induction-principle-cumulative-hierarchy :
+    induction-principle-cumulative-hierarchy l2 cumulative-hierarchy)
+  (cumulative-hierarchy-has-compute-induction-principle-cumulative-hierarchy :
+    compute-induction-principle-cumulative-hierarchy l2 cumulative-hierarchy
+      cumulative-hierarchy-has-induction-principle-cumulative-hierarchy)
+  where
```
fredrik-bakke, line 130: you could give the cumulative hierarchy variable a single character as a name, e.g. `C` or your favorite letter, and then change `cumulative-hierarchy-has-induction-principle-cumulative-hierarchy` to
`induction-principle-cumulative-hierarchy-C` and similarly for the other variables.
Note the usage of prefix notation and the removal of the verb "has".

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Cumulative Hierarchy
+# Cumulative hierarchy
 
 ```agda
 module set-theory.cumulative-hierarchy where
```
fredrik-bakke, line 4: File/module names and titles are conventionally in plural form unless there is only one unique cumulative hierarchy, which I'm guessing there is not.
I.e.
```
# Cumulative hierarchies
```
and
```
module set-theory.cumulative-hierarchies where
```
and then also change the file name to `cumulative-hierarchies.lagda.md`.

shiranaiyo: Hmm there is only one in the same sense there is only one circle, it's unique up to equivalence

fredrik-bakke: From my very shallow understanding. It would make sense to consider a cumulative hierarchy of for instance 0-truncated types, but also a cumulative hierarchy for 1-truncated types, correct?
If so, then these would surely not be equivalent.

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: sounds great!

shiranaiyo:
Made the first change, not so sure about the second

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,137 @@
...
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
+  (A → X) → (B → X) → UU (l1 ⊔ l2 ⊔ l3)
+has-smaller-image {l1} {l2} {l3} {A} {B} {X} f g =
+  (a : A) → Σ B (λ b → g b ＝ f a)
+
+have-equal-images :
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
+  (A → X) → (B → X) → UU (l1 ⊔ l2 ⊔ l3)
+have-equal-images {l1} {l2} {l3} {A} {B} {X} f g =
+  has-smaller-image f g × has-smaller-image g f
```
fredrik-bakke, line 36: I don't feel particularly strongly about this, but at least I have a few comments on terminology here.
The term "small" is generally used in the library to refer to the essential universe level of a construction. A better formulation of `has-smaller-image` could thus be `is-onto-image`.
Moreover, the verb `is` is more prevalent than `has`, so `have-equal-images` should perhaps be renamed to `is-equal-images`.

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke: I guess since you're working in set theory, a better second example would be infinite 0-truncated types.

fredrik-bakke: or -1-truncated types.

shiranaiyo: I'm not sure I agree. While there could be a cumulative hierarchy of -1-truncated types, *the* cumulative hierarchy is the one for sets. I think that "cumulative hierarchy" acts as a proper name. Since Egbert correct my "h" https://github.com/UniMath/agda-unimath/pull/454#discussion_r1110931200 , I think he is ok with the current name.

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Just checked and there is indeed no `have-X` instances, and I don't want to be the first one. `is-equal-images` sounds ungrammatical though, unlike most of the other definitions I have seen. `is-onto-image` sounds like just a surjection to me. Wonder what Egbert thinks.

EgbertRijke: This condition means that `f` lifts against `g`. 

I suggest that we ditch the terminology `smaller-image` because it is not about small in the sense of the library and neither is it about images. 

I think we can use the definition `lift` from the file `orthogonal-factorization-systems.lifts-of-maps` wherever we use `smaller-images` in the present work. 

Would you be willing to try this out Fernando?

src/set-theory/cumulative-hierarchy.lagda.md
EgbertRijke: Won't there be a cumulative hierarchy for each universe level?

shiranaiyo: That's an interesting point, I'm not so sure about that, but I think they would still be equivalent. I think the key constructor
> (i) For every A : U and f : A → V, there is an element set(A, f ) : V.

doesn't really makes a difference between universe levels (at least for how things have been set up), but maybe there's something I'm missing. In any case, that's a theorem I'll try to (dis?)prove.

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Ok I'll try to implement that! (and read about that)

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke: That would be interesting. If that is the case, I agree that the file can keep its name. :)
I read a little bit more about the cumulative hierarchy and agree that my examples aren't directly applicable. Still, the issue of sizes remains.

EgbertRijke: I also agree, let's keep the name.

src/set-theory/cumulative-hierarchy.lagda.md
EgbertRijke: Great! If you get stuck let us know. The pull request is already excellent!

shiranaiyo: Yeah I think I'm stuck, I've tried to elaborate by hand to match the terms, and also used agda as in
```
test :
  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
  (f : A → X) → (g : B → X) → has-smaller-image f g ＝ (lift ? ?)
test f g = refl
```
or 
```
test :
  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
  (f : A → X) → (g : B → X) → has-smaller-image f g ＝ ((a : A) → (lift ? ?))
test f g = refl
```
but I've failed. Since we need a witness for `B` in `has-smaller-image`, I think that forces `(X → A)` to be `unit → B` below, but then the other things do not match.
```
lift :
  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} (i : A → B)
  {X : UU l3} → (X → B) → UU (l1 ⊔ l2 ⊔ l3)
lift {X = X} i f = Σ (X → A) (λ g → f ~ (i ∘ g))
```

fredrik-bakke: Hi Fernando,
I think the best approach here is to define `has-smaller-image` in terms of `lift` instead of trying to prove that they are equal. I suggest the following definition
```agda
has-smaller-image :
  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3} →
  (A → X) → (B → X) → UU (l1 ⊔ l2 ⊔ l3)
has-smaller-image f g = lift g f
```
This states that "`f` has a smaller image than `g`" if `f` lifts against `g`. This means that there is a function `f' : A → B` such that `g ∘ f' ~ f`. If we elaborate a little further on what this, it states that for every `a : A`, there is a `f' a := b : B` such that `g b = f a`. I.e. we have recovered your original definition.

fredrik-bakke: As stated earlier though, the name `has-smaller-image` should also be changed.

shiranaiyo: you are right, my bad, I'm using lift throughout now and removed `have-equal-images`.

fredrik-bakke: Just to elaborate on my last comment, note that your previous definition gave "for every `a : A`, a `b : B` such that `g b = f a`. I.e. it was _a function_ from `A` to dependent pairs such and such... Meanwhile, `lift` is defined as _a dependent pair_ where the first component is a function `f' : A → B` and the other component is a proof that for every `a : A`, we have an equality `g (f' a) = f a`. These are equivalent types, although they are not equal by `refl`.

EgbertRijke:
I think this is looking very good now, and it is ready to be merged. Fredrik, will you do the honors?

EgbertRijke:
I think this is looking very good now, and it is ready to be merged. Fredrik, will you do the honors?

EgbertRijke:
I think this is looking very good now, and it is ready to be merged. Fredrik, will you do the honors?

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -0,0 +1,128 @@
...
+
+  compute-induction-principle-cumulative-hierarchy :
+    induction-principle-cumulative-hierarchy → UU (lsuc (l1 ⊔ l2))
+  compute-induction-principle-cumulative-hierarchy IP =
+    ( P : type-pseudo-cumulative-hierarchy V → UU l2)
+    → ( σ : (x : type-pseudo-cumulative-hierarchy V) → is-set (P x))
+    → ( ρ : {A : UU l1} (f : A → type-pseudo-cumulative-hierarchy V )
+      → ((a : A) → P (f a)) → P (set-pseudo-cumulative-hierarchy V f))
+    → ( τ : {A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V)
+        ( g : B → type-pseudo-cumulative-hierarchy V) (e : lift f g × lift g f)
```
fredrik-bakke, line 99: I think it was good when you had `lift f g × lift g f` defined as a separate predicate on maps. If no better name comes to mind, this could perhaps be defined as a `mutual-lift`, although I personally don't find the name `is-equal-images` that bad either. However, I only think it makes sense for sets. Regarding your comment on grammar earlier, I should note that this is how we usually phrase predicates, even when the verb is not conjugated properly with respect to spoken English.

fredrik-bakke:
Sure, I will merge it once the checks have passed! 😄

src/set-theory/cumulative-hierarchy.lagda.md
EgbertRijke: I think it's fine as it is, using `lift f g × lift g f`. No-one who is familiar with our naming scheme will think that `is-equal-images` should mean that `f` lifts against `g` and `g` lifts against `f`.

PR #455, opened 2023-02-16
Update CONVENTIONS.lagda.md

fredrik-bakke:
Add instructions for adding links and misc. small improvements.

### DIFF
--- a/src/CONVENTIONS.lagda.md
+++ b/src/CONVENTIONS.lagda.md
@@ -6,7 +6,7 @@
 3. Files that are part of the formalisation should be in literate agda using markdown. They should have the file extension `.lagda.md`.
 4. Every file should begin with a header in the following format
 ```md
-# (The title of the file)
+# The title of the file
 ```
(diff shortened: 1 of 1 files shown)


fredrik-bakke:
Also, I think it is time to decide a convention for adding references to articles and other research material.

fredrik-bakke:
Actually, should other files be referenced as `.md`'s or `.html`'s?

EgbertRijke:
I think other files should be referenced as `.md`

fredrik-bakke:
Alright, I will create a second PR to fix that.

src/CONVENTIONS.lagda.md
```diff
@@ -6,17 +6,20 @@
 3. Files that are part of the formalisation should be in literate agda using markdown. They should have the file extension `.lagda.md`.
 4. Every file should begin with a header in the following format
 ```md
-# [The title of the file]
+# (The title of the file)
```
EgbertRijke, line 9: Perhaps even without the parentheses

PR #457, opened 2023-02-18
Coherence

fredrik-bakke:
Rename commutative squares and variants and refactor `foundation.squares-of-identifications`.

The new naming scheme is as follows.
- Coherence-proofs for commuting diagrams of maps are called `coherence-[diagram shape]-maps`. E.g. `coherence-square-maps`.
- Coherence-proofs for commuting diagrams of identifications are called `coherence-[diagram shape]-identifications`. E.g. `coherence-square-identifications`.
- Coherence-proofs for commuting diagrams of homotopies are called `coherence-[diagram shape]-homotopies`. E.g. `coherence-triangle-homotopies`.


EgbertRijke:
I would say that the data of a commuting square includes the four objects, the four maps, and the homotopies. This was the reason why I called it `coherence` and not simply `commuting-square`.

EgbertRijke:
Similarly, I would say that a commuting diagram of shape `D` in a category C consists of a functor `F : D -> C` that factors through the preorder reflection of `D`. This is just another way of saying that all parallel arrows in `D` are mapped to the same arrow in `C`.

EgbertRijke:
So I prefer the old naming scheme

src/foundation/commuting-squares-of-identifications.lagda.md
```diff
@@ -22,20 +22,21 @@ A square of identifications
   y1 ------ z
 ```
 
-is said to _commute_ if there is an identification `left ∙ bottom ＝ top ∙ right`. Such an identification may be called a _coherence_ or _filler_ of the square.
+is said to _commute_ if there is an identification `left ∙ bottom ＝ top ∙ right`.
+Such an identification may be called a _coherence_ or _filler_ of the square.
```
EgbertRijke, line 26: The old sentence said "filler", but I think that is inaccurate. The n-lab uses the word "filler" in this way:

https://ncatlab.org/nlab/show/filler

I think we can simply omit mentioning the word filler here.

fredrik-bakke:
Oh, I see. I misunderstood when you thumbs-up'ed my suggestion regarding this.
How about naming them `htpy-commuting-square`? Or does this conflict with potential identification characterizations of what you would define as `commuting-square`? In my mind, the basic problem here is that a square of identifications has the non-descript name `square`.

src/foundation/commuting-squares-of-identifications.lagda.md
fredrik-bakke: Wow, I have never heard anyone use just _filler_ for diagonal fillers before I think. That is IMO a bad usage of the word. But I've omitted the word as you suggested.

EgbertRijke:
Ah, perhaps I misunderstood what you proposed in your discord message. My bad. I'm sorry that I've put you through extra work. Let's chat on discord.

PR #458, opened 2023-02-18
Refactor identity types

fredrik-bakke:
Some minor refactoring to `foundation.identity-types` and `foundation-core.identity-types`. I realized after I started that there wasn't really that much to refactor after all, but here are a few improvements anyways.

### Highlights
- Fix some references
- Mov things to be in a more natural order
- Add a few headings and some explanation
- Move `lift-eq-Σ` to `foundation-core.equality-dependent-pair-types`

The structure of `foundation-core.identity-types` still feels a little off to me with how the sections are structured, but I don't know exactly how I would want it organized.


fredrik-bakke:
Although it does not follow convention, one possible way to organize this file would be to have sections along the lines of
```
## Groupoidal structure
## Functorial action
## Properties
```

EgbertRijke:
I think the organization is currently ok. Thank you for the cleanup!

PR #460, opened 2023-02-18
[WIP] Modalities

fredrik-bakke:


src/orthogonal-factorization-systems/factorizations-of-maps.lagda.md
```diff
@@ -0,0 +1,107 @@
...
+    Σ ( X → B)
+      ( λ g →
+        Σ ( A → X)
+          ( is-factorization g))
+
+  factorization : (l3 : Level) → UU (l1 ⊔ l2 ⊔ lsuc l3)
+  factorization l3 = Σ (UU l3) (factorization-through)
+```
+
+### Destructors for factorizations
```
fredrik-bakke, line 66: What do you think about this way of organizing the definitions?
I'm not convinced that using "destructor" here is good, perhaps "projections" is better?

fredrik-bakke:
Should "modalities" be specified to _monadic_ modalities?

PR #468, opened 2023-02-20
Creating a page presenting the maintainers

EgbertRijke:


src/SUMMARY.lagda.md
```diff
@@ -5,6 +5,7 @@
 - [Agda-UniMath](README.md)
   - [Home](README.md)
```
fredrik-bakke, line 6: I think these should link to `SUMMARY.md`. Could you fix this in the same PR? :)

EgbertRijke: Are you sure? The summary page only contains the list of files, but the home page (README) displays a few more things before it displays SUMMARY.

fredrik-bakke: At least these links don't work for me. Do they work for you? Maybe the issue is that the README file is not properly uploaded?

EgbertRijke: They just bring me back to the home page, as they should. @jonaprieto when you have time can you take a quick look to see if there is any issue? It's not urgent.

fredrik-bakke: Oh, I see. Then I apologize for bringing it up here.

EgbertRijke: No need to apologize. Thanks for bringing it up!

jonaprieto:
It looks good to me.

jonaprieto (approved):

PR #469, opened 2023-02-21
two loop path-over

morphismz:
computes path-over in a type of two loops


src/synthetic-homotopy-theory/double-loop-spaces.lagda.md
```diff
@@ -103,3 +105,20 @@ eckmann-hilton-Ω² α β =
...
+
+### Transport in a family of double loops
+
+```agda
+module _
+  {l : Level} {X : UU l} {x0 x1 : X}
+  where
+
+  compute-path-over-Ω² :
+    (p : x0 ＝ x1) (α : (refl {x = x0}) ＝ refl) (β : (refl {x = x1}) ＝ refl) →
```
EgbertRijke, line 117: It will be much cleaner if you would use `type-Ω² (X , x0)` and `type-Ω² (X , x1)` rather `(refl {x = x0}) ＝ refl` and `(refl {x = x1}) ＝ refl`

src/synthetic-homotopy-theory/double-loop-spaces.lagda.md
```diff
@@ -103,3 +105,20 @@ eckmann-hilton-Ω² α β =
...
+    (path-over (λ x → (refl {x = x}) ＝ refl) p α β) ≃
+    (((inv left-unit ∙ (horizontal-concat-Id² α (refl {x = p}))) ∙ left-unit ) ＝
```
EgbertRijke, line 119: On the left of an equation I like to associate to the left, and on the right to the right. This makes it a bit more likely that the number of times you need to associate gets minimized. But feel free to ignore this comment:)

EgbertRijke:
One more comment: It is really better to have the computations of pathover go in the reverse direction, i.e., `... ≃ pathover`. The reason is that this is the direction in which you want to apply your computation.

EgbertRijke:
This PR seems to be stale, and the comments have not been addressed. I'm closing this PR for now, but it can be reopened if the comments are addressed.

PR #481, opened 2023-03-03
Axiom L generalizes univalence and axiom K

fredrik-bakke:
Adds proofs that axiom L generalizes univalence and axiom K.


fredrik-bakke:
Should `axiom-L` and `axiom-K` be capitalized as `AXIOM-L` and `AXIOM-K` to match with `UNIVALENCE` and `FUNEXT`?

EgbertRijke:
I'm not sure. Actually I don't like the all-caps very much.

PR #482, opened 2023-03-03
Define orthogonal factorization systems

fredrik-bakke:
Defines orthogonal factorization systems using the intermediate concepts of function classes and factorization operations.

Also does some refactoring on `extensions-of-maps` and `lifts-of-maps`.


src/orthogonal-factorization-systems/orthogonal-factorization-systems.lagda.md
```diff
@@ -0,0 +1,106 @@
...
+```agda
+is-orthogonal-factorization-system :
+  {l lL lR : Level}
+  (L : function-class l l lL)
+  (R : function-class l l lR)
+  → UU (lsuc l ⊔ lL ⊔ lR)
+is-orthogonal-factorization-system {l} L R =
+  ( is-composition-closed-function-class L × is-equiv-closed-function-class L) ×
+  ( ( is-composition-closed-function-class R × is-equiv-closed-function-class R) ×
+  ((A B : UU l) (f : A → B) → is-contr (function-class-factorization L R f)))
```
fredrik-bakke, line 66: I don't like that this definition is halfway univalence ignorant by on the one hand requiring equivalence closure as opposed to identity closure, and on the other hand, requiring a contractible choice of factorizations rather than "essentially unique factorizations" (which would be ugly in a library about univalent math).

This is kind of bad, as I think I want to use some of this theory without invoking univalence.

src/orthogonal-factorization-systems/function-classes.lagda.md
```diff
@@ -0,0 +1,150 @@
...
+  type-Prop (c f) → type-Prop (c g) →
+  type-Prop (c (g ∘ f))
+
+composition-closed-function-class :
+  (l1 l2 : Level) → UU (lsuc l1 ⊔ lsuc l2)
+composition-closed-function-class l1 l2 =
+  Σ (function-class l1 l1 l2) (is-composition-closed-function-class)
+```
+
+A function class that has identities and is composition closed
```
EgbertRijke, line 72: I would put wide function classes in a separate file.

EgbertRijke:
I had a file `wide-function-classes` in mind.

Why the sad smiley in your commit message?

EgbertRijke:
The name `precat` should only be used when constructing elements of type `Precat`, or when working with elements lf type `Precat`

fredrik-bakke:
Is this more like what you wanted?

fredrik-bakke:
If so, then the PR is ready to be merged :)

PR #488, opened 2023-03-06
Show module declarations

jonaprieto:
Review this one with extra care


fredrik-bakke:
> Review this one with extra care

Since you say that, did you do something dubious? 😬

src/category-theory/endomorphisms-of-objects-categories.lagda.md
```diff
@@ -1,13 +1,16 @@
...
+<details><summary>Imports</summary>
+```agda
 open import foundation.dependent-pair-types
 open import foundation.identity-types
 open import foundation.sets
 open import foundation.universe-levels
+open import category-theory.categories
 open import group-theory.monoids
 open import group-theory.semigroups
 ```
```
fredrik-bakke, line 16: I just realized, when inside a certain namespace (here `category-theory`), one usually has focus on other modules in that namespace, which is why we see here that a module from `category-theory` was imported before the others. Usually, these import statements carry more information and thus deserve special emphasis.

fredrik-bakke: We see this happen in very many of the other files as well. I am tempted to say that this should be the convention.

fredrik-bakke (approved):

fredrik-bakke:
I'm happy with this for now. Although as I've said, I expect that Egbert will want some changes.
One thing that is worth pointing out is that this ordering will be hard for users to follow when making contributions, although it's something they can get used to. I would say it is perhaps unnecessary to have the convention that non-public foundation(-core) imports are sorted before other non-public imports.

fredrik-bakke:
On second thought, this would simplify the logic a bit, would it not?

I.e.
1. public imports
2. Agda primitive
3. modules in the same namespace
4. others, sorted

PR #493, opened 2023-03-08
Recursive principles for the Cumulative hierarchy

shiranaiyo:
I've followed very closely [Tom's work](https://www.cs.bham.ac.uk/~mhe/agda/UF.CumulativeHierarchy.html) as those recursion principles are very useful. However I've also noticed that no recursion principles have been stated in agda-unimath (at least not with that name), so I'm unsure if this is the correct naming scheme.

I left the last one in a comment as I wanted to check the name too. It also made me realize I needed `is-set (Prop l)`. Which can be seen as a particular case of
image

Just double checking, this isn't in the library yet, right?


fredrik-bakke:
Hi Fernando! I'll quickly answer your questions.

> I've followed very closely [Tom's work](https://www.cs.bham.ac.uk/~mhe/agda/UF.CumulativeHierarchy.html) as those recursion principles are very useful. However I've also noticed that no recursion principles have been stated in agda-unimath (at least not with that name), so I'm unsure if this is the correct naming scheme.

I'm unsure why this is the case other than that recursion principles are just special cases of induction principles. My feeling is it would not be wrong to add a recursion principle kind of statement to the library. Indeed, in one of my newest PRs I am also doing this.

> I left the last one in a comment as I wanted to check the name too. It also made me realize I needed `is-set (Prop l)`. Which can be seen as a particular case of image
> 
> Just double checking, this isn't in the library yet, right?

This is formalized  as `is-set-Prop` on lines 88-90 in `foundation.sets`.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -117,14 +122,113 @@ module _
...
+  --       (g : B → type-pseudo-cumulative-hierarchy V)
+  --       (e : lift f g × lift g f)
+  --     → (IH₁ : A → Prop l2)
+  --     → (IH₂ : B → Prop l2)
+  --     → ( (a : A) → ∃ B (λ b → Σ (f a ＝ g b)
+  --         ( λ _ → IH₁ a ＝ IH₂ b)))
+  --     → type-Prop (ρ f IH₁) → type-Prop (ρ g IH₂))
+  --   → type-pseudo-cumulative-hierarchy V → Prop l2
+  -- prop-recursion-principle-cumulative-hierarchy-V {l2} ρ τ =
+  --   recursion-principle-cumulative-hierarchy-V (Prop l2) {!!} ρ {!!}
```
fredrik-bakke, line 231: commented code should be removed

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -117,14 +122,113 @@ module _
...
+        (g : B → type-pseudo-cumulative-hierarchy V)
+        (e : lift f g × lift g f)
+      → (IH₁ : A → X)
+      → (IH₂ : B → X)
+      → ( (a : A) → ∃ B (λ b → Σ (f a ＝ g b)
+          ( λ _ → IH₁ a ＝ IH₂ b)))
+      → ( (b : B) → ∃ A (λ a → Σ (g b ＝ f a)
+          ( λ _ → IH₂ b ＝ IH₁ a)))
+      → ρ f IH₁ ＝ ρ g IH₂)
+      → type-pseudo-cumulative-hierarchy V → X
```
fredrik-bakke, line 158: This is a very minor point, but our convention is to place arrows at the end of lines instead of the beginning in cases like this.
Cf. <https://unimath.github.io/agda-unimath/CODINGSTYLE.html> (under Indentation)

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -117,14 +122,113 @@ module _
...
+        ( λ A → Σ (A → type-pseudo-cumulative-hierarchy V)
+          ( λ f → set-pseudo-cumulative-hierarchy V f ＝ v))
+  underlying-function-cumulative-hierarchy =
+    induction-principle-cumulative-hierarchy-V
+      ( lsuc l1) _
+      ( λ - → is-trunc-type-Truncated-Type (set-trunc-Prop _))
+      ( λ {A} f H → unit-trunc-Prop (pair A (pair f refl)))
+      ( λ f g e IH₁ IH₂ hIH₁ hIH₂ → eq-is-prop is-prop-type-trunc-Prop)
+
+  recursion-principle-cumulative-hierarchy-V : {l2 : Level}
```
fredrik-bakke, line 145: The name of this construction should not reference variable names (i.e. here `V`). I suggest `recursion-principle-cumulative-hierarchy` instead.
Cf. <https://unimath.github.io/agda-unimath/CODINGSTYLE.html> (under Names)

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -117,14 +122,113 @@ module _
...
+        where
+          path-f-g = set-ext-pseudo-cumulative-hierarchy V f g e
+          hIH₁' : (a : A) → ∃ B (λ b → Σ (f a ＝ g b) (λ _ → IH₁ a ＝ IH₂ b))
+          hIH₁' a = map-trunc-Prop
+            (λ (b , p , q) → (b , p , (inv (tr-const p _) ∙ q))) (hIH₁ a)
+          hIH₂' : (b : B) → ∃ A (λ a → Σ (g b ＝ f a) (λ _ → IH₂ b ＝ IH₁ a))
+          hIH₂' b = map-trunc-Prop
+            (λ (a , p , q) → (a , p , (inv (tr-const p _) ∙ q))) (hIH₂ b)
+
+  compute-recursion-principle-cumulative-hierarchy-V : {l2 : Level}
```
fredrik-bakke, line 181: this name also references a variable

fredrik-bakke (changes requested):

shiranaiyo:
@fredrik-bakke thank you! I'll be implementing the suggestions soon. With respect to `is-set (Prop l)`, I think it is very different from `is-set-Prop` in `foundation.sets`. The latter just being that for any `A` we have a Prop, that of `is-set A`. So it's kind of the other way around.

fredrik-bakke:
Oh, yeah, you're absolutely right! I didn't even bother reading what `is-set-Prop` actually did before I responded. My bad!

fredrik-bakke:
I am also unable to find Theorem 7.1.11 formalized in the library.
I can note that
- Theorem 7.1.6 is `is-trunc-emb`
- the statement `(A ≃ B) ↪ (A → B)` is `emb-map-equiv`
- And Lemma 3.5.1 should be easy to prove if it's not already in the library

shiranaiyo:
Haha np, funny how `is-set-Prop` can be interpreted both ways. I think I've found 7.1.11 as `is-trunc-Truncated-Type` in `foundations.truncated-types`, but since `Prop` is not judgementally equal to the type of propositional truncations the lemma `is-set (Prop l)` is still needed. I'd appreciate name suggestions for it.

fredrik-bakke:
Would you look at that.

Also, I'm unsure what you mean. The following type checks for me

```agda
is-set-Prop' : (l : Level) → is-set (Prop l)
is-set-Prop' l = is-trunc-Truncated-Type (neg-one-𝕋)
```

shiranaiyo:
Yeah I found `is-trunc-Truncated-Type` because I was about to implemented it but agda complained it already existed. Yay for consistent naming schemes :)

Forgot `Prop` here is defined differently than in the HoTT book, thanks! I'll continue with the cumulative hierarchy then, just using `is-trunc-Truncated-Type (neg-one-𝕋)` when needed.

fredrik-bakke:
Sounds like a good solution for now. I am a little annoyed with `is-set-Prop` at the moment.

EgbertRijke:
The file `foundation.propositional-extensionality` contains the statement you want. It is called `is-set-type-Prop`

EgbertRijke:
I think the confusing naming comes from when we converted to `--no-import-sorts`. This allowed us to rename `UU-Prop` to `Prop`. The names were previously `is-set-Prop` for the property that something is a set and `is-set-UU-Prop` for the theorem that `UU-Prop` is a set. After converting from `UU-Prop` to `Prop` this name clash came to the surface, and apparently I went with the subobtimal `is-set-type-Prop`. I'm open to other suggestions.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +82,141 @@ module _
...
-      induction-principle-cumulative-hierarchy-V)
+    (l2 : Level) → compute-induction-principle-cumulative-hierarchy l2 V
+      (induction-principle-cumulative-hierarchy-V l2))
   where
 
+  underlying-function-cumulative-hierarchy :
+    (v : type-pseudo-cumulative-hierarchy V) →
+    ∃ ( UU l1)
+      ( λ A → Σ (A → type-pseudo-cumulative-hierarchy V)
+        ( λ f → set-pseudo-cumulative-hierarchy V f ＝ v))
```
EgbertRijke, line 139: If you want to follow our line-breaks conventions here, this should be:

```
    ∃ ( UU l1)
       ( λ A → 
         Σ ( A → type-pseudo-cumulative-hierarchy V)
           ( λ f → set-pseudo-cumulative-hierarchy V f ＝ v))
```

If a lambda-expression doesn't fit on a line, make a line break immediately after the arrow.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +82,141 @@ module _
...
+  underlying-function-cumulative-hierarchy =
+    induction-principle-cumulative-hierarchy-V
+      ( lsuc l1) _
+      ( λ - → is-trunc-type-Truncated-Type (set-trunc-Prop _))
```
EgbertRijke, line 143: The dash symbol of this lambda should be `_`, not `-`

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Thanks, I didn't understand well the style guide but I think I do now. I've modified it now.

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: Corrected.

shiranaiyo:
This last commit adds some other useful principles, corrects that the `lift` properties should be propositionally truncated, and implements the membership relation (finally!).

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +82,141 @@ module _
...
+          ((a : A) → P (f a)) → P (set-pseudo-cumulative-hierarchy V f)) →
+    ( { A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V) →
+      ( g : B → type-pseudo-cumulative-hierarchy V)
+      ( e : lift f g × lift g f)
+      ( IH₁ : (a : A) → P (f a))
+      ( IH₂ : (b : B) → P (g b)) →
+      ( (a : A) → ∃ B (λ b → Σ (f a ＝ g b)
+          ( λ p → tr P p (IH₁ a) ＝ IH₂ b))) →
+      ( (b : B) → ∃ A (λ a → Σ (g b ＝ f a)
+                       (λ p → tr P p (IH₂ b) ＝ IH₁ a))) →
```
fredrik-bakke, line 100: indentation

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +85,332 @@ module _
...
-    compute-induction-principle-cumulative-hierarchy l2 V
-      induction-principle-cumulative-hierarchy-V)
+    (l2 : Level) → compute-induction-principle-cumulative-hierarchy l2 V
+      (induction-principle-cumulative-hierarchy-V l2))
   where
+```
+
+### Every element of the cumulative hierarchy is given by a function into the cumulative hierarchy.
+
+```
```
fredrik-bakke, line 141:
``````
```agda
``````

File: src/set-theory/cumulative-hierarchy.lagda.md [hunk 9]
```diff
@@ -78,53 +85,332 @@ module _
...
+    ( ρ : {A : UU l1} → (A → type-pseudo-cumulative-hierarchy V) → (A → X) → X)
+    ( τ : {A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V)
+      ( g : B → type-pseudo-cumulative-hierarchy V)
+      ( e : type-trunc-Prop (lift f g) × type-trunc-Prop (lift g f))
+      ( IH₁ : A → X)
+      ( IH₂ : B → X) →
+      ( (a : A) → ∃ B (λ b → Σ (f a ＝ g b)
+        ( λ _ → IH₁ a ＝ IH₂ b))) →
+      ( (b : B) → ∃ A (λ a → Σ (g b ＝ f a)
+        ( λ _ → IH₂ b ＝ IH₁ a))) →
```
fredrik-bakke, line 239: indentation/product

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +85,332 @@ module _
...
+```agda
+  prop-recursion-principle-cumulative-hierarchy : {l2 : Level}
+    ( ρ : {A : UU l1} → (A → type-pseudo-cumulative-hierarchy V) →
+          (A → Prop l2) → Prop l2)
+    ( τ : {A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V)
+      ( g : B → type-pseudo-cumulative-hierarchy V)
+      ( e : type-trunc-Prop (lift g f))
+      ( IH₁ : A → Prop l2)
+      ( IH₂ : B → Prop l2) →
+      ( (a : A) → ∃ B (λ b → Σ (f a ＝ g b)
```
fredrik-bakke, line 279: indentation/product

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +85,332 @@ module _
...
+    ( type-pseudo-cumulative-hierarchy V) →
+    Prop (lsuc l1)
+  ∈-cumulative-hierarchy-Prop x =
+    simple-prop-recursion-principle-cumulative-hierarchy
+      (λ {A} f → (∃ A (λ a → f a ＝ x)) , is-prop-∃ _ _) e
+    where
+      e : {A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V)
+        ( g : B → type-pseudo-cumulative-hierarchy V) →
+        ( type-trunc-Prop (lift g f)) →
+        ( ∃ A (λ a → f a ＝ x) → ∃ B (λ b → g b ＝ x))
```
fredrik-bakke, line 366: Add a line break to keep things aligned
```agda
e :
  {A B UU l1} ...
```
Cf. <https://unimath.github.io/agda-unimath/CODINGSTYLE.html> (under Indentation)

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +85,332 @@ module _
...
+      ( (b : B) → ∃ A (λ a → Σ (g b ＝ f a)
+        ( λ _ → IH₂ b ＝ IH₁ a))) →
+      ρ f IH₁ ＝ ρ g IH₂) →
+    {A : UU l1} → (f : A → type-pseudo-cumulative-hierarchy V) → (IH : A → X) →
+    recursion-principle-cumulative-hierarchy X σ ρ τ
+      ( set-pseudo-cumulative-hierarchy V f) ＝ ρ f IH
+  compute-recursion-principle-cumulative-hierarchy {l2} X σ ρ τ =
+      compute-induction-principle-cumulative-hierarchy-V l2 (λ _ → X) (λ _ → σ) ρ τ'
+    where
+      τ' : {A B : UU l1} (f : A → pr1 V) (g : B → pr1 V)
```
fredrik-bakke, line 247: line break after `τ' :`

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -78,53 +85,332 @@ module _
...
+            (λ (a , p , q) → (a , p , (inv (tr-const p _) ∙ q))) (hIH₂ b)
+```
+
+A simplification of the recursion principle, when the codomain is `Prop l2`.
+
+```agda
+  prop-recursion-principle-cumulative-hierarchy : {l2 : Level}
+    ( ρ : {A : UU l1} → (A → type-pseudo-cumulative-hierarchy V) →
+          (A → Prop l2) → Prop l2)
+    ( τ : {A B : UU l1} (f : A → type-pseudo-cumulative-hierarchy V)
```
fredrik-bakke, line 274: line break after `τ :`

shiranaiyo:
Thank you @fredrik-bakke, I apologize for not following the guidelines earlier. I believe I'm following all of them now.

fredrik-bakke:
You mustn't apologize, that's why we review! 🙂

To fix CI, install pre-commit (see how-to-install) and then run `make pre-commit` in the terminal

shiranaiyo:
Yup, I got it passing now, thanks.

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke: Missed a spot 😬

fredrik-bakke:
This is certainly good enough to merge now. Excellent work! 😁

PR #499, opened 2023-03-10
Gitattributes

shiranaiyo:
Adds the `.gitattributes` file so that github recognizes the agda code. Other than updating the languages used info
image
it makes the project more discoverable, I believe it is currently not appearing in a search for agda projects on github https://github.com/search?l=Agda&p=1&q=agda&type=Repositories .

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.lagda.md linguist-language=Agda


jonaprieto (approved):
Thanks!

PR #500, opened 2023-03-11
Extensionality

shiranaiyo:
Other than proving extensionality, I added `has-smaller-image` as the previous one was wrong (at least without assuming choice).


fredrik-bakke (approved):
An excellent PR! :)

Issue #195, opened 2022-05-31
Prepare library for first release


elisabethstenholm:
### TODO
- [ ] #714
- [ ] Clean up commented code
- [x] Clean up unused imports (?)
- [x] #479
- [x] Add symbol for coproducts
- [x] Syntax for Id (?)
- [x] Change name of src/README (?)
- [ ] Complete refactoring of synthetic homotopy theory
- [ ] Make sure all files conform to guidelines

EgbertRijke:
Some more suggestions for a release:
- Write documentation
- Write an article presenting the library

EgbertRijke:
website version that corresponds statically to the release (suggested by Leo)

elisabethbonnevier:
Do we want to keep or remove src/labels.lagda.md?

EgbertRijke:
Yes

fredrik-bakke:
Item 3 is resolved by #502, which adds the simplistic script `scripts/remove_unused_imports.py` that finds and removes unused imports. However, this script is not automatic, (it's a bit hacky and takes quite a long time to complete, e.g. 6.4 hours on my Mac for that PR) thus it will have to be occasionally rerun.

fredrik-bakke:
Regarding items 1 and 2, it would be possible (and not very hard) to ban comments and holes with a pre-commit script. Additionally, we could ban `--allow-unsolved-metas`.

PR #501, opened 2023-03-11
Generate module indexes

fredrik-bakke:
Automatically generate both namespace index modules and the main index file (meaning the long list in `SUMMARY.md`) as part of `pre-commit`.

The namespace index modules are generated by
- Create a file if it does not exist
- If there is no title, use the namespace but capitalize the first word and replace dashes with spaces. E.g. `orthogonal-factorization-systems` becomes "Orthogonal factorization systems"
- If there is no Agda-block, create one, otherwise, capture the last Agda block in the file. (this is to preserve possible `OPTION` pragmas as in the case of `type-theories`.
- Declare the namespace module, and import all files in the namespace folder publically and in alphabetic ordering.

This assures that if there is a custom title or other text in the namespace index module from before (as in the case of `univalent-combinatorics`), then this is preserved.

The main index file is generated as follows
- Overwrite the previous file
- Add title "Formalisation in Agda"
- Add entry lists for each namespace. The title of a module is used for the entry, and the lists are sorted by title as well.

Resolves #448. 

Written in collaboration with @jonaprieto.

PR #502, opened 2023-03-12
Remove all unused imports

fredrik-bakke:
Adds a script that goes through each and every import statement of the library and checks if it is used by (simplistically) removing it from the file and then checking if it still typechecks. This script is very slow and is not intended to be used regularly. For instance, it took 6.4 hours for the script to finish on my Mac. Also, it is quite hacky, so users should be cautious.

Also adds a script that, in `foundation` and `foundation-core` modules, checks if every import to a `foundation` module can be changed to an import of its `foundation-core` module. This script is significantly faster, but still takes some time and is not intended to be used regularly.

Progresses #195 (item 3).


fredrik-bakke:
Note: there is some refactoring I want to do in the python files that will have to wait until _after_ #501 is merged.

fredrik-bakke:
Jonathan pointed out to me that we can consider in the future adding targets in the makefile for these scripts, so that they can be applied to only files touched by a branch/commit/PR/something in this direction.

PR #503, opened 2023-03-13
Add `MODULE-INDEX.md` to metafiles

fredrik-bakke:


fredrik-bakke:
This is being worked on in #504 instead.

PR #509, opened 2023-03-14
Restore CONTRIBUTORS.md while finding a CI fix

jonaprieto:
The CI fails due to a call to the GitHub API to generate the contributor's list. While
we find a solution for it, let's restore the corresponding file.

Btw, two additional scripts were added to an ongoing discussion regarding the space convention so that we can keep them for the record and experiment locally. See the history of this description and the comments below to see all the details of why there are not running by default.


EgbertRijke:
I mentioned this in chat already, but for the record let me say it here too. The old convention displayed clearly the structure of the code. This "fix" breaks some of that. For example, in the old convention

```
function
  ( ( ( fist-path x y) .
       ( second-path x y)) .
    ( third-path x y))
```

The structure of the code is clearly visible: The opening parentheses of each part of the definition are exactly below each other when they are at the same level. In the new convention, this got converted to:

```
function
  (((fist-path x y) .
       (second-path x y)) .
    (third-path x y))
```

This is much less readable, because the indentation levels of the arguments of the path concatenation are inconsistent.

Any convention that we are going to settle on must be such that the structure of the code is displayed clearly. It is really helpful to see what is going on in a complicated definition if the convention is helping you see the structure of the definition, so I don't want to accept a PR that removes this.

EgbertRijke:
Of course, we can discuss other possible conventions, but to me the requirement that the formatting displays the structure clearly is a must, and it is more important to me than consistency between our use of parentheses in variable declarations and in function applications.

fredrik-bakke:
I haven't looked too closely at the PR yet, but let me make a comment anyways.
I think it is a pretty general rule with our coding style that before any opening parenthesis `(`, single opening curly brace `{` or double opening curly brace `{{` there should be a whitespace character. Are you considering enforcing this/adding a warning?

src/foundation-core/homotopies.lagda.md
```diff
@@ -27,38 +27,38 @@ functions.
 
 ```agda
 module _
-  {l1 l2 : Level} {X : UU l1} {P : X → UU l2} (f g : (x : X) → P x)
+  { l1 l2 : Level } { X : UU l1 } { P : X → UU l2 } ( f g : ( x : X ) → P x )
   where
 
   eq-value : X → UU l2
-  eq-value x = (f x ＝ g x)
+  eq-value x = ( f x ＝ g x )
```
fredrik-bakke, line 34: To me, none of these additional spaces makes sense, as they're not contributing to displaying any nested tree structure.

Makefile
```diff
@@ -67,13 +67,12 @@ agda-html: src/everything.lagda.md
 SUMMARY.md:
 	@python3 scripts/generate_main_index_file.py
 
-CONTRIBUTORS.md:
-	@python3 scripts/update_contributors.py
+# CONTRIBUTORS.md:
+# 	@python3 scripts/update_contributors.py
```
fredrik-bakke, line 71: Should this be removed instead of just commented out?

scripts/check_agdablock_line_length.py
```diff
@@ -0,0 +1,31 @@
...
+
+MSG = """{fpath}:{j}
+Warning: Agda code block line exceeds 80 characters in width."""
+
+if __name__ == "__main__":
+    for fpath in utils.get_agda_files(sys.argv[1:]):
+        with open(fpath, 'r') as f:
+            contents = f.read()
+        block_start = r'```agda\b'
+        block_end = r'```'
```
fredrik-bakke, line 17: This will match for instance "```md"

CONTRIBUTORS.md
```diff
@@ -0,0 +1,37 @@
+## Contributors
+
+We are grateful to the following people for their contributions to the library.
+In alphabetical order, by GitHub username:
```
EgbertRijke, line 4: This is now ordered by number of commits

scripts/update_contributors.py
```diff
@@ -48,13 +50,10 @@ def get_github_user_name(username):
...
 
-for contributor in sorted_contributors:
+for contributor in contributors:
     login = contributor['login']
-    name = get_github_user_name(login)
-    if name is None or name == "":
-        name = login
     html_url = contributor['html_url']
-    contributors_list.append(f"- [{name} ({login})]({html_url})")
+    contributors_list.append(f"- [{login}]({html_url})")
```
EgbertRijke, line 56: I think it is better if we use `{name}` rather than `{login}`

fredrik-bakke (approved):
Looks great now!

PR #512, opened 2023-03-17
Make `unused_imports_remover` faster and safer

### LINKED ISSUE #523: Some files erroneously end the details block with `<details>` instead of `<\details>`
Some files erroneously end the details block with `<details>` instead of `<\details>`. However, I don't know the regular expression to search and replace this throughout the library. All occurrences of `<details>` followed by a new line character should be replaced by `<\details>`

fredrik-bakke:
- Builds on #502.

So I wanted to try using GPT-4 to improve my programs and chose the script `unused_imports_remover` as a test case.

This PR improves the speed and safety of the program. Now, the original Agda files are only overwritten by working code, so I deem it no longer unsafe for others to use. Moreover, it uses multithreading and a couple of other tricks to make the overall run time significantly shorter. To compare, the previous version took 6.4 hours to finish on my computer while this one only took 55 minutes.

Also gives the same treatment to `foundation_demoter`. It is now safer and takes only 70 seconds to run.

- Also resolves #523.


fredrik-bakke:
Please advice on how I can make this script in `Makefile` ignore files in the folder `src/temp`:

```Makefile
.PHONY : agdaFiles
agdaFiles :
	@rm -rf $@
	@rm -rf src/everything.lagda.md
	@find src -type f \( -name "*.agda" -o -name "*.lagda"  -o -name  "*.lagda.md" \) > $@
	@sort -o $@ $@
	@wc -l $@
	@echo "$(shell (find src -name '*.lagda.md' -print0 | xargs -0 cat ) | wc -l) LOC"
```

fredrik-bakke:
Nevermind! GPT-4 suggested this solution:
```Makefile
.PHONY : agdaFiles
agdaFiles :
	@rm -rf $@
	@rm -rf src/everything.lagda.md
	@find src -type f \( -name "*.agda" -o -name "*.lagda"  -o -name  "*.lagda.md" \) -not -path "src/temp/*" > $@
	@sort -o $@ $@
	@wc -l $@
	@echo "$(shell (find src -name '*.lagda.md' -print0 | xargs -0 cat ) | wc -l) LOC"
```

scripts/demote_foundation_imports.py
```diff
@@ -0,0 +1,169 @@
...
+import shutil
+import utils
+from concurrent.futures import ThreadPoolExecutor
+from fix_imports import *
+
+
+def process_agda_file(agda_file, agda_options, root, temp_dir):
+
+    def get_agda_module_name(file_path):
+        return file_path[len(root) + 1:file_path.rfind('.lagda.md')].replace('/', '.').replace('\\', '.')
```
jonaprieto, line 12: I'd prefer to find this in utils.py

scripts/utils.py
```diff
@@ -102,3 +103,11 @@ def get_equivalence_classes(equivalence_relation, iterable):
...
             partitions.append([e])
     return partitions
+
+
+print_lock = threading.Lock()
+
+
+def thread_safe_print(*args, **kwargs):
+    with print_lock:
+        print(*args, **kwargs)
```
jonaprieto, line 113: Do we want to have this `print_lock = threading.Lock()` here? I'd prefer to see this and the `thread_safe_print` in a separate file.

fredrik-bakke: `multithread_utils.py` for instance?

scripts/demote_foundation_imports.py
fredrik-bakke: Sorry for being a dirty coder 😔

scripts/utils.py
jonaprieto: Perhaps, create a file called multithread, inside a folder called utils. Remember to put the __init__ file in that folder.

fredrik-bakke: What do you want me to do with `utils.py` in this case? I cannot have a folder and a file with the same name.

VojtechStep:
> Nevermind! GPT-4 suggested this solution:
> 
> ```makefile
> .PHONY : agdaFiles
> agdaFiles :
> 	@rm -rf $@
> 	@rm -rf src/everything.lagda.md
> 	@find src -type f \( -name "*.agda" -o -name "*.lagda"  -o -name  "*.lagda.md" \) -not -path "src/temp/*" > $@
> 	@sort -o $@ $@
> 	@wc -l $@
> 	@echo "$(shell (find src -name '*.lagda.md' -print0 | xargs -0 cat ) | wc -l) LOC"
> ```

Actually the manpage says "To ignore a whole directory tree, use -prune rather than checking every file in the tree.", so this should probably be
`@find src -name temp -prune -o -type f \( -name "*.agda" -o -name "*.lagda" -o -name "*.lagda.md" \)`

scripts/remove_unused_imports.py
```diff
@@ -0,0 +1,114 @@
...
+            f"'{agda_module}' Could not find imports. Skipping.")
+        return
+
+    # # We can assume the file would compile, as this is usually the case.
+    # # Otherwise, no harm is done, it will just determine that the file does
+    # # not compile with any of the imports removed either, so the result is the same
+    # # If file doesn't compile, skip
+    # elif (utils.call_agda(agda_options, agda_file) != 0):
+    #     utils.multithread.thread_safe_print(f"'{agda_file}': ERROR! did not typecheck. Skipping.")
+    #     return
```
jonaprieto, line 33: It seems you can remove this.

fredrik-bakke: Fine fine

fredrik-bakke:
> Contributor

This does not seem to work properly. With your suggestion, `everything.lagda.md` still has the import statement `open import temp`, If I on the other hand prune "src/temp" then this import statement is not included, but every file inside temp is.

VojtechStep:
> This does not seem to work properly. With your suggestion, `everything.lagda.md` still has the import statement `open import temp`, If I on the other hand prune "src/temp" then this import statement is not included, but every file inside temp is.

My bad, apparently for `prune` you also need `-print` at the end, otherwise it behaves funky, so the whole command should read
```make
@find src -name temp -prune -o -type f \( -name "*.agda" -o -name "*.lagda"  -o -name  "*.lagda.md" \) -print > $@

```

A propos, since the library only consists of literal Agda markdown files, and the scripts seem to work with that assumption (for detecting titles etc), do we maybe want to drop the *.agda and *.lagda patterns from there?

EgbertRijke:
Is this PR done?

fredrik-bakke:
Give me 15 minutes, I'll finish it up.

fredrik-bakke:
> A propos, since the library only consists of literal Agda markdown files, and the scripts seem to work with that assumption (for detecting titles etc), do we maybe want to drop the *.agda and *.lagda patterns from there?

We could do this, but I personally always lean towards preferring a little redundancy with things like this, for increased robustness. By including these patterns, a person won't get away with committing a `*.agda` file with code that doesn't typecheck at least. :)

fredrik-bakke:
Should be done now if the typechecker agrees. I also fixed #523 quickly, as I was planning on sneaking this in anyways 😅

PR #518, opened 2023-03-17
Cumulative hierarchy

shiranaiyo:
Last work on the cumulative hierarchy. There's still some missing axioms but I'll stop here for now.


shiranaiyo:
Oh also I tried to prove that the CH is independent of universe level but I think this is not provable (though I still think it's true). The problem is that when trying to do recursion, we need an element of `V` from a function `A -> f` but the `A` may be at a higher level than what `V` allows for. Still, I think that an appropriate `A'` with the same image should live in a lower universe, but this is not provable, I think.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -555,6 +585,161 @@ needed.
...
+
+### ∈-induction axiom
+
+```agda
+  ∈-induction-cumulative-hierarchy :
+    {l2 : Level}
+    ( P : type-pseudo-cumulative-hierarchy V → UU l2) →
+    (( x : type-pseudo-cumulative-hierarchy V) → is-prop (P x)) →
+    (( x : type-pseudo-cumulative-hierarchy V) →
+      (( y : type-pseudo-cumulative-hierarchy V) →
```
fredrik-bakke, line 686: `( ( `

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -555,6 +585,161 @@ needed.
...
+                          ( mere-preimage-∈-cumulative-hierarchy
+                            ( unit-trunc-Prop (a , refl)))
+                          ( refl))})
+                    ( ∈-cumulative-hierarchy-mere-preimage H))
+                ( λ H → mere-preimage-∈-cumulative-hierarchy
+                  ( map-universal-property-trunc-Prop
+                    ( ∃-Prop A _)
+                    ( λ {(z , K , refl) →
+                      ( map-trunc-Prop
+                        (λ {(a , refl) → (a , refl)})
```
fredrik-bakke, line 737: `( `

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -555,6 +585,161 @@ needed.
...
+  infinity-axiom-cumulative-hierarchy :
+    ∈-cumulative-hierarchy
+      empty-set-cumulative-hierarchy
+      infinity-cumulative-hierarchy ×
+    ( ( x : type-pseudo-cumulative-hierarchy V) →
+      ∈-cumulative-hierarchy x infinity-cumulative-hierarchy →
+      ∈-cumulative-hierarchy
+        ( pair-cumulative-hierarchy x
+          ( singleton-cumulative-hierarchy x))
+        ( infinity-cumulative-hierarchy))
```
fredrik-bakke, line 665: The indentation looks off here after the product formation.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -555,6 +585,161 @@ needed.
           f x H₁ H₂)
 ```
 
+### Empty set axiom
```
fredrik-bakke, line 588: It could be made more clear that these are not just formulations of axioms, but in fact proofs that "any cumulative hierarchy satisfies 'the empty set axiom'".
This could for instance be accomplished by writing

```md
### Cumulative hierarchies satisfy the empty set axiom
```

fredrik-bakke (approved):
An expertly written PR! :)

I only have a couple of minor comments.

src/set-theory/cumulative-hierarchy.lagda.md
fredrik-bakke: and similarly for the other headers of course.

src/set-theory/cumulative-hierarchy.lagda.md
shiranaiyo: I've done all other suggestions except this one. Could you tell me which the correct way to do this would be? I'm not sure what you meant.

fredrik-bakke: Ah, I read it again now and see that what is missing is actually a parenthesis
```agda
  ( ∈-cumulative-hierarchy
      empty-set-cumulative-hierarchy
      infinity-cumulative-hierarchy) ×
    ( ( x : type-pseudo-cumulative-hierarchy V) →
      ∈-cumulative-hierarchy x infinity-cumulative-hierarchy →
      ∈-cumulative-hierarchy
        ( pair-cumulative-hierarchy x
          ( singleton-cumulative-hierarchy x))
        ( infinity-cumulative-hierarchy))
```

shiranaiyo: you're right! that's clearer, all suggestions are done now

fredrik-bakke: Excellent! I will merge this PR once the checks have passed :)

PR #525, opened 2023-03-19
Refactoring ideals in semirings, rings, commutative semirings, and commutative rings; refactoring a corollary of the binomial theorem; constructing the nilradical of an ideal in a commutative ring

EgbertRijke:


EgbertRijke:
This PR is ready for review

src/commutative-algebra/radical-ideals-commutative-rings.lagda.md
```diff
@@ -0,0 +1,128 @@
...
+open import foundation.propositions
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+An ideal `I` in a commutative ring is said to be **radical** if for every
+element `f : A` such that there exists an `n` such that `fⁿ ∈ I`, we have
```
fredrik-bakke, line 27: Just a thought, but maybe it's nice to mention how this relates to that `I` is closed under taking `n`'th roots when this makes sense?

src/ring-theory/nilpotent-elements-semirings.lagda.md
```diff
@@ -86,123 +86,13 @@ is-nilpotent-add-Semiring R x y H f h =
...
-                      ( sum-zero-Semiring R (succ-ℕ m)))) ∙
-                  ( left-unit-law-add-Semiring R
-                    ( zero-Semiring R)))))))
+            ( ( is-linear-combination-power-add-Semiring R n m x y H) ∙
+              ( ( ap-add-Semiring R
+                  ( ( ap (mul-Semiring' R _) q) ∙
+                    ( left-zero-law-mul-Semiring R _))
+                  ( ( ap (mul-Semiring' R _) p) ∙
+                    ( left-zero-law-mul-Semiring R _))) ∙
+                ( left-unit-law-add-Semiring R (zero-Semiring R))))))
```
fredrik-bakke, line 95: You found a shorter proof! 🥳

src/ring-theory/radical-ideals-rings.lagda.md
```diff
@@ -19,30 +19,30 @@ open import ring-theory.rings
 
 ## Idea
 
-A radical ideal in a ring R is a two-sided ideal I such that `1 + x` is a
-multiplicative unit for every `x ∈ I`.
+A radical ideal in a ring R is an ideal I such that `1 + x` is a multiplicative
+unit for every `x ∈ I`.
```
fredrik-bakke, line 23: Just a curious comment again, but are you considering showing/have you shown any of its equivalents? Are some of the other formulations weaker/stronger in the constructive setting? E.g.
1. The radical ideal is the intersection of all prime ideals containing it
2. A radical ideal is closed under taking radicals
3. I may be misremembering this one, but something like "every element is invertible or nilpotent"

fredrik-bakke (approved):
Looks good to me! I made some comments, but feel free to ignore them.

src/commutative-algebra/radicals-of-ideals-commutative-rings.lagda.md
```diff
@@ -0,0 +1,152 @@
+# Radicals of ideals in commutative rings
```
fredrik-bakke, line 1: Maybe define the radical of a (non-commutative) ring as well?

fredrik-bakke: Could also prove that taking the radical is an idempotent operation.

fredrik-bakke:
Just a thought, but shouldn't "radical ideals" make sense in any monoid? Is this uninteresting?

src/ring-theory/radical-ideals-rings.lagda.md
fredrik-bakke: The first one feels a little choice'y to me...

fredrik-bakke: Please forgive my ignorance if any of what I'm saying is just obvious in some way.

EgbertRijke:
I don't think radical ideals make sense in an arbitrary monoid, because you need multiplication and an additive unit to state it. You can sensibly talk about nilpotent elements in a semiring, which I have already formalized. But in that case you only get that `(x+y)` is nilpotent for two nilpotent elements `x` and `y` if `x` and `y` commute with each other. That brings us to commutative semirings. Here the nilpotents indeed form an ideal, but to my understanding the notion of ideals in a commutative semiring is not ideal (pun intended). What bugs me about ideals in a semiring is that they don't correspond uniquely to congruence relations on semirings. There is an obvious fix: consider a notion of "normal ideal". I intend to formalize this later.

But then the problem is that it is unknown to me whether the nilpotent elements form such a normal ideal. Essentially, the condition that is missing is that if `x` and `x+y` are nilpotent, then `y` should also be nilpotent. But I don't know whether this is true in an arbitrary semiring. On the other hand, we can always normalize an ideal in a semiring, and call the normalization of the ideal of nilpotents the nilradical. 

Here I'm starting to get a bit out of my depth. I am not familiar enough with semigroup theory.

EgbertRijke:
Oh wait, maybe you don't need an additive unit to state what a *radical ideal* is. I was thinking of nilpotent elements when I said that they don't make sense in a monoid.

But ideals in a monoid are submonoids closed under multiplication. Since they're submonoids, they must contain 1. This is unlike ideals in (semi)rings, which are more interesting when the don't contain 1.

src/commutative-algebra/radicals-of-ideals-commutative-rings.lagda.md
EgbertRijke: The radical of a nocommutative ring doesn't make sense, because to be closed under addition you need commutativity.

src/ring-theory/radical-ideals-rings.lagda.md
EgbertRijke: Yes, my student will do this sort of work. I just took over this part of the setup, since it got quite involved and required us to go too deep down into the library. Now that radicals of ideals are defined, she can get started. This is definitely part of our plan

EgbertRijke: 3. Cannot be true, because there are rings with idempotent elements, such as boolean rings.

fredrik-bakke:
So one definition of "radical ideal" is that $1+x$ is invertible for every $x \in R$. However, another definition (I'm unsure about when these agree) is that "if $r\in R$ and $x^n = r$ for some $n$, then $ x \in R$". This should make sense in any monoid, right?

src/ring-theory/nilpotent-elements-semirings.lagda.md
EgbertRijke: Well, I found a shorter proof by moving the long part over to another file (and lengthening it), but yeah:)

src/commutative-algebra/radicals-of-ideals-commutative-rings.lagda.md
fredrik-bakke: Ah, I see.

EgbertRijke:
That's the Jacobson radical. It is the intersection of all maximal ideals, whereas this radical is the intersection of all prime ideals.

src/commutative-algebra/radicals-of-ideals-commutative-rings.lagda.md
fredrik-bakke: I mean, one generalization would be as the intersection of all prime ideals, but I guess this isn't considered "the radical" of a ring anymore.

fredrik-bakke:
Okay okay, I'm out of my depth here. Apologies for the ignorance.

EgbertRijke:
That's ok. I'm also learning at the same time.

src/ring-theory/radical-ideals-rings.lagda.md
EgbertRijke: > The first one feels a little choice'y to me... and the third one a little LEM'y...

Yes, you need choice to show that the nilradical contains the intersection of all prime ideals. The proof is quite nice.

EgbertRijke: With choice you can also show that the locale of radical ideals is equivalent to the locale of opens in the Zariski topology.

PR #527, opened 2023-03-20
Recording OEIS sequences that are formalized in the library

EgbertRijke:
This is a small pull request starting a file in which we can record which OEIS sequences are formalized. I also thought of a nice challenge, that I will perhaps record as an issue: Formalize the OEIS sequence with the smallest index that isn't already formalized.

### DIFF
--- a/src/online-encyclopedia-of-integer-sequences.lagda.md
+++ b/src/online-encyclopedia-of-integer-sequences.lagda.md
@@ -1,0 +1,7 @@
+# Online encyclopedia of integer sequences
+
+```agda
+module online-encyclopedia-of-integer-sequences where
+
+open import online-encyclopedia-of-integer-sequences.oeis public
+```
--- a/src/online-encyclopedia-of-integer-sequences/oeis.lagda.md
+++ b/src/online-encyclopedia-of-integer-sequences/oeis.lagda.md
@@ -11,6 +11,7 @@ open import elementary-number-theory.eulers-totient-function
 open import elementary-number-theory.exponentiation-natural-numbers
 open import elementary-number-theory.factorials
 open import elementary-number-theory.fibonacci-sequence
+open import elementary-number-theory.infinitude-of-primes
 open import elementary-number-theory.natural-numbers
 
 open import finite-group-theory.finite-groups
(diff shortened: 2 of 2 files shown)


src/online-encyclopedia-of-integer-sequences/oeis.lagda.md
```diff
@@ -0,0 +1,51 @@
...
+
+A000244 : ℕ → ℕ
+A000244 = exp-ℕ 3
+
+A000720 : ℕ → ℕ
+A000720 = prime-counting-ℕ
+
+A001477 : ℕ → ℕ
+A001477 = id
+```
```
fredrik-bakke, line 51: Please add a link to the OEIS

src/online-encyclopedia-of-integer-sequences.lagda.md
```diff
@@ -0,0 +1,7 @@
+# Online encyclopedia of integer sequences
+
+```agda
+module online-encyclopedia-of-integer-sequences where
+
+open import online-encyclopedia-of-integer-sequences.oeis public
+```
```
fredrik-bakke, line 7: Should this really have its own namespace? I'm not sure I like that. What about placing it in `elementary-number-theory` for now, or placing `oeis.lagda.md` plainly in `src`?

fredrik-bakke (approved):
This is a cool addition, and I like the idea!
It's a bit underwhelming that currently A000002 is not formalized though, although I don't expect this to be formalized anytime soon.

EgbertRijke:
> This is a cool addition, and I like the idea!
> It's a bit underwhelming that currently A000002 is not formalized though, although I don't expect this to be formalized anytime soon.

Formalizing it should be a nice exercise. It shouldn't be very hard.

src/online-encyclopedia-of-integer-sequences.lagda.md
fredrik-bakke: Alright, I agree it's not best practice to shoehorn this into `elementary-number-theory`. Perhaps individual files can be created in `online-encyclopedia-of-integer-sequences` when defining the number sequence isn't a one-liner and the sequence doesn't fit in another namespace.

fredrik-bakke:
> > This is a cool addition, and I like the idea!
> > It's a bit underwhelming that currently A000002 is not formalized though, although I don't expect this to be formalized anytime soon.
> 
> Formalizing it should be a nice and relaxing exercise. It shouldn't be very hard.

I guess so, for those that are interested. 😄

PR #528, opened 2023-03-21
First definitions towards n-ary functoriality of set quotients

shiranaiyo:
I defined n-ary operations using the functional vector approach, i.e. by indexing with finite types. The problem with this is that then `preserves-sim-n-ary-map-Eq-Rel` has as codomain `UUω` since agda takes a pi type
```agda
( as : (i : Fin n) → As i) → _
```
to live in `UUω`. It seems it doesn't realize that `Fin n` is finite so that `As i` is always bounded. This makes it so that I can't generate the corresponding subtype of n-ary-operations.

I'd like to hear your opinions/suggestions.


EgbertRijke:
Can you call this file `multivariable-functoriality-set-quotients`? Names of files should not have variables in them.

EgbertRijke:
It lands in `UUω` because you assume `Fin n -> Level`

EgbertRijke:
Perhaps working with a finite family of types in one single universe is fine. It is probably the most workable solution. Even more so if the agda-developers will break functionality around universe levels.

shiranaiyo:
I'm following the structure of `foundation.binary-homotopies`. Though I'm having some trouble proving the equivalent of `is-equiv-binary-htpy-eq`. I.e.
```agda
is-equiv-multivariable-htpy-eq :
  { l1 : Level}
  ( n : ℕ)
  ( As : functional-vec (UU l1) n)
  ( X : UU l1)
  ( f g : multivariable-operation n As X) →
  is-equiv (multivariable-htpy-eq n As X f g)
```
Wonder if you have any suggestions to do it in a somewhat direct way.

shiranaiyo:
After struggling for almost all day with "n-ary homotopies" I've just realized that I can just use normal homotopies on an n-dimensional vector :)

shiranaiyo:
Done! Didn't know that was possible.

shiranaiyo:
I finished up to the function I needed (`inv-precomp-set-quotient-vector`), though I didn't prove uniqueness.

As I side note I'm very proud of these few lines, there were many ways to proceed and a lot of them failed. I think this one is probably the fastest.

src/foundation/multivariable-functoriality-set-quotients.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+  inv-precomp-set-quotient-vector n
+    ( tail-functional-vec n As)
+    ( λ i → Rs (inl i))
+    ( function-Set (set-quotient (Rs (inr star))) X)
+    ( pair
+      ( λ as qa →
+        ( curry-once-multivariable-reflecting-map
+          n As Rs X (f , H) qa as))
+      ( λ {bs bs'} K → eq-htpy λ qa →
+          htpy-eq
```
fredrik-bakke, line 174: The `eq-htpy` instantly followed by a `htpy-eq` here looked a bit odd to me, but I couldn't immediately find a solution that omit them.

src/foundation/multivariable-functoriality-set-quotients.lagda.md
```diff
@@ -0,0 +1,186 @@
+# Multivariable functoriality of set quotients
+
+```agda
+{-# OPTIONS --lossy-unification #-}
```
fredrik-bakke, line 4: Is this file slow to compile without this pragma?

fredrik-bakke (approved):

fredrik-bakke:
Although I didn't take the time to fully understand it, I'm glad you're proud of your proof. Keep up the good work!

shiranaiyo:
Actually I've realized I made a mistake, this is not functoriality... I'm putting it on draft again, my bad.

src/foundation/multivariable-functoriality-set-quotients.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+
+open import univalent-combinatorics.standard-finite-types
+```
+
+</details>
+
+## Idea
+
+Say we have a family of types `A1`, ..., `An` each equipped with an equivalence
+relation `Ri`, as well as a type `X` equipped with an equivalence relation `X`,
```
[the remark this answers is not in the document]
shiranaiyo: done

src/foundation/multivariable-functoriality-set-quotients.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+         ( is-symmetric-all-sim-Rel-Prop )
+         ( is-transitive-all-sim-Rel-Prop )))
+
+  set-quotient-vector : UU (l1 ⊔ l2)
+  set-quotient-vector =
+    (i : Fin n) → set-quotient (Rs i)
+
+  is-set-qARs :
+    is-set set-quotient-vector
+  is-set-qARs = is-set-Π (λ i → (is-set-set-quotient (Rs i)))
```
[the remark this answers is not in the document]
shiranaiyo: done

src/foundation/multivariable-functoriality-set-quotients.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+  ( As : functional-vec (UU l1) n)
+  ( Rs : (i : Fin n) → Eq-Rel l2 (As i))
+  where
+
+  all-sim-Rel :
+    ( as a's : multivariable-input n As) →
+    UU l2
+  all-sim-Rel as a's =
+    (i : Fin n) →
+    sim-Eq-Rel (Rs i) (as i) (a's i)
```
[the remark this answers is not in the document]
shiranaiyo: While I agree that is nice, a lot of times when doing induction I also get one single relation `R`. These `s`s help me keep track which is which.

src/foundation/multivariable-functoriality-set-quotients.lagda.md
shiranaiyo: agree it seems weird but I'm swapping the order of arguments, I think funext (or univalence) is definitely needed here

shiranaiyo:
Proved the equivalence $A^n/R^n \cong (A/R)^n$, though not that the left side is an actual set quotient. Uniqueness of functoriality is therefore missing, but I think I won't need that (hopefully). I know the proof goes way too long instead of splitting it up, but all those lemmas are completely useless/uninteresting in my opinion.

src/foundation/cartesian-products-set-quotients.lagda.md
```diff
@@ -0,0 +1,177 @@
...
+## Definition
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  {A : UU l1} (R : Eq-Rel l2 A)
+  {B : UU l3} (S : Eq-Rel l4 B)
+  where
+
+  prod-Eq-Rel :
```
EgbertRijke, line 45: Make separate definitions here:
`sim-prod-Eq-Rel`
`rel-prod-Eq-Rel`
`refl-prod-Eq-Rel`
`symmetric-prod-Eq-Rel`
`transitive-prod-Eq-Rel`

src/foundation/multivariable-functoriality-set-quotients.lagda.md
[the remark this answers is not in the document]
EgbertRijke: > I'd just like to point out that if you had named `Rs` as just `R`, then the `Ri`s in your explanation text would correspond to `R i` in the code, which I personally find more notationally elegant.

I also like it much more without the -s. Perhaps we could convince you to change it?:)

src/foundation/multivariable-functoriality-set-quotients.lagda.md
fredrik-bakke: You may be right. This is funext btw.

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+  inv-map-equivalence-classes :
+    equivalence-class (all-sim-Eq-Rel (succ-ℕ n) As Rs) →
+    equivalence-class-prod-R0-Rs
+  inv-map-equivalence-classes (P , H) =
+    pair
+      ( λ (a , as) → P (cons-multivariable-input' n As a as))
+      ( map-trunc-Prop
+        ( λ (as , p) →
+          pair
+            ( pair (as (inr star)) (λ i → as (inl i)))
```
fredrik-bakke, line 265: action on maps

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+    eq-pair-Σ
+      ( eq-htpy
+        ( λ as →
+          ( eq-pair-Σ
+            ( ap
+              ( pr1 ∘ P)
+              ( eq-htpy
+                ( λ { (inl x) → refl;
+                      (inr star) → refl})))
+            ( eq-is-prop' (is-prop-is-prop _) _ _))))
```
fredrik-bakke, line 299: `( eq-is-prop (is-prop-is-prop _)))))`

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+      ( eq-is-prop' is-prop-type-trunc-Prop  _ _)
```
fredrik-bakke, line 300: `( eq-is-prop is-prop-type-trunc-Prop)`

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+      ≃ (set-quotient (Rs (inr star)) ×
+          ( set-quotient-vector n
+            ( tail-functional-vec n As)
+            ( λ x → Rs (inl x))))
+  lemma =
+    pair
+      ( λ {qas →
+        ( pair
+           ( qas (inr star))
+           ( λ i → qas (inl i)))})
```
fredrik-bakke, line 337: action on maps

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+      ( pair
+        ( pair
+          ( λ { (qa , qas) (inl i) → qas i ;
+                (qa , qas) (inr star) → qa})
```
fredrik-bakke, line 341: It feels to me like this function should have a common name (if not, then please add it). I think it would be better to use that than explicit pattern matching.

I don't know the best terminology here, but this is something like a "cons" right?

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+          ( λ (qa , qas) → eq-pair-Σ refl refl))
+        ( pair
+          ( λ { (qa , qas) (inl i) → qas i ;
+                (qa , qas) (inr star) → qa})
```
fredrik-bakke, line 345: cons

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+                          ( h))})
+              ( λ h →
+                ( backward-implication
+                  ( p
+                    ( pair
+                      ( a's (inr star))
+                      ( tail-multivariable-input n As a's))))
+                ( pair
+                  ( h (inr star))
+                  ( λ i → h (inl i))))))
```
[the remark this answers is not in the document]
fredrik-bakke: Maybe it's not precisely vectors, but an equivalent type to vectors anyways (since it's not a list ending with an empty vecotr). Either your code should use that definition, or you should define a file with your "vector-like" objects.

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+  all-sim-Eq-Rel :
+    ( Eq-Rel l2 (multivariable-input n As))
+  all-sim-Eq-Rel =
+    pair
+     ( all-sim-Rel-Prop)
+     ( pair
+       ( is-reflexive-all-sim-Rel-Prop)
+       ( pair
+         ( is-symmetric-all-sim-Rel-Prop)
+         ( is-transitive-all-sim-Rel-Prop)))
```
fredrik-bakke, line 113: for performance reasons we prefer to pattern match definitions.
I.e.
```agda
pr1 all-sim-Eq-Rel = all-sim-Rel-Prop
pr1 (pr2 all-sim-Eq-Rel) = ...
```

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+  is-prop-all-sim-Rel as a's =
+    is-prop-Π λ i → is-prop-sim-Eq-Rel (Rs i) (as i) (a's i)
+
+  all-sim-Rel-Prop :
+    ( as a's : multivariable-input n As) →
+    Prop l2
+  all-sim-Rel-Prop as a's =
+    pair
+     (all-sim-Rel as a's)
+     (is-prop-all-sim-Rel as a's)
```
fredrik-bakke, line 88: pattern match definitions

fredrik-bakke (commented):
Please read my comments in chronological order 😅

shiranaiyo:
As a side note, I'm no longer proud of this mess haha

shiranaiyo:
Finished the set quotients, this PR is finished. Some comments
- I hope variable names are ok in `where` clauses. Those lemmas really are completely uninteresting, I'm forced to state them because otherwise the typechecker doesn't allow the proof.
- A lot of `ap`s together looks weird, but there's no clean way to merge them

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -59,292 +61,372 @@ of the set quotients of each `Ai`.
...
+      ( all-sim-Eq-Rel n A R)
+      ( set-quotient-vector-Set n A R)
+      ( reflecting-map-quotient-vector-map n A R)
+      ( X)))
+pr1 (isretr-inv-precomp-vector-set-quotient n A R X) =
+  inv-precomp-vector-set-quotient n A R X
+pr2 (isretr-inv-precomp-vector-set-quotient zero-ℕ A R X) f =
+  eq-htpy (λ {(map-raise star) → refl})
+pr2 (isretr-inv-precomp-vector-set-quotient (succ-ℕ n) A R X) f =
+  ap (λ - → - ∘ set-quotient-vector-prod-set-quotient)
```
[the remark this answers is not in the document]
fredrik-bakke: This is a bit of special Agda syntax that I quite like :)

src/foundation-core/equivalence-relations.lagda.md
```diff
@@ -143,3 +144,16 @@ equiv-trans-Eq-Rel' :
 equiv-trans-Eq-Rel' R r =
   equiv-iff' (prop-Eq-Rel R _ _) (prop-Eq-Rel R _ _) (iff-trans-Eq-Rel' R r)
 ```
+
+## Examples
+
+### The trivial equivalence relation that always holds
```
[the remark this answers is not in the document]
fredrik-bakke: Actually, I think a better definition for the discrete equivalence relation is $x\sim y := (x = y)$ but only define it on sets.

src/foundation/vectors-set-quotients.lagda.md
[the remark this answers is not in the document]
shiranaiyo: Oh yeah that's nice!

shiranaiyo:
I've implemented the suggestions now.

src/foundation/vectors-set-quotients.lagda.md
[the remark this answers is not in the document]
EgbertRijke: Also, it is dangerous to use `λ - →`. Note that `-` is not part of the agda-syntax, so `-` is interpreted as a variable!! If you want to avoid naming the variable, use the underscore `_`.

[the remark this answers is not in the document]
fredrik-bakke: That is true, but if he had written
`λ _ → _ ∘ set-quotient-vector-prod-set-quotient`, then this probably would not type check as Agda would be unable to infer the value of the second `_`.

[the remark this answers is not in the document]
EgbertRijke: In that case it is better to use a variable name that looks like a variable. There's no reason to fool the reader.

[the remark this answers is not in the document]
fredrik-bakke: Agreed.

src/foundation/cartesian-products-set-quotients.lagda.md
```diff
@@ -0,0 +1,242 @@
...
+
+  isretr-inv-precomp-set-quotient-prod-set-quotient :
+    { l : Level}
+    ( X : Set l ) →
+    ( ( inv-precomp-set-quotient-prod-set-quotient X) ∘
+      precomp-Set-Quotient
+      prod-Eq-Rel
+      prod-set-quotient-Set
+      reflecting-map-prod-quotient-map X
+      ) ~
```
fredrik-bakke, line 168:
```agda
    ( ( inv-precomp-set-quotient-prod-set-quotient X) ∘
      precomp-Set-Quotient
        ( prod-Eq-Rel)
        ( prod-set-quotient-Set)
        ( reflecting-map-prod-quotient-map X)) ~
```

src/foundation/cartesian-products-set-quotients.lagda.md
```diff
@@ -0,0 +1,242 @@
...
+              (λ b → f (a , b))
+              (λ p → H (pair (refl-Eq-Rel R) p)))
+            qb')
+        ( λ {a1} {a2} p →
+          ( ap (inv-precomp-set-quotient S X)
+            ( eq-pair-Σ
+              ( eq-htpy (λ b → H (pair p (refl-Eq-Rel S))))
+              ( eq-is-prop
+                ( is-prop-reflects-Eq-Rel S X _ ))))))
+      ( qa) ( qb)
```
fredrik-bakke, line 137: add a line break between these two, or remove the parentheses

src/foundation/cartesian-products-set-quotients.lagda.md
```diff
@@ -0,0 +1,242 @@
...
+            ( pair
+              (λ b → f (a , b))
+              (λ p → H (pair (refl-Eq-Rel R) p)))
+            qb')
+        ( λ {a1} {a2} p →
+          ( ap (inv-precomp-set-quotient S X)
+            ( eq-pair-Σ
+              ( eq-htpy (λ b → H (pair p (refl-Eq-Rel S))))
+              ( eq-is-prop
+                ( is-prop-reflects-Eq-Rel S X _ ))))))
```
fredrik-bakke, line 136: the convention is to write `_)`. I see you've written `_ )` in a couple of places.

fredrik-bakke (approved):
After these last comments, I think the PR is ready to be merged, unless @EgbertRijke wants to add anything.

fredrik-bakke:
I have another comment regarding prepositions in file names which I mentioned in a previous PR of yours. I realized after that the library is actually quite inconsistent with regard to this. This is why I haven't mentioned it in this PR.

shiranaiyo:
Implemented the changes.
> I have another comment regarding prepositions in file names which I mentioned in a previous PR of yours. I realized after that the library is actually quite inconsistent with regard to this. This is why I haven't mentioned it in this PR.

Yeah I was thinking about that, since I'm basically generalizing previous results, I tried to just mimic every name for functions/proofs/files related to set quotients.

src/foundation/cartesian-products-set-quotients.lagda.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Idea
+
+Given two types `A` and `B`, equipped with equivalence relations `R` and `S`,
+respectively, the cartesian product of their set quotients is the set quotient
+of their product.
+
+## Definition
+
+### The product of two relations
```
EgbertRijke, line 44: This should probably go into separate files `products-binary-relations` and `products-equivalence-relations`. Not every time one has binary relations over two types we want to refer to quotients, so it is not natural to be looking for products of binary relations in a file about quotients.

EgbertRijke:
> After these last comments, I think the PR is ready to be merged, unless @EgbertRijke wants to add anything.

Sorry to spoil the party. I saw another thing that I wanted to comment on:)

shiranaiyo:
That's a good point! I've separated them.

src/foundation/cartesian-products-set-quotients.lagda.md
fredrik-bakke: Doesn't look to me like you did the parentheses and indentation correctly on this one. You could also do
```agda
    ( ( inv-precomp-set-quotient-prod-set-quotient X) ∘
      ( precomp-Set-Quotient
        ( prod-Eq-Rel)
        ( prod-set-quotient-Set)
        ( reflecting-map-prod-quotient-map X))) ~
```
if you prefer.

fredrik-bakke: No, wait, this wasn't right. The X must be in its own parenthesis. I.e.
```agda
    ( ( inv-precomp-set-quotient-prod-set-quotient X) ∘
      ( precomp-Set-Quotient
        ( prod-Eq-Rel)
        ( prod-set-quotient-Set)
        ( reflecting-map-prod-quotient-map)
        ( X))) ~
```
Sorry about that.

PR #529, opened 2023-03-21
Some additions to and refactoring large types

fredrik-bakke:
I was having a look around the library with regard to large types and started making a few additions and doing a bit of refactoring. Ultimately, I decided against going further, but here is some of what I did before stopping.

In particular, I may note that the definition of large dependent pair types should allow us to avoid using records for large precategories and so on.


EgbertRijke:
I will look at this PR tomorrow. I'm not sure if I like the idea of changing the records to Sigma-types.

EgbertRijke:
I assigned this PR to myself. Please don't merge it before we agree on what is going on in it.

fredrik-bakke:
I'll revert the changes regarding record types. I think there is some other refactoring in this PR I still would like to get merged, however. Do you want large dependent pair types and so on completely removed?

PR #530, opened 2023-03-21
Formatting fixes

fredrik-bakke:
Just a bunch of miscellaneous formatting.

- Fix capitalization of module titles
- Make comment headers into actual headers
- Remove some superfluous whitespace
- Organize VSCode settings
- Add VSCode settings that nest ".agdai" and ".agda~" files under the associated ".lagda.md" file to declutter the file view.
- Fix a breaking bug with `generate_main_index` when there are duplicate titles
- Add `<details>` blocks to all files (except `foundation-core.universe-levels`)
- Remove unused imports


fredrik-bakke:
I'll spare you from having to go through these changes, but I would like to point out one issue that remains.

There is a big bunch of references to theorems and exercises in the library without referencing any book. I am okay with having a section header state

```md
### Theorem 12.3
```

But then the source should be linked. Therefore, I've left these headers like this as comments for now.

fredrik-bakke:
I will merge this PR once the check has finished.

PR #526, opened 2023-03-20
Improvements to file conventions and other documentation

fredrik-bakke:
- Reformulates file conventions
- Adds a template agda file
- misc. documentation improvements and small fixes

PR #521, opened 2023-03-18
Refactorization for the delooping of the sign homomorphism

EleonoreMangel:

PR #537, opened 2023-03-22
Improvements to file conventions and other documentation

fredrik-bakke:
This is a rerun of #526.

- Reformulates file conventions
- Adds a template agda file
- misc. documentation improvements and small fixes

Currently, there seems to be an issue with the link to `template.lagda.md`.


jonaprieto (approved):

EgbertRijke:
What was the issue with the original #526?

EgbertRijke:
If this happens again in the future, wouldn't it be better just to make a PR fixing the broken link, instead of reverting an entire PR? For example, I did this in #539. I don't think it would have been right to revert #521 just for a broken link.

fredrik-bakke:
I agree that fixing a broken link with a separate PR is the better approach in principle. However, in this specific case, a few reasons led me to revert the PR instead:

1. I was informed that Jonathan had not wanted the original PR merged at that time.
2. I wasn't sure how to fix the broken link, which is why I asked Jonathan for assistance in the first place.
3. It wasn't clear to me if the broken link was the only issue with the original PR.
4. Neither Jonathan nor I had the time to address the issue at the time.

As for reverting a PR in general, I don't see a significant issue with it, except for the potential downside of polluting the commit history.

EgbertRijke:
The main issue is that reverting a PR undoes all the good work of that PR. We shouldn't do this lightly.

PR #538, opened 2023-03-23
Kolakoski sequence

EgbertRijke:
This PR adds A000002 to the library. The proof that it is correct is left as an exercise to the reader.


fredrik-bakke:
Also, I'd just like to point out that my original opinion wasn't that it wasn't in principle "easy" to formalize this sequence, just that I didn't quite see the point other than recreation and learning.

EgbertRijke:
> Another one down!
> 
> I'd just like to ask one question as I'm still a bit confused about the `online-encyclopedia-of-integer-sequences` namespace. Do you consider the Kolakoski sequence elementary enough to be placed in `elementary-number-theory` instead of this namespace, or is there another reason?

I'm not sure what the best location is for the kolakoski sequence. This sequence gives more CS vibes than elementary number theory vibes. My reason for putting it in elementary number theory was just that I wanted to keep the OEIS file free of definitions. My idea was that it is just there to record what sequences have been formalized somewhere in the library.

PR #540, opened 2023-03-23
Commutativity and unit laws for cauchy products and dirichlet products of species

VictorBlanchi:


src/species/dirichlet-product-species-of-subuniverse.lagda.md
```diff
@@ -212,3 +214,133 @@ module _
...
+  (C2 : is-in-subuniverse P (raise-unit l1))
+  (C3 :
+    (X : type-subuniverse P) →
+    is-in-subuniverse
+      ( subuniverse-global-subuniverse Q l1)
+      ( is-contr (inclusion-subuniverse P X)))
+  (S : species-subuniverse P ( subuniverse-global-subuniverse Q l2))
+  where
+
+  unit-dirichlet-product-species-subuniverse :
```
EgbertRijke, line 284: This definition is not using all the parameters of the module. It is better to only assume what is used, so that whan you apply this definition, you don't have to provide unnecessary arguments. It's therefore better to take this definition out of the module.

PR #545, opened 2023-03-27
Change integral domains definition

shiranaiyo:
Adds the nonzero condition to integral domains.


EgbertRijke:
Excellent PR. Note that there was already a file `nontrivial-rings`, which I think you missed. 

I was tempted to say that the name `is-zero-commutative-ring-Commutative-Ring` can be simplified to `is-zero-Commutative-Ring`, but then I realized that the latter is already defined to be the assertion that an element is zero. Perhaps we should speak of trivial (commutative) rings and nontrivial (commutative) rings, to avoid name clashes. But then the file names should be changed accordingly. What do you think of this idea?

shiranaiyo:
Yeah I missed that haha, I will search with more detail in the future. I'm using `(non)trivial-X` as suggested now, grouping them both in a `trivial-X`.lagda.md` file.

PR #550, opened 2023-04-06
Rational addition and format for future work on rationals

shiranaiyo:
Refactored some stuff, importantly separating integer fractions from the rationals. My idea is that all future operations on rationals be defined on integer fractions first, and then carried to rationals; as working with the gcd directly is horrible. I did the first part of the additions to show how it'd look, I'd use this schema for the other operations and properties too.

I'd like confirmation that this approach is fine, though it duplicates the number of files.

### DIFF
--- a/src/elementary-number-theory.lagda.md
+++ b/src/elementary-number-theory.lagda.md
@@ -1,9 +1,10 @@
-# Elementary Number Theory
+# Elementary number theory
 
 ```agda
 module elementary-number-theory where
 
 open import elementary-number-theory.absolute-value-integers public
+open import elementary-number-theory.addition-integer-fractions public
 open import elementary-number-theory.addition-integers public
 open import elementary-number-theory.addition-natural-numbers public
 open import elementary-number-theory.addition-rationals public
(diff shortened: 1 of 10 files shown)


EgbertRijke:
These changes look good to me, overall.

In #546 I'm also refactoring many files of elementary number theory, including some files you have been working on here. Therefore I would like to merge this as soon as possible, so that I can resolve any conflicts.

shiranaiyo:
Great! I think I may take some time to finish it though. I'd guess you won't touch the rationals so I think this could be merged to resolve any conflicts now. My changes from now on will touch only fractions and rationals, most likely. I'm marking it as ready for review then.

shiranaiyo:
I realized real numbers could be define without further theory, so I did it, following HoTT book's dedekind real numbers.

I will rest for a week so you can merge now to avoid conflicts with https://github.com/UniMath/agda-unimath/pull/546, if you want.

EgbertRijke:
That's very nice! Indeed, we were planning on following this route with the Dedekind real numbers, because it is the most straightforward in my opinion.

PR #553, opened 2023-04-08
parity of naturals

ElifUskuplu:
I've just added a minor lemma about parity of natural numbers which is useful for computation. If `x` is even/odd, then `x + 1` is odd/even.

### DIFF
--- a/src/elementary-number-theory/parity-natural-numbers.lagda.md
+++ b/src/elementary-number-theory/parity-natural-numbers.lagda.md
@@ -87,7 +87,7 @@ is-odd-succ-succ-is-odd-ℕ :
 is-odd-succ-succ-is-odd-ℕ n = map-neg (is-even-is-even-succ-succ-ℕ n)
 ```
 
-### If a natural number `x` is odd, then `x + 1` is even. Similarly, if `x` is even, then `x + 1` is odd.
+### If a natural number `x` is odd, then `x + 1` is even
 
 ```agda
 is-even-succ-is-odd-ℕ :
@@ -98,7 +98,11 @@ is-even-succ-is-odd-ℕ (succ-ℕ (succ-ℕ n)) p =
   is-even-succ-succ-is-even-ℕ
     (succ-ℕ n)
     (is-even-succ-is-odd-ℕ n (is-odd-is-odd-succ-succ-ℕ n p))
+```
 
+### If a natural number `x` is even, then `x + 1` is odd
+
+```agda
 is-odd-succ-is-even-ℕ :
   (n : ℕ) → is-even-ℕ n → is-odd-ℕ (succ-ℕ n)
 is-odd-succ-is-even-ℕ zero-ℕ p = is-odd-one-ℕ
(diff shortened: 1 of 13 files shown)


src/elementary-number-theory/parity-natural-numbers.lagda.md
```diff
@@ -85,3 +86,26 @@ is-odd-succ-succ-is-odd-ℕ :
   (n : ℕ) → is-odd-ℕ n → is-odd-ℕ (succ-ℕ (succ-ℕ n))
 is-odd-succ-succ-is-odd-ℕ n = map-neg (is-even-is-even-succ-succ-ℕ n)
 ```
+
+### If a natural number `x` is odd, then `x + 1` is even. Similarly,
+### if `x` is even, then `x + 1` is odd.
```
EgbertRijke: We should have only one section title per code block. Perhaps it is best to split the code block, one showing

```
### If a natural number `x` is odd, then `x + 1` is even
```

and the other showing

```
### If `x` is even, then `x + 1` is odd
```

Note also that section titles shouldn't have punctuation at the end.

ElifUskuplu: Got it. I changed as you suggested.

src/elementary-number-theory/parity-natural-numbers.lagda.md
```diff
@@ -85,3 +86,25 @@ is-odd-succ-succ-is-odd-ℕ :
...
+
+```agda
+is-even-succ-is-odd-ℕ :
+  (n : ℕ) → is-odd-ℕ n → is-even-ℕ (succ-ℕ n)
+is-even-succ-is-odd-ℕ zero-ℕ p = ex-falso (p is-even-zero-ℕ)
+is-even-succ-is-odd-ℕ (succ-ℕ zero-ℕ) p = (1 , refl)
+is-even-succ-is-odd-ℕ (succ-ℕ (succ-ℕ n)) p =
+  is-even-succ-succ-is-even-ℕ
+    (succ-ℕ n)
+    (is-even-succ-is-odd-ℕ n (is-odd-is-odd-succ-succ-ℕ n p))
```
EgbertRijke, line 100: One other minor remark. Our convention is to have a space after opening parentheses on a new line. When this is done consistently, the code clearly displays the syntaxtree-structure of a definition (note that all definitions in agda form syntax trees). We like to display this syntax tree clearly, because it helps reading more complicated code.

In your case here, the properly formatted code would look as follows:

```
is-even-succ-is-odd-ℕ :
   (n : ℕ) → is-odd-ℕ n → is-even-ℕ (succ-ℕ n)
 is-even-succ-is-odd-ℕ zero-ℕ p = ex-falso (p is-even-zero-ℕ)
 is-even-succ-is-odd-ℕ (succ-ℕ zero-ℕ) p = (1 , refl)
 is-even-succ-is-odd-ℕ (succ-ℕ (succ-ℕ n)) p =
   is-even-succ-succ-is-even-ℕ
     ( succ-ℕ n)
     ( is-even-succ-is-odd-ℕ n (is-odd-is-odd-succ-succ-ℕ n p))
```

Similarly for your definition below (and also your code on baire spaces, but in that case I didn't comment on this point).

ElifUskuplu: Got it, sorry for bothering you with these details. I should have read the instruction about the conventions more carefully.

PR #560, opened 2023-04-11
Some countability results and auxilary things

ElifUskuplu:
Basically, I've proved that countability is preserved under product and coproduct, and `Z` is countable. For this aim, I've done the followings:

* An equivalence between `N + N` and `N`
* An equivalence between `N x N` and `N` which is called pairing map. I thought this deserves a particular file because it is about the fact that any natural number `n` is of the form `(2^u(2v+1)) - 1` for a unique pair `(u , v)`
* An equivalence between `Z` and `N`
* Functoriality of product and coproduct preserve surjectivity
* `exp m n` is always nonzero if m is nonzero
* If `d` is nonzero and `d` divides `m`, then `(m/d)<=m`
* `x` is even iff `x+1` is odd, `x` is odd iff `x+1` is even
* `x` is odd iff there is `y` such that `2y+1=x`

Let me know what you think. I've tried to follow the conventions, but I'm not sure about some of the term names.


divisibility-natural-numbers.lagda.md
```diff
@@ -0,0 +1,488 @@
+# Divisibility of natural numbers
```
EgbertRijke, line 1: It looks like this file got copied directly into `src` by accident.

exponentiation-natural-numbers.lagda.md
```diff
@@ -0,0 +1,80 @@
+# Exponentiation of natural numbers
```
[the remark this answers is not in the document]
ElifUskuplu: Deleted.

EgbertRijke:
I still have to properly review this PR, but my initial thought is that the result that every nonzero natural number `n` can uniquely be written as a product of a power of two and an odd number is should perhaps have its own file, with naming that doesn't refer to the pairing function. My reason for thinking this is that someone might need this result without needing the pairing function, and they wouldn't think to look for it in a file called `pairing-function`. Perhaps we should have a separate file called `powers-of-two` where such result would be more easily findable.

src/foundation/type-arithmetic-natural-numbers.lagda.md
```diff
@@ -0,0 +1,173 @@
+# Type arithmetic with natural numbers
```
EgbertRijke, line 1: It's probably best if this file is moved to `elementary-number-theory`, where the other stuff about the natural numbers is.

ElifUskuplu: Got it, also I think it is better to add the equivalences among `N x [n]`, `N + [n]`, and `N` for any standard finite type `[n]`. I'll do it soon.

src/elementary-number-theory/parity-natural-numbers.lagda.md
```diff
@@ -112,3 +113,47 @@ is-odd-succ-is-even-ℕ (succ-ℕ (succ-ℕ n)) p =
...
+```
+
+### A natural number `x` is odd if and only if there is a natural number `y` such that `succ-ℕ (mul-ℕ y 2) ＝ x`
+
+```agda
+has-odd-expansion : ℕ → UU lzero
+has-odd-expansion x = Σ ℕ (λ y → (succ-ℕ (mul-ℕ y 2)) ＝ x)
+
+is-odd-has-odd-expansion : (n : ℕ) → has-odd-expansion n → is-odd-ℕ n
+is-odd-has-odd-expansion n (m , p) = tr is-odd-ℕ p s
```
EgbertRijke, line 146: If you use induction on `p` here then the definition can just be `is-odd-succ-is-even-ℕ (mul-ℕ m 2) (m , refl)` and you won't need the `where` clause.

ElifUskuplu:
@EgbertRijke 
Do you have any other comments? If no, then I can edit and finalize.

EgbertRijke:
Are you sure that you will be able to prove that R is uncountable?

Andrej Bauer constructed a topos in which the dedekind reals are countable, so I don't expect us to be able to formalize any proof that R is uncountable unless you will assume some choice such as countable choice.

ElifUskuplu:
> Are you sure that you will be able to prove that R is uncountable?
> 
> Andrej Bauer constructed a topos in which the dedekind reals are countable, so I don't expect us to be able to formalize any proof that R is uncountable unless you will assume some choice such as countable choice.

Yes, I know his work about this. I meant the latter as you guessed. I should have explained better. But of course, it will be a trying, I'm not sure whether I succeed or not. In any way, it is reasonable to add a proof of `R` is a set.

EgbertRijke:
BTW it would be best if you would do the new work on a new branch so that we can finish this pull request independently, but I see that you are working from your master branch.

For the future, the best workflow is this:
- Keep your `master` branch clean and in sync with the master branch of the repository. Never make any changes in your master branch.
- When you start a new line of work, create a branch from your master branch. In this case it would make sense to have a branch called `countable-sets` and for your new work a branch called `real-numbers`
- If you have a new idea that is independent of the line of work in the previous point, create a new branch for your new idea. This way you will be able to work on it independently from the other changes you were making
- After you made a pull request, it is likely that you will want to continue working on something while we review your PR. As in this case, it takes us some time to do this. So if you start working on a new idea in the same branch, it creates the problem that we won't be able to finish with the PR until your work on the new idea is finished. That's why it is best to work with branches in git.

Anyway, since you are on your master branch currently it is a bit complicated to start with this workflow. So I will mark this PR as a draft until you bring the PR to a point where we can review and merge it.

Sorry for the long message. Your work is very good. Keep it up!

ElifUskuplu:
I understand. I didn't know these details. For a fresh start, I can fork the repository again and make my master branch in sync with the master branch. Then I can create such branches as you suggested. Since I'm working on my computer to write codes, I will not lose any previous work. Does it make sense?

ElifUskuplu:
> @EgbertRijke I understand. I didn't know these details. For a fresh start, I can fork the repository again and make my master branch in sync with the master branch. Then I can create such branches as you suggested. Since I'm working on my computer to write codes, I will not lose any previous work. Does it make sense?

I think there is no harm to do this. There are already many unnecessary commits in this PR. I'll do it properly as you suggested and create new PRs.

PR #562, opened 2023-04-12
New Agda syntax highlighting extension for VSCode

fredrik-bakke:
I've written an improved Agda syntax-highlighting extension for VSCode called _agda-syntax_ ([GitHub](https://github.com/fredrik-bakke/agda-syntax-vscode), [VSCode Marketplace](https://marketplace.visualstudio.com/items?itemName=FredrikBakke.agda-syntax)). Although it is still in preview, my opinion is that it is already a significant improvement over the previously used extension. Therefore, I propose that we migrate our development environment (for VSCode users) to use this new extension.

### Some highlights of the extension
Compared to the previously used extension, this new extension
- injects into markdown syntax, so that the markdown code can be highlighted as markdown code as well
- highlights all variable declarations (with some bugs still), module names, wildcard symbols, all reserved keywords (and only recognizes reserved keywords as reserved keywords)
- Recognizes the appropriate token-boundaries
- Highlights line comments properly

Please understand that the grammar framework that has to be used to write the extension is highly limited, so not all highlighting functionality can be implemented. For instance, the parsing must be done in a single pass, and the functionality to match over multiple lines is very limited. Hence, for example, matching the left-hand side of an equals sign is very gnarly (although I have one idea left to try with regard to this).

Still, I would greatly appreciate any feedback, either if it is a bug or a feature request, which is another reason why I want to introduce it into our defined development environment at this point.

If you want to try out the extension right now, follow the VSCode Marketplace link: https://marketplace.visualstudio.com/items?itemName=FredrikBakke.agda-syntax


EgbertRijke:
Cool! Can you also include some instructions on how to obtain this VSCode extension?

fredrik-bakke:
> Cool! Can you also include some instructions on how to obtain this VSCode extension?

The added line to `.vscode/extensions.json` will make VSCode notify the user that they should install the extension when they open their IDE :)

EgbertRijke:
Fancy:)

fredrik-bakke:
Hey!

I was hoping to get Jonathan's feedback before merging this PR, hence the review request. I fear these changes, if not done properly, can be more frustrating than helpful to contributors. Although now I guess we'll know sooner! 😅

jonaprieto:
One thing, I just opened a new instance of vscode in a clean clone of the library
and I lost the feature go-to-definition, is this due to the new setup? I'm running with:
agda-mode v0.3.11 and agda-syntax v0.2.3, vscode 1.75.0.

image

Also, related to the extension, it looks like we lose some colours, isn't it? everything seems to be black or blue now.

image

fredrik-bakke:
> One thing, I just opened a new instance of vscode in a clean clone of the library
> and I lost the feature go-to-definition, is this due to the new setup? I'm running with:
> agda-mode v0.3.11 and agda-syntax v0.2.3, vscode 1.75.0.
> 
> image
> 
> Also, related to the extension, it looks like we lose some colours, isn't it? everything seems to be black or blue now.
> 
> image
> 

Thanks for the feedback!

The go-to-definition feature has never worked for me, so I'm curious to know how you've ever gotten it to work. As for the colors, this is highly theme dependent. Do you still get only 2 colors if you change theme?

fredrik-bakke:
It seems in your example that vscode is confused about where the boundary of the token should be. It seems to be trying to access `Preorder` when you in reality wanted to access `element-Preorder`.

PR #564, opened 2023-04-12
Fixes #534

### LINKED ISSUE #534: Add mdbook build to check-docs-gen
As the title says. This would prevent introducing of broken links, for example.

jonaprieto:
Remove the check-docs job as we no longer use pandoc. Instead, we use the `pages.yml` workflow to check the build for the mdbook. I also included another workflow to clean the cache.


jonaprieto:
I need to merge this to add the check for the next PRs.

PR #571, opened 2023-04-19
Reflection and solvers

FernandoChu:
# Solvers!

Add the necessary infrastructure to use solvers and add one for precategories (see the examples at the end of `src/category-theory/precategory-solver.lagda.md`). There are some important design decisions I'd like to highlight (and recieve comments on):

- Primitive types are imported into `foundation`. As discussed with Fredrik, maybe they could be in their own `primitive` folder. Though I'd suggest universe levels to be there too if that would be the case.
- Since the reflection mechanism is somewhat esoteric, I've named all the primitives just as in `Agda.Builtin`, so that it can be easierly maintained (i.e. by just comparing with `Agda.Builtin`). This however breaks the rule of unique identifiers, e.g. multiple constructors are called `lit` in `src/reflection/terms.lagda.md`. Since all the modules in `reflection` should only be used for solvers, this shouldn't cause any problems, though.
- I've been using private blocks to add list syntax as otherwise it just becomes unmanageable. This won't pollute the namespace though, as they're private.

## Additional comments

- I haven't commented on what some primitives do because either it's obvious from the name, they are very niche and probably won't be used, or I don't know what they do.
- There is commented code [here](https://github.com/UniMath/agda-unimath/pull/571/files#diff-cc5e1b4d5733084679a96b922f92f351a4fffb3b348c440701b19786a62e45a4R161), which will be usable in Agda 2.6.4., but not on 2.6.3. So that I (and others) don't forget to update it, I've left them there.
- ~~The precategories solver is not as general as it could be, it won't work on types that compute to equations using composition and identities. What I'd suggest in those cases is using equational reasoning to state the equality using only composition and identities and only then apply the solver. (Interestingly, this is a consequence of using Sigma types instead of records, although it can be worked around).~~ It now works with modulo beta terms! (E.g. should work with categories too)


fredrik-bakke:
Actually, looking a little further into this, I'm guessing you've copied the definitions from `Agda.Builtin`, correct?

If this is the case, why not just use those with
```agda
open import Agda.Builtin.Float public
```
etc.

shiranaiyo:
Sorry for the delay, my pc died (and I lost my non-pushed changes rip).

I don't import them as some of them have their primitive definitions done with the builtin bool/nats/Id; which would conflict with the `BUILTIN` pragmas that agda-unimath already uses for their defined types.

shiranaiyo:
On the topic of where they should be, I don't think that `reflection` should be their home, as, say, strings have use outside of reflection, I was thinking of either `foundation` or a new `primitive` folder.

fredrik-bakke:
I like the idea of having a `primitive` namespace. Also, one solution to the other issue is changing the definition of booleans but with renaming.

shiranaiyo:
I personally actually like not importing anything from `builtin`, to make the library as self-contained as possible. Currently the only thing from builtin we import is the [universe levels](https://unimath.github.io/agda-unimath/foundation-core.universe-levels.html). If we go with a `primitive` folder then that could be rewritten to directly define them there, as well as `string`s, `char`s, etc. This is the approach of [1lab](https://1lab.dev/Prim.Type.html#primitives-sorts), for example (though not of the stdlib).

shiranaiyo:
I'm done! Please see the first post for details on this PR :)

shiranaiyo:
The pre-commit is failing despide havin run the command, I think it's because the link is too long? I'm not sure, I'd appreciate help with that.

EgbertRijke:
I agree, we need to give some thought about where to put things. I also feel like foundation is not the right place for many of these types. Note that not even the natural numbers or lists are in foundation.

At the moment my best suggestion would be to put them all in the folder `reflection` and rename this folder to `syntax` or `meta-theory`, or perhaps someone has some other suggestion.

src/foundation/unit-type.lagda.md
```diff
@@ -34,8 +34,10 @@ The unit type is inductively generated by one point.
 ### The unit type
 
 ```agda
-data unit : UU lzero where
-  star : unit
+record unit : UU lzero where
+  instance constructor star
```
fredrik-bakke, line 38: I just want to remark that this introduces eta-equality for the unit type. I wonder, why didn't we have this before?

fredrik-bakke (commented):
I'm out of my depth to review most of this. But I will say I think the code looks well-structured.

fredrik-bakke:
> I agree, we need to give some thought about where to put things. I also feel like foundation is not the right place for many of these types. Note that not even the natural numbers or lists are in foundation.
> 
> At the moment my best suggestion would be to put them all in the folder `reflection` and rename this folder to `syntax`, or perhaps someone has some other suggestion.

Personally, I think I prefer Fernando's suggestion to have `reflection` and `primitive` namespaces. Again though, I'm not very familiar with this domain.

src/category-theory/precategory-solver.lagda.md
```diff
@@ -0,0 +1,248 @@
...
+  solve-Precat! cat hole = do
+    goal <- inferType hole
+    (lhs , rhs) <- boundary-TCM goal
+    built-lhs ← (returnTC ∘ build-Precat-Expr) lhs
+    built-rhs ← (returnTC ∘ build-Precat-Expr) rhs
+    unify hole (apply-solve-Precat-Expr cat built-lhs built-rhs)
+```
+
+## Examples
+```
```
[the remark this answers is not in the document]
shiranaiyo: corrected :)

src/foundation/unit-type.lagda.md
shiranaiyo: indeed haha

src/category-theory/precategory-solver.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Doesn't look like it to me? My initial comment was ill-formatted, sorry about that.

shiranaiyo:
> I agree, we need to give some thought about where to put things. I also feel like foundation is not the right place for many of these types. Note that not even the natural numbers or lists are in foundation.
> 
> At the moment my best suggestion would be to put them all in the folder `reflection` and rename this folder to `syntax`, or perhaps someone has some other suggestion.

There's nothing inherent about, say, `strings` that links it to reflection, syntax, or meta-properties though, it's just a builtin type. One could prove things about it and manipulate it like any other type. I think that the `reflection` and `primitive` namespaces approach is better.

fredrik-bakke:
Oh, sorry, I'm not sure about the link thing. I can have a look at it tomorrow evening when I'm at my desk, though 😊

shiranaiyo:
thanks!

src/foundation/unit-type.lagda.md
EgbertRijke: Probably we didn't have it because my book introduces the unit type as an inductive type without eta equality, and because we never needed eta for the unit type. But I'm fine with having eta for the unit type.

EgbertRijke:
I do think strings are about syntax, and the other builtin types that you introduced in this PR too. Note that they are postulated types. If I understand the code correctly, they're not even inductive or coinductive, so I wouldn't be sure how to prove things about these strings.

One thing that bothers me a bit is that we used to be able to be very clear about what is postulated in the library, and it bothers me a bit that with this PR this is out of the window. The current postulates are listed [here](https://unimath.github.io/agda-unimath/DESIGN-PRINCIPLES.html), and these postulates are carefully chosen because the semantics of them is known. We shouldn't be adding random new postulates just because we want extra functionality, because it changes the semantics of the theory in the library.

fredrik-bakke:
This is an interesting point. Is it even possible to do tactics without these primitives?

shiranaiyo:
But we can prove things about `strings` using the primitive functions it comes with, see [the stdlib](https://agda.github.io/agda-stdlib/Data.String.Properties.html) for example. For an easier example, machine words come directly with a injectivity primitive, from which we can prove more stuff.
```agda
  primWord64ToNatInjective :
    ∀ a b → primWord64ToNat a ＝ primWord64ToNat b → a ＝ b
```

About the second point, I agree changing semantics is very important, however I think none of these new postulates/primitives change it. More precisely, no new statements involving the already admitted terms can be proved. The only thing new is the ease with which we can write terms.

My informal argument is as follows:
1. The new primitive functions have as codomain these new types, so no new term of the previous terms can be created
2. No (non-refl) equivalence can be made with these new types so there's no interaction with univalance or HITs
3. The reflection mechanism only modify the typechecker conservatively

And that's it, I think that covers all the possible semantical changes.

shiranaiyo:
> This is an interesting point. Is it even possible to do tactics without these primitives?

Nope, all of them are needed. They aren't even doing anything fancy though, they are just there so we can reflect agda's syntax.
E.g. so that I can reflect the term `ℕ` into an element `def (quote ℕ) nil` of type `Term`.

I think that some of the push back is because it may seem somewhat complicated but it's not, all of them are there (among other things), to just allow this reflection process of `written agda term` to `Term`.

src/category-theory/precategory-solver.lagda.md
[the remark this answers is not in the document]
shiranaiyo: ohhh my bad, thanks! I thought it was about the lack of a line break, that also fixed the pre-commit :)

shiranaiyo:
A much better argument for consistency provided by @plt-amy (if I understood it correctly) is that since we have types that have definable functions of the same type as the primitives, then all of them are safe. E.g.
> a string is just a list of characters, a character is just an element of Fin 1114112, etc

Only the TCM is different but again the codomain of all the primitives is the TCM, so it's essentially aislated from the rest of the theory.

EgbertRijke:
I am not worried about consistency. I do believe that these postulates don't change the consistency.

The point is that with the current set of postulates, we can claim that all things proven in agda-unimath are true in all Grothendieck higher toposes. This is not known for cubical type theory, and this also breaks when we add new postulates to the library. Probably the postulates that are added here are fine, but again, "probably" is also kind of iffy. It would actually need to be checked.

EgbertRijke:
Also I'm not sure that the argument that these new types are isolated from the rest of the type theory. If I understand the premise correctly, you will want to use these new types to construct identifications in the more traditional types. But then the new type cannot be "isolated". What if they allow for an identification to be constructed that wouldn't hold in every higher topos?

shiranaiyo:
To the first comment, I think Amy's argument covers it. We can interpret the postulated `Char` to be whatever models `Fin 1114112`, and so on; so even if they were not isolated it is no problem.

The only exception is the [TCM](https://github.com/UniMath/agda-unimath/pull/571/files#diff-cc5e1b4d5733084679a96b922f92f351a4fffb3b348c440701b19786a62e45a4R52). But this type is (and should be) completely isolated. This type is used to construct identifications indeed, but it does so meta-theoretically; not by interacting with other terms but interacting with agda's typechecker directly. It allows us to manipulate the syntax from inside agda, one could (in principle) move the TCM outside of agda to opperate completely on the emacs level (though not that I'd suggest that). For a concrete example:
```agda
    _ :
      {x : obj-Precat C} →
      id-hom-Precat C {x} ＝ id-hom-Precat C {x}
    _ = solve-Precat! C
```
This typechecks, but `solve-Precat! C` "evaluates" to a concrete term not involving the TCM, `solve-Precat-Expr C ? ? refl`. The TCM essentially lets me aid the typechecker to tell what terms should be in the holes. Hence, why I say `solve-Precat! C` "evaluates", it's not evaluating in the sense of eta or beta reduction, but some meta-evaluation at the typechecking level.

If you'd like, I think I can explain myself better in a call.

plt-amy:
> This is an interesting point. Is it even possible to do tactics without these primitives?

I am going to be annoying about Agda, if you'll excuse me: It's totally possible to do them without the _primitives_, but not without the postulated types. `String`, `Word64`, etc., all need to be postulated, and bound to their corresponding builtin thing, for the reflection primitives to be well-typed.

> The point is that with the current set of postulates, we can claim that all things proven in agda-unimath are true in all Grothendieck higher toposes.

This will not change. Reflection is best understood as an extension of text editing: you do not (in fact, you _can_ not) prove anything non-trivial about TCM.

The other primitives --- e.g. Word64, QName, etc --- _can_ be reasoned about, using the included conversion functions into inductive types, and the proofs that these are injective. But, more likely than not, they won't. If you are not doing program verification, these types only exist to reflect Agda syntax, and were it not for those aforementioned proofs of injectivity, it would be fine for you to place them in some silly universe like SSetω₁₂₃ to indicate their meta-ness.

> If I understand the premise correctly, you will want to use these new types to construct identifications in the more traditional types.

That is not correct. 

Strictly speaking, macros are not constructions; they don't even have values. A macro is executed _only_ for its side effects, and it can only "produce a value" by solving a meta. Of course, the meta solution must also be well-typed.

---

NbE solvers, like the one @shiranaiyo is introducing here, are very cleanly split into type theory and... text editing (reflection). Let's focus on solving identity + associativity in categories.

This is all internal: Given a category, you can construct a type of "trees of morphisms" (with constructors for arbitrary maps, identity, and composition), together with a type of normal forms — it could be lists of morphisms, or it could be something Yoneda-y. Doesn't matter.
In either case, you have functions `embed : Expr a b → Hom a b`, `normal : Expr a b → Nf a b`, and `solve : (f g : Expr a b) → normal f ≡ normal g → embed f ≡ embed g`.
For the solver to be useful, these should be set up so that `normal f ≡ normal g` is _definitionally_ equal for `Expr`s which `embed` to the same thing.

You could stop here, and it would be fine, it would just be useless. Writing `Expr`s by hand is annoying, but not impossible, and in all possible identity+associativity problems, the path argument to `solve` will be `refl`. But the effort put into writing those Exprs by hand is the same as, if not less than, just equating the maps yourself.
To make this useful, we wrap it in reflection. Now at the meta level, we're allowed to pattern-match _on the goal type_ to find the endpoints, compute `Term`s representing the corresponding `Expr`s, and generate the application `solve ... ... refl`, which Agda will type-check as if you had written it.

There are actual semantic concerns to be had about the reflection interface, like `declarePostulate`, but... this ain't one of them. Asking about what the models of `Word64`&co are, at this level, is essentially equivalent to asking about the models of copy+paste.

Oh, and for what it's worth, `declarePostulate` can only be used from `unquoteDecl`, and it's impossible to generate new names using the reflection machinery --- in other words, to use `declarePostulate`, you have to come up with the name yourself. Put more directly, it's impossible to write a macro that just postulates whatever the goal type is.

shiranaiyo:
Also, any macro could easily be modified so that it prints the constructed term to emac, and then we could copy-paste it to agda; the TCM is really isolated.

fredrik-bakke:
Did you forget to push it?

shiranaiyo:
yeah oops sorry haha

DESIGN-PRINCIPLES.md
```diff
@@ -28,14 +28,20 @@ makes use of several postulates.
    [`synthetic-homotopy-theory.circle`](https://unimath.github.io/agda-unimath/synthetic-homotopy-theory.circle.html)
 9. **Pushouts** are postulated in
    [`synthetic-homotopy-theory.pushouts`](https://unimath.github.io/agda-unimath/synthetic-homotopy-theory.pushouts.html)
+10. **Agda built-in types** are postulate in
```
fredrik-bakke, line 31: postulated*

DESIGN-PRINCIPLES.md
```diff
@@ -28,14 +28,20 @@ makes use of several postulates.
...
 [univalence axiom implies function extensionality](https://unimath.github.io/agda-unimath/foundation.univalence-implies-function-extensionality.html),
 but we still assume function extensionality separately. Furthermore, The
 interval type is contractible, so there is no need at all to postulate it. The
 circle can be constructed as the type of `ℤ`-torsors, and the replacement axiom
-can be used to prove there is a circle in `UU lzero`. Finally, the replacement
-axiom can be proven by the join construction, which only uses pushouts.
+can be used to prove there is a circle in `UU lzero`. Adittionally, the
+replacement axiom can be proven by the join construction, which only uses
+pushouts. Finally, the Agda built-in types do not change the semantics of the
+theory, they only give convenience to the user of the library.
```
fredrik-bakke, line 44: Are we completely 100% sure about this?

src/primitive-types.lagda.md
```diff
@@ -0,0 +1,10 @@
+# Primitive types
```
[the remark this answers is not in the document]
fredrik-bakke: Not to insist on changing it, I'm just curious.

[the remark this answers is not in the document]
shiranaiyo: hmm yeah you're right I'm reverting to the other one

DESIGN-PRINCIPLES.md
shiranaiyo: done, ty

DESIGN-PRINCIPLES.md
shiranaiyo: I am, at least. There's two cases
- For all primitives except the TCM: We can interpret the postulated `Char` to be whatever models `Fin 1114112`, and so on.
- For the TCM, all the codomains of the postulated functions are of the form `TC X`, for some `X`, so as Amy said
> you do not (in fact, you can not) prove anything non-trivial about TCM.

So we might as well interpret `TC X` as unit and done.

EgbertRijke:
I think this PR is ready now. Thank you for your amazing work, Fernando!

PR #573, opened 2023-04-25
fixing bug

EgbertRijke:
The website didn't compile because a title wasn't formatted correctly. I fixed the title, but this error should probably have been caught by the checks.


fredrik-bakke:
Right, thanks for the remark. I'll maybe make a PR fixing this later today

PR #579, opened 2023-04-27
Miscellaneous refactoring

fredrik-bakke:


src/finite-group-theory/delooping-sign-homomorphism.lagda.md
```diff
@@ -107,1954 +108,1431 @@ module _
...
+      eq-equiv
+        ( equivalence-class (R (n +ℕ 2) (Fin-UU-Fin l1 (n +ℕ 2))))
+        ( raise l4 (Fin 2))
+        ( compute-raise-Fin l4 2 ∘e
+          inv-equiv
+            ( equiv-D/R-fin-2-equiv
+              ( n +ℕ 2)
+              ( Fin-UU-Fin l1 (n +ℕ 2))
+              ( star)
+              ( compute-raise-Fin l1 (n +ℕ 2))))
```
fredrik-bakke, line 162: Please advise on what this thing should be named.

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
```diff
@@ -107,1954 +108,1431 @@ module _
...
+    that-thing :
+      (n : ℕ) →
+      Σ ( Fin 2 → equivalence-class (R (n +ℕ 2) (raise-UU-Fin-Fin (n +ℕ 2))))
+        ( is-equiv)
+    that-thing n =
+      equiv-D/R-fin-2-equiv
+        ( n +ℕ 2)
+        ( raise-UU-Fin-Fin (n +ℕ 2))
+        ( star)
+        ( compute-raise-Fin l1 (n +ℕ 2))
```
fredrik-bakke, line 208: Please advise on what that thing should be named.

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
```diff
@@ -107,1954 +108,1431 @@ module _
...
+      (n : ℕ) (p : type-Group (loop-group-Set (raise-Fin-Set l1 (n +ℕ 2)))) →
+      D ( n +ℕ 2)
+        ( raise-Fin l1 (n +ℕ 2) ,
+          unit-trunc-Prop (compute-raise-Fin l1 (n +ℕ 2)))
+    this-third-thing n p =
+      quotient-aut-succ-succ-Fin n
+        ( map-hom-symmetric-group-loop-group-Set
+          ( raise-Fin-Set l1 (n +ℕ 2))
+          ( raise-Fin-Set l1 (n +ℕ 2))
+          ( p))
```
fredrik-bakke, line 544: Please advise on what this third thing should be named.

fredrik-bakke (commented):
@EleonoreMangel
I hope I'm not stepping on your toes here, but I found there is quite a bit of duplicate code in the delooping sign homomorphism files. I tried to refactor some of it, but need some help naming the terms. Please see the specific comments.

There is still things that can be done with these files, but I'm throwing in the towel here.

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
EgbertRijke: Can't you write that Sigma-type as a type of equivalences?

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
EgbertRijke: In the language of `univalent-combinatorics.counting` equivalences from a standard finite type to a type `A` are called countings of `A`. So we could use that in the name, or perhaps `eq-counting-...`.

But the naming is made a bit more difficult because the equivalence relation `R` isn't named. The naming scheme of agda-unimath relies on the fact that everything is named. If that is done consistently, then naming subsequent things is often a lot easier. Here I wouldn't really know what to name it either, unless we can come up for a name of `R`.

src/foundation-core/equivalences.lagda.md
```diff
@@ -46,14 +46,37 @@ module _
...
   is-equiv f = sec f × retr f
 
-_≃_ :
-  {i j : Level} (A : UU i) (B : UU j) → UU (i ⊔ j)
+_≃_ : {l1 l2 : Level} (A : UU l1) (B : UU l2) → UU (l1 ⊔ l2)
 A ≃ B = Σ (A → B) is-equiv
 ```
 
-### Immediate structure of equivalences
+### Projections for equivalences
```
EgbertRijke, line 53: Would it be better to say that this is `### Infrastructure for equivalences`?

src/foundation/preidempotent-maps.lagda.md
```diff
@@ -0,0 +1,53 @@
+# Preidempotent maps
```
EgbertRijke, line 1: Nice! I've been wanting to formalize this paper of Mike. Thanks for making a start!

src/foundation/pullback-squares.lagda.md
```diff
@@ -0,0 +1,74 @@
...
+  pullback-square =
+    Σ ( A → C)
+      ( λ f →
+        Σ ( B → C)
+          ( λ g →
+            Σ ( cone f g X)
+              (universal-property-pullback l f g)))
+```
+
+## Projections
```
EgbertRijke, line 51: I wouldn't make a separate section `## Projections` for this. Most of the time I put this right in the same agda-block with the definition.

src/structured-types/pointed-cartesian-product-types.lagda.md
```diff
@@ -0,0 +1,37 @@
...
+module _
+  {l1 l2 : Level}
+  where
+
+  prod-Pointed-Type :
+    (A : Pointed-Type l1) (B : Pointed-Type l2) → Pointed-Type (l1 ⊔ l2)
+  pr1 (prod-Pointed-Type (A , a) (B , b)) = A × B
+  pr2 (prod-Pointed-Type (A , a) (B , b)) = a , b
+
+  _×*_ = prod-Pointed-Type
```
EgbertRijke, line 36: Cheeky! :)

src/synthetic-homotopy-theory/wedges-of-pointed-types.lagda.md
```diff
@@ -24,9 +24,10 @@ open import synthetic-homotopy-theory.pushouts
 </details>
 
 ```agda
-_∨_ :
-  {l1 l2 : Level} (A : Pointed-Type l1) (B : Pointed-Type l2) → Pointed-Type (l1 ⊔ l2)
-A ∨ B =
+_∨*_ :
```
EgbertRijke, line 27: The old notation was more standard, but I like this change for consistency with the other pointed operations.

EgbertRijke (commented):
Amazing PR. Thanks so much for doing this.

Since this is a rather large PR changing many files in the library, let's aim to merge this quickly.

src/category-theory/products-precategories.lagda.md
```diff
@@ -44,20 +44,20 @@ module _ {l1 l2 : Level} (C : Precat l1 l2) where
     type-hom-Precat C p x →
     type-hom-Precat C p y →
     UU (l1 ⊔ l2)
-  is-product x y p proj₁ proj₂ =
+  is-product x y p pr1 pr2 =
```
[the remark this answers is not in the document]
fredrik-bakke: Oops! That's a rogue search-and-replace. Well spotted!

src/foundation-core/equivalences.lagda.md
fredrik-bakke: I think "infrastructure" is a little unspecific. Surely we can find a better word!

src/structured-types/pointed-cartesian-product-types.lagda.md
fredrik-bakke: I think this one looks quite ugly. Do you want to keep it?

fredrik-bakke:
> BTW, I didn't see your changes in the design-principles file among the files that were changed. Did I miss it?

I think this was merged as part of an earlier PR. This branch is an amalgamation of some old and new code. 😅

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
fredrik-bakke: Since `this-thing` is a private term, I think we should get away with referring to the variable `R` in the name.
Is `eq-counting-R` a passable name?

EgbertRijke: `eq-counting-equivalence-class-R`

src/foundation-core/equivalences.lagda.md
EgbertRijke: How about not giving it a header, and just put it in the same agda-block as the definition of equivalences? That's fairly standard practice for me anyways, to put the infrastructure right there with the definition.

src/structured-types/pointed-cartesian-product-types.lagda.md
EgbertRijke: I agree it looks very odd. On the other hand, it fits in perfectly with the other operations. Why don't we just try this one for now, and if we don't like it we can always change it later.

EgbertRijke:
> > BTW, I didn't see your changes in the design-principles file among the files that were changed. Did I miss it?
> 
> I think this was merged as part of an earlier PR. This branch is an amalgamation of some old and new code. 😅

I've had that issue a couple of times too. My "solution" was to do a `git reset --hard upstream/master` on an existing branch that should not have changes wrt upstream/master. It's a bit of a dangerous solution:)

src/category-theory/products-precategories.lagda.md
```diff
@@ -25,11 +25,11 @@ open import foundation.universe-levels
...
 - an object `p`
-- morphisms `proj₁ : hom p x` and `proj₂ : hom p y` such that for every object
-  `z` and morphisms `f : hom z x` and `g : hom z y` there exists a unique
-  morphism `h : hom z p` such that
-- `comp proj₁ h = f`
-- `comp proj₂ h = g`.
+- morphisms `pr1 : hom p x` and `pr2 : hom p y` such that for every object `z`
+  and morphisms `f : hom z x` and `g : hom z y` there exists a unique morphism
+  `h : hom z p` such that
+- `comp pr1 h = f`
```
EgbertRijke, line 31: Now this line has `pr1` and `pr2` in it. Is that intentional?

src/elementary-number-theory/based-strong-induction-natural-numbers.lagda.md
```diff
@@ -381,9 +381,8 @@ comp-succ-based-strong-ind-ℕ k P p0 pS n N N' =
       ( pS n N)
       ( eq-htpy
         ( λ m →
-          eq-htpy
-            ( λ M →
-              eq-htpy
-                ( eq-inductive-step-comp-succ-based-strong-ind-ℕ
-                  k P p0 pS n N m M))))))
+        eq-htpy
```
EgbertRijke, line 384: Shouldn't the indentation level be 2 spaces further right?

fredrik-bakke: Thanks

src/foundation-core/equivalences.lagda.md
fredrik-bakke: That could work. I had this issue with `higher-modalities` though, that the projections section became very large and it was difficult to navigate. The extra headers really help me out in this regard.

fredrik-bakke: What about `### Components of an equivalence`?

EgbertRijke: That's true. I also make those parts very large sometimes. "Components" sounds good. But note that in some of the more complicated cases the basic infrastructure can contain more than just components or iterated projections.

fredrik-bakke:
Alright, I think it only remains to determine names for `that-thing` and `this-third-thing` now. 😊

src/category-theory/groupoids.lagda.md
```diff
@@ -37,8 +37,8 @@ A groupoid is a category in which every morphism is an isomorphism.
 ## Definition
 
 ```agda
-is-groupoid-cat-Prop : {l1 l2 : Level} (C : Cat l1 l2) → Prop (l1 ⊔ l2)
-is-groupoid-cat-Prop C = is-groupoid-precat-Prop (precat-Cat C)
+is-groupoid-Cat-Prop : {l1 l2 : Level} (C : Cat l1 l2) → Prop (l1 ⊔ l2)
+is-groupoid-Cat-Prop C = is-groupoid-Precat-Prop (precat-Cat C)
```
EgbertRijke, line 41: I have not been consistent with capitalization. Sometimes I think that only one part of the name should be capitalized, but other times I capitalized more parts of the name. Perhaps we can come up with a decent capitalization convention.

src/category-theory/isomorphisms-categories.lagda.md
```diff
@@ -155,32 +155,31 @@ module _
...
             ( hom-inv-iso-Cat C g))
         by
-        ap
-          ( comp-hom-Cat C (hom-iso-Cat C g))
-          ( inv
-            ( assoc-comp-hom-Cat C
-              ( hom-iso-Cat C f)
-              ( hom-inv-iso-Cat C f)
-              ( hom-inv-iso-Cat C g)))
+          ap
```
EgbertRijke, line 158: This increase in indentation level looks a bit strange to me. I don't see why the identification should be more deeply indented then the target of the identification. On the other hand, since the wide equals is slightly longer there is no way to nicely align them anyway.

src/category-theory/groupoids.lagda.md
fredrik-bakke: Sure, I'll think a bit about it.

src/category-theory/isomorphisms-categories.lagda.md
fredrik-bakke: I guess this is just a personal preference for me. I think it is worth it to create guidelines for this syntax though, as there are a lot of different flavors of `equational-reasoning`-code in the library.

fredrik-bakke:
Thank you for all your prompt and attentive feedback by the way!

src/category-theory/groupoids.lagda.md
fredrik-bakke: Okay, I've thought a bit more about it, and to me, capitalizing both words makes the most sense. This is consistent with the convention of just appending names, so that I, the user, can be sure that `is-groupoid-Cat-Prop` is the `Prop`-inhabitant of `is-groupoid-Cat`. Moreover, this convention would make it so we don't have to introduce an exception to another convention.

I don't really see the argument for writing `is-groupoid-cat-Prop`, but would love to hear any thoughts you have.

fredrik-bakke: Oh, I guess actually the highlighted code gives another good example, `precat-Cat`. Here, my previous comment would suggest that we write instead `Precat-Cat`, but then the name would suggest that it took values in types. I.e. that we have a notion of "Precategory Categories" or some such thing. Perhaps it would be better to name this term `underlying-Precat-Cat` but again I'm not sure I like this, as in this case there are many other constructions that should be renamed as well.

EgbertRijke: The way I think about this is that we're working in the category `Cat`. This appends `-Cat` to everything that we do, but otherwise all the constructions defined on things from the category `Cat` should be lowercase. This makes the name `precat-Cat` logical. I wouldn't want to have a naming scheme where we write `underlying-` in front of everything, because it will make many names really long. E.g. would we want to write `x : underlying-type-Group` every time we want to assume an element of the underlying type of a group?

EgbertRijke: On the other hand, with this explanation of naming and capitalization, perhaps it would be more natural to write `is-groupoid-prop-Precat` for the proposition that a precategory is a groupoid.

I've been appending `-Prop` at the end most of the time, when something is a proposition.

src/category-theory/natural-transformations-precategories.lagda.md
```diff
@@ -65,7 +66,8 @@ module _
     type-hom-Precat D (obj-functor-Precat C D F x) (obj-functor-Precat C D G x)
   components-nat-trans-Precat = pr1
 
-  squares-nat-trans-Precat : (γ : nat-trans-Precat) → is-nat-trans-Precat (components-nat-trans-Precat γ)
+  squares-nat-trans-Precat :
```
EgbertRijke, line 69: We could call this `coherence-square-nat-trans-Precat`.

I'm not such a fan of `nat-trans` by the way, but I'm not suggestion that you have to change it. I'd be happier to write it out as `natural-transformation`, because I don't like code that is riddled with abbreviations. When too many things are abbreviated, it looks so bureaucratese.

src/category-theory/groupoids.lagda.md
fredrik-bakke: Also, addressing your second to last comment. I guess `precat-Cat` doesn't actually violate my proposed convention. Thus, I think this is the way to go; i.e. keep words capitalized.

EgbertRijke: Ah, perhaps I remember the reason why I write `-Prop` at the end. Whenever we construct a set, a group, a category, or a proposition, we always write `-Set`, `-Group`, `-Cat`, or `-Prop` at the end. This is done consistently, with the exception if it is an underlying object.

EgbertRijke: > Just to address your last comment, the formulation `is-groupoid-prop-Precat` makes very little sense to me. I guess we could do `Prop-is-groupoid-Precat` if we _really_ want to, however.

I think names look nicer if we only capitalize at the end for really important things. Ideally, Agda would have namespaces and then we would have a namespace for what we are capitalizing now. Namespaces are much more flexible than modules, because modules are restricted to single files and once you close them nothing can be added to them. I don't know why Agda doesn't have them.

fredrik-bakke: Ohh, I see. I didn't know the distinction, thanks.

EgbertRijke: There is one form of namespace that Agda has: folders. But I'm not sure I would like to have folders called `Set`, `Prop`, `Cat`, and so on. As much as possible, I prefer to go with folders that describe names of fields of study (although we already have some exceptions).

fredrik-bakke: Just to get back on topic, are we in agreement that it should be called `is-groupoid-Cat-Prop`?

EgbertRijke: > Just to get back on topic, are we in agreement that it should be called `is-groupoid-Cat-Prop`?

Sure, we can capitalize propositions twice. But let's keep `precat-Cat` as it is.

fredrik-bakke: > This is also the reason why within files we mostly work with anonymous modules. Naming modules would be really annoying, because the nice names that you would choose for them should actually be namespaces that you can add stuff to in later files.

I see, thanks for sharing!

src/category-theory/natural-transformations-precategories.lagda.md
fredrik-bakke: Would you also want to rename `nat-iso` to `natural-isomorphism`?

EgbertRijke: Yes

EgbertRijke:
Is this PR finished modulo `that-thing` and `this-third-thing`?

fredrik-bakke:
Sure. I have some more refactoring in mind, but I can save it for another PR.

EgbertRijke:
Great, then I'll merge this one and let you do further refactoring in a new PR.

PR #561, opened 2023-04-11
Enriched directed trees and elements of W-types

EgbertRijke:
This pull request develops the theory of directed and undirected trees.

- moves polynomial endofunctors and algebras for polynomial endofunctors to `trees`
- introduces enriched directed trees
- defines some operations on directed trees and enriched directed trees
- defines enriched directed trees for each element of a coalgebra of a polynomial endofunctor

PR #581, opened 2023-04-28
Refactor 2, the sequel to refactor

fredrik-bakke:
### Line wrapping
- Add statistics in `max_line_length.py`s output.
- Make `max_line_length.py` ignore commented code.
- Make `max_line_length.py` forgiving to lines that just cannot be shorter than the max length.
  - A long line is forgiven if it is a named module declaration, an open import statement, or a term name by itself (with possible parenthization or colon or equals sign).
  - Add documentation explaining this.
- Add script for line wrapping some easily fixable long lines: `scripts/easily_fixable_long_lines.py`.
  - It follows some very conservative rules where I was pretty sure it would work. It still has a lot of potential for improvement.
- Fix all violations of the character limit rule.
- **Make character limit rule mandatory** _- should this be kept?_
- Wrap the first line for definitions that span multiple lines. (This could conceivably also be automated)

### Spacing
- Add script automatically enforcing (mostly) uncontroversial spacing conventions: `spaces_convention_simple.py`.
  - Always whitespace before and after semicolon (`;`)
  - Never whitespace before closing bracket (`)` or `}`)
  - No consecutive whitespace together anywhere after the first non-whitespace character in a line. (see the point below)
- The commits (b47a80e0f16d32c17aed86ca33c79590e3668413 and 43cb2d90ebbf68b2bd668d9013370d61e6d950ce) are slightly more devious. They are based on a suggestion by Jonathan to disallow consecutive whitespace inside lines (i.e. after the first non-whitespace character) in Agda code. This is useful as such whitespace is usually a typo. _However, this also disallows aligning code with whitespace. I think this is beneficial in the project, as artificially aligned code is less maintainable, and does not fit with the style of our library._

### Renaming
- `nat-trans` -> `natural-transformation`
- `nat-iso` -> `natural-isomorphism`
- `respects` -> `preserves`
- `pt-Pointed-Type` -> `point-Pointed-Type`
- `assoc` -> `associative`, except for the proof of associativity of identification concatenation.
- The abbreviation `comp` for `computation` is elaborated to `compute`. Now `comp` is reserved for `composition`.

### Miscellaneous
- Format all non-code comments properly with markdown.
- Add miscellaneous pre-commit hooks: `check-ast`, `double-quote-string-fixer`, `mixed-line-ending`.
- Format files in `type-theories`.
- Remove more unused imports.


fredrik-bakke:
Would you want `iso` in general to be elaborated to `isomorphism`? This feels a little overkill to me.

fredrik-bakke:
However, I'm not against the idea of disambiguation `comp` into `compute` and `compose` or whatever else it can stand for.

EgbertRijke:
> Would you want `iso` in general to be elaborated to `isomorphism`? This feels a little overkill to me.

No, I think stuff like `iso-Group` and even `iso-Cat` reads just fine.

EgbertRijke:
> However, I'm not against the idea of disambiguation `comp` into `compute` and `compose` or whatever else it can stand for.

I already write out `compute` in several places. If you see any places where it would be better to expand it, feel free to go ahead.

EgbertRijke:
Note that there is a big PR on trees coming up. In this PR I also touch pretty much every file in trees extensively. I hope to finish it up soon, but perhaps it is not the best time to do an extensive refactoring here as well, as it will cause a complicated merge conflict.

fredrik-bakke:
> Note that there is a big PR #561 on trees coming up. In this PR I also touch pretty much every file in trees extensively. I hope to finish it up soon, but perhaps it is not the best time to do an extensive refactoring here as well, as it will cause a complicated merge conflict.

Thanks for the heads up! I can revert the changes in that namespace, they are very minor anyways.

fredrik-bakke:
In some places the abbreviation `Eq-Rel` is used, while in other places, `Equivalence-Relation` is used. What do you prefer?

fredrik-bakke:
> > However, I'm not against the idea of disambiguation `comp` into `compute` and `compose` or whatever else it can stand for.
> 
> I already write out `compute` in several places. If you see any places where it would be better to expand it, feel free to go ahead.

I went ahead and made the distinction everywhere I could. This way, the notation is more consistent across the board, and there cannot be confusion with regard to this.

FernandoChu:
All the `comp` stuff in the category folder is for `compose`

fredrik-bakke:
Thanks! But I figured this out 😅 You can see what's left by cloning my branch :)

src/foundation/universal-property-truncation.lagda.md
```diff
@@ -410,7 +410,7 @@ map-id-trunc-Set {l1} {A} =
         { pair (map-trunc-Set id) (naturality-trunc-Set id)}
         { pair id refl-htpy}))
 
-map-comp-trunc-Set :
+map--trunc-Set :
```
FernandoChu, line 413: maybe a mistake?

FernandoChu:
> Thanks! But I figured this out sweat_smile You can see what's left by cloning my branch :)

Oops, I dled directly from your fork but forgot to change branch haha. It is somewhat weird that `comp` is being removed but not `precomp` and `postcomp`; I'm personally not sure if there are worth expanding too.

src/foundation/universal-property-truncation.lagda.md
fredrik-bakke: Oops, thanks!

fredrik-bakke:
> > Thanks! But I figured this out sweat_smile You can see what's left by cloning my branch :)
> 
> Oops, I dled directly from your fork but forgot to change branch haha. It is somewhat weird that `comp` is being removed but not `precomp` and `postcomp`; I'm personally not sure if these are worth expanding too.

I agree, and I see three possible solutions to this.
1. We let `precomp` be `precomp` and use `compose` when this is a standalone word.
2. We elaborate `precomp` to `precompose` and so on.
3. We reserve the word `comp` for `compose` and always elaborate `compute`.

At the moment I'm thinking the second option would be the most consistent one. The first one is kind of iffy as one would expect `precomp` and `compose` to agree. The third one is a little tricky to enforce, and would be a source of confusion I think.

EgbertRijke:
I know I have been inconsistent with this, but can we have `associative` instead of `assoc`?

src/category-theory/functors-large-precategories.lagda.md
```diff
@@ -47,12 +47,12 @@ module _
         type-hom-Large-Precat D
           ( obj-functor-Large-Precat X)
           ( obj-functor-Large-Precat Y)
-      preserves-comp-functor-Large-Precat :
+      preserves-compose-functor-Large-Precat :
```
EgbertRijke, line 50: "preserves-compose" doesn't read good. This should be "preserves-composition"

src/category-theory/functors-precategories.lagda.md
```diff
@@ -61,12 +62,12 @@ module _
     type-hom-Precat D (obj-functor-Precat F x) (obj-functor-Precat F y)
   hom-functor-Precat F = pr1 (pr2 F)
 
-  respects-comp-functor-Precat :
+  respects-compose-functor-Precat :
```
EgbertRijke, line 65: `respects-composition`... But actually I would prefer if it said `preserves-composition`

src/category-theory/isomorphisms-categories.lagda.md
```diff
@@ -121,131 +121,135 @@ module _
...
           ap
-            ( comp-hom-Cat C (hom-inv-iso-Cat C f))
-            ( left-unit-law-comp-hom-Cat C (hom-iso-Cat C f))
+            ( compose-hom-Cat C (hom-inv-iso-Cat C f))
+            ( left-unit-law-compose-hom-Cat C (hom-iso-Cat C f))
       ＝ id-hom-Cat C
         by isretr-hom-inv-iso-Cat C f
 
-  is-iso-hom-comp-iso-Cat :
+  is-iso-hom-compose-iso-Cat :
```
EgbertRijke, line 239: This also doesn't read well anymore. I'm starting to wonder whether in the case of composition the change of `comp` to `compose` is really worthwhile.

src/category-theory/isomorphisms-categories.lagda.md
```diff
@@ -421,20 +428,20 @@ module _
     {x y : obj-Cat C} (p : x ＝ y) → eq-iso-Cat (iso-eq-Cat C p) ＝ p
   isretr-eq-iso-Cat {x} {y} = isretr-map-inv-equiv (extensionality-obj-Cat x y)
 
-  preserves-comp-eq-iso-Cat :
+  preserves-composite-eq-iso-Cat :
```
EgbertRijke, line 431: `preserves-composition-eq-iso-Cat` would be better in my opinion.

fredrik-bakke:
Found an issue with one of the python functions. Trying to fix it currently

EgbertRijke (commented):
I reviewed some of the files in the category theory folder. Some of the names don't as well with `compose` as they did with `comp`, so I'm not sure if `compose` is such a good choice for a general naming scheme. We could try `composition` everywhere, but I wouldn't be against keeping `comp` either.

src/category-theory/isomorphisms-categories.lagda.md
fredrik-bakke: I am also not so sure.

> I know I have been inconsistent with this, but can we have `associative` instead of `assoc`?

Do you also want the associativity proof for identifications to be called just `associative`? To me, this reads a bit oddly.

fredrik-bakke:
> I know I have been inconsistent with this, but can we have `associative` instead of `assoc`?

Do you also want this for the associativity proof for identifications? I.e. just `associative`? This reads a bit oddly to me.

src/category-theory/isomorphisms-categories.lagda.md
fredrik-bakke: I am also not so sure. Maybe if we consistently use `compute` for computations, and then reserve `comp` for compositions?

fredrik-bakke: This would also resolve the issue that Fernando posed.

fredrik-bakke:
Okay, so if we are to change `compose` to `comp`, then I would think it best if we found out which terms actually use `comp` for something other than a composite. Currently, I've found the following terms to be ambiguous:

```agda
list-comp-f-g :
  ( h : Fin n ≃ type-UU-Fin n X) →
  list
    ( Σ ( (type-UU-Fin n X) → Decidable-Prop l)
        ( λ P →
          has-cardinality 2
            ( Σ (type-UU-Fin n X) (type-Decidable-Prop ∘ P))))
```
```agda
sign-comp-Eq-Rel : Eq-Rel lzero (Fin n ≃ type-UU-Fin n X)
```
and all the other `comp` terms in `simpson-delooping-sign-homomorphism`
```agda
comp-htpy-eq-square-refl-htpy :
  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {C : UU l4}
  (f : A → X) (g : B → X) (c c' : cone f g C) →
  ( (htpy-eq-square-refl-htpy f g c c') ∘
    (concat (tr-tr-refl-htpy-cone f g c) c')) ~
  ( htpy-eq-square f g c c')
```
```agda
  comp-htpy-parallel-cone-eq' :
    {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} {C : UU l4}
    (f : A → X) (g : B → X) (c c' : cone f g C) →
    ( ( htpy-parallel-cone-eq' f refl-htpy c c') ∘
      ( concat (tr-tr-refl-htpy-cone f g c) c')) ~
    ( htpy-eq-square f g c c')
```
```agda
precomp-comp-Set-Quotient :
  {l1 l2 l3 l4 l5 : Level} {A : UU l1} (R : Eq-Rel l2 A)
  (B : Set l3) (f : reflecting-map-Eq-Rel R (type-Set B))
  (C : Set l4) (g : type-hom-Set B C)
  (D : Set l5) (h : type-hom-Set C D) →
  ( precomp-Set-Quotient R B f D (h ∘ g)) ＝
  ( precomp-Set-Quotient R C (precomp-Set-Quotient R B f C g) D h)
```
```agda
cocone-map-comp :
  {l1 l2 l3 l4 l5 l6 : Level} {S : UU l1} {A : UU l2} {B : UU l3}
  (f : S → A) (g : S → B) {X : UU l4} (c : cocone f g X)
  {Y : UU l5} (h : X → Y) {Z : UU l6} (k : Y → Z) →
  Id (cocone-map f g c (k ∘ h)) (cocone-map f g (cocone-map f g c h) k)
```
Help with any of these would be greatly appreciated!

src/category-theory/functors-precategories.lagda.md
fredrik-bakke: I've changed it to `preserves-compose` currently. Awaiting with the second half until we have resolved the main issue at hand.

FernandoChu:
`comp` in `precomp-comp-Set-Quotient` is `compose`. I'm glad we're going with option 3 btw.

Since this is already a refactor PR, I'd also like to point out the arbitraryness between using `Id` or `＝`.

fredrik-bakke:
Suggestions are very welcome, although I won't guarantee I will address them!

You are right that the usage of `Id` and `＝` in the library is quite arbitrary. The same goes for `pair`, `,`, `triple`, and `pair'` and so on. Also, some people like writing lots of lambdas, while I personally prefer using underscore notation and `∘`.

Although I do often replace lambdas, I don't really see the need to enforce either use. After all, the different notations have slightly different uses.

FernandoChu:
Personally, I like `＝` and `,` more than their counterparts, at least for things that fit in one line, though I've been told that pair is prefered. I don't have any objective reason though, I think everyone could undersstand both ways easily.

fredrik-bakke:
Although I also _tend_ to prefer `_,_`, there is a good argument for `pair`. In larger constructions, it usually makes the structure of the code more clear. I think the same can be said for `Id` as opposed to `＝`.

fredrik-bakke:
Unrelated question: how do we feel about renaming `pt-Pointed-Type` to `point-Pointed-Type`?

EgbertRijke:
> > I know I have been inconsistent with this, but can we have `associative` instead of `assoc`?
> 
> Do you also want this for the associativity proof for identifications? I.e. just `associative`? This reads a bit oddly to me.

Good catch. We can keep calling this `assoc`. Operations on identifications are important enough to give them the most primitive names.

EgbertRijke:
> Although I also _tend_ to prefer `_,_`, there is a good argument for `pair`. In larger constructions, it usually makes the structure of the code more clear. I think the same can be said for `Id` as opposed to `＝`.

I used to prefer `pair` for this reason, but people kept bringing the comma up and eventually I gave in. Now I'm neutral about this. I'm ok with `_,_`.

EgbertRijke:
> Unrelated question: how do we feel about renaming `pt-Pointed-Type` to `point-Pointed-Type`?

Sure!

EgbertRijke:
Actually I have a story about the comma as a mathematical operator:)

For me, the comma is just a punctuation symbol. It is just there to aid readability. Then someone told me that my formulas didn't type check, because I am used to writing functions that take multiple arguments with a comma separating their arguments, even when I really intend them to have that number of arguments and not a single argument from an iterated Sigma-type. We kept having arguments about it, and it was really hard to figure out what the problem was. It turned out that they made the assumption that the comma was a mathematical operation, i.e., that it was the pairing function of the Sigma-type. Under that assumption I agreed that my formulas didn't typecheck, but I was stunned to find out that people assume that the comma is a mathematical operation and not a punctuation symbol. When you silently make that assumption, then it is bound to lead to confusion with people who are not trained to think of `_,_` as a mathematical operation. So then I was even more opposed to use the `_,_` in the agda-unimath library. I didn't want to contribute to the confusion. But by now I have accepted `_,_` in the library, because it does make some pairs more readable.

End of the story:)

FernandoChu:
Nice story!

In defense the other person, using `_,_` as a mathematical operation is *conservative*. It restricts what typechecks, but does not expands it, so it won't confuse anyone no matter their interpretation of `_,_`. What generates confusion is stead using it as syntax.

EgbertRijke:
I don't think it is conservative in mathematical text. Having the comma as a punctuation symbol allows the comma to be used far more often than when you stipulate that the comma is an operation of type `(a : A) -> B a -> Sigma A B`. So you must restrict your use of the comma by quite a lot, when writing mathematics. The friction point was mostly that people tend to not tell you whether they assume that the comma is a mathematical operation. My opinion is then if you do assume it is a mathematical operation, then it is your responsibility to be clear about it, because for the rest of the world the comma is just a punctuation symbol.

So, within this library it won't confuse people, but if you go out and talk to people you have to be aware that the comma is not this pairing operation for everyone.

FernandoChu:
Good point, going one step back, I was talking a conservativity w.r.t. unambiguous constructions such as `pair` and curried functions, i.e. replacing pairs with `,` will always be understood by both interpretations of `,`. On the other hand, it breaks conservativity w.r.t. any mathematical book. Which should be the reference is not an objective decision. I'm currently reading how Coq people manage these things, hopefully I'll be back with some helpful insight.

fredrik-bakke:
Okay, I think this is a good point to stop making additions to the PR.

fredrik-bakke:
I understand it will be difficult to properly review this PR, so please at least take some time to reflect on the points I highlight in the summary, if these are actually wanted in the library.

fredrik-bakke:
Actually, I will add documentation explaining the line limit in a little more detail, but this will be the last thing for this PR

EgbertRijke:
> I understand it will be difficult to properly review this PR, so please at least take some time to reflect on the points I highlight in the summary, if these are actually wanted in the library.

On the other hand I think it will definitely be necessary to review this PR, since you made a lot of changes in bulk, and might have missed some unforeseen applications of your scripts. I'll try to do it tomorrow.

EgbertRijke:
BTW, the character limit rule cannot be mandatory. Some names in our library are too long:)

fredrik-bakke:
Hehe yes, it can! Notice how `pre-commit` is happy with this branch. 😇

EgbertRijke:
Oh! That's very surprising! What did you do with those long names?

fredrik-bakke:
Nothing! I just made exceptions to the rule. Give me 2 minutes, I'll push the documentation that explains this.

fredrik-bakke:
From the commit:

## <a name="character-limit"></a>80-character limit

The library employs an 80-character line limit to improve readability and
maintainability. However, there are a few notable exceptions to this rule that
enable us to have names that are more than 80 characters long. The exceptions to
the rule are as follows:

- Named module declarations
- Open import statements
- Lines consisting of a single, possibly parenthesized (`(){}`), token that is
  potentially followed by one of the symbols `;`, `:`, `=`, or `→`.

src/foundation/singleton-induction.lagda.md
```diff
@@ -37,10 +37,10 @@ ind-is-singleton :
   B a → (x : A) → B x
 ind-is-singleton a is-sing-A B = pr1 (is-sing-A B)
 
-comp-is-singleton :
+compute-is-singleton :
```
EgbertRijke, line 40: `compute-ind-is-singleton`

src/foundation/symmetric-operations.lagda.md
```diff
@@ -76,7 +76,7 @@ is-commutative-symmetric-operation f x y =
   ap f (is-commutative-standard-unordered-pair x y)
 ```
 
-### A binary operation from $A$ to $B$ is commutative if and only if it extends to the unordered pairs in $A$.
+### A binary operation from $A$ to $B$ is commutative if and only if it extends to the unordered pairs in $A$
```
EgbertRijke, line 79: ``### A binary operation from `A` to `B` is commutative if and only if it extends to the unordered pairs in `A` ``

src/foundation/transport.lagda.md
```diff
@@ -82,7 +82,8 @@ module _
 ```agda
 tr-eq-pair-diagonal :
   {l1 l2 : Level} {A : UU l1} {a0 a1 : A} (C : A × A → UU l2)
-  (p : a0 ＝ a1) (u : C (a0 , a0)) → (tr C (eq-pair p p) u) ＝ (tr (λ a → C (a , a)) p u)
+  (p : a0 ＝ a1) (u : C (a0 , a0)) →
+  (tr C (eq-pair p p) u) ＝ (tr (λ a → C (a , a)) p u)
```
EgbertRijke, line 86: I guess the parentheses around the arguments both sides of the equation aren't necessary here

src/foundation/univalence.lagda.md
```diff
@@ -106,30 +136,38 @@ eq-equiv-fam :
 eq-equiv-fam {B = B} {C} = map-inv-is-equiv (is-equiv-equiv-eq-fam B C)
 ```
 
-## Properties
+## Computations with univalence
```
EgbertRijke, line 139: I think I prefer `## Properties` here, if it is intended as a level `##` section. Otherwise make it a `###` section.

src/foundation/univalence.lagda.md
```diff
@@ -142,13 +180,3 @@ commutativity-inv-eq-equiv A B f =
           ( λ e → map-equiv e (inv-equiv f))
           ( inv (right-inverse-law-equiv equiv-univalence)))))
 ```
-
-### eq-equiv on id is refl
```
EgbertRijke, line 146: Why was this removed?

src/orthogonal-factorization-systems/higher-modalities.lagda.md
```diff
@@ -51,27 +51,27 @@ module _
   (unit-○ : modal-unit ○)
   where
 
-  modal-ind : UU (lsuc l1 ⊔ l2)
-  modal-ind =
+  ind-modal : UU (lsuc l1 ⊔ l2)
```
EgbertRijke, line 54: Perhaps we could make the naming scheme with `modality` rather than `modal`. What do you think?

src/set-theory/cardinalities.lagda.md
```diff
@@ -161,7 +163,8 @@ is-effective-cardinality X Y =
   ( is-effective-unit-trunc-Set (Set _) X Y)
 ```
 
-### Using this and assuming excluded middle , we can show `≤-cardinality` is a partial order by showing that it is antisymmetric.
+Using this and assuming excluded middle , we can show `≤-cardinality` is a
```
EgbertRijke, line 166: I think this was fine as a subsection header. We often have such descriptive titles as subsection headers. Then the will show up on the RHS of the website.

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
```diff
@@ -276,118 +276,110 @@ module _
...
-          ( equiv-cauchy-product-Σ-extension-species-subuniverse
-            ( P)
-            ( Q)
-            ( C3)
-            ( C5)
-            ( cauchy-exponential-species-subuniverse P Q C1 S)
-            ( cauchy-exponential-species-subuniverse P Q C1 T)
-            ( inclusion-subuniverse P X))) ∘e
-        ( ( equiv-tot
+    ( inv-equiv
```
EgbertRijke, line 287: I think the parentheses that you removed should be put back. With these parentheses it will be easier to spot that the `inv-equiv ...` is the first argument of a binary operation. Note that the second argument also has parentheses. I.e. it should have the following form:

```agda
( ( inv-equiv
    ( some-equivalence)) ∘e
  ( second-equivalence)) ∘e
( third-equivalence)
```

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
```diff
@@ -276,118 +276,110 @@ module _
...
+      ( equiv-Σ-extension-species-subuniverse
+        ( P)
+        ( subuniverse-global-subuniverse Q (lsuc l1 ⊔ l2 ⊔ l3 ⊔ l4))
+        ( cauchy-product-species-subuniverse P Q C3
+          ( cauchy-exponential-species-subuniverse P Q C1 S)
+          ( cauchy-exponential-species-subuniverse P Q C1 T))
+        ( X)) ∘e
+      ( inv-equiv
```
EgbertRijke, line 295: I think the parentheses that you removed should be put back. With these parentheses it will be easier to spot that the `inv-equiv ...` is the first argument of a binary operation. Note that the second argument also has parentheses. I.e. it should have the following form:

```agda
( ( inv-equiv
    ( some-equivalence)) ∘e
  ( second-equivalence)) ∘e
( third-equivalence)
```

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
```diff
@@ -276,118 +276,110 @@ module _
...
+        ( equiv-cauchy-product-Σ-extension-species-subuniverse
+          ( P)
+          ( Q)
+          ( C3)
+          ( C5)
+          ( cauchy-exponential-species-subuniverse P Q C1 S)
+          ( cauchy-exponential-species-subuniverse P Q C1 T)
+          ( inclusion-subuniverse P X)) ∘e
+        ( equiv-tot
```
EgbertRijke, line 304: I think the parentheses that you removed should be put back. With these parentheses it will be easier to spot that the `inv-equiv ...` is the first argument of a binary operation. Note that the second argument also has parentheses. I.e. it should have the following form:

```agda
( ( inv-equiv
    ( some-equivalence)) ∘e
  ( second-equivalence)) ∘e
( third-equivalence)
```

(now with equiv-tot)

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
```diff
@@ -276,118 +276,110 @@ module _
...
+            ( Σ-extension-species-subuniverse
+              ( P)
+              ( subuniverse-global-subuniverse Q l3)
+              ( S))
+            ( Σ-extension-species-subuniverse
+              ( P)
+              ( subuniverse-global-subuniverse Q l4)
+              ( T))
+            ( pr1 X) ∘e
+            ( equiv-tot
```
EgbertRijke, line 335: Probably it's best to revert this change entirely. I've formatted it properly in my earlier round of refactoring.

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
```diff
@@ -276,118 +276,110 @@ module _
...
-                          ( T)
-                          ( cotype-Relaxed-Σ-Decomposition d x)))) ∘e
-              ( ( equiv-cauchy-exponential-Σ-extension-species-subuniverse
-                    ( P)
-                    ( Q)
-                    ( C1)
-                    ( C4)
-                    ( coproduct-species-subuniverse P Q C2 S T)
-                    ( X)) ∘e
+                equiv-Π
```
EgbertRijke, line 337: Probably it's best to revert this change. It was already properly formatted.

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -38,45 +38,57 @@ open import synthetic-homotopy-theory.universal-property-pushouts
...
   ( (x : X) → P x) → dep-cocone f g c P
 dep-cocone-map f g c P h =
-  pair (λ a → h (pr1 c a)) (pair (λ b → h (pr1 (pr2 c) b)) (λ s → apd h (pr2 (pr2 c) s)))
+  ( λ a → h (pr1 c a)) ,
+  ( λ b → h (pr1 (pr2 c) b)) ,
+  ( λ s → apd h (pr2 (pr2 c) s))
+```
 
-{- Definition 18.1.1 The induction principle of pushouts -}
+### Definition 18.1.1 The induction principle of pushouts
```
EgbertRijke, line 64: `### The induction principle of pushouts`

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -38,45 +38,57 @@ open import synthetic-homotopy-theory.universal-property-pushouts
...
   { S : UU l1} {A : UU l2} {B : UU l3} {X : UU l4} →
-  ( f : S → A) (g : S → B) (c : cocone f g X) → UU (lsuc l ⊔ l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  ( f : S → A) (g : S → B) (c : cocone f g X) →
+  UU (lsuc l ⊔ l1 ⊔ l2 ⊔ l3 ⊔ l4)
 Ind-pushout l {X = X} f g c =
   (P : X → UU l) → sec (dep-cocone-map f g c P)
+```
 
-{- Definition 18.1.2 The dependent universal property of pushouts -}
+### Definition 18.1.2 The dependent universal property of pushouts
```
EgbertRijke, line 76: `### The dependent universal property of pushouts`

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -38,45 +38,57 @@ open import synthetic-homotopy-theory.universal-property-pushouts
...
 
+```agda
 dependent-universal-property-pushout :
   {l1 l2 l3 l4 : Level} (l : Level) {S : UU l1} {A : UU l2} {B : UU l3}
   (f : S → A) (g : S → B) {X : UU l4} (c : cocone f g X) → UU _
 dependent-universal-property-pushout l f g {X} c =
   (P : X → UU l) → is-equiv (dep-cocone-map f g c P)
+```
+
+### Remark 18.1.3 Computation of the identity type of `dep-cocone`
```
EgbertRijke, line 87: ``### Computation of the identity type of `dep-cocone` ``

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -249,12 +261,14 @@ abstract
         ( λ k → is-equiv-htpy-dep-cocone-eq f g c P
           ( dep-cocone-map f g c P k) h))
       ( is-contr-map-is-equiv (dup-c P) h)
+```
 
-{- This finishes the formalization of remark 18.1.3. -}
+This finishes the formalization of remark 18.1.3.
```
EgbertRijke, line 267: We can just remove this remark

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -284,46 +298,36 @@ dependent-pullback-property-pushout l {S} {A} {B} f g {X}
...
-
-   Proof of 2 → 1
-   universal-property-pushout-pullback-property-pushout
-
-   Proof of 2 → 3
-   dependent-pullback-property-pullback-property-pushout
-
-   Proof of 3 → 2
-   pullback-property-dependent-pullback-property-pushout
+## Theorem 18.1.4
```
EgbertRijke, line 304: `### Equivalent formulations of the universal property of pushouts`

src/synthetic-homotopy-theory/26-descent.lagda.md
```diff
@@ -410,9 +418,11 @@ dependent-pullback-property-dependent-universal-property-pushout
       ( λ h → is-equiv-tot-is-fiberwise-equiv
         ( λ h' → funext (λ x → tr P (H x) (h (f x))) (λ x → h' (g x)))))
     ( I l P)
+```
 
-{- Proof of Theorem 18.1.4, (iv) implies (iii). -}
+### Proof of Theorem 18.1.4, (4) implies (3)
```
EgbertRijke, line 424: I think I'll just have to go over this file another day. There's lots of leftovers from the old formalization of my book here, but you don't have to address that in this PR.

src/univalent-combinatorics/equality-standard-finite-types.lagda.md
```diff
@@ -131,7 +131,7 @@ is-prop-is-zero-or-one-Fin-two-ℕ x =
     ( is-prop-is-one-Fin 1 x)
 ```
 
-### Every element in the standard two-element type is either 0 or 1.
+### Every element in the standard two-element type is either 0 or 1
```
EgbertRijke, line 134: ``### Every element in the standard two-element type is either `0` or `1` ``

EgbertRijke (commented):
A big review for a big PR:)

EgbertRijke (commented):
A big review for a big PR:)

src/elementary-number-theory/bezouts-lemma-natural-numbers.lagda.md
```diff
@@ -707,12 +707,20 @@ is-distance-between-multiples-div-mod-ℕ (succ-ℕ x) y z (u , p) =
              (int-ℕ (succ-ℕ x))) (mul-ℤ p (int-ℕ y)))
           (int-abs-is-nonnegative-ℤ (add-ℤ (int-ℕ (succ-ℕ x))
             (neg-ℤ (int-ℤ-Mod (succ-ℕ x) u))) (int-ℤ-Mod-bounded x u))
-        ＝ dist-ℤ (mul-ℤ (add-ℤ (int-ℕ (abs-ℤ (neg-ℤ a))) (int-ℕ y)) (int-ℕ (succ-ℕ x)))
-          (mul-ℤ (add-ℤ (int-ℕ (succ-ℕ x))
+        ＝ dist-ℤ
```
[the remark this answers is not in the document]
fredrik-bakke: I should note that the only refactoring I did for this file was to wrap lines that exceeded the 80-character limit.

src/elementary-number-theory/divisibility-integers.lagda.md
```diff
@@ -230,10 +230,10 @@ pr1 (neg-div-ℤ x y (pair d p)) = neg-ℤ d
 pr2 (neg-div-ℤ x y (pair d p)) =
   equational-reasoning
     mul-ℤ (neg-ℤ d) (neg-ℤ x)
-    ＝ neg-ℤ (mul-ℤ d (neg-ℤ x))   by left-negative-law-mul-ℤ d (neg-ℤ x)
-    ＝ neg-ℤ (neg-ℤ (mul-ℤ d x))   by ap neg-ℤ (right-negative-law-mul-ℤ d x)
-    ＝ (mul-ℤ d x)                 by neg-neg-ℤ (mul-ℤ d x)
-    ＝ y                           by p
+    ＝ neg-ℤ (mul-ℤ d (neg-ℤ x)) by left-negative-law-mul-ℤ d (neg-ℤ x)
```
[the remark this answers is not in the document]
fredrik-bakke: This alignment was modified by the `spaces_convention_simple.py` script. I'll quote the PR description:

> The commits (b47a80e0f16d32c17aed86ca33c79590e3668413 and 43cb2d90ebbf68b2bd668d9013370d61e6d950ce) are slightly more devious. They are based on a suggestion by Jonathan to disallow consecutive whitespace inside lines (i.e. after the first non-whitespace character) in Agda code. This is useful as such whitespace is usually a typo. However, this also disallows aligning code with whitespace. I think this is beneficial in the project, as artificially aligned code is less maintainable, and does not fit with the style of our library.

src/elementary-number-theory/well-ordering-principle-standard-finite-types.lagda.md
```diff
@@ -86,7 +86,10 @@ exists-not-not-forall-count {l1} {l2} {X} P p e =
   f nf f' =
     nf
       ( λ x →
-        tr P (htpy-eq-equiv (right-inverse-law-equiv (pr2 e)) x) (f' (map-inv-equiv (pr2 e) x)))
+        tr
+          ( P)
```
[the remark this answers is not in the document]
fredrik-bakke: It's not so difficult to implement an exception rule for this if you want. How many 1-character arguments can be allowed before a line break is warranted would you say?

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -31,7 +31,7 @@ postulated in
 [`foundation.function-extensionality`](foundation.function-extensionality.md) as
 `funext`.
 
-## Definition
+## Statement
```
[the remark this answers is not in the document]
fredrik-bakke: True. I changed the wording to make the distinction between _statement_ and _postulate_ more clear.

src/foundation-core/fundamental-theorem-of-identity-types.lagda.md
```diff
@@ -101,7 +101,7 @@ module _
           ( is-contr-retract-of (Σ _ (λ y → a ＝ y))
             ( pair (tot i)
               ( pair (tot λ x → pr1 (R x))
-                ( ( inv-htpy (tot-comp i (λ x → pr1 (R x)))) ∙h
+                ( ( inv-htpy (preserves-comp-tot i (λ x → pr1 (R x)))) ∙h
```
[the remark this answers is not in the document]
fredrik-bakke: Any other instances that come to mind? Which convention would you prefer? I'm not sure I like the `op1-op2` convention, as it reads like the actual composition of the two operations to me, and not a law about the composition.

src/foundation-core/universe-levels.lagda.md
```diff
@@ -6,8 +6,8 @@
 module foundation-core.universe-levels where
 
 open import Agda.Primitive
-  using (Level; lzero; lsuc; _⊔_)
-  renaming (Set to UU; Setω to UUω)
+  using (Level ; lzero ; lsuc ; _⊔_)
```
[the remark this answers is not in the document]
fredrik-bakke: Ah, I didn't consider this instance when I wrote the convention.
The idea was that with the extra spaces, the notation becomes more consistent with how we use `:` `,` `λ` `=` `→`, and so on.
The main case I had in mind was when semicolons are used in a pattern matching lambda definition.

src/elementary-number-theory/well-ordering-principle-standard-finite-types.lagda.md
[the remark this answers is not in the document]
EgbertRijke: Any number

src/foundation-core/universe-levels.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: I can revert this change if you want.

src/foundation/cantor-schroder-bernstein-escardo.lagda.md
```diff
@@ -28,12 +28,12 @@ open import foundation-core.universe-levels
...
 ## Idea
 
-The classical Cantor-Schröder-Bernstein theorem asserts that from any pair of
-injective maps `f : A → B` and `g : B → A` we can construct a bijection between
-`A` and `B`. In a recent generalization, Escardó proved that the
-Cantor-Schröder-Bernstein theorem also holds for ∞-groupoids. His generalization
-asserts that from given embeddings of two types into each other, we can
-construct an equivalence between them.
+The classical Cantor–Schröder–Bernstein theorem asserts that from any pair of
+injective maps $f : A → B$ and $g : B → A$ we can construct a bijection between
```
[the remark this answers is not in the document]
fredrik-bakke: VSCode has an inbuilt feature to easily compile markdown files... It's called the `Markdown: Open Preview` command. :)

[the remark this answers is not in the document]
fredrik-bakke: Also, about using unicode in the latex code, I just find that the raw code is more readable this way. And latex _should_ 🤞 treat it properly (in my opinion).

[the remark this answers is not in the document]
EgbertRijke: Unfortunately emacs users like me don't have access to this nice feature:)

src/foundation/dependent-paths.lagda.md
```diff
@@ -161,77 +207,134 @@ identities are not well typed. So these identities involve transporting.
...
     tr-path-over-path-over² B (assoc p01 p12 p23)
     (d-concat B (p01 ∙ p12) (d-concat B p01 q01 p12 q12) p23 q23)
     (d-concat B p01 q01 (p12 ∙ p23) (d-concat B p12 q12 p23 q23))
-    (d-assoc p01 q01 p12 q12 p23 q23)
-
-  d-right-unit : (p : a0 ＝ a1) (q : path-over B p b0 b1) →
-    path-over² B (right-unit {p = p}) (d-concat B p q refl (refl-path-over B a1 b1)) q
+    (d-associative p01 q01 p12 q12 p23 q23)
+
+  d-right-unit :
```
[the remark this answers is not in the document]
fredrik-bakke: To build on this, in`26-descent` and `26-id-pushout` the letter combination `dep` is used.

src/foundation/double-negation.lagda.md
```diff
@@ -40,8 +40,7 @@ of double negation.
 intro-dn : {l : Level} {P : UU l} → P → ¬¬ P
 intro-dn p f = f p
 
-map-dn : {l1 l2 : Level} {P : UU l1} {Q : UU l2} →
-  (P → Q) → (¬¬ P → ¬¬ Q)
+map-dn : {l1 l2 : Level} {P : UU l1} {Q : UU l2} → (P → Q) → (¬¬ P → ¬¬ Q)
```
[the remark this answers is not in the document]
fredrik-bakke: But then I assume we would also want to change notation for single negation...

src/foundation/dependent-paths.lagda.md
```diff
@@ -82,50 +88,83 @@ module _
...
-    pr1 (pr1 (is-equiv-inv-con (inv (ap (λ t → tr B t b0) α)) q0 q1))
-    (inv-con (inv (ap (λ t → tr B t b0) α)) q0 q1 (inv (tr-path-over B α q0) ∙ z) ∙ x ))
-    (right-inv (ap (λ t → t ∙ q1) (inv-inv (ap (λ t → tr B t b0) α))))) ∙
+    ( assoc
+      ( inv-con
+        ( inv (ap (λ t → tr B t b0) α))
+        ( q0)
+        ( q1)
+        ( inv (tr-path-over B α q0) ∙ z))
+    (ap (_∙ q1) (inv-inv (ap (λ t → tr B t b0) α)))
```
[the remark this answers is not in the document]
fredrik-bakke: Sorry, I just didn't want to deal with this file anymore. I'll fix it now though.

src/foundation/coproduct-decompositions.lagda.md
```diff
@@ -608,29 +646,33 @@ module _
...
-                   equiv-tot
-                     ( λ _ → extensionality-Fin 2 (inr star) (inr star)))) ∘e
-               ( ( right-distributive-Σ-coprod
-                   ( left-summand-binary-coproduct-Decomposition d)
-                   ( right-summand-binary-coproduct-Decomposition d)
-                   ( λ y →
-                     map-inv-equiv-binary-coproduct-Decomposition-map-into-Fin-Two-ℕ-helper
-                       d y ＝ inr star) )))))
+          ( map-equiv
+            ( ( ( left-unit-law-coprod
```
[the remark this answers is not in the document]
fredrik-bakke: nicely spotted!

src/foundation/fibered-maps.lagda.md
```diff
@@ -132,7 +136,8 @@ module _
       (λ g G → coherence-htpy-map-over m (g , G))
       (is-contr-total-htpy (map-total-map-over f g i m))
       (map-total-map-over f g i m , refl-htpy)
-      (is-contr-total-htpy (is-map-over-map-total-map-over f g i m ∙h refl-htpy))
+      (is-contr-total-htpy
+        ( is-map-over-map-total-map-over f g i m ∙h refl-htpy))
```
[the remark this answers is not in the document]
fredrik-bakke: Looks like that's it.

src/foundation/double-negation.lagda.md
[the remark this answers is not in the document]
EgbertRijke: I prefer my suggestion. Right now the naming scheme is using `dn`. This would look better if we replace all of them with `double-negation`. But some of them will start to look really ugly if we would replace `dn` with `¬¬`. So for a consistent naming scheme it will be better to use `double-negation`.

src/foundation-core/commuting-3-simplices-of-homotopies.lagda.md
```diff
@@ -48,16 +48,20 @@ module _
   coherence-3-simplex-homotopies : UU (l1 ⊔ l2)
   coherence-3-simplex-homotopies =
     ( upper-right ∙h
-      left-whisk-htpy-coherence-triangle-homotopies diagonal-up right upper-left) ~
+      left-whisk-htpy-coherence-triangle-homotopies
+        diagonal-up right upper-left) ~
```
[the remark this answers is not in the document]
fredrik-bakke: Just a remark. I find that generally, the extra parentheses add to the readability even in cases like this.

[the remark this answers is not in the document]
EgbertRijke: Sure, you can also add the parentheses:)

src/foundation/dependent-paths.lagda.md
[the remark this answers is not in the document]
EgbertRijke: I feel you! The amount of work you put into refactoring so many files is really impressive. Some of them I struggled with myself, and I didn't manage to come back to them yet. So I'm really grateful for this PR!

src/foundation/univalence.lagda.md
fredrik-bakke: Whoops! That was a mistake

src/foundation/univalence.lagda.md
fredrik-bakke: It is now `eq-equiv-id-equiv`.

src/orthogonal-factorization-systems/higher-modalities.lagda.md
fredrik-bakke: Yes, that's great! Thanks for the suggestion

src/set-theory/cardinalities.lagda.md
fredrik-bakke: Thanks! I added the following instead:


```md
### Assuming excluded middle we can show that `≤-cardinality` is a partial order

Using the previous result and assuming excluded middle, we can show
`≤-cardinality` is a partial order by showing that it is antisymmetric.
```

src/foundation/univalence.lagda.md
EgbertRijke: Here is an example of the `op1-op2` naming scheme:)

fredrik-bakke: Hahaha! I was about to say that I didn't like the new name. I'll change it to `compute-eq-equiv-id-equiv`

src/set-theory/cardinalities.lagda.md
EgbertRijke: It would be better to have

```md
### `≤-cardinality` is a partial order, assuming the law of excluded middle
```

or some other variation that doesn't involve "we can show", because that shouldn't really be in a section title.

src/foundation/cantor-schroder-bernstein-escardo.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Well, I'm not going to say anything about that :)

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
fredrik-bakke: The issue with that is that it did not adhere to the 80-character limit

EgbertRijke: Doesn't it only go over the 80-character limit if it is a single token?

Personally, I think well-structured code is more important than the 80-character limit.

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
fredrik-bakke: I think the file is in a good state now.

EgbertRijke: I'll have a look

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
fredrik-bakke: Sure, I can agree with that. However, my opinion is that it is better structured now than before.

src/species/cauchy-exponentials-species-of-types-in-subuniverses.lagda.md
EgbertRijke: It looks good to me now

src/foundation-core/0-maps.lagda.md
```diff
@@ -127,7 +127,7 @@ module _
     is-0-map-tot = is-trunc-map-tot zero-𝕋
 ```
 
-### For any type family over the codomain, a 0-map induces a 0-map on total spaces.
+### For any type family over the codomain, a 0-map induces a 0-map on total spaces
```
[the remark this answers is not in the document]
fredrik-bakke: Actually, are you sure this is what we want? It will display the `0` as if it is code, which sounds kind of odd to me.

src/foundation-core/singleton-induction.lagda.md
```diff
@@ -54,18 +54,18 @@ ind-singleton-is-contr :
 ind-singleton-is-contr a is-contr-A B b x =
   tr B ((inv (contraction is-contr-A a)) ∙ (contraction is-contr-A x)) b
 
-comp-singleton-is-contr :
+compute-singleton-is-contr :
```
EgbertRijke, line 57: Would this be better if it said `compute-ind-singleton-is-contr`?

src/foundation-core/0-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: This is why I think formatting with latex instead of code snippets would be better in some cases.

src/synthetic-homotopy-theory/26-descent.lagda.md
fredrik-bakke: Alright, then I won't address your remaining remarks about the headers in this file. I will say however that I think having references to the theorems somehow would be fine.

src/foundation-core/0-maps.lagda.md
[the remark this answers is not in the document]
EgbertRijke: For now, I think it is the most consistent option.

src/elementary-number-theory/well-ordering-principle-standard-finite-types.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Hmmmm, I don't like this description. If we allow arbitrary 1-character arguments, then why not a few 2-character arguments, and so on. Thus, either we should have a fixed limit, or just not allow them at all.

src/foundation/path-algebra.lagda.md
```diff
@@ -235,18 +241,37 @@ module _
...
-        ap (λ t → (horizontal-concat-Id² t (horizontal-inv-Id² α)) ∙ right-inv p') (ap-id α)
+    ( ( ( horizontal-concat-Id² refl (inv (ap-const refl α))) ∙
+        ( nat-htpy right-inv α)) ∙
+      ( horizontal-concat-Id²
+        ( ap-binary-comp-diagonal (_∙_) id inv α)
+        ( refl))) ∙
+        ap
+          ( λ t →
+            ( horizontal-concat-Id² t (horizontal-inv-Id² α)) ∙ right-inv p')
+          ( ap-id α)
```
EgbertRijke, line 258: I would format this as follows:

```md
( ( ( horizontal-concat-Id² refl (inv (ap-const refl α))) ∙
    ( nat-htpy right-inv α)) ∙
  ( horizontal-concat-Id²
    ( ap-binary-comp-diagonal (_∙_) id inv α)
    ( refl))) ∙
( ap
  ( λ t → horizontal-concat-Id² t (horizontal-inv-Id² α) ∙ right-inv p')
  ( ap-id α))
```

Note that in particular the `ap` identification at the end shouldn't be so deeply indented.

src/foundation/path-algebra.lagda.md
```diff
@@ -235,18 +241,37 @@ module _
...
-        ap (λ t → (horizontal-concat-Id² (horizontal-inv-Id² α) t) ∙ left-inv p') (ap-id α))
+    ( ( ( ( horizontal-concat-Id² refl (inv (ap-const refl α))) ∙
+          ( nat-htpy left-inv α)) ∙
+        ( horizontal-concat-Id²
+          ( ap-binary-comp-diagonal _∙_ inv id α)
+          ( refl))) ∙
+      ap
+        ( λ t → (horizontal-concat-Id² (horizontal-inv-Id² α) t) ∙ left-inv p')
+        ( ap-id α))
 ```
```
EgbertRijke, line 275: I would format this as follow:

```md
( ( ( horizontal-concat-Id² refl (inv (ap-const refl α))) ∙
    ( nat-htpy left-inv α)) ∙
  ( horizontal-concat-Id²
    ( ap-binary-comp-diagonal _∙_ inv id α)
    ( refl))) ∙
( ap
  ( λ t → (horizontal-concat-Id² (horizontal-inv-Id² α) t) ∙ left-inv p')
  ( ap-id α))
```

(also this term looks suspiciously similar to the previous term. Probably something can be generalized and both are an instance of the same construction. That's for another day to figure out, though)

src/foundation/univalence.lagda.md
```diff
@@ -108,28 +138,38 @@ eq-equiv-fam {B = B} {C} = map-inv-is-equiv (is-equiv-equiv-eq-fam B C)
 
 ## Properties
```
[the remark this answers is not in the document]
fredrik-bakke: 🤦‍♂️

fredrik-bakke:
I think I've adequately addressed all of your comments at this point, so if you are happy with the code in its current state, I think it is ready to be merged.
There is just one comment left that I haven't marked as resolved, which is with regard to exceptions to the line character limit. If you still want such an exception, maybe we can implement it later?

EgbertRijke:
> I think I've adequately addressed all of your comments at this point, so if you are happy with the code in its current state, I think it is ready to be merged.
> There is just one comment left that I haven't marked as resolved, which is with regard to exceptions to the line character limit. If you still want such an exception, maybe we can implement it later?

Sounds good! Thank you for all your work! This PR is very nice!

PR #582, opened 2023-05-01
Define the smash product of pointed types

fredrik-bakke:
### Summary
- Define the smash product of pointed types and derive some basic properties
- Small additions to pointed types
- Replace `pt` with `point` in general.


EgbertRijke:
I think you need to clean up your local git history. Otherwise this is unreviewable:)

fredrik-bakke:
Working on it! your previous tip of doing `git reset --hard agda-unimath/master` didn't work :(

EgbertRijke:
> Working on it! your previous tip of doing `git reset --hard agda-unimath/master` didn't work :(

Don't do that in branches where you have any work. Even committed work will be thrown away by a `git reset --hard`

PR #583, opened 2023-05-01
Define the smash product of pointed types and refactor pointed types

fredrik-bakke:
### Summary
- Add a module on the pointed unit type.
- Add a module on pushouts of pointed types.
- Add a module on cocones under spans of pointed types.
- Define the smash product of pointed types and derive some basic properties.
- Redefine the wedge sum of pointed types in terms of pushouts of pointed types.
- Small additions to pushouts and misc.
- Replace `pt` with `point` in general.
- Remove duplicate definition of `ev-point`.
- Add an optimization to `remove_unused_imports.py` by prioritizing the files in order of most recently changed.
- Add textual names to the different pointed constructions.
- ~Change the notation for the join from `_*_` to `_⋆_`, as the asterisk is already quite heavily overloaded.~
- Add notes on symbol usage.
- Change notation for pointed type constructions to using the asterisk operator `∗` instead of the asterisk `*`.

### DIFF
--- a/src/group-theory/group-solver.lagda.md
+++ b/src/group-theory/group-solver.lagda.md
@@ -7,7 +7,21 @@ module group-theory.group-solver where
 <details><summary>Imports</summary>
 
 ```agda
+open import elementary-number-theory.natural-numbers
 
+open import foundation.coproduct-types
+open import foundation.decidable-types
+open import foundation.identity-types
+open import foundation.universe-levels
+
+open import group-theory.groups
+
+open import linear-algebra.vectors
+
+open import lists.concatenation-lists
+open import lists.functoriality-lists
+open import lists.lists
+open import lists.reversing-lists
 ```
 
 </details>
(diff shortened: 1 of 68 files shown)


fredrik-bakke:
Here we go! I really need to learn how to rebase properly some time. 😅

EgbertRijke:
Would `smash-product` be nicer for a naming scheme than `smash-prod`?

fredrik-bakke:
I guess, but `prod` is already very prevalent in the library...

EgbertRijke:
> Here we go! I really need to learn how to rebase properly some time. 😅

If you figure out how to do it properly, let me know:)

src/structured-types/pointed-unit-type.lagda.md
```diff
@@ -0,0 +1,43 @@
...
+## Properties
+
+```agda
+gap-pointed-unit : {l : Level} (X : Pointed-Type l) → X →* pointed-unit
+pr1 (gap-pointed-unit X) _ = star
+pr2 (gap-pointed-unit X) = refl
+
+cogap-pointed-unit : {l : Level} (X : Pointed-Type l) → pointed-unit →* X
+pr1 (cogap-pointed-unit X) _ = point-Pointed-Type X
+pr2 (cogap-pointed-unit X) = refl
```
fredrik-bakke, line 42: Do you have a better name suggestion for these maps?

EgbertRijke: `pointed-unit` -> `unit-Pointed-Type`

fredrik-bakke: Thanks, that's a lot better.

EgbertRijke:
For notation in agda-unimath I usually go by one of the following two options:
- Either go by the standard notation
- Use a descriptive name
I don't like to make up notation too much. We make up enough notation already if we "don't" do it.

fredrik-bakke:
I understand. My concern is that very many things in the library use the asterisk. E.g. pointed types, groups, and multiplication. Moreover the star operator _is_ used for the join in some places. For instance [nlab:join of simplicial spaces](https://ncatlab.org/nlab/show/join+of+simplicial+sets), [nlab:join of topological spaces](https://ncatlab.org/nlab/show/join+of+topological+spaces).

EgbertRijke:
But I'd still say that if all the time we choose to employ substandard notation in the library, by the argument that the standard notation is already used somewhere else, then we will quickly get a library riddled with strange-looking symbols. I would like to prevent that, if possible.

fredrik-bakke:
Actually, come to think of it, it would make sense to use the asterisk operator instead of the asterisk for pointed types.
Just compare the following two expressions
```agda
(A ∨∗ B) →∗ (A ×∗ B)
```
as opposed to the current
```agda
(A ∨* B) →* (A ×* B)
```
This is more in line with the "standard" notation for pointed things, which usually uses a subscript asterisk in my experience.

EgbertRijke:
> Actually, come to think of it, it would make sense to use the asterisk operator instead of the asterisk for pointed types. Just compare the following two expressions
> 
> ```agda
> (A ∨∗ B) →∗ (A ×∗ B)
> ```
> 
> as opposed to the current
> 
> ```agda
> (A ∨* B) →* (A ×* B)
> ```
> 
> This is more in line with the "standard" notation for pointed things, which usually uses a subscript asterisk in my experience.

How did you get the subscript asterisks?

fredrik-bakke:
`\ast`

EgbertRijke:
> \ast

I agree, it looks better!

src/group-theory/group-solver.lagda.md
```diff
@@ -19,67 +33,95 @@ way and removes units and inverses next to the original.
...
-data Fin : ℕ → UU where zero-Fin : ∀ {n} → Fin (succ-ℕ n) succ-Fin : ∀ {n} → Fin
-n → Fin (succ-ℕ n)
-
-finEq : ∀ {n} → (a b : Fin n) → is-decidable (Id a b) finEq zero-Fin zero-Fin =
-inl refl finEq zero-Fin (succ-Fin b) = inr (λ ()) finEq (succ-Fin a) zero-Fin =
-inr (λ ()) finEq (succ-Fin a) (succ-Fin b) with finEq a b ... | inl eq = inl (ap
-succ-Fin eq) ... | inr neq = inr (λ { refl → neq refl})
-
-getVec : ∀ {n} {l} {A : UU l} → vec A n → Fin n → A getVec (x ∷ v) zero-Fin = x
+```agda
```
fredrik-bakke, line 36: This opening tag seemed to be missing. Also, I had to comment out some code due to the definition of `in-list` being missing.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,128 @@
...
+  pr1 (inl-pushout-Pointed-Type f g) =
+    inl-pushout (map-pointed-map S A f) (map-pointed-map S B g)
+  pr2 (inl-pushout-Pointed-Type f g) = refl
+
+  inr-pushout-Pointed-Type :
+      (f : S →∗ A) (g : S →∗ B) → B →∗ pushout-Pointed-Type f g
+  pr1 (inr-pushout-Pointed-Type f g) =
+    inr-pushout (map-pointed-map S A f) (map-pointed-map S B g)
+  pr2 (inr-pushout-Pointed-Type f g) =
+    equational-reasoning
```
EgbertRijke, line 64: If we need to have any control over this specific equation, which I imagine we might need at some point, then equational reasoning isn't the best option.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,128 @@
...
+
+Given a span of pointed maps, then the pushout is in fact a pushout diagram in
+the category of pointed types.
+
+## Definition
+
+```agda
+module _
+  {l1 l2 l3 : Level}
+  {S : Pointed-Type l1} {A : Pointed-Type l2} {B : Pointed-Type l3}
```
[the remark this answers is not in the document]
fredrik-bakke: Hmmm I think so. The pointed maps will have to carry the information of which points their domain and codomain is pointed at for the `preserves-point-pointed-map` witness to make sense I think.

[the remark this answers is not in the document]
fredrik-bakke: Also, it seems to be working for the wedge and smash product currently at the very least. Let's change it later, if it turns out not to be sufficient.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,128 @@
...
+      ( map-pointed-map S B g)
+      ( point-Pointed-Type A)
+```
+
+## Properties
+
+### The pushout of a pointed map along a pointed map is pointed
+
+```agda
+  inl-pushout-Pointed-Type :
```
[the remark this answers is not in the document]
fredrik-bakke: I'm with you on that.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
fredrik-bakke: A side remark: if we were to define the pushout to be pointed at `inr` instead, we could use one less `inv` in this proof.

src/synthetic-homotopy-theory/cocones-under-spans-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+
+  vertical-pointed-map-cocone-Pointed-Type : B →∗ X
+  vertical-pointed-map-cocone-Pointed-Type = pr1 (pr2 c)
+
+  vertical-map-cocone-Pointed-Type :
+    type-Pointed-Type B → type-Pointed-Type X
+  vertical-map-cocone-Pointed-Type =
+    pr1 vertical-pointed-map-cocone-Pointed-Type
+
+  coherence-square-cocone-Pointed-Type :
```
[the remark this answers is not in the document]
fredrik-bakke: Aaaaahh, so if you have a coherence square

```md
A --i--> B
|        |
f        g
v        v
X --j--> Y
```
with pointedness preservation proofs `F, G, I, J` and unpointed coherence square proof `H`, then you want an additional coherence proof of type

```agda
ap j F ∙ inv (ap g I) ＝ H a
```
?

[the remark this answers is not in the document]
EgbertRijke: I think it we say that a pointed type is a pointed pushout, then that would be expected, yes. But there should also be a theorem that a pointed type is a pointed pushout if and only if its underlying type is an ordinary pushout. And the same for pullbacks. You'll surely need those theorems at some point, along with the pointed universal property of pointed pushouts, and so on, but it might be a lot of work so it doesn't need to happen in this PR.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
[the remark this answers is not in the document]
EgbertRijke: The reason why we will probably need to make them explicit is that our Sigma-types don't have eta-equality. I.e., not every element of a Sigma-type is definitionally equal to a pair.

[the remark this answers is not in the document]
EgbertRijke: On the other hand, if it seems to work, then why not have them implicit?:)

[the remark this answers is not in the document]
fredrik-bakke: Hmmm, let's fix it when/if it becomes an issue 🙃

src/synthetic-homotopy-theory/cocones-under-spans-of-pointed-types.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: I realize now that this is precisely the purpose of pointed homotopies. Thanks for pointing it out!

src/group-theory/group-solver.lagda.md
EgbertRijke: The group solver broke because at some point I changed things in lists, and while I was able to fix the other parts of the library that used lists, I wasn't able to fix the group solver. The code here is not at the standards of our library, and I've been unable to understand it.

fredrik-bakke: Alright. At least this way it's properly formatted markdown. Let's just leave it like this 😅

src/synthetic-homotopy-theory/cocones-under-spans-of-pointed-types.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Okay, I've switched to coherence squares of pointed types now. Let me know if this is good.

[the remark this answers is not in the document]
EgbertRijke: It's very good!

fredrik-bakke:
Sorry, I see I forgot to `make pre-commit`. However, I figured out how to do `compute-glue-cogap`, so I'll add that before wrapping up the PR later today.

fredrik-bakke:
And by later today I meant right now apparently 😅

src/synthetic-homotopy-theory/cocones-under-spans-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,115 @@
...
+
+  type-cocone-Pointed-Type :
+    {l4 : Level} → Pointed-Type l4 → UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  type-cocone-Pointed-Type X =
+    Σ ( A →∗ X)
+      ( λ i →
+        Σ ( B →∗ X)
+          ( λ j → coherence-square-pointed-maps g f j i))
+
+  cocone-Pointed-Type :
```
EgbertRijke, line 48: Is the idea behind this definition that `map-cocone` becomes a pointed map?

fredrik-bakke: The idea is that cocones in the category of pointed types is a pointed object

src/synthetic-homotopy-theory/pushouts.lagda.md
```diff
@@ -89,3 +84,68 @@ is-pushout :
...
+        ( _∙ inv (compute-inr-cogap f g c (g s)))
+        ( pr2
+          ( pr2
+            ( htpy-cocone-map-universal-property-pushout
+              ( f)
+              ( g)
+              ( cocone-pushout f g)
+              ( up-pushout f g)
+              ( c)))
+          ( s)))))
```
fredrik-bakke, line 150: I found a slicker argument for this, give me a couple of minutes

fredrik-bakke: Okay done! 😄

src/synthetic-homotopy-theory/pushouts.lagda.md
```diff
@@ -128,24 +128,17 @@ compute-glue-cogap :
...
-        ( pr2
-          ( pr2
-            ( htpy-cocone-map-universal-property-pushout
-              ( f)
-              ( g)
-              ( cocone-pushout f g)
-              ( up-pushout f g)
-              ( c)))
-          ( s)))))
+  con-inv
```
EgbertRijke, line 131: Oh yeah `con-inv` can be a life saver:)

PR #584, opened 2023-05-02
Cavallo's trick

EgbertRijke:
This PR adds Cavallo's trick to the library.


fredrik-bakke:
Oh, that's neat. Thanks for adding this!

PR #586, opened 2023-05-04
The product of precategories

FernandoChu:
Small PR. note that we already have `product-Precat` for the product of two objects, maybe that should be renamed?

Btw my hands are bad, I won't contribute for a while, sorry for that.


EgbertRijke:
Excellent! Thanks so much for this addition, Fernando!

PR #590, opened 2023-05-04
VSCode Literate Agda Markdown file association

fredrik-bakke:
Sets up file association with literate agda markdown files, to accommodate the newest `agda-syntax` version which defines this file format.


EgbertRijke:
Cool! Thank you so much for doing this. I know you've been busy!

PR #594, opened 2023-05-05
Big cleanup throughout library

EgbertRijke:
- Moving files to more natural folders
- Renaming files
- Rearranging the contents of a file over other files, if that is more natural
- Precat -> Precategory


fredrik-bakke:
Since you elaborated `Precat` to `Precategory`, why not also elaborate `Cat` to `Category`.

EgbertRijke:
It was getting late yesterday, so I just merged it and went to sleep:)

PR #601, opened 2023-05-07
Fix: TOC small tweaks

jonaprieto:
TOC:

- [x] Add an extra margin on the right
- [x] Bigger font size

Now:

image

Before:

image


EgbertRijke:
Beautiful!

PR #606, opened 2023-05-08
Some half-finished additions to modalities

fredrik-bakke:
Since the library is currently undergoing a makeover, I figured I should push what finished work I had to avoid a larger merge hell later on. Also contains some corrections which I thought would be good to have sooner rather than later.

### Summary
- Define Σ-closed reflective subuniverses
- Define `Y`-null types
  I've defined these as being types such that `const Y A : A → (Y → A)` is an equivalence and shown this is equivalent to being local at the terminal projection of `Y`. Would you rather have this as the definition?
- Define stable orthogonal factorization operations
- Refactor factorization operations
- Prove that being uniquely eliminating is a property
- Prove `○ X` is modal for uniquely eliminating modalities
- Add a start of an alternative characterization of identifications for higher modalities
- Correct references to equivalent notions of higher modalities


src/orthogonal-factorization-systems/modal-operators.lagda.md
```diff
@@ -160,9 +161,9 @@ every term of `○ X` and for every family `P` over `X` equipped with a section
 conventional terminology.
 
 ```agda
-is-Σ-closed-modal-operator :
-  {l1 l2 : Level} → modal-operator l1 l2 → UU (lsuc l1 ⊔ l2)
-is-Σ-closed-modal-operator {l1} ○ =
+is-Σ-closed-operator-modality :
```
EgbertRijke, line 164: Wouldn't it make more sense to move this definition to `dependent-pair-closed-reflective-subunvierses`?

src/orthogonal-factorization-systems/function-class-factorization-operations.lagda.md
```diff
@@ -0,0 +1,103 @@
...
+
+```agda
+module _
+  {l1 l2 lF lL lR : Level}
+  (L : function-class l1 lF lL)
+  (R : function-class lF l2 lR)
+  (A : UU l1) (B : UU l2)
+  where
+
+  function-class-factorization-operation : UU (l1 ⊔ l2 ⊔ lsuc lF ⊔ lL ⊔ lR)
```
EgbertRijke, line 46: It would be more fitting with our naming scheme to call this `factorization-operation-function-class`, but I'm thinking that perhaps we could simply call it `factorization-operation`

The next definitions should then be either `mere-factorization-property-function-class` or `mere-factorization-property`, and `unique-factorization-property-function-class` or `unique-factorization-property`.

fredrik-bakke: This conflicts with factorization operations without a specified left and right class as defined in `orthogonal-factorization-systems.factorization-operations`, but I think it is fine to specify with a `function-class` suffix.

fredrik-bakke: Then again, `is-function-class-factorization` and `function-class-factorization` reads better to me than `is-factorization-function-class` and `factorization-function-class`, which reads as function classes with extra properties...

fredrik-bakke: By the way, what do you think of the naming scheme I'm using for universe levels? As far as I'm aware, it's not used many places in the library. However, it's a bit easier to use when writing code in my experience.

fredrik-bakke: Do you still want `is-factorization-function-class` and so on, or can we keep it like it is?

fredrik-bakke: Alright, done! Would you also want to change the module name to
`orthogonal-factorization-systems.factorization-operations-function-classes` to reflect this?
In this case we can change the title of the file to something like
```md
# Factorization operations into function classes
```

EgbertRijke: Sure!
After that I'll merge it

PR #610, opened 2023-05-09
large galois connections, nuclei, order preserving maps, and update o…

EgbertRijke:
…f locales to include explicit top element

### DIFF
--- a/src/foundation/large-locale-of-propositions.lagda.md
+++ b/src/foundation/large-locale-of-propositions.lagda.md
@@ -12,6 +12,7 @@ open import foundation.existential-quantification
 open import foundation.functions
 open import foundation.propositional-extensionality
 open import foundation.propositions
+open import foundation.unit-type
 open import foundation.universe-levels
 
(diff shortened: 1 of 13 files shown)


src/foundation/large-locale-of-propositions.lagda.md
```diff
@@ -61,6 +63,31 @@ is-greatest-binary-lower-bound-meet-has-meets-Large-Poset
...
+### The largest element in the large poset of propositions
+
+```agda
+has-largest-element-Prop-Large-Locale :
+  has-largest-element-Large-Poset Prop-Large-Poset
+top-has-largest-element-Large-Poset
+  has-largest-element-Prop-Large-Locale = unit-Prop
+is-largest-element-top-has-largest-element-Large-Poset
+  has-largest-element-Prop-Large-Locale P p =
+  star
```
fredrik-bakke, line 75: can `has-greatest-element` be used instead?

fredrik-bakke:
You're just blazing through with your additions! 🏎️💨

PR #618, opened 2023-05-13
Define binary operator notations for `mul-ℕ`, `mul-ℤ`, and `exp-ℕ`.

fredrik-bakke:
They are `_*ℕ_`, `_*ℤ_`, and `_^ℕ_` respectively.

In addition to what the title says, I've also refactored the files
- `multiplication-natural-numbers`
- `multiplication-integers`
- `exponentiation-natural-numbers`
- `absolute-value-integers`

to use the binary operators where it's natural, to get a sense for how this would look. It should be possible to change most usages easily enough* with a nice enough regex, but I don't want to do this before I get some feedback on whether it is preferable.

Also, pairs like `is-injective-mul-succ-ℕ` and `is-injective-mul-succ-ℕ'` should perhaps be renamed to something like `is-injective-left-mul-succ-ℕ` and `is-injective-right-mul-succ-ℕ`.

### DIFF
--- a/src/elementary-number-theory/absolute-value-integers.lagda.md
+++ b/src/elementary-number-theory/absolute-value-integers.lagda.md
@@ -179,19 +179,19 @@ is-nonzero-abs-ℤ (inr (inr x)) H = is-nonzero-succ-ℕ x
 
 ```agda
 multiplicative-abs-ℤ' :
-  (x y : ℤ) → abs-ℤ (explicit-mul-ℤ x y) ＝ mul-ℕ (abs-ℤ x) (abs-ℤ y)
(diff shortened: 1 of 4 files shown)


EgbertRijke:
I think it is very nice! Would you like to go ahead with the larger search/replace, or would you like me to merge this one before you do that?

fredrik-bakke:
Thanks! I guess I'll do it in a second PR

EgbertRijke:
I also like the suggestion of renaming injectivity on the left/(right) to `is-injective-left-mul-succ-ℕ`. It is indeed more clear.

PR #620, opened 2023-05-13
Refactor to use infix binary operators for arithmetic

fredrik-bakke:
- Refactor to use infix binary operators for arithmetic.
- Define infix binary operators for arithmetic operations on the Eisenstein integers, Gaussian integers,half integers,  and truncation levels.
- Swap to using left/right instead of a `'` for different laws for binary arithmetic operators.
- Some additional refactoring for Eisenstein integers and Gaussian integers

Note that the non-infix variants of the operators are still used some places, matching how `Id` and `pair` are used.


EgbertRijke:
Very nice!

fredrik-bakke:
Someone should format the Bezout's lemma files properly someday...

PR #621, opened 2023-05-13
Remove unused imports and fix some unaddressed comments

fredrik-bakke:
Sorry, my script wasn't finished when I made the last commit in #620, and here are the remaining unused imports. This PR also fixes some comments I made earlier that went unaddressed.

I'm wondering, are the imports scripts interfering with people's merging process a lot? If so, we don't have to merge those changes now.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -18,8 +18,8 @@
       "problemMatcher": []
     },
     {
-      "label": "imports",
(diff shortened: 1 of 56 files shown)


EgbertRijke:
I think it's ok to merge this. There weren't many merge conflicts for my students even when we refactored a lot a week or so ago.

fredrik-bakke:
Good to hear!

PR #623, opened 2023-05-15
Beyond finite sets

EgbertRijke:
In this PR I add some concepts that came up during the Beyond finite sets workshop.

1. Omega-finite types

### DIFF
--- a/src/univalent-combinatorics/omega-finite-types.lagda.md
+++ b/src/univalent-combinatorics/omega-finite-types.lagda.md
@@ -1,0 +1,72 @@
+# ω-Finite types
+
+```agda
+module univalent-combinatorics.omega-finite-types where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.contractible-types
+open import foundation.dependent-pair-types
+open import foundation.empty-types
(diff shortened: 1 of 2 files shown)


src/univalent-combinatorics/omega-finite-types.lagda.md
fredrik-bakke, line 1: Could "ω-Finite types" as you define them here also faithfully be called "profinite types"?

EgbertRijke: You can only call something a pro-object if it is an object of the procompletion of your category, in this case types. Pro types should have many nice properties which omega-finite types don’t have, such as being closed under retracts.

In general in Agda-unimath, we shouldn’t try to be too fancy with terminology. When we define a concept, it should be obvious that it is defined in the expected way.

fredrik-bakke: You mean the procompletion of finite types and maps of finite types, right?

fredrik-bakke: My comment was not meant as a suggestion, but merely as a question arising from curiosity.

PR #627, opened 2023-05-21
Pre-commit stuff

fredrik-bakke:
Tried out a bunch of ideas while I was on the plane, this is mostly a result of that.

### Summary
- Add a script `markdown_conventions.py` that enforces some markdown conventions. This script
  - Prohibits completely empty sections (I.e. when a header is immediately succeeded by a header of the same or less deep level). Often this is a mistake, but if a section is empty because it is left for future work, I suggest noting this with a sentence like "This remains to be shown." instead of leaving the section empty.
  - Prohibits top-level headers after the first line
- Fix current violators of these conventions.
- Try out an `indentation_conventions.py` script that currently only detects unevenly indented code. I.e. when code is indented with an odd number of spaces. There are about 1000 violations at the moment so the script does not run by default, but it is something we can consider enabling in the future.
- Rename some pre-commit scripts
- Define infix binary operator `_-ℤ_` for the difference of integers, and use it where possible.


EgbertRijke:
Awesome!

fredrik-bakke:
I also wanted to add code to blank_line_conventiond that ensures a blank line after a module declaration, but I couldn't get it to work properly. Will have to look it again some other time.

EgbertRijke:
Yeah, I think it's ok to ban them

jonaprieto:
Wild! Cool. I think people can benefit from some of these hooks in their own projects. Maybe we can consider documenting how.

PR #632, opened 2023-05-26
Yoneda

emilyriehl:
In two separate files we

(i) defined covariant representable functors for precategories and representable natural transformations between them
(ii) proved the Yoneda lemma for precategories.


emilyriehl:
I messed around with my local version of python and now the latest commit on this branch passes make pre-commit. Can I add a new commit to an open pull request? Or should I delete this and start over? (Though I've forgotten the steps to make a pull request.)

VojtechStep:
A pull request is associated with a branch, so as long as you keep pushing to the same branch, the PR will be automatically updated. If you're talking about the commit titled "passed make pre-commit", then that's already included in the PR.

src/category-theory/representable-functors-precategories.lagda.md
```diff
@@ -0,0 +1,65 @@
...
+```agda
+open import category-theory.functors-precategories
+open import category-theory.natural-transformations-precategories
+open import category-theory.precategories
+
+open import foundation.dependent-pair-types
+open import foundation.function-extensionality
+open import foundation.universe-levels
+
+open import foundation-core.homotopies
```
VojtechStep, line 18: Modules outside of `foundation` and `foundation-core` usually don't depend directly on `foundation-core` - every module from core has a corresponding module in just `foundation`, which re-exports it, and possibly adds more on top.

If there is no compelling reason to depend on `foundation-core` specifically, I would suggest changing this to `open import foundation.homotopies`.

src/category-theory/yoneda-lemma-precategories.lagda.md
```diff
@@ -0,0 +1,99 @@
...
+open import category-theory.natural-transformations-precategories
+open import category-theory.precategories
+open import category-theory.representable-functors-precategories
+
+open import foundation.dependent-pair-types
+open import foundation.function-extensionality
+open import foundation.sets
+open import foundation.universe-levels
+
+open import foundation-core.equality-dependent-pair-types
```
VojtechStep, line 20: Same thing here - if possible, these should probably all be imports from `foundation`.

fredrik-bakke:
Thank you for your contribution! I took the liberty to make some formatting fixes to your PR. You can have a look at my commits for details.

emilyriehl:
Thanks everyone. Could someone explain the difference between foundation and foundation-core, so I'll know next time?

EgbertRijke:
I explained it on discord, because I'm not sure how easy it is to find messages on closed PRs

VojtechStep:
Sure thing! Basically `foundation-core` is an implementation detail - the objective is to have the library split into modules by topic, with one module for some "foundational" stuff. However at this low a level, one would often encounter cyclic dependencies, and these "pre-foundation" things are put into `foundation-core`. The last three paragraphs in the Design principles sum it up quite well IMO: https://unimath.github.io/agda-unimath/DESIGN-PRINCIPLES.html.

The general rule of thumb is to pretend that `foundation-core` doesn't exist. I would even go as far as to say that we should have a pre-commit hook to check that it's never explicitly imported outside of itself and `foundation`.

PR #635, opened 2023-05-28
Enforce even indentation and automate some conventions

fredrik-bakke:
Wraps up the last PR on pre-commit stuff.

### Summary
- add detection for empty sections at the end of a file
- add autoremoval of empty code blocks
- add autoremoval of punctuation at the end of headers
- fix all uneven indentation in the library
- enable enforcing of even indentation
- pre-commit now properly fails when a misplaced top-level header is detected

### DIFF
--- a/src/univalent-combinatorics/orientations-complete-undirected-graph.lagda.md
+++ b/src/univalent-combinatorics/orientations-complete-undirected-graph.lagda.md
@@ -364,23 +364,23 @@ module _
                       ( 2-Element-Decidable-Subtype-subtype-pointwise-difference
                           d2 d3))
                     ( is-finite-2-Element-Decidable-Subtype n X)))
-               ( pair
-                 ( number-of-elements-is-finite
-                   ( is-finite-type-decidable-subtype
-                     ( 2-Element-Decidable-Subtype-subtype-pointwise-difference
-                        d1 d3)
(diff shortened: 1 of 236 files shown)


src/univalent-combinatorics/orientations-complete-undirected-graph.lagda.md
```diff
@@ -2439,136 +2439,137 @@ module _
...
+                    ( map-inv-equiv
+                      ( transposition
+                        ( standard-2-Element-Decidable-Subtype
+                          ( has-decidable-equality-count eX)
+                          ( np)))))
+                  ( x)))
+            ( P)
+            ( inl
               ( inv
-                ( right-computation-standard-transposition
```
fredrik-bakke, line 2555: Did I override a renaming here?

EgbertRijke: I don't know, but it looks like `right-computation-standard-transposition` and `is-fixed-point-standard-transposition` do the same thing, which means that they should be merged.

EgbertRijke:
How confident are you that you increased/decreased the indentation level of the oddly indented code correctly?:)

fredrik-bakke:
I did go through all the changes a second time, but I won't guarantee that everything is correct now. The main goal wasn't to make everything correct anyways, but to avoid introducing more mistakes later.

PR #642, opened 2023-06-05
Move public imports before "Imports" block

fredrik-bakke:
### Summary
- Move public imports before "Imports" block
- Make start of file formatting consistent
- Update file conventions
- make nonpublic two rogue public imports that should not have been
- Prohibit public imports inside "Imports" block as part of the `fix_imports.py` hook

### DIFF
--- a/src/elementary-number-theory/decidable-types.lagda.md
+++ b/src/elementary-number-theory/decidable-types.lagda.md
@@ -7,8 +7,6 @@ module elementary-number-theory.decidable-types where
 <details><summary>Imports</summary>
 
 ```agda
-open import foundation.decidable-dependent-pair-types public
-
 open import elementary-number-theory.inequality-natural-numbers
 open import elementary-number-theory.natural-numbers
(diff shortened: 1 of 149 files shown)


src/elementary-number-theory/decidable-types.lagda.md
```diff
@@ -7,15 +7,14 @@ module elementary-number-theory.decidable-types where
 <details><summary>Imports</summary>
 
 ```agda
-open import foundation.decidable-dependent-pair-types public
```
fredrik-bakke, line 10: This import statement looked like it should not have been here

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -2,17 +2,13 @@
...
-```
 
-```agda
 module foundation.vectors-set-quotients where
 ```
 
 <details><summary>Imports</summary>
 
 ```agda
-open import foundation-core.identity-types public
```
fredrik-bakke, line 14: This public import statement looked like it should not have been here

PR #650, opened 2023-06-10
cleaning up transport and dependent identifications files

EgbertRijke:
- correcting typos in the tables of files related to identity types
- Moving facts about transport from `foundation.identity-types` to `foundation.transport`
- Renaming dependent paths to dependent identifications
- Renaming path-overs to dependent identifications
- Factoring out the computations of transport to their specific files, eg transport in cartesian products to cartesian products, transport in identity types to identity types, and so on.
- Moving `functions` to `function-types` to more clearly support both operations on function types as well as other lemmas about function types.
- Cleaning up dependent identifications file. There is still some conceptual work to be done here. Refactoring this is highly nontrivial.


src/foundation/dependent-pair-types.lagda.md
```diff
@@ -19,58 +26,27 @@ When `B` is a family of types over `A`, then we can form the type of pairs
...
-  (a : A) (b : B a) → C (pair a b) → Σ (Σ A B) C
-pr1 (pr1 (triple' a b c)) = a
-pr2 (pr1 (triple' a b c)) = b
-pr2 (triple' a b c) = c
+tr-Σ :
+  {l1 l2 l3 : Level} {A : UU l1} {a0 a1 : A} {B : A → UU l2}
+  (C : (x : A) → B x → UU l3) (p : a0 ＝ a1) (z : Σ (B a0) (λ x → C a0 x)) →
+  tr (λ a → (Σ (B a) (C a))) p z ＝
+  pair (tr B p (pr1 z)) (tr (ind-Σ C) (eq-pair-Σ p refl) (pr2 z))
+tr-Σ C refl z = refl
```
fredrik-bakke, line 39: Are you trying to move away from having a separate `equality-dependent-pair-types` file (and so on)? Otherwise, I believe this would fit better there.

EgbertRijke: Excellent point! No I just didn't think of putting it there, but it makes a lot of sense

PR #649, opened 2023-06-09
Cleaning up synthetic homotopy theory

EgbertRijke:
- Remove `24-pushouts.lagda.md`
- Add links to `cofiber.lagda.md`
- Add explanatory text in `pushouts.lagda.md`
- Add explanatory text in `universal-property-pushouts.lagda.md`
- Factoring out dependent cocones under spans
- Factoring out the pullback property of pushouts
- Factoring out the dependent universal property of pushouts
- Factoring out the induction principle of pushouts
- Factoring out the dependent pullback property of pushouts
- Moving a fact about `eq-htpy` to `function-extensionality`
- Factoring out constant type families
- Factoring out transport
- Factoring out action on identifications of functions
- Factoring out action on identifications of dependent functions
- Factoring out action on identifications of binary functions
- Factoring out path-overs

PR #654, opened 2023-06-11
Refactor synthetic homotopy theory

EgbertRijke:
- Hyperlinks throughout `synthetic-homotopy-theory`
- Using `coherence-square-identifications` and `dependent-identifications` in `interval`
- Changing the order of arguments of `coherence-square-identifications` to be the same as for maps

This PR builds on #649 and #650, and works toward progressing item 8 of #195.


src/foundation-core/function-extensionality.lagda.md
```diff
@@ -43,6 +40,48 @@ module _
...
 
+## Postulate
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2}
+  where
+
+  postulate funext : (f : (x : A) → B x) → FUNEXT f
+```
```
fredrik-bakke, line 51: I see you reverted an earlier change of mine. I'm not sure I disagree with this, I'm just sad there isn't an easy way to exclude certain axioms and their consequences. I.e. if one could make the type checker error if an axiom is invoked in a certain file. TypeTopology resolves this in a different way by defining axioms via variables instead of as postulates, although I don't think this is the best solution either as it clutters the context.

EgbertRijke: I needed to be able to import files that use function extensionality in foundation/function-extensionality. Having the postulate in foundation was leading to cyclic dependencies. So I moved it to foundation-core. It was just a practical consideration, and I apologize for reverting your change. I definitely don't want to make you sad:)

fredrik-bakke: You are definitely not making me sad! It's more that the limitation of the language makes libraries in general less useful to people. For instance, if I don't want to subscribe to univalence, `agda-unimath` suddenly becomes very hard to use. I think I'm looking for something like a pragma
```agda
{-# SUSPEND univalence #-}
```
that would make the type checker complain if I somehow inadvertently invoke univalence through a lemma I'm applying in the file. Maybe I can talk to the Agda wizards about this at some point.

fredrik-bakke: My suggestion feels a bit hacky though, so I'm not sure this is a good approach.

EgbertRijke: It is true that for agda-unimath it was an early design decision that it shouldn't be agnostic about foundations, but committed to univalent foundations. For me, this library is about univalent mathematics. There is a development of the foundations of mathematics in the library, but in the grand scheme of things the library is not about foundations or logic. Having univalence postulated just clarifies that point of view: We want to do univalent mathematics.

People can do logic and study what lemmas depend on what axioms in other excellent libraries, such as TypeTopology:)

Note that this is just my point of view right now. Perhaps the point of view taken for the library development will change over time.

EgbertRijke: > You are definitely not making me sad! It's more that the limitation of the language makes libraries in general less useful to people. For instance, if I don't want to subscribe to univalence, `agda-unimath` suddenly becomes very hard to use. I think I'm looking for something like a pragma
> 
> ```agda
> {-# SUSPEND univalence #-}
> ```
> 
> that would make the type checker complain if I somehow inadvertently invoke univalence through a lemma I'm applying in the file. Maybe I can talk to the Agda wizards about this at some point.

That's a very interesting suggestion. I think you should definitely bring it up with the agda developers, and if functionality like that would be implemented we can definitely consider using it in agda-unimath.

EgbertRijke: Your suggestion made me think about the `--safe` pragma. Right now it can only apply at the top level of a file, but actually we want to declare modules and definitions to be `--safe` within files that might also contain definitions that use postulates.

EgbertRijke: And actually what should really happen is that Agda marks all safe definitions as safe at compilation time, so that the humans don't have to do it for all definitions and clutter the code during that process. Humans might then still want to mark definitions as safe, but agda should then just check by itself that it indeed doesn't use definitions that rely on postulates. If agda keeps track of safe definitions during type-checking, this should be super straightforward. Humans shouldn't be required to mark anything at all as safe.

fredrik-bakke: Ah, that makes a lot of sense. I agree with your position!
Regarding your suggestion, I guess the same goes with many of Agda's flags, that they could in concept apply on a per-module or per-definition basis, as opposed to only on a per-file basis.

fredrik-bakke: In the same vein, one could perhaps decorate a definition with
```agda
{-# SAFE + funext #-}
```
to verify that this is all it uses, although the type checker would somehow keep a ledger of this regardless of the decoration.

fredrik-bakke: > That's a very interesting suggestion. I think you should definitely bring it up with the agda developers, and if functionality like that would be implemented we can definitely consider using it in agda-unimath.

Just as a remark, one instance where it could be useful inside the library, as opposed to the use case I originally described, is in `foundation.univalence-implies-function-extensionality`, where it could be used to verify that we are not applying univalence in the process of proving univalence.

fredrik-bakke:
I just noticed cofibers are called `cofiber`. I think it would make sense to call them `cofib` in line with the definition of fibers. Alternatively, we could call fibers `fiber`, but I think the shortened name looks better and doesn't cause much confusion. Can you make the renaming as part of this PR?

fredrik-bakke:
> Sure! I would prefer `fiber` in this case. It's just 2 characters longer, but it spells out the whole word so I consider it an improvement.

Alright, I'm not opposed to that. 😄

EgbertRijke:
And... action! 🎬

fredrik-bakke:
🍿🥱🥤

EgbertRijke:
🥲 I'm back at the code that traumatized me:)

fredrik-bakke:
> 🥲 I'm back at the code that traumatized me:)

Which was...

EgbertRijke:
This code:

https://github.com/UniMath/agda-unimath/pull/654/files#diff-c5b7c6ff105aa32eac32958ef7dde576c52600d7fd94a99868a7d8d49c8772e4

wich originates from `26-descent`

src/foundation/lifts-families-of-elements.lagda.md
```diff
@@ -0,0 +1,266 @@
...
+
+  lift-family-of-elements :
+    (A → X) → UU (l1 ⊔ l3)
+  lift-family-of-elements h = (a : A) → P (h a)
+```
+
+### Precomposition of lifts by ordinary functions
+
+```agda
+precompose-lifts :
```
fredrik-bakke, line 52: precomp

src/foundation/lifts-families-of-elements.lagda.md
```diff
@@ -0,0 +1,266 @@
...
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {X : UU l3} (P : X → UU l4)
+  (h : B → X)
+  where
+
+  tr-lift-family-of-elements :
+    {f g : A → B} (H : f ~ g) →
+    lift-family-of-elements A P (h ∘ f) → lift-family-of-elements A P (h ∘ g)
+  tr-lift-family-of-elements = tr-htpy (λ _ → P ∘ h)
+
+  COMPUTE-TR-LIFT-FAMILY-OF-ELEMENTS :
```
fredrik-bakke, line 74: oh god

fredrik-bakke: `statement` to the rescue? :)

fredrik-bakke: or `instance`

EgbertRijke: I wish my problems were just about the name:)

fredrik-bakke: You can always leave it as future work for the next sucker to pick up again!

EgbertRijke: I'm testing if the library compiles without that file

src/foundation/function-extensionality.lagda.md
```diff
@@ -24,24 +22,83 @@ open import foundation-core.identity-types
...
+### Based function extensionality
+
+This property asserts that, _given_ a function `f`, the map
+
+```text
+  htpy-eq : f ＝ g → f ~ g
 ```
 
-### Components of `funext`
+is an equivalence is an equivalence for any function `g` of the same type.
```
fredrik-bakke, line 71: something went wrong here

fredrik-bakke: never mind

src/synthetic-homotopy-theory/cavallos-trick.lagda.md
```diff
@@ -24,10 +24,13 @@ open import structured-types.pointed-types
 
 ## Idea
 
-Cavallo's trick is a way of upgrading an unpointed homotopy between pointed maps
-to a pointed homotopy. Originally, this trick was formulated by Evan Cavallo for
-homogeneous spaces, but it works as soon as the evaluation map `(id ~ id) → Ω B`
-has a section.
+Cavallo's trick is a way of upgrading an unpointed
```
fredrik-bakke, line 27:
```suggestion
**Cavallo's trick** is a way of upgrading an unpointed
```

src/synthetic-homotopy-theory/circle.lagda.md
```diff
@@ -43,6 +43,11 @@ open import univalent-combinatorics.standard-finite-types
 
 </details>
 
+## Idea
+
+The circle is the initial type equipped with a point and a
```
fredrik-bakke, line 48:
```suggestion
**The circle** is the initial type equipped with a base point and a
```

src/synthetic-homotopy-theory/cocones-under-spans-of-pointed-types.lagda.md
```diff
@@ -24,8 +24,11 @@ open import synthetic-homotopy-theory.cocones-under-spans
 
 ## Idea
 
-A cocone under a span of pointed types is a **pointed cocone** if it consists of
-pointed maps such that the proofs of point-preservation cohere.
+A [cocone under a span](synthetic-homotopy-theory.cocones-under-spans.md) of
+[pointed types](structured-types.pointed-types.md) is a **pointed cocone** if it
+consists of [pointed maps](structured-types.pointed-maps.md) equipped with a
+[pointed homotopy](structured-types.pointed-homotopies.md) witnessing that the
+naturality square commutes.
```
fredrik-bakke, line 31: link commutes to commuting squares of pointed maps

src/synthetic-homotopy-theory/cocones-under-spans.lagda.md
```diff
@@ -32,7 +32,8 @@ open import foundation-core.torsorial-type-families
 ## Idea
 
 A cocone under a span `A <-f- S -g-> B` with vertex `X` consists of two maps
```
fredrik-bakke, line 34:
```suggestion
A **cocone under a [span](foundation.spans.md)** `A <-f- S -g-> B` with vertex `X` consists of two maps
```

fredrik-bakke: Also, I like to say "codomain" instead of "vertex", to reinforce which way a cocone goes.

src/synthetic-homotopy-theory/free-loops.lagda.md
```diff
@@ -25,9 +25,9 @@ open import foundation.universe-levels
 
 ## Idea
 
-A free loop in a type `X` consists of a point `x : X` and an identification
-`x ＝ x`. The type of free loops in `X` is equivalent to the type of maps
-`𝕊¹ → X`.
+A **free loop** in a type `X` consists of a point `x : X` and an
+[identification](foundation.identity-types.md) `x ＝ x`. The type of free loops
+in `X` is equivalent to the type of maps `𝕊¹ → X`.
```
fredrik-bakke, line 30:
```suggestion
in `X` is [equivalent](foundation-core.equivalences.md) to the type of maps `𝕊¹ → X`.
```

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
```diff
@@ -38,9 +38,13 @@ open import synthetic-homotopy-theory.powers-of-loops
 
 ## Idea
 
-**Hatcher's** example of an acyclic type is a higher inductive type equipped
-with a point and two loops `a` and `b`, and identifications witnessing that
-`a⁵ ＝ b³` and `b³ = (ab)²`.
+**Hatcher's** example of an
+[acyclic type](synthetic-homotopy-theory.acyclic-types.md) is a higher inductive
```
fredrik-bakke, line 42:
```suggestion
**Hatcher's [acyclic type](synthetic-homotopy-theory.acyclic-types.md)** is a higher inductive
```

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
```diff
@@ -38,9 +38,13 @@ open import synthetic-homotopy-theory.powers-of-loops
...
+type equipped with a point and two
```
fredrik-bakke, line 43:
```suggestion
type equipped with a base point and two
```

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
```diff
@@ -80,19 +84,19 @@ hom-algebra-Hatcher-Acyclic-Type
...
+                ( s1)) ×
+              ( coherence-square-identifications
                 ( ap (map-Ω f) r2)
+                ( map-power-nat-Ω 3 f a2 ∙ ap (power-nat-Ω 3 B) v)
                 ( ( map-power-nat-Ω 2 f (a1 ∙ a2)) ∙
                   ( ap
                     ( power-nat-Ω 2 B)
                     ( ( preserves-mul-map-Ω f a1 a2) ∙
-                      ( ap-binary _∙_ u v)))))))
+                      ( ap-binary _∙_ u v))))
```
fredrik-bakke, line 98:
```suggestion
                      ( ap-binary (_∙_) u v))))
```

src/synthetic-homotopy-theory/interval-type.lagda.md
```diff
@@ -24,7 +27,7 @@ open import foundation.universe-levels
 ## Idea
 
 The interval type is a higher inductive type with two points and an
```
fredrik-bakke, line 29:
```suggestion
**The interval type** is a higher inductive type with two points and an
```

EgbertRijke: One sec dude, I'm trying to make this PR work, not to get more on my plate

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
```diff
@@ -23,8 +23,9 @@ open import synthetic-homotopy-theory.pushouts
 
 ## Idea
 
-Given a span of pointed maps, then the pushout is in fact a pushout diagram in
-the category of pointed types.
+Given a span of [pointed maps](structured-types.pointed-maps.md), then the
+[pushout](synthetic-homotopy-theory.pushouts.md) is in fact a pushout diagram in
+the (wild) category of [pointed types](structured-types.pointed-types.md).
```
fredrik-bakke, line 28: in fact, the (∞-)category

fredrik-bakke:
Sorry, I was just trying to help you get this PR out the door. Do you mind if I post my final comments now?

EgbertRijke:
Sure, sorry it just stressed me out a little, but it's ok you can continue reviewing.

EgbertRijke:
Ready for review!

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
EgbertRijke: In fact, the conjectured oo-category according to a conjecturally conjectured definition of oo-categories inside type theory.

src/synthetic-homotopy-theory/powers-of-loops.lagda.md
```diff
@@ -30,8 +30,9 @@ open import synthetic-homotopy-theory.loop-spaces
 
 ## Idea
 
-The `n`-th power of a loop `ω` in a type `A` is defined by iterated
-concatenation of `ω` with itself.
+The **`n`-th power** of a [loop](synthetic-homotopy-theory.loop-spaces.md) `ω`
```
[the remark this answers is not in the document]
EgbertRijke: Sure:)

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
EgbertRijke: Actually, this should be a horizontal concatenation of identifications

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
EgbertRijke: Do you want the library to mention oo-categories even though we can't say what they are?

I mean, the set theoretic definition of oo-categories is not really what we mean, so I don't even know what to cite if we would mention oo-categories.

EgbertRijke:
One conversation is left open, where I need further instructions if you want me to do something.

src/synthetic-homotopy-theory/pushouts-of-pointed-types.lagda.md
fredrik-bakke: I am fine with just calling them categories and not linking anywhere. To be honest, oo-categories should be what we call categories, and what we call categories should be called 0-truncated categories. I just find it dissatisfying to say it's a wild category. Wild categories is just an implementation detail because we can't say what categories are.

EgbertRijke: Great! Then I addressed all the comments :)

fredrik-bakke: I.e. I'd remove `(wild) ` in the above text

EgbertRijke: Even given that in our library "category" refers to 1-category? I feel like given our current library as context, the disclaimer wild adds some information.

src/synthetic-homotopy-theory/cocones-under-spans.lagda.md
```diff
@@ -31,8 +31,9 @@ open import foundation-core.torsorial-type-families
 
 ## Idea
 
-A cocone under a span `A <-f- S -g-> B` with vertex `X` consists of two maps
-`i : A → X` and `j : B → X` equipped with a homotopy witnessing that the square
+A **cocone under a [span](foundation.spans.md)** `A <-f- S -g-> B` with vertex
+`X` consists of two maps `i : A → X` and `j : B → X` equipped with a
```
fredrik-bakke, line 35: Did you disagree with the following change or just miss it?
```suggestion
A **cocone under a [span](foundation.spans.md)** `A <-f- S -g-> B` with codomain
`X` consists of two maps `i : A → X` and `j : B → X` equipped with a
```

EgbertRijke: Oh yeah, I missed it because conversations get automatically resolved if you commit a suggestion.

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
```diff
@@ -38,9 +39,12 @@ open import synthetic-homotopy-theory.powers-of-loops
 
 ## Idea
 
-**Hatcher's** example of an acyclic type is a higher inductive type equipped
-with a point and two loops `a` and `b`, and identifications witnessing that
-`a⁵ ＝ b³` and `b³ = (ab)²`.
+**Hatcher's [acyclic type](synthetic-homotopy-theory.acyclic-types.md)** is a
+higher inductive type equipped with a base point and two
```
fredrik-bakke, line 43:
```suggestion
higher inductive type [equipped](foundation.structure.md) with a base point and two
```
At some point, we should probably touch up the structure file and make it into a small exposition

fredrik-bakke (approved):

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
EgbertRijke: Yeah, and we should put it in `structured-types`

By the way, do you want to link to `structure` in every file where we say `equipped with`? Isn't that exaggerated?

EgbertRijke: I think in this pull request I will just accept this suggestion, but if you're going to link to `structure` at "equipped with" in all other files then I'm gonna ask you to do that in another pull request.

fredrik-bakke: Is it? It's not like it's a big distraction to have the link there. If we made the structure file into a nice exposition about structures in constructive math, then I think it could be a nice way to introduce curious people to type theory

EgbertRijke: Yes, perhaps you're right, but I'll still ask you to do it in another pull request if you want to link more files to `structure`

fredrik-bakke: So you would prefer if I don't make suggestions on adding links to that particular file anymore?

fredrik-bakke:
⏰

fredrik-bakke:
This one was a long time coming. Thank you again for making the effort to finally finish it, Egbert!

src/synthetic-homotopy-theory/hatchers-acyclic-type.lagda.md
EgbertRijke: Don't get me wrong, I think it is good to have such links, but if you're now going to comprehensively add links throughout the library then it would make more sense to me if we merge this one and you do that in vscode, where you will be much faster than on the github website.

EgbertRijke:
Thank you too for all your swift help with this!

PR #655, opened 2023-06-11
Localizations and other things

### LINKED ISSUE #652: Detect ill-formed import blocks
This file had an ill-typed `<details>` block which wasn't closed, but it wasn't detected by `make pre-commit`.

https://unimath.github.io/agda-unimath/finite-group-theory.transpositions-standard-finite-types.html

fredrik-bakke:
### Summary
- Define localizations
- Organize sigma-closed subuniverses and reflective subuniverses
- Detect ill-formed imports blocks
- Add `unused-imports` `make` target
- Add `.editorconfig` and misc. settings tweaks. The `.editorconfig` file allows for a few shared settings between vscode, emacs, and other editors, making the experience of using the library slightly more homogenous. To use the `.editorconfig` file in emacs, please install this extension: https://github.com/editorconfig/editorconfig-emacs. It would be beneficial to note this in the install guide as well. Maybe someone with more emacs knowledge can do that?

Resolves #652.


src/orthogonal-factorization-systems/reflective-modalities.lagda.md
```diff
@@ -0,0 +1,45 @@
...
+open import orthogonal-factorization-systems.local-types
+open import orthogonal-factorization-systems.reflective-subuniverses
+open import orthogonal-factorization-systems.modal-operators
+```
+
+</details>
+
+## Idea
+
+A [modal operator](foundation.modal-operators.md) with unit is **reflective** if
```
fredrik-bakke, line 24: I'm tempted to define a modal operator and unit pair as a "modality". I.e.

```agda
modality : (l1 l2 : Level) → UU (lsuc l1 ⊔ lsuc l2)
modality l1 l2 = Σ (operator-modality l1 l2) (unit-modality)
```
Is this a good definition? I fear it will be conflated with the notion of "higher modality" somehow

EgbertRijke: I think that's ok. When you write good infrastructure for it, you will be able to conveniently use it in whatever way you want.

fredrik-bakke: Alright, great! 😄 Then I can generally phrase things in terms of a _modality_ instead of a modal unit, as I have until now

EgbertRijke: Btw I assume you'll want to put the universal property in that sigma-type too, because that's what is part of being a modality. Do you make sure that the universal property holds with respect to types of arbitrary universe levels?

fredrik-bakke: What universal property? A "modality" in my proposed sense does not have a universal property yet.

EgbertRijke: Well, the dependent universal property of course:) If you're implementing something else than the modalities from the paper, that's ok too of course, but then you have to be very clear about it. And if you want to do the kind of general modalities that are neither monadic or idempotent, then it would make sense to drop the unit of the modality as well.

fredrik-bakke: Right. I guess I should just call it operator-unit pairs for now to avoid confusion. This is precisely what my initial question was probing at.

EgbertRijke:
I wonder if it would make sense to call them unital species of types

https://unimath.github.io/agda-unimath/species.species-of-types.html

fredrik-bakke:
Is that generally accepted terminology though? It seems to me that _species_ are usually understood to regard finite sets in some way, although I know very little about them.

EgbertRijke:
I'm not aware of any generally accepted terminology for operations on types that return types. We called them species of types because we did the same things with them as typical things one does with species. Many such constructions generalise readily to general types, so there seemed little reason to only do them for finite types. So as far as I'm concerned it's at least accepted terminology within agda-unimath.

fredrik-bakke:
Thank you for the input! We could also steal terminology from multimodal adjoint type theory and call them _positive species_. Although, I'm still leaning towards keeping the term "modal operator", and "positive modal operator" doesn't sound right to me.

fredrik-bakke:
I could call it `positive-modality` and then be very explicit that it does not have a universal property. However, I would actually want to use the word "comodality" to describe the dual, and not "negative modality"...

src/commutative-algebra/nilradicals-commutative-semirings.lagda.md
```diff
@@ -67,20 +67,20 @@ module _
   {l : Level} (A : Commutative-Semiring l)
   where
 
-  is-closed-under-mul-right-nilradical-Commutative-Semiring :
+  is-closed-under-multiplication-right-nilradical-Commutative-Semiring :
```
EgbertRijke, line 70: It would be better to use `left-multiplication` and `right-multiplication` in the names to match the corresponding names of the closure properties

fredrik-bakke: Ah, thanks, I didn't notice those!

fredrik-bakke:
This PR leaves a lot of future work, but I think it is currently in a reasonable state to be reviewed/merged.

fredrik-bakke:
I'll add the new spacing formatting in a separate PR. I realized it could be generalized substantially.

src/foundation-core/contractible-types.lagda.md
```diff
@@ -224,13 +224,20 @@ module _
...
+      is-contr A → (a : A) → is-contr (B a) → is-contr (Σ A B)
     pr1 (pr1 (is-contr-Σ H a K)) = a
     pr2 (pr1 (is-contr-Σ H a K)) = center K
     pr2 (is-contr-Σ H a K) (pair x y) =
       eq-pair-Σ
         ( inv (eq-is-contr H))
         ( eq-transpose-tr (eq-is-contr H) (eq-is-contr K))
+
+  abstract
+    is-contr-Σ-center :
```
EgbertRijke, line 236: Not sure if I like the name, because usually the second part would refer to hypotheses. The following is more verbose, but more clearly conveys what the definition is doing.

`is-contr-Σ-is-contr-base-is-contr-center`

src/foundation/subuniverses.lagda.md
```diff
@@ -99,27 +99,6 @@ module _
...
-  (Y : inclusion-subuniverse P X → type-subuniverse Q) →
-  is-in-subuniverse R
-    ( Σ (inclusion-subuniverse P X) (λ x → inclusion-subuniverse Q (Y x)))
-
-is-closed-under-Σ-subuniverse :
-  {l1 l2 : Level} (P : subuniverse l1 l2) → UU (lsuc l1 ⊔ l2)
-is-closed-under-Σ-subuniverse P = is-closed-under-Σ-subuniverses P P P
-```
-
 ## Properties
```
EgbertRijke, line 102: Maybe this file could have a `## See also` reference to the sigma-closed subuniverses

src/orthogonal-factorization-systems/closed-modalities.lagda.md
```diff
@@ -30,7 +31,7 @@ open import synthetic-homotopy-theory.joins-of-types
 ## Idea
 
 Given any proposition `Q`, the
```
EgbertRijke, line 33: The word proposition could be hyperlinked

src/orthogonal-factorization-systems/higher-modalities.lagda.md
```diff
@@ -232,7 +236,7 @@ module _
     {X : UU l1} {x' y' : ○ X} →
     ○ (type-is-small (is-locally-small-○ X x' y')) → x' ＝ y'
   elim-Id-higher-modality {X} {x'} {y'} =
-    map-inv-unit-is-modal-is-small unit-○
+    map-inv-unit-is-modal-small unit-○
```
EgbertRijke, line 239: Is `is-modal-small` something we can factor out?

src/orthogonal-factorization-systems/local-types.lagda.md
```diff
@@ -40,7 +40,7 @@ A type family `A` over `X` is said to be **local at** `f : Y → X`, or
 
 is an equivalence.
 
-Likewise, a type `A` is said to be `f`-local if the precomposition map
+Likewise, a _type_ `A` is said to be `f`-local if the precomposition map
```
EgbertRijke, line 43: `**`f`-local**`

This section could also have more hyperlinks, for words such as equivalence and precomposition map

src/orthogonal-factorization-systems/localizations-maps.lagda.md
```diff
@@ -0,0 +1,112 @@
...
+
+open import orthogonal-factorization-systems.local-types
+open import orthogonal-factorization-systems.localizations-subuniverses
+```
+
+</details>
+
+## Idea
+
+Let `f` be a map of type `X → Y` and let `A` be a type. The **localization** of
```
EgbertRijke, line 39: I like to have left maps be `f : A -> B` and right maps `g : X -> Y`, but here you use the opposite convention. I would suggest having `f : A -> B` here and say what it means to localize a type `X`, at least that's how I would be used to doing it.

src/orthogonal-factorization-systems/localizations-subuniverses.lagda.md
```diff
@@ -0,0 +1,137 @@
...
+**localization** at `P`, or **`P`-localization**, is a type `B` in `P` and a map
+`η : A → B` such that every type in `P` is
+`η`[-local](orthogonal-factorization-systems.local-types.md). I.e. for every `X`
+in `P`, the precomposition map
+
+```text
+  _∘ η : (B → X) → (A → X)
+```
+
+is an equivalence.
```
EgbertRijke, line 48: hyperlink equivalences?

src/orthogonal-factorization-systems/modal-operators.lagda.md
```diff
@@ -61,60 +66,69 @@ module _
...
+  modal-type-subuniverse : subuniverse l1 (l1 ⊔ l2)
+  modal-type-subuniverse = is-modal-Prop
 ```
 
 ### Modal small types
 
 A small type is said to be modal if its small equivalent is modal.
 
 ```agda
+is-modal-small :
```
EgbertRijke, line 78: Ah, it is factored out here! Why didn't those previous definition use this definition of `is-modal-small`?

EgbertRijke: Would the name be clearer if it was `is-modal-type-is-small`? Right now I find it hard to know what `is-modal-small` would mean just by looking at the definition, and I would guess something like that it is modal and small. That's not how it is defined though, so this indicates to me that the name `is-modal-small` is not as clear as it could be.

src/orthogonal-factorization-systems/reflective-subuniverses.lagda.md
```diff
@@ -9,50 +9,191 @@ module orthogonal-factorization-systems.reflective-subuniverses where
...
     ( λ ○ →
       Σ ( unit-modality ○)
         ( λ unit-○ →
-          Σ ( UU l → Prop lM)
-            ( is-reflective-subuniverse unit-○)))
+          ( (X : UU l) → is-in-subuniverse P (○ X)) ×
+          ( (X Y : UU l) → is-in-subuniverse P X → is-local (unit-○ {Y}) X)))
 ```
 
+```agda
```
EgbertRijke, line 50: Should this not be in the same code block with the previous definition?

src/foundation-core/contractible-types.lagda.md
fredrik-bakke: I think the name was fine, as it refers to applying `is-contr-Σ` at the center, but I can change it. Shouldn't it be in the other order, though? I.e. `is-contr-Σ-is-contr-center-is-contr-base`

src/orthogonal-factorization-systems/reflective-subuniverses.lagda.md
fredrik-bakke: Sometimes I end and start code blocks like this because it "seems right". It adds a small separator where I would mentally have a separator. I don't know how it looks on the website, however. Should I change it?

src/orthogonal-factorization-systems/higher-modalities.lagda.md
fredrik-bakke: Sorry, I'm not quite sure what to factor out here. Can you elaborate?

src/foundation-core/contractible-types.lagda.md
EgbertRijke: I actually think that the rationale behind this definition is questionable. A proof of `is-contr` is often abstract, because they can be complicated proofs that are irrelevant. So I don't think `is-contr-Σ-center` should get much use, and I would prefer to use `is-contr-Σ` even if it involves specifying one more argument.

EgbertRijke: Note that even if a proof of contractibility is abstract, we should not ask Agda to evaluate it by asking what the center of this proof of contraction is. This slows down the library too much. The definition `is-contr-Σ` on the other hand would never ask Agda to evaluate a proof of contractibility in order to get access to the center.

EgbertRijke: I just thought about this over lunch, and I think it is better to remove this definition to not encourage it to be used.

fredrik-bakke: Interesting, I didn't think this definition would be controversial. I will remove it.

EgbertRijke: Perhaps we could add a short explanation along the following lines, of our choice of definition:

"We prove that `Σ A B` is contractible whenever `A` is contractible and whenever `B a` is contractible for a given element `a : A`. Note that we _could_ choose this element `a` to be the center of contraction. However, it turns out to be better to not do so in the construction of `is-contr-Σ`. The reason is that proofs of contractibility could be quite complicated and difficult to normalize. If we would require in the definition of `is-contr-Σ` that `B (center c)` is contractible at the center of contraction of provided by a proof `c` of contractibility of `A`, then the type-inference algorithm of Agda will be forced to normalize the proof `c` including the contraction. By providing a center of contraction by hand, we avoid this unnecessary load on the type-inference algorithm, and hence any instance of `is-contr-Σ` will type-check more efficiently"

EgbertRijke: There were a few more instances where I opted for a general user-provided argument instead of one that could be inferred from a previous hypothesis, but I don't recall right now where this happened. But this issue comes up more often than just here. The general theme is that it may be computationally expensive to extract information from proofs of propositions, such as the center of contraction of a proof of contractibility.

fredrik-bakke: Thanks! I rephrased your text slightly and added it as a note directly after `is-contr-Σ`. I think it is a nice addition to inform the user of this, even if it's a bit of a digression. Also, on the use of hyphens in "type inference" and "type checking". I'm not entirely sure what is correct. But I think we can't go wrong with space separation. I think also "typechecking" is correct, but I'm not entirely sure.

The note:

> **Note**: In the previous construction, we showed that `Σ A B` is contractible
> whenever `A` is contractible and whenever `B a` is contractible for a specified
> term `a : A`. We _could_ have chosen this element `a` to be the center of
> contraction of `A`. However, it turns out to be better not to do so in the
> construction of `is-contr-Σ`. The reason is that proofs of contractibility could
> be quite complicated and difficult to normalize. If we would require in the
> definition of `is-contr-Σ` that `B (center c)` is contractible, given the proof
> `c` of contractibility of `A`, then the type inference algorithm of Agda will be
> forced to normalize the proof `c` including the contraction. By instead
> providing a center of contraction by hand, we avoid this unnecessary load on the
> type inference algorithm, and hence any instance of `is-contr-Σ` will type check
> more efficiently.

EgbertRijke: Perfect! Many thanks!

EgbertRijke: Perhaps you could add this too. I think it would be good to record it somewhere, and here is fine for now:

"The general theme is that it may be computationally expensive to extract information from proofs of propositions, such as the center of contraction of a proof of contractibility. The reason for that is that when Agda normalizes an element (as it inevitably will do sometimes) then in such cases it will not just normalize the extracted information (in this case the first projection of the proof of contractibility), but it will normalize the entire proof of contractibility first, and then apply the projection."

src/orthogonal-factorization-systems/reflective-subuniverses.lagda.md
EgbertRijke: I don't know... I never put those in separate code blocks, because for me it all belongs together:)

fredrik-bakke: Then how about keeping it for now, then we can see how it feels. It is not hard to change it later. Just do a simple text search for

````
```

```agda
````

src/orthogonal-factorization-systems/higher-modalities.lagda.md
EgbertRijke: On second inspection it looks fine to me now:)

EgbertRijke:
Looks good to me now! I'll merge it when the checks are done.

PR #657, opened 2023-06-14
typo

EgbertRijke:


EgbertRijke:
I thought I commited the fix of that typo, but apparently I hadn't.

PR #659, opened 2023-06-15
Replacing isretr with is-retraction and issec with is-section

EgbertRijke:


fredrik-bakke:
Are you also doing `retr` -> `retraction` and `sec` -> `section`? 😁

EgbertRijke:
Thanks for the suggestion. I did those too now

EgbertRijke:
Since there are only trivial changes here, I'm going to merge this as soon as the checks are done. Then I can work with these new conventions in my other PRs.

If there's anything you'd like me to change, I will do it promptly in another PR.

PR #660, opened 2023-06-15
typo in standard-finite-types.lagda.md

MatthiasHu:


EgbertRijke:
Awesome! Thank you for spotting and correcting it!

PR #665, opened 2023-06-22
Posets are categories, and refactor binary relations

fredrik-bakke:
### Summary
- Define large binary relations
- Refactor use of reflexivity, symmetry, and transitivity to use common definition found in (large) binary relations.
  **Note**: In general, all transitivity proofs will now be of the form `R y z → R x y → R x z`, and all reflexivity, symmetry and transitivity proofs will take the endpoints _explicitly_.
- Renamings:
  - `symm` -> `symmetric`
  - `trans` -> `transitive`
  - `Rel` -> `Relation`
  - `Eq-Rel` -> `Equivalence-Relation`
  - (Reflexivities remain as `refl` everywhere)
- Basic additions to `large-categories`
- Prove that there are categories of small sets, groups, and semigroups
- Prove that posets are categories whose hom-sets are propositions

Co-authored-by: louismntnu <louis.o.martini@ntnu.no>
Co-authored-by: fernabnor <fernando.a.garcia@ntnu.no>


fredrik-bakke:
I'm currently refactoring binary relations and preorders, but there's inconsistency in
1. where reflexivities take their arguments implicitly or explicitly
2. where transitivities take the endpoints implicitly or explicitly
3. use of variance for transitivity

Ideally, I would like the variance of transitivity to be the same as the variance for composition in categories, but I also prefer covariance over contravariance. What do you think?

EgbertRijke:
Since this is together with Fernando and Luis, can you add their github handles in the PR description. Here is a link with instructions:

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

EgbertRijke:
> I'm currently refactoring binary relations and preorders, but there's inconsistency in
> 
> 1. where reflexivities take their arguments implicitly or explicitly
> 2. where transitivities take the endpoints implicitly or explicitly
> 3. use of variance for transitivity
> 
> Ideally, I would like the variance of transitivity to be the same as the variance for composition in categories, but I also prefer covariance over contravariance. What do you think?

There are some different needs for transitivity of an arbitrary relation and composition in a category. In a category, the objects are often composed with Sigma and Pi-types, and cannot be automatically inferred as implicit arguments. In posets and arbitrary relations, the situation is often simpler and the arguments can be inferred. Probably it is best to make conventions according to how it is used, not to how it is most uniform and symmetric.

EgbertRijke:
That said, you are right that there are inconsistencies in how we currently do it, and it can probably be improved.

EgbertRijke:
By the way, I've been doing some work on relations too in the "Beyond finite sets" PR. But I don't think the current changes are causing any conflicts or issues.

fredrik-bakke:
Thanks! I'll try and get their handles :)

It turns out that for transitivity, the first endpoint can't generally be inferred for some reason. Therefore I'm leaning towards having two definitions `is-transitive` when explicit, and `transitive` when implicit, since the latter is commonly used in this case. However, I know this conflicts with another naming convention. Likewise, we could have `is-reflexive` and `refl`, and so on. Another suggestion would be to have `is-transitive` and `is-transitive'`, but I don't like this too much. What do you think? Also, regarding merge conflicts, I haven't pushed most of the changes yet, but I'll have a look at what you've done in your PR.

src/foundation/category-of-sets.lagda.md
```diff
@@ -25,10 +36,49 @@ sets for each universe level, and there is a large category of sets.
 
 ## Definitions
 
-### The large category of sets
+### The precategory of small sets
 
 ```agda
-Set-Large-Precategory : Large-Precategory lsuc (λ l1 l2 → l1 ⊔ l2)
+Set-Precategory : (l : Level) → Precategory (lsuc l) l
```
EgbertRijke, line 42: Note that there is a function that turns a large precategory and a universe level l into a small precategory

src/foundation/category-of-sets.lagda.md
```diff
@@ -37,3 +87,6 @@ associative-comp-hom-Large-Precategory Set-Large-Precategory h g f = refl
 left-unit-law-comp-hom-Large-Precategory Set-Large-Precategory f = refl
 right-unit-law-comp-hom-Large-Precategory Set-Large-Precategory f = refl
 ```
+
+Note that to prove that the large precategory of sets is a large _category_, we
```
EgbertRijke, line 91: Note that large categories are level-wise univalent, so we don't need univalence for large types.

src/foundation/category-of-sets.lagda.md
fredrik-bakke: Thanks, I'll refactor to use that! This was just copy-pasted

src/foundation/category-of-sets.lagda.md
fredrik-bakke: Thanks for the correction!

fredrik-bakke:
This idea of starting with large constructions and later extracting small ones from them is so nice :)

src/category-theory/isomorphisms-categories.lagda.md
```diff
@@ -442,7 +442,7 @@ module _
   extensionality-obj-Category :
     (x y : obj-Category C) → (x ＝ y) ≃ iso-Category C x y
   pr1 (extensionality-obj-Category x y) = iso-eq-Category C
-  pr2 (extensionality-obj-Category x y) = is-category-Category C x y
+  pr2 (extensionality-obj-Category x y) = is-category-precategory-Category C x y
```
EgbertRijke, line 445: I like `is-category-Category` more. I have used this shorter naming scheme in a lot of places, such as `is-ideal-ideal-Ring` instead of `is-ideal-subset-ideal-Ring`.

src/category-theory/isomorphisms-large-precategories.lagda.md
```diff
@@ -119,6 +121,14 @@ iso-eq-Large-Precategory :
...
   X ＝ Y → iso-Large-Precategory C X Y
 iso-eq-Large-Precategory C X .X refl = id-iso-Large-Precategory C
+
+htpy-iso-eq-precategory-Large-Precategory :
+  {α : Level → Level} {β : Level → Level → Level} →
+  (C : Large-Precategory α β) {l1 : Level}
+  (X : obj-Large-Precategory C l1) (Y : obj-Large-Precategory C l1) →
+  iso-eq-Precategory (precategory-Large-Precategory C l1) X Y ~
+  iso-eq-Large-Precategory C X Y
+htpy-iso-eq-precategory-Large-Precategory C X .X refl = refl
```
EgbertRijke, line 131: I would be surprised if this cannot be defined directly as `htpy-iso-eq-Large-Precategory`. This has the advantage that you can reuse the proof that `htpy-iso-eq-Large-Precategory` is an equivalence, whereas if you define it anew by induction you will have to do some unnecessary argument to show that it is an equivalence.

fredrik-bakke: Is `htpy-iso-eq-Large-Precategory` supposed to be an already defined term? I cannot find it

src/category-theory/large-categories.lagda.md
```diff
@@ -41,10 +47,136 @@ record
...
+module _
+  {α : Level → Level} {β : Level → Level → Level}
+  (C : Large-Category α β)
+  where
+
+  precategory-Large-Category : (l : Level) → Precategory (α l) (β l l)
+  precategory-Large-Category =
+    precategory-Large-Precategory (large-precategory-Large-Category C)
+
+  is-category-precategory-Large-Category :
```
EgbertRijke, line 170: I would call this `is-large-category-Large-Category`

src/category-theory/large-categories.lagda.md
```diff
@@ -41,10 +47,136 @@ record
...
+    (l : Level) → is-category-Precategory (precategory-Large-Category l)
```
EgbertRijke, line 171: There should be a predicate `is-large-category-Precategory` that expresses exactly the above line of code.

src/category-theory/large-categories.lagda.md
fredrik-bakke: Sorry I messed up some of the namings, but I'm not sure this is right. Please have a look if the new names make more sense.

src/category-theory/isomorphisms-large-precategories.lagda.md
```diff
@@ -119,6 +121,14 @@ iso-eq-Large-Precategory :
   (X : obj-Large-Precategory C l1) (Y : obj-Large-Precategory C l1) →
   X ＝ Y → iso-Large-Precategory C X Y
 iso-eq-Large-Precategory C X .X refl = id-iso-Large-Precategory C
+
+htpy-iso-eq-Large-Precategory :
```
EgbertRijke, line 125: Hm.. perhaps my previous comment on this function was not so accurate, because I think at that time I misunderstood the type of this construction.

It asserts that `iso-eq` for the small precategory obtained from a large precategory is homotopic to the `iso-eq` taken directly from the large precategory.

In that case I think the name `htpy-iso-eq` is not so good, because under our naming scheme it would assert be the construction that turns an equality of isomorphisms into a homotopy of isomorphisms, which would be the underlying function for the extensionality principle of isomorphisms.

How about `compute-iso-eq-precategory-Large-Precategory`?

And my apologies for my earlier comment, which was off.

fredrik-bakke: `compute` is good! And no worries, I value your input even when it's wrong :)

fredrik-bakke: I went with the slightly shorter `compute-iso-eq-Large-Precategory`

EgbertRijke: Excellent!

EgbertRijke (commented):
This is a large PR. I skimmed over all the changes, and I liked removing `dependent-pair-types` from `foundation-core` and the new file `large-binary-relations`. Overall the changes look good to me. Thanks for doing this!

src/category-theory/isomorphisms-precategories.lagda.md
```diff
@@ -298,3 +322,50 @@ module _
...
+
+```agda
+module _
+  {l1 l2 : Level} (C : Precategory l1 l2)
+  (x y : obj-Precategory C)
+
+  where
+
+  is-iso-is-prop-hom-Precategory' :
+    is-prop (type-hom-Precategory C x x) →
```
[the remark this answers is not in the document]
fredrik-bakke: I like this; very colorful language! Do you want me to create a small file on `unyielding-objects-precategories` or are you undecided on the name? Another possible naming is `obstinate`, following the principle of "not naming things by what they are not".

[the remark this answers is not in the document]
EgbertRijke: I'm not really decided on the name. Perhaps just "superrigid" would be more easily understood by mathematicians than any creative terminology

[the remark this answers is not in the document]
fredrik-bakke: Good call; I think "superrigid" is better.

src/category-theory/large-categories.lagda.md
```diff
@@ -41,10 +47,137 @@ record
...
+  right-unit-law-comp-hom-Large-Category :
+    {l1 l2 : Level}
+    {X : obj-Large-Category l1} {Y : obj-Large-Category l2}
+    (f : type-Set (hom-Large-Category X Y)) →
+    ( comp-hom-Large-Category f id-hom-Large-Category) ＝ f
+  right-unit-law-comp-hom-Large-Category =
+    right-unit-law-comp-hom-Large-Precategory
+      ( large-precategory-Large-Category C)
+
+  type-hom-Large-Category :
```
EgbertRijke, line 124: This and the next definition should come earlier, and the definitions above should use `type-hom-Large-Category`

src/category-theory/large-categories.lagda.md
EgbertRijke: It looks like this comment was perhaps not accurate, because I misunderstood what this code was about. It seems to be about proving that the small precategory of objects of universe level `l` obtained from a large category is a category. The thing that confused me was the quantification over `Level` in the type specification, but on a second look it actually looks alright.

Perhaps the old name `is-category-precategory-Large-Category` is the most accurate name for this construction. Sorry that it took me a while to understand it:)

src/category-theory/isomorphisms-precategories.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: The inclusion of `superrigid-objects-precategories` feels a bit unmotivated with the current state of `category-theory`, so perhaps it is best to leave this for now. Can we keep the naming `is-iso-is-prop-hom-Precategory'`in the meanwhile?

src/category-theory/large-categories.lagda.md
fredrik-bakke: I'm a bit confused by this comment. Is the current state of this code good?

src/category-theory/large-categories.lagda.md
fredrik-bakke: Thanks! Sorry for my slopiness

fredrik-bakke:
Since every comment has been resolved, I will merge the PR.

EgbertRijke:
Thank you for all your work on this PR!

Issue #638, opened 2023-06-04
Make our pre-commit hooks usable in third-party projects


fredrik-bakke:
Following a suggestion made by @jonaprieto (<https://github.com/UniMath/agda-unimath/issues/627#issuecomment-1561203428>) I'll record here an issue about potential ways to make our pre-commit hooks usable by third-parties.

fredrik-bakke:
At a minimum, we will have to abstract away our folder structure and support nested namespaces, increase the robustness of our scripts, and make most constants configurable from a common location. I think it is also worthwhile to try to iron out more edge cases, add more detailed error messages, and port to using a well-established markdown parsing library.

Other than just making the scripts available as pre-commit hooks, we could also consider compiling them into an autoformatting/linting tool. If we are to make an autoformatting tool, I think we should try and make a script that automatically indents and line-breaks code correctly as well (although I worry this can be rather difficult).

fredrik-bakke:
A thought: does there exist a markdown autoformatting tool that lets us inject autoformatting for specific code blocks?

PR #670, opened 2023-06-23
Fix alignment `where` blocks

### LINKED ISSUE #667: Inconsistencies with regards to `where` convention
Just noting that the library is not particularly consistent with respect to this convention currently.

_Originally posted by @fredrik-bakke in https://github.com/UniMath/agda-unimath/pull/511#discussion_r1234323191_

fredrik-bakke:
Resolves #667.


fredrik-bakke:
I didn't really find a natural place to put the code for detecting misalignments such as these, so I've left it out of the PR for now. For future me, something like the following pattern should do the trick:

```python
where_indentation_pattern = re.compile(
    r'^((where\n)|(\s{2,2}where\n\s{2,2})|(\s{4,4}where\n\s{4,4})|(\s{6,6}where\n\s{8,8})|(\s{10,10}where\n\s{10,10})|(\s{12,12}where\n\s{12,12})|(\s{14,14}where\n\s{14,14})|(\s{16,16}where\n\s{16,16}))\s\s\S')
```

VojtechStep:
> ```python
> where_indentation_pattern = re.compile(
>     r'^((where\n)|(\s{2,2}where\n\s{2,2})|(\s{4,4}where\n\s{4,4})|(\s{6,6}where\n\s{6,6})|(\s{8,8}where\n\s{8,8})|(\s{10,10}where\n\s{10,10})|(\s{12,12}where\n\s{12,12})|(\s{14,14}where\n\s{14,14})|(\s{16,16}where\n\s{16,16}))\s\s\S')
> ```

FYI @fredrik-bakke 
- `pat{n,n}` is equivalent to `pat{n}`
- Python regexes support references to previously matched strings, so I believe the expression could be simplified to something like `r'^((?:\s\s)*)where\n\1\s\s\S'` (it might need a little twiddling, I'm on my phone). Also since we're enforcing indentation by spaces, I think it would be more correct to use ` ` instead of `\s`, since we don't want people indenting by tabs or form feeds

fredrik-bakke:
Thanks for the feedback! I was unsure if references worked within a single regex, but I will try it. Also, I think the best practice would be to define an `AGDA_INDENT` constant as I've actually done in other scripts. This also makes it easier to port the code to other repositories, in line with #638.

src/category-theory/discrete-categories.lagda.md
```diff
@@ -30,7 +30,7 @@ module _
   discrete-precategory-Set : Precategory l l
   discrete-precategory-Set =
     type-Set X , disc-Hom , composition-structure , id-structure
-    where
+      where
```
EgbertRijke, line 33: I'm not sure why this entire where block is indented when it doesn't need to be.

src/category-theory/slice-precategories.lagda.md
```diff
@@ -223,7 +223,7 @@ module _
     map-is-pullback-is-product-Slice-Precategory
       ϕ (Z , .(comp-hom-Precategory C f h₁)) (h₁ , refl) (h₂ , β₂) =
       is-contr-Σ-is-prop c d q σ
-      where
+        where
```
EgbertRijke, line 226: I'm not sure why this entire where block is indented when it doesn't need to be.

src/category-theory/slice-precategories.lagda.md
```diff
@@ -284,7 +284,7 @@ module _
       is-pullback-Precategory C A X Y f g W p₁ p₂ α
     map-inv-is-pullback-is-product-Slice-Precategory ψ W' p₁' p₂' α' =
       is-contr-Σ-is-prop k γ q σ
-      where
+        where
```
EgbertRijke, line 287: I'm not sure why this entire where block is indented when it doesn't need to be.

EgbertRijke (commented):
I guess this goes for all of the `where` blocks that are modified in this PR. I'm not sure why they are indented more than they need to be.

src/category-theory/discrete-categories.lagda.md
fredrik-bakke: Oh, then I've misunderstood the convention. What would you want the code to actually look like?

EgbertRijke: I like it if we don't unnecessarily indent stuff more than necessary. I guess what was confusing here was that the contents of the `where` blocks were also erroneously more indented than they needed to be. Would it be ok with you to make all of these `where` blocks including their contents two spaces less indented?

EgbertRijke: Perhaps for clarity we could also change the guideline for `where` blocks to:

```
statement : Statement
statement =
  some-possibly-long-proof a
  where
  a : type-of-a
  a = construction-of-a
```

fredrik-bakke: Okay, that makes sense. I'll fix it now. Then a more fitting pattern to search for would be something along the lines of
```regex
^\s{2}(?!record|data|module)\S.*\n\s{2}\s\swhere
```

EgbertRijke: Sorry, I can't read patterns, but I'll check the result of applying them:)

fredrik-bakke: No worries, I'm just recording it for future me :)

fredrik-bakke:
I'm not such a fan of `where` blocks in general. It usually (always?) means a bit of logic should be refactored into its own definition. I won't suggest banning them, but I will refactor `cantor-schroder-bernstein` to not use it, as we use that file to demonstrate "exemplary style".

fredrik-bakke:
Actually, I don't have time for this. In the meantime, I will remove it as an exemplary file.

EgbertRijke:
> I'm not such a fan of `where` blocks in general. It usually (always?) means a bit of logic should be refactored into its own definition. I won't suggest banning them, but I will refactor `cantor-schroder-bernstein` to not use it, as we use that file to demonstrate "exemplary style".

I agree with you. The use of `where` blocks is usually a clear sign that refactoring needs to be done. I also don't have time to refactor that CBSE argument though

EgbertRijke:
> Actually, I don't have time for this. In the meantime, I will remove it as an exemplary file.

The Galois connections file could be a great alternative exemplary file

src/elementary-number-theory/addition-integer-fractions.lagda.md
```diff
@@ -131,7 +131,7 @@ associative-add-fraction-ℤ (nx , dx , dxp) (ny , dy , dyp) (nz , dz , dzp) =
   ap-mul-ℤ
     eq-num
     (inv (associative-mul-ℤ dx dy dz))
-  where
+    where
```
EgbertRijke, line 134: This `where` block is still indented more than it should be

src/elementary-number-theory/type-arithmetic-natural-numbers.lagda.md
```diff
@@ -81,18 +81,19 @@ is-split-surjective-map-ℕ+ℕ-to-ℕ (succ-ℕ (succ-ℕ b)) =
...
+  s = concatenate-div-eq-ℕ (x , commutative-mul-ℕ x 2) p
+
+  t : ¬ (div-ℕ 2 (succ-ℕ (2 *ℕ y)))
+  t =
+    ( is-odd-succ-is-even-ℕ
+      ( 2 *ℕ y)
+      ( y , commutative-mul-ℕ y 2))
 is-injective-map-ℕ+ℕ-to-ℕ {inr x} {inl y} p = ex-falso (t s)
-  where
+    where
```
EgbertRijke, line 96: This `where` block is still indented more than it should be

src/elementary-number-theory/type-arithmetic-natural-numbers.lagda.md
```diff
@@ -81,18 +81,19 @@ is-split-surjective-map-ℕ+ℕ-to-ℕ (succ-ℕ (succ-ℕ b)) =
...
-        ( y , commutative-mul-ℕ y 2))
+  s : (div-ℕ 2 (succ-ℕ (2 *ℕ y)))
+  s = concatenate-div-eq-ℕ (x , commutative-mul-ℕ x 2) p
+
+  t : ¬ (div-ℕ 2 (succ-ℕ (2 *ℕ y)))
+  t =
+    ( is-odd-succ-is-even-ℕ
+      ( 2 *ℕ y)
+      ( y , commutative-mul-ℕ y 2))
 is-injective-map-ℕ+ℕ-to-ℕ {inr x} {inl y} p = ex-falso (t s)
```
EgbertRijke, line 95: `ex-falso (t s)` should be on a new line

src/foundation/decidable-equivalence-relations.lagda.md
```diff
@@ -280,7 +280,7 @@ module _
             ( class-Decidable-Equivalence-Relation R a)
             ( pair P p))
           ( α))
-      where
+        where
```
EgbertRijke, line 283: This `where` block is still indented more than it should be

src/foundation/equivalence-extensionality.lagda.md
```diff
@@ -45,7 +45,7 @@ module _
       ( pr2 f)
       ( refl-htpy' (pr1 f))
       ( λ g → equiv-funext)
-    where
+      where
```
EgbertRijke, line 48: This `where` block is still indented more than it should be

src/set-theory/baire-space.lagda.md
```diff
@@ -78,6 +78,6 @@ is-uncountable-baire-space P =
         ( λ F →
           reductio-ad-absurdum
             (pr2 F) (has-no-fixed-points-succ-ℕ (pr1 F))))
-  where
+    where
```
EgbertRijke, line 81: This `where` block is still indented more than it should be

src/set-theory/countable-sets.lagda.md
```diff
@@ -135,7 +135,7 @@ module _
           ( pair
             ( f ∘ (pr1 P))
             ( is-surjective-comp is-surjective-f (pr2 P))))
-    where
+      where
```
EgbertRijke, line 138: This `where` block is still indented more than it should be

src/universal-algebra/algebraic-theory-of-groups.lagda.md
```diff
@@ -88,7 +88,7 @@ group-Theory =
             pair
               (op mul-group-op (var 0 ∷ op unit-group-op empty-vec ∷ empty-vec))
               (var 0)})
-  where
+    where
```
EgbertRijke, line 91: This `where` block is still indented more than it should be

src/universal-algebra/terms-over-signatures.lagda.md
```diff
@@ -140,7 +140,7 @@ element of `A`.
       pair
         ( pr1 (is-nil-lemma p))
         ( all-vec-lemma v (pr2 (is-nil-lemma p)))
-      where
+        where
```
EgbertRijke, line 143: This `where` block is still indented more than it should be

EgbertRijke (commented):
Since there were a lot of `where` blocks that were still indented more than they should be, I was wondering whether your script actually works as intended.

fredrik-bakke:
Thanks for the review, and apologies for being sloppy.

> Since there were a lot of `where` blocks that were still indented more than they should be, I was wondering whether your script actually works as intended.

There's no script currently. But the code I suggested would find the cases that I previously indented correctly (after your first review).

EgbertRijke:
Perhaps I meant regular expression instead of script:)

fredrik-bakke:
By the way, what about `let`-`in` constructions? How should they be indented?

EgbertRijke:
Idk... with common sense?:)

EgbertRijke:
There are also `with` constructions, where we _could_ have indentation conventions... but I feel like here too just common sense is fine at the moment. We don't use them very often.

PR #671, opened 2023-06-24
Whitespace after closing braces and before opening braces

fredrik-bakke:
Adds autoformatting that ensures whitespace after closing braces and before opening braces except for in special cases. From the updated documentation:

> - **Agda space conventions**: Corrects some common spacing mistakes according to
>   our conventions. This includes removing repeat whitespace characters between
>   two non-whitespace characters in a line, having whitespace before an opening
>   brace or semicolon if it is not preceeded by a dot or another opening brace,
>   and having whitespace after a closing brace or semicolon if it is not
>   succeeded by a closing brace.

There are also special cases for the characters `"'@` that I didn't think was necessary to mention there.


EgbertRijke:
Great! Easy merge.

PR #692, opened 2023-08-07
Guidelines for forking the repository and creating branches within clone

EgbertRijke:


HOWTO-INSTALL.md
```diff
@@ -10,17 +10,57 @@ environment for making contributions to the library.
 
 ## Getting a copy of the library
 
-Get a copy of our library on your machine with `git` using
+### Creating a fork of the library
 
-```shell
-git clone --depth 1 git@github.com:UniMath/agda-unimath.git
```
fredrik-bakke, line 16: I think this should be kept. When only using the library, it should not be necessary to have to create one's own fork.

EgbertRijke: Alright, I will add it back, but in my opinion the main guideline should remain to create a fork first:)

PR #694, opened 2023-08-07
Recovering the guidelines for direct clones

EgbertRijke:


EgbertRijke:
@fredrik-bakke this one is specially for you :-)

fredrik-bakke:
In the first case, I guess it is fine without merging this PR, but in the second case, I would actually suggest moving the part about forking and branching to `CONTRIBUTING.MD` instead.

fredrik-bakke:
I guess what stands out to me is that cloning is less of a committing act than forking. Cloning is something that I do to get code locally to use or look at it. Forking is a much more deliberate act that I do when I have a plan for a contribution. To me, just cloning fits the bill better in this instance.

EgbertRijke:
Maybe I'm wrong about it, but I consider the main users base potential contributors. At least for now I don't see many people merely using the library, but of course that could change over time.

My experience with young students who are not very well-versed with git is that they just clone the library, and then get confused and have to change their setup. They have no reference point for how it should be done. I would say that these guidelines should be right there with the installation guidelines, because the initial setup is important and especially inexperienced users need to get the right information instead of some information that also works but that would require them to change their setup later.

EgbertRijke:
> I guess what stands out to me is that cloning is less of a committing act than forking. Cloning is something that I do to get code locally to use or look at it. Forking is a much more deliberate act that I do when I have a plan for a contribution. To me, just cloning fits the bill better in this instance.

But isn't what I have written in this PR clear about this?

fredrik-bakke:
> But isn't what I have written in this PR clear about this?

It's clear, but maybe the ordering of things can be improved a little. How about starting _Getting a copy of the library_ with a brief comment that we recommend the user create a fork immediately to avoid headaches at a later point, but that they are certainly welcome to clone the library without forking if they prefer.

HOWTO-INSTALL.md
```diff
@@ -36,6 +36,19 @@ size.
...
+
+```bash
+git clone --depth 1 git@github.com:UniMath/agda-unimath.git
+```
+
+This is not our first recommendation, however, because if you later decide to
+use this clone for contributions, you will need some proficiency in using `git`
+to adjust the remote addresses of your clone.
+
 ### Creating a branch within your clone of the library
```
fredrik-bakke, line 52: Given my other suggestion, I think this part can be removed.

````suggestion
```

### Creating a branch within your clone of the library
````

fredrik-bakke: BTW, did you know you could insert suggestions like this directly in GitHubs web interface by clicking this button:  
image
😄

EgbertRijke: No I didn't know:)

I added a few sentences at the start of this section, but given what I added I chose to keep the sentence about git remotes where it was.

EgbertRijke:
I noticed that too, but let's do that in another PR. Thanks for your suggestions!

PR #695, opened 2023-08-07
adding overview to installation guide

EgbertRijke:


EgbertRijke:
Question for @VojtechStep. When you set the up library with nix, does it install only the right version of Agda or also all the other prerequisites. Right now, nix is only mentioned in the `Installing agda` section, but if it is not only about installing agda, then it is perhaps not the best place for that mention of nix.

HOWTO-INSTALL.md
```diff
@@ -1,14 +1,57 @@
...
-step-by-step directions, you will create a tailored working environment for
-using the library. Additionally, we provide instructions for setting up your
-environment for making contributions to the library.
+## Overview
 
-## Getting a copy of the library
+### Quick setup
+
+To work or experiment with the `agda-unimath` library on your machine, you will
+need to have `agda` version 2.6.3 installed on your machine, and a suitable
```
VojtechStep, line 8: The repeating of "on your machine" sounds unnatural, I'd leave out the second occurrence (I think it should be clear from context).

VojtechStep:
> Question for @VojtechStep. When you set the up library with nix, does it install only the right version of Agda or also all the other prerequisites. Right now, nix is only mentioned in the `Installing agda` section, but if it is not only about installing agda, then it is perhaps not the best place for that mention of nix.

The Nix devshell makes the following tools available:
- Agda 2.6.3
- Python 3.8 with the `requests` module
- `pre-commit`
- `make`
- mdbook, along with the `katex`, `pagetoc`, `linkcheck` and `catppuccin` plugins
- (some standard command-line utilities such as `grep`, `time`, `sed`, ... - these are included automatically in every Nix devshell, so they are not specified in the flake.nix file; `make` is actually one of those)

which should be all the tools necessary for working on the library.

EgbertRijke:
So it doesn't install git?

HOWTO-INSTALL.md
EgbertRijke: Thanks for noticing that! I removed the second mention.

HOWTO-INSTALL.md
```diff
@@ -1,14 +1,56 @@
...
+   ```shell
+   python3 -m pip install -r scripts/requirements.txt
+   ```
+5. `rust` version 1.68 or newer
+6. `mdbook` version 4.34 or newer, along with the following plugins:
+
+   - `mdbook-linkcheck`
+   - `mdbook-katex`
+   - `mdbook-pagetoc`
+   - `mdbook-catppuccin`
```
EgbertRijke, line 34: Issue #696 came up here, where I would like to use the double whitespace trick to prevent the renderer from adding extra space between all the items. The way this enumerated list is written now, it will output as

```text
1. ...

2. ...

3. ...
```

which is unfortunately very ugly. But there's no way around it, because double whitespaces are removed by the `pre-commit` hook.

VojtechStep:
No, it doesn't. I guess it didn't make sense to me to include git there, because the devshell is for contributors, and they already use git to obtain the library (technically they can use the devshell via `nix shell github:UniMath/agda-unimath` without actually cloning the library, but you typically wouldn't do that).

HOWTO-INSTALL.md
EgbertRijke: Screenshot 2023-08-07 at 23 55 43

EgbertRijke:
That's totally fine! It was just to be clear about what is included and what isn't. Many thanks!

PR #710, opened 2023-08-24
Allow misc. binary operators over max line length

fredrik-bakke:
Relaxes the max line length checker to allow for lines ending with miscellaneous binary operators. The criteria I have placed is that the binary operator is at most two characters, and starts with a character that is not alphanumeric. To be precise, it must satisfy:
```regexp
[^a-zA-Z\d\s"_({⦃][^\s"_.;(){}⦃@]?
```
This describes that the first character is not a character between `a` and `z`, not a character between `A` and `Z`, not a digit (`\d`), not whitespace (`\s`) and none of the characters `"_({⦃`. The `[^\s"_.;(){}@⦃]?`-part says that the second character is optional, and must be non-whitespace and none of the characters `"_.;(){}⦃@`.


fredrik-bakke:
Short answer: yes.

PR #711, opened 2023-08-27
Some computations for different notions of equivalence

fredrik-bakke:
### Overview
- Move lemmas regarding `precomp` and `postcomp` in `equivalences` to `functoriality-function-types`
- Fill in lemmas so that `functoriality-dependent-function-types` and `functoriality-function-types` are similar
- Factor invertible maps into their own file `invertible-maps`
- Rename `has-inverse` to `is-invertible` (this makes the naming scheme consistent for all the different notions of equivalence. I feel like we've had a discussion about this before, and if you disagree with this change it is easy to revert.)
- Clean up `logical-equivalences`
- Compute the fibers of `iff-equiv`
- Characterize equality of logical equivalences
- Characterize equality of invertible maps
- Characterize equality of involutions
- Characterize equality of fibered equivalences and equivalences over a map
- Show that `htpy-lifting-square` is reflexive (it remains to show that the total space is contractible, but I'm at a bit of a loss for how to proceed with that)
- Add some miscellaneous path-algebra lemmas (this file needs some refactoring)
- Define local maps in `orthogonal-factorization-systems.local-maps`
- Redefine the pullback-hom in terms of fibered maps instead
- Compute the fibers of `pullback-hom`

Builds on #432.


src/foundation-core/functoriality-function-types.lagda.md
```diff
@@ -108,6 +160,140 @@ pr2 (equiv-postcomp A e) =
...
+    is-equiv-is-equiv-precomp f is-equiv-precomp-f =
+      is-equiv-is-equiv-precomp-subuniverse
+        ( λ l → l1 ⊔ l2)
+        ( λ l X → A → B)
+        ( A , f)
+        ( B , f)
+        ( f)
+        ( λ l C → is-equiv-precomp-f l (pr1 C))
+```
+
```
EgbertRijke, line 268: This code block has no section title.

Also, it feels like the things in the following code block don't really belong in this file, because they are about propositions, sets, and truncated types.

src/foundation-core/invertible-maps.lagda.md
```diff
@@ -0,0 +1,145 @@
...
+## Idea
+
+An **inverse** for a map `f : A → B` is a map `g : B → A` equipped with
+[homotopies](foundation-core.homotopies.md) ` (f ∘ g) ~ id` and `(g ∘ f) ~ id`.
+Such data, however is [structure](foundation.structure.md) on the map `f`, and
+not generally a [property](foundation-core.propositions.md).
+
+## Definition
+
+### The type of inverse proofs
```
EgbertRijke, line 31: I would say: ``### The predicate that a map `g` is an inverse of a map `f` ``

The phrase "The type of inverse proofs" looks so clumsy

EgbertRijke, line 31:
```suggestion
### The predicate that a map `g` is an inverse of a map `f`
```

src/foundation-core/invertible-maps.lagda.md
```diff
@@ -0,0 +1,145 @@
...
+  is-section-is-inverse :
+    {f : A → B} {g : B → A} → is-inverse f g → (f ∘ g) ~ id
+  is-section-is-inverse = pr1
+
+  is-retraction-is-inverse :
+    {f : A → B} {g : B → A} → is-inverse f g → (g ∘ f) ~ id
+  is-retraction-is-inverse = pr2
+```
+
+### The type of inverses of a map
```
EgbertRijke, line 50:
```suggestion
### The predicate that a map `f` is invertible
```

src/foundation-core/invertible-maps.lagda.md
```diff
@@ -0,0 +1,145 @@
...
+  is-retraction-map-invertible-map :
+    (f : invertible-map A B) →
+    (map-invertible-map f ∘ map-inv-invertible-map f) ~ id
+  is-retraction-map-invertible-map =
+    is-retraction-is-invertible ∘ is-invertible-map-invertible-map
+```
+
+## Properties
+
+### The inverse invertible map
```
EgbertRijke, line 118:
```suggestion
### The invertible inverse of an invertible map
```

src/foundation/equivalences.lagda.md
```diff
@@ -424,13 +296,19 @@ module _
 
 ### The groupoid laws for equivalences
 
+#### Equivalence composition is associative
```
EgbertRijke, line 299:
```suggestion
#### Composition of equivalences is associative
```

src/foundation/equivalences.lagda.md
```diff
@@ -424,13 +296,19 @@ module _
...
+
 ```agda
 associative-comp-equiv :
   {l1 l2 l3 l4 : Level} {A : UU l1} {B : UU l2} {C : UU l3} {D : UU l4} →
   (e : A ≃ B) (f : B ≃ C) (g : C ≃ D) →
   ((g ∘e f) ∘e e) ＝ (g ∘e (f ∘e e))
 associative-comp-equiv e f g = eq-equiv-eq-map-equiv refl
+```
+
+#### Unit laws for equivalence composition
```
EgbertRijke, line 309:
```suggestion
#### Unit laws for composition of equivalences
```

src/foundation/equivalences.lagda.md
```diff
@@ -440,15 +318,23 @@ module _
 
   right-unit-law-equiv : (e : X ≃ Y) → (e ∘e id-equiv) ＝ e
   right-unit-law-equiv e = eq-equiv-eq-map-equiv refl
+```
 
+#### Inverse laws for equivalence composition
```
EgbertRijke, line 323:
```suggestion
#### Inverse laws for composition of equivalences
```

src/foundation/equivalences.lagda.md
```diff
@@ -457,21 +343,29 @@ module _
...
       ( inv-equiv)
       ( inv-inv-equiv')
       ( inv-inv-equiv)
 
   equiv-inv-equiv : (X ≃ Y) ≃ (Y ≃ X)
   pr1 equiv-inv-equiv = inv-equiv
   pr2 equiv-inv-equiv = is-equiv-inv-equiv
+```
+
+#### Higher coherence between the unit laws for equivalence composition
```
EgbertRijke, line 356:
```suggestion
#### A coherence law for the unit laws for composition of equivalences
```

src/foundation/equivalences.lagda.md
```diff
@@ -488,51 +382,73 @@ module _
...
+
+#### Iterated inverse laws for equivalence composition
 
-comp-inv-equiv-comp-equiv :
+```agda
+left-left-inverse-law-equiv :
   {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {C : UU l3}
   (f : B ≃ C) (e : A ≃ B) → (inv-equiv f ∘e (f ∘e e)) ＝ e
-comp-inv-equiv-comp-equiv f e =
+left-left-inverse-law-equiv f e =
```
EgbertRijke, line 393: I don't really understand the rationale behind this name change. The previous name was clearer in my opinion, and I don't think `left-left-inverse-law-equiv` is part of any existing naming scheme.

How about `is-retraction-postcomp-inv-equiv`? This fits in a naming scheme with `is-section-postcomp-inv-equiv`, `is-retraction-precomp-inv-equiv` and `is-section-precomp-inv-equiv`.

src/foundation/equivalences.lagda.md
```diff
@@ -488,51 +382,73 @@ module _
...
+  (f : B ≃ C) (e : B ≃ A) →
+  ((f ∘e inv-equiv e) ∘e e) ＝ f
+right-left-inverse-law-equiv f e =
+  eq-htpy-equiv (λ x → ap (map-equiv f) (is-retraction-map-inv-equiv e x))
+```
+
+### Left and right composition by an equivalence defines equivalences
+
+```agda
+is-equiv-left-comp-equiv :
```
EgbertRijke, line 421: I don't think we should speak of "left composition" and "right composition". The usual terminology is "postcomposition" and "precomposition"

src/foundation/equivalences.lagda.md
```diff
@@ -488,51 +382,73 @@ module _
...
+
+right-left-inverse-law-equiv :
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {C : UU l3}
+  (f : B ≃ C) (e : B ≃ A) →
+  ((f ∘e inv-equiv e) ∘e e) ＝ f
+right-left-inverse-law-equiv f e =
+  eq-htpy-equiv (λ x → ap (map-equiv f) (is-retraction-map-inv-equiv e x))
+```
+
+### Left and right composition by an equivalence defines equivalences
```
EgbertRijke, line 418:
```suggestion
### The Postcomposition and precomposition operations by an equivalance are equivalences
```

src/foundation/equivalences.lagda.md
```diff
@@ -488,51 +382,73 @@ module _
...
+    ( _∘e inv-equiv e)
+    ( λ f → right-left-inverse-law-equiv f e)
+    ( λ f → right-right-inverse-law-equiv f e)
 
 equiv-postcomp-equiv :
   {l1 l2 l3 : Level} {B : UU l2} {C : UU l3} →
   (f : B ≃ C) → (A : UU l1) → (A ≃ B) ≃ (A ≃ C)
 pr1 (equiv-postcomp-equiv f A) e = f ∘e e
-pr2 (equiv-postcomp-equiv f A) = is-equiv-comp-equiv f A
+pr2 (equiv-postcomp-equiv f A) = is-equiv-left-comp-equiv f A
```
EgbertRijke, line 443: Here it shows that the name is clearly off

src/foundation/equivalences.lagda.md
fredrik-bakke: Sure, that sounds like a more reasonable naming scheme. Sorry, this was the first scheme I could think of. I introduced similar lemmas for identification concatenation as well, but I guess preconcat and postconcat make sense too

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -131,8 +132,11 @@ module _
       ( emb-subtype (is-equiv-Prop ∘ pr1))
       ( emb-equiv equiv-Σ-is-equiv-equiv-over)
 
-  map-emb-map-over-equiv-over : equiv-over f g i → map-over f g i
-  map-emb-map-over-equiv-over = map-emb emb-map-over-equiv-over
+  map-map-over-equiv-over : equiv-over f g i → map-over f g i
```
EgbertRijke, line 135: In the previous name I liked that by the prefix `map-emb` it clearly showed the intention that it was the underlying map of an embedding. If we're going to shorten this name, I think I like `map-over-equiv-over` better. We never use `map-abc-xyz` as a naming scheme for maps from `xyz` to `abc`. We'd simply use the name `abc-xyz` for such cases.

src/foundation-core/functoriality-function-types.lagda.md
fredrik-bakke: I agree. They didn't really fit in `equivalences` either, where I moved them from. Perhaps some file about subuniverses would be better for these theorems

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -218,6 +222,86 @@ module _
 
   map-fibered-map-fibered-equiv : fibered-equiv f g → fibered-map f g
```
EgbertRijke, line 223: In line with my previous comment, this should probably be called `fibered-map-fibered-equiv`

src/foundation/functoriality-function-types.lagda.md
```diff
@@ -25,40 +29,88 @@ open import foundation-core.truncation-levels
...
+      ( is-equiv-postcomp-equiv f A')
+      ( is-equiv-precomp-equiv (inv-equiv e) B)
+
+  equiv-function-type : (A' → B') ≃ (A → B)
+  pr1 equiv-function-type = map-equiv-function-type
+  pr2 equiv-function-type = is-equiv-map-equiv-function-type
 ```
 
-### The precomposition function preserves homotopies
+### A map is truncated iff postcomposition by it is truncated
```
EgbertRijke, line 63: This feels more like something that belongs in the file about truncated maps

src/foundation/invertible-maps.lagda.md
```diff
@@ -0,0 +1,291 @@
...
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f : A → B}
+  where
+
+  coherence-htpy-is-invertible :
+    (s t : is-invertible f) →
+    map-inv-is-invertible s ~ map-inv-is-invertible t → UU (l1 ⊔ l2)
+  coherence-htpy-is-invertible s t H =
+    ( ( is-retraction-is-invertible s) ~
```
EgbertRijke, line 70: Actually, there should be `coherence-htpy-is-section` and `coherence-htpy-is-retraction` in the files about sections and retractions, and this definition should be reusing those characterizations of equality.

src/foundation/invertible-maps.lagda.md
```diff
@@ -0,0 +1,291 @@
...
+      ( is-contr-total-htpy (map-invertible-map s))
+      ( map-invertible-map s , refl-htpy)
+      ( is-contr-total-Eq-structure
+        ( λ x z →
+          coherence-htpy-invertible-map s
+            ( map-invertible-map s , x , z)
+            ( refl-htpy))
+        ( is-contr-total-htpy (map-inv-invertible-map s))
+        ( map-inv-invertible-map s , refl-htpy)
+        (is-contr-total-Eq-structure
```
EgbertRijke, line 173:
```suggestion
        ( is-contr-total-Eq-structure
```

fredrik-bakke:
Thank you for the comments! I will try and address them tomorrow

src/foundation/path-algebra.lagda.md
```diff
@@ -45,7 +45,7 @@ coherences on the boundary 1-paths.
 
 ## Properties
 
-### the unit laws of concatination induce homotopies
+### the unit laws of concatenation induced homotopies
```
EgbertRijke, line 48:
```suggestion
### The unit laws of concatenation induce homotopies
```

Present tense conveys the meaning better.

src/foundation/path-algebra.lagda.md
```diff
@@ -117,30 +117,72 @@ vertical-concat-square {a = a} {f = f}
...
+```
+
+### Iterated inverse laws
+
+```agda
+module _
+  {l : Level} {A : UU l}
+  where
+
+  left-left-inv :
```
EgbertRijke, line 171:
```suggestion
  is-retraction-left-concat-inv :
```

src/foundation/path-algebra.lagda.md
```diff
@@ -117,30 +117,72 @@ vertical-concat-square {a = a} {f = f}
...
+    {x y z : A} (p : x ＝ y) (q : y ＝ z) → (inv p ∙ (p ∙ q)) ＝ q
+  left-left-inv refl q = refl
+
+  left-right-inv :
```
EgbertRijke, line 175:
```suggestion
  is-section-left-concat-inv :
```

src/foundation/path-algebra.lagda.md
```diff
@@ -117,30 +117,72 @@ vertical-concat-square {a = a} {f = f}
...
+    {x y z : A} (p : x ＝ y) (q : x ＝ z) → (p ∙ (inv p ∙ q)) ＝ q
+  left-right-inv refl q = refl
+
+  right-left-inv :
```
EgbertRijke, line 179:
```suggestion
  is-section-right-concat-inv :
```

src/foundation/path-algebra.lagda.md
```diff
@@ -117,30 +117,72 @@ vertical-concat-square {a = a} {f = f}
...
+    {x y z : A} (p : x ＝ y) (q : z ＝ y) → ((p ∙ inv q) ∙ q) ＝ p
+  right-left-inv refl refl = refl
+
+  right-right-inv :
```
EgbertRijke, line 183:
```suggestion
  is-retraction-right-concat-inv :
```

EgbertRijke (changes requested):
There is a lot of good work in this PR. I reviewed 100 files so far, but it is taking my quite long so I leave it here for now.

fredrik-bakke:
Thank you so much! I'm really sorry for making you review so many files. In the other PR, I can revert the naming change of `foundation.transport` to make the size more manageable

src/foundation/path-algebra.lagda.md
EgbertRijke: I'm also fine with `preconcat` and `postconcat` although we don't use those words yet in the library.

src/foundation-core/invertible-maps.lagda.md
fredrik-bakke: I wanted to avoid "predicate", but I guess it won't really cause confusion

src/foundation/equivalences.lagda.md
EgbertRijke: Good suggestion! I like this even more than my original suggestion.

src/foundation/invertible-maps.lagda.md
fredrik-bakke: So, homotopies of retractions are defined in terms of homotopies in the coslice. Do you prefer if homotopies of sections are similarly characterized in terms of homotopies in the slice, or do you want an independent definition of homotopies of retractions?

Personally, I find retractions to be a more central notion than homomorphisms in the coslice, so perhaps we can swap directions and define homotopies in the coslice in terms of homotopies of retractions?

EgbertRijke: I'm not sure what you have in mind. How would you define homotopies in the coslice as homotopies of retractions?

src/foundation/fibered-equivalences.lagda.md
fredrik-bakke: Since these maps can also be seen as projections, I think I prefer the name `map-over-equiv-over`.

EgbertRijke: Great! Then we agree that `map-over-equiv-over` is to be preferred over `map-map-over-equiv-over`

src/foundation/invertible-maps.lagda.md
fredrik-bakke:
```agda
module _
  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f : A → B}
  where

  htpy-retraction : retraction f → retraction f → UU (l1 ⊔ l2)
  htpy-retraction = htpy-hom-coslice
```

fredrik-bakke: Ah, I guess the other way around doesn't really make sense. Sorry, I'm writing to you while attending a lecture

EgbertRijke: That's defining `htpy-retraction` in terms of `htpy-hom-coslice`. I understand this direction, but I thought you were suggesting the other direction, i.e., defining `htpy-hom-coslice` in terms of `htpy-retraction`.

EgbertRijke: Must be an interesting lecture:)

fredrik-bakke: My apologies, I am just wrong. So do you want `htpy-section` defined in terms of `htpy-hom-slice`, or to have independent definitions for both sections and retractions?

fredrik-bakke: I think it is, I'm just too stupid to follow it

EgbertRijke: I would not be very sure if that isn't going to create cyclic dependencies. In my above comment I was just suggesting to reuse the characterization of identity types of sections in the characterization of identity types of invertible maps. That would require factoring out `coherence-htpy-section` from the definition of `htpy-section`, but once that is done it can be reused in the definition of `coherence-htpy-is-invertible`. 

So I think I'm suggesting something simpler, but indeed it is also possible to do it in terms of `htpy-hom-coslice`. If that goes through I would be happy with that idea as well.

fredrik-bakke: Okay, I will have to look into this another day

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -131,8 +132,11 @@ module _
       ( emb-subtype (is-equiv-Prop ∘ pr1))
       ( emb-equiv equiv-Σ-is-equiv-equiv-over)
 
-  map-emb-map-over-equiv-over : equiv-over f g i → map-over f g i
-  map-emb-map-over-equiv-over = map-emb emb-map-over-equiv-over
+  map-over-equiv-over : equiv-over f g i → map-over f g i
+  map-over-equiv-over = map-emb emb-map-over-equiv-over
+
+  is-emb-map-over-equiv-over : is-emb map-over-equiv-over
+  is-emb-map-over-equiv-over = is-emb-map-emb emb-map-over-equiv-over
```
EgbertRijke, line 139: It would be reasonable to also include `emb-map-over-equiv-over` below, for the embedding.

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -216,8 +220,88 @@ module _
       ( emb-subtype is-fibered-equiv-fibered-map-Prop)
       ( emb-equiv equiv-Σ-is-fibered-equiv-fibered-map-fibered-equiv)
 
-  map-fibered-map-fibered-equiv : fibered-equiv f g → fibered-map f g
-  map-fibered-map-fibered-equiv = map-emb emb-fibered-map-fibered-equiv
+  fibered-map-fibered-equiv : fibered-equiv f g → fibered-map f g
+  fibered-map-fibered-equiv = map-emb emb-fibered-map-fibered-equiv
+
+  is-emb-fibered-map-fibered-equiv : is-emb fibered-map-fibered-equiv
```
EgbertRijke, line 226: You could also include `emb-fibered-map-fibered-equiv` below

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -216,8 +220,88 @@ module _
...
+      ( map-over-equiv-over f g i e)
+      ( map-over-equiv-over f g i e'))
+  htpy-eq-equiv-over e e' = map-equiv (extensionality-equiv-over e e')
+
+  eq-htpy-equiv-over :
+    (e e' : equiv-over f g i) →
+    ( htpy-map-over f g i
+      ( map-over-equiv-over f g i e)
+      ( map-over-equiv-over f g i e')) →
+    ( e ＝ e')
```
EgbertRijke, line 265:
```suggestion
    htpy-map-over f g i
      ( map-over-equiv-over f g i e)
      ( map-over-equiv-over f g i e') →
    e ＝ e'
```

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -216,8 +220,88 @@ module _
...
+    (e e' : fibered-equiv f g) →
+    ( e ＝ e') ≃
+    ( htpy-fibered-map f g
+      ( fibered-map-fibered-equiv f g e)
+      ( fibered-map-fibered-equiv f g e'))
+  extensionality-fibered-equiv e e' =
+    ( extensionality-fibered-map f g
+      ( fibered-map-fibered-equiv f g e)
+      ( fibered-map-fibered-equiv f g e')) ∘e
+    ( ap (fibered-map-fibered-equiv f g) ,
```
EgbertRijke, line 287: This could be `equiv-ap-emb (emb-fibered-map-fibered-equiv f g)` instead of the explicit pair, if you would define `emb-fibered-map-fibered-equiv`. That would slightly simplify this proof term, because having an explicit pair looks odd here.

src/foundation/fibered-equivalences.lagda.md
```diff
@@ -216,8 +220,88 @@ module _
...
+      ( fibered-map-fibered-equiv f g e)
+      ( fibered-map-fibered-equiv f g e'))
+  htpy-eq-fibered-equiv e e' = map-equiv (extensionality-fibered-equiv e e')
+
+  eq-htpy-fibered-equiv :
+    (e e' : fibered-equiv f g) →
+    ( htpy-fibered-map f g
+      ( fibered-map-fibered-equiv f g e)
+      ( fibered-map-fibered-equiv f g e')) →
+    ( e ＝ e')
```
EgbertRijke, line 303:
```suggestion
    htpy-fibered-map f g
      ( fibered-map-fibered-equiv f g e)
      ( fibered-map-fibered-equiv f g e') →
    e ＝ e'
```

src/foundation/path-algebra.lagda.md
```diff
@@ -117,30 +117,72 @@ vertical-concat-square {a = a} {f = f}
...
+    {x y z : A} (p : x ＝ y) (q : y ＝ z) →
+    assoc p refl q ＝ ap (_∙ q) (right-unit)
+  assoc-middle-refl refl q = refl
+
+  assoc-right-refl :
+    {x y z : A} (p : x ＝ y) (q : y ＝ z) →
+    assoc p q refl ＝ (right-unit ∙ ap (p ∙_) (inv right-unit))
+  assoc-right-refl refl refl = refl
+
+  unit-law-assoc-101' :
```
EgbertRijke, line 139: It feels odd to have half a naming scheme `assoc-[left/middle/right]-refl` and half a naming scheme `unit-law-assoc-xxx`. At least it was more consistent previously.

src/foundation/logical-equivalences.lagda.md
```diff
@@ -63,13 +192,55 @@ pr1 (equiv-equiv-iff P Q) = equiv-iff' P Q
...
-  {l1 l2 : Level} (P : Prop l1) (Q : Prop l2) → is-prop (P ⇔ Q)
-is-prop-logical-equivalence P Q =
-  is-prop-prod
-    ( is-prop-function-type (is-prop-type-Prop Q))
-    ( is-prop-function-type (is-prop-type-Prop P))
+module _
+  {l1 l2 l3 l4 : Level} {I : UU l1} {J : UU l2} {A : I → UU l3} {B : J → UU l4}
+  where
+
+  iff-Π :
```
EgbertRijke, line 202: The specification of this operation looks a little strange to me, and I'm not convinced it is in its most useful form. Could you explain why you gave it this specification?

src/orthogonal-factorization-systems/extensions-of-maps.lagda.md
```diff
@@ -365,25 +376,25 @@ module _
   equiv-fib-precomp-extension-Π f =
     (equiv-fib'-precomp-extension-Π f) ∘e (equiv-fib (precomp-Π i P) f)
 
-  equiv-is-contr-extension-Π-is-local-family :
-    is-local-family i P ≃
+  equiv-is-contr-extension-Π-is-local-Π :
```
EgbertRijke, line 379: What was wrong with the terminology `is-local-family`?

src/foundation/logical-equivalences.lagda.md
fredrik-bakke: The idea here was to mimic `equiv-Pi` if I recall correctly

src/orthogonal-factorization-systems/local-types.lagda.md
```diff
@@ -53,11 +51,11 @@ module _
   {l1 l2 : Level} {Y : UU l1} {X : UU l2} (f : Y → X)
   where
 
-  is-local-family : {l : Level} → (X → UU l) → UU (l1 ⊔ l2 ⊔ l)
-  is-local-family A = is-equiv (precomp-Π f A)
+  is-local-Π : {l : Level} → (X → UU l) → UU (l1 ⊔ l2 ⊔ l)
```
EgbertRijke, line 54: Given how we name things in agda-unimath, `is-local-Π` reads like the property that a dependent function type is local. That is not what is asserted here, so the name `is-local-Π` doesn't look right to me.

fredrik-bakke: I agree, I just didn't come up with a better name. Originally it was called `is-local'` I believe. Maybe the best name would be `is-local-dependent-type`. (Then we should make a similar renaming for extension types)

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -99,26 +189,57 @@ module _
...
-      ( postcomp A g)
-      ( precomp f' Y')
-      ( postcomp A' g')
+    map-is-pullback
+      ( precomp f Y) (postcomp A g)
+      ( precomp f' Y') (postcomp A' g')
+      ( cone-pullback-hom f g)
+      ( cone-pullback-hom f' g')
+      ( is-pullback-fibered-map f g)
+      ( is-pullback-fibered-map f' g')
```
EgbertRijke, line 203:
```suggestion
    map-is-pullback
      ( precomp f Y)
      ( postcomp A g)
      ( precomp f' Y')
      ( postcomp A' g')
      ( cone-pullback-hom f g)
      ( cone-pullback-hom f' g')
      ( is-pullback-fibered-map f g)
      ( is-pullback-fibered-map f' g')
```

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -99,26 +189,57 @@ module _
   where
 
   map-pullback-hom :
-      hom-cospan (precomp f' Y') (postcomp A' g') (precomp f Y) (postcomp A g) →
-      type-pullback-hom f' g' → type-pullback-hom f g
+    hom-cospan
+      ( precomp f' Y') (postcomp A' g')
+      ( precomp f Y) (postcomp A g) →
```
EgbertRijke, line 194:
```suggestion
    hom-cospan
      ( precomp f' Y')
      ( postcomp A' g')
      ( precomp f Y)
      ( postcomp A g) →
```

src/foundation/fibered-equivalences.lagda.md
fredrik-bakke: That's nice. Thanks for the tip!

src/foundation/path-algebra.lagda.md
fredrik-bakke: I agree, I just didn't want to touch the old stuff. To be honest I want to remove it

src/orthogonal-factorization-systems/local-maps.lagda.md
```diff
@@ -0,0 +1,103 @@
...
+
+  is-local-map-Prop :
+    {l3 l4 : Level} {A : UU l3} {B : UU l4}
+    (g : A → B) → Prop (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  is-local-map-Prop g = is-local-family-Prop (fib g)
+```
+
+### A type `B` is `f`-local if and only if the terminal projection `B → unit` is `f`-local
+
+This remains to be formalized.
```
EgbertRijke, line 94: We should start the habit of opening issues for this:)

EgbertRijke:
I reviewed all the files now. If we address all the issues, this PR will be ready to be merged. Thanks for doing all this work!

src/foundation/path-algebra.lagda.md
EgbertRijke: Such laws are actually coming up in A2-spaces. I don't think they should be removed.

fredrik-bakke:
Great! I'll address the remaining ones as soon as I have time. Thanks again for taking the time

src/foundation/logical-equivalences.lagda.md
EgbertRijke: The problem is that with logical equivalences, there are a lot more nonequivalent ways to construct equivalences between Pi-types. I don't think it is clear that this is the preferred/canonical choice. With equivalences, things are a bit more restrictive because you have a 1-1 correspondence between elements of the domain and codomain.

I also wonder if this form of `iff-Π` is actually used somewhere. I don't recall seeing it showing up in this PR, but I might just have missed it. If there are no uses for `iff-Π` then I might have a preference to rename `iff-Π-iff-fam` back to `iff-Π` because that one actually sees some use.

src/orthogonal-factorization-systems/local-maps.lagda.md
fredrik-bakke: That's actually not a terribly bad idea! We'll generate a lot of good first issues this way, and it might also give a hint as to what people are working on/interested in

EgbertRijke: I always thought that "good first issue" means that it is a good issue and it is the first issue from a new contributor. But now you seem to say that it is more like a good first task to work on for a new contributor. Perhaps we should have a clearer label than "good first issue" because to be honest I thought about removing that label altogether because I thought it was so useless:)

EgbertRijke: Is there also a way to link back to issues in agda-unimath website? Then we could mention that in those "remains to be formalized" remarks.

fredrik-bakke: > I always thought that "good first issue" means that it is a good issue and it is the first issue from a new contributor. But now you seem to say that it is more like a good first task to work on for a new contributor. Perhaps we should have a clearer label than "good first issue" because to be honest I thought about removing that label altogether because I thought it was so useless:)

Oh, haha, the latter is the generally accepted meaning I think. 😅 I've been a bit sloppy with how I use it, so sorry if that's caused further confusion. What do you propose instead? Something like `newcomer friendly`? For what it's worth,  I do believe `good first issue` is pretty widely used.

fredrik-bakke: Appending `[#xyz](link)` should suffice

src/foundation/logical-equivalences.lagda.md
fredrik-bakke: So, just to make sure. The definition `iff-Π-iff-fam` imitates `equiv-Π-equiv-fam`. Do you want to break this pattern as well? I understand that logical equivalences and equivalences are very different notions, but still, given its name and usage, I was under the impression that it should imitate equivalences. The principle I was following when making these changes was that every construction in logical equivalences with a corresponding construction in equivalences is mapped to it under `iff-equiv`.

src/foundation/fibered-equivalences.lagda.md
fredrik-bakke: This is defined 10 lines further up.

src/foundation/fibered-equivalences.lagda.md
fredrik-bakke: This is defined 9 lines further up.

src/foundation/path-algebra.lagda.md
fredrik-bakke: Alright, how do you feel about the current names? I don't like that `transpose-assoc-middle-refl` introduces a new naming scheme, but I didn't find the alternative `assoc-middle-refl'` to be much better.

fredrik-bakke: Also, I just realized that the `con` in `inv-con` stands for concatenation. However, we don't use the abbreviation `con` for concatenation anywhere else (`(?<!inv-)\bcon\b(?!-inv)`). Perhaps we can give `inv-con` and `con-inv` new names? Something like `left-transpose` and `right-transpose`

EgbertRijke: I agree, but please in another PR that is just for that purpose.

EgbertRijke: I still like to call them unit laws for associators, and I don't see why it is a good idea to give that up. It aligns well with unit laws for binary operations, and to me there is no reason why multivariable operators such as associators shouldn't have unit laws. But I like your [left/middle/right], so my preferred naming would be `[left/middle/right]-unit-law-assoc`.

EgbertRijke: But actually the point with the binary number notation had a purpose. Actually there are 7 unit laws for the associator: Three of them have refl in one of the arguments, 3 of them have refl in two of the arguments, and one of them is a coherence that puts them all together and it has refl in all three of the arguments. The old naming scheme was supposed to make naming the other 4 unit laws (or coherences) easy to name.

EgbertRijke: With the pentagonator there will be 15 unit laws. More generally, n-ary operations may have 2^n -1 unit laws.

fredrik-bakke: The binary numbering scheme makes sense to me. However, currently, there are 6 unit laws recorded for the associator where only one of the arguments is reflexivity, not just 3. The general naming scheme seems to have been to just append primes for different variants of the same unit law. It makes sense that there are more unit laws for some of the positions since concatenation computes on the left and not on the right, but I don't see the pattern for which variant has fewer primes than the other. In particular, I have a problem with what was defined as "the" right unit law. What previously had the name `unit-law-assoc-110` I expected to have type `assoc p q refl ＝ ...`. However, instead, it was an equality of type `(assoc p q refl ∙ ap (concat p z) right-unit) ＝ ...`, and the unit law that I was looking for was not recorded at all.

EgbertRijke: That's perhaps because I transposed them in order to state them free of inverses.

EgbertRijke: Probably the reason I did that was because then it will be easier to state the coherences between them (with refl twice or three times in the arguments of assoc)

EgbertRijke: When it would have been stated using `coherence-triangle-identifications` it would have been clearer, because `assoc p q refl` would just be one of the sides of the triangle.

fredrik-bakke: Ah, okay, that makes sense. So then maybe it _is_ a good idea to use two different naming schemes. One for the collection of unit laws and their coherences, and one for computing with the associator.

EgbertRijke: Lol, but then we would both win

EgbertRijke: Anyway, that sounds good. Perhaps keep then near each other and add a short explanation of what is going on.

fredrik-bakke: Oh, shoot, we can't have that

fredrik-bakke: Yep, sounds good!

src/foundation/logical-equivalences.lagda.md
EgbertRijke: Is `equiv-Π-equiv-fam` the same as `equiv-map-Π`?

fredrik-bakke: Ohh, whoops, I see I reverted that change. Sorry for the confusion. Then perhaps `iff-Π-iff-fam` should be named `iff-map-Π`?

EgbertRijke: How about `equiv-Π-family-of-equivalences` and `iff-Π-family-of-logical-equivalences`? We also spell out "family of elements" elsewhere in the library, and I think it reads better than `equiv-fam`. The old name `equiv-map-Π` also doesn't seem quite right to me.

fredrik-bakke: I think I prefer `equiv-Π-equiv-family`. I don't like how your suggestions seem to use two different names for the same notions. E.g. `equiv` and `equivalence`, and `iff` and `logical-equivalence`. Moreover, the `-of-` binding in definition names doesn't seem to fit too well with our general naming scheme.

src/orthogonal-factorization-systems/local-maps.lagda.md
fredrik-bakke: If it is okay with you, I will create issues for empty sections like this in a separate PR.

EgbertRijke: Sure!

src/foundation/logical-equivalences.lagda.md
EgbertRijke: We should probably define `equiv-family` somewhere, so that `equiv-Π-equiv-family` reads as the function that takes an element of type `equiv-family` to an equivalence of dependent function types.

EgbertRijke: Also, I think we already have `is-family-of-equivalences`

VojtechStep: > We should probably define equiv-family somewhere

If by that you mean give a name to `(x :  X) -> A x ≃ B x` the you're in luck, because it's not only already defined, it's even defined twice 😃. Once in univalence (`equiv-fam`) and once in equivalences (`fam-equiv`).

EgbertRijke: Lol! Thanks for spotting that. It should probably go in a separate file called `families of equivalences`, and then we can have the definition of `is-family-of-equivalences` there too.

fredrik-bakke: I hope you are not expecting me to do that as part of this PR

EgbertRijke: Sure, let me create an issue then.

src/foundation/slice.lagda.md
```diff
@@ -67,7 +68,7 @@ module _
 
   triangle-hom-slice :
     (f : A → X) (g : B → X) (h : hom-slice f g) →
-    f ~ (g ∘ (map-hom-slice f g h))
+    f ~ (g ∘ map-hom-slice f g h)
```
EgbertRijke, line 71: And that last set of parentheses isn't needed anymore either:) Do we want to omit them in general, or do you prefer these homotopies to be stated with parentheses?

fredrik-bakke: I think in cases like this we can omit the parentheses

src/foundation-core/functoriality-function-types.lagda.md
fredrik-bakke: Would it make sense to move these corollaries and the main theorem to its own file? I'm struggling to find anywhere natural to place them that doesn't introduce cyclic dependencies. If so, what should the file be called?

fredrik-bakke: Placing it in subuniverses didn't really make sense to me, unless every corollary is placed elsewhere.

src/foundation/functoriality-function-types.lagda.md
fredrik-bakke: I'm struggling to find a place to put this that doesn't introduce cyclic dependencies, and this is taking a lot of my time. Is it okay if we leave it here for now?

fredrik-bakke: Maybe a passable argument for keeping it here for the moment is that truncated maps is a generalization of equivalences, and this theorem for equivalences is something that belongs naturally in this file? 🙃

src/foundation-core/functoriality-function-types.lagda.md
EgbertRijke: I guess one can argue that even though these corollaries mention the universe of propositions, the universe of sets, and the universe of truncated types, they are still about the functoriality of Pi-types. Perhaps it is ok to have them here.

EgbertRijke: The other option would be to have separate files for `precomposition-dependent-functions` and `postcomposition-dependent-functions`, but this would require some thought before we start doing that. For example, is it really sensible?

In `functoriality-dependent-function-types` we currently deal with precomposition, postcomposition, and two-sided composition. If we would make a split, would we also want to split the `functoriality-function-types` files? Can we make this split on conceptual grounds?

fredrik-bakke: I would hope that wouldn't be necessary, but let me think about it a little. Can we perhaps merge this PR and rather make further changes to this when the need arises again?

fredrik-bakke: Or a better solution is found

EgbertRijke: Let's at least agree that it is not necessary for this PR. But every time I've factored out an operation from a larger file into its own I've actually been happy with the result, and I could see that happening here too. So I think it's worth some consideration whether we want files for post and precomposition.

src/foundation/functoriality-function-types.lagda.md
EgbertRijke: Yeah, let's keep it here.

src/foundation-core/coherently-invertible-maps.lagda.md
```diff
@@ -48,30 +49,46 @@ module _
...
-  inv-is-coherently-invertible : is-coherently-invertible f → B → A
-  inv-is-coherently-invertible = pr1
+  map-inv-is-coherently-invertible : B → A
+  map-inv-is-coherently-invertible = pr1 H
 
-  is-section-inv-is-coherently-invertible :
-    (H : is-coherently-invertible f) → (f ∘ inv-is-coherently-invertible H) ~ id
-  is-section-inv-is-coherently-invertible H = pr1 (pr2 H)
+  is-retraction-is-coherently-invertible :
+    (f ∘ map-inv-is-coherently-invertible) ~ id
```
EgbertRijke, line 63:
```suggestion
    f ∘ map-inv-is-coherently-invertible ~ id
```

src/foundation-core/coherently-invertible-maps.lagda.md
```diff
@@ -48,30 +49,46 @@ module _
...
+  is-retraction-is-coherently-invertible = pr1 (pr2 H)
 
-  is-retraction-inv-is-coherently-invertible :
-    (H : is-coherently-invertible f) → (inv-is-coherently-invertible H ∘ f) ~ id
-  is-retraction-inv-is-coherently-invertible H = pr1 (pr2 (pr2 H))
+  is-section-is-coherently-invertible :
+    (map-inv-is-coherently-invertible ∘ f) ~ id
```
EgbertRijke, line 67:
```suggestion
    map-inv-is-coherently-invertible ∘ f ~ id
```

src/foundation-core/coherently-invertible-maps.lagda.md
```diff
@@ -97,51 +114,60 @@ module _
...
-      ( H : has-inverse f) →
-      ( is-section-inv-has-inverse H ·r f) ~
-      ( f ·l is-retraction-inv-has-inverse H)
-    coherence-inv-has-inverse H x =
+    is-section-map-inv-is-invertible :
+      (H : is-invertible f) → (f ∘ map-inv-is-invertible H) ~ id
+    is-section-map-inv-is-invertible H y =
+      ( inv (is-retraction-is-invertible H (f (map-inv-is-invertible H y)))) ∙
+      ( ( ap f (is-section-is-invertible H (map-inv-is-invertible H y))) ∙
+        ( is-retraction-is-invertible H y))
```
EgbertRijke, line 123:
```suggestion
      ( inv (is-retraction-is-invertible H (f (map-inv-is-invertible H y)))) ∙
      ( ap f (is-section-is-invertible H (map-inv-is-invertible H y))) ∙
      ( is-retraction-is-invertible H y)
```

src/foundation-core/coherently-invertible-maps.lagda.md
fredrik-bakke: Already reaping benefits from #712 😊

src/foundation-core/coherently-invertible-maps.lagda.md
fredrik-bakke: turns out it didn't type check with this change

EgbertRijke (approved):

EgbertRijke:
Looks good! I went over it one more time, and I'm happy to hit that green button again:)

src/foundation-core/coherently-invertible-maps.lagda.md
EgbertRijke: Ah, sorry! My bad

src/foundation-core/coherently-invertible-maps.lagda.md
EgbertRijke: Oh yes!

fredrik-bakke:
Thanks for all the discussions! I'm sorry not everything could be resolved. Still, I think this PR introduces some good improvements overall

EgbertRijke:
They're definitely improvements! Maybe things are not yet ready to be printed on t-shirts, but hey, you made it better than it was before!

EgbertRijke:
Brr some error again

fredrik-bakke:
working on it

fredrik-bakke:
Some definitions had randomly disappeared in the merging process. Git isn't always perfect when it comes to this.

EgbertRijke:
That's the first time I hear of such a bug. Quite a bad one!

fredrik-bakke:
I guess it is also possible that I wasn't paying enough attention at some point 😅

src/foundation/homotopies.lagda.md
```diff
@@ -222,6 +222,24 @@ module _
   equiv-right-transpose-htpy-concat : ((H ∙h K) ~ L) ≃ (H ~ (L ∙h (inv-htpy K)))
   pr1 equiv-right-transpose-htpy-concat = right-transpose-htpy-concat H K L
   pr2 equiv-right-transpose-htpy-concat = is-equiv-right-transpose-htpy-concat
+
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2} {f g h : (x : A) → B x}
+  (H : f ~ h) (K : f ~ g) (L : g ~ h)
+  where
+
+  equiv-left-transpose-htpy-concat' : (H ~ (K ∙h L)) ≃ (((inv-htpy K) ∙h H) ~ L)
```
EgbertRijke, line 231: Doesn't this line have way too many parentheses?

EgbertRijke: I see 3 pairs that can be removed:)

src/foundation/homotopies.lagda.md
```diff
@@ -222,6 +222,24 @@ module _
...
+  equiv-left-transpose-htpy-concat' =
+    ( equiv-inv-htpy L ((inv-htpy K) ∙h H)) ∘e
+    ( equiv-left-transpose-htpy-concat K L H) ∘e
+    ( equiv-inv-htpy H (K ∙h L))
+
+  equiv-right-transpose-htpy-concat' :
+    (H ~ (K ∙h L)) ≃ ((H ∙h (inv-htpy L)) ~ K)
```
EgbertRijke, line 238: Same here

fredrik-bakke:
Sorry, I had only restored what was there before.

EgbertRijke:
Great! Let's see if the checks will be done before I find something else that can be changed XD

fredrik-bakke:
I'm glad you're eager about the new operator system, but let's save it for newer PRs :)

fredrik-bakke:
😂 @EgbertRijke

EgbertRijke:
What's up?:)

PR #712, opened 2023-08-27
Define precedence levels and associativities for all binary operators

### LINKED ISSUE #479: Add precedence and fixity to (standard) binary operators
I figured we could compile a list of operators and propose fixities and constraints on what should precede what if we indeed want to define this in the library.

Some of my initial suggestions are as follows.

#### Fixity
- Function composition should associate to the right.
`f ∘ g ∘ h` means `f ∘ (g ∘ h)`.
- ~Identification concatenation should associate to the left as concatenation is done in diagrammatic order.
`p ∙ q ∙ r` means `(p ∙ q) ∙ r`.~
- ~Homotopy concatenation should associate to the left as concatenation is done in diagrammatic order.
`p ∙h q ∙h r` means `(p ∙h q) ∙h r`.~
- Homotopy left whiskering should associate to the right.
`f ·l g ·l H` only makes sense as `f ·l (g ·l H)`.
Although in this instance it would also make sense to just write `(f ∘ g) ·l H`.
- Homotopy right whiskering should associate to the left.
`H ·r g ·r f` only makes sense as `(H ·r g) ·r f`.
- Products and coproducts should associate to the right.
`A × B × C` means `A × (B × C)`, and `A + B + C` means `A + (B + C)`.
- Identity and homotopy type formation should not associate to either side.
`x ＝ y ＝ z` is ambiguous, `f ~ g ~ h` is always ill-typed.

#### Precedence
- Function composition, homotopy concatenation, homotopy left whiskering, and homotopy right whiskering should have precedence over identity and homotopy type formation.
E.g. `f ∘ g ~ h` means `(f ∘ g) ~ h`.
- Identification concatenation should have precedence over identity type formation.
`p ∙ q ＝ r` means `(p ∙ q) ＝ r`.
- Identification concatenation, homotopy concatenation, homotopy left whiskering, and homotopy right whiskering should have the same precedence.
- Function composition should have the same precedence level as homotopy concatenation, and homotopy right whiskering.
e.g. `H ∙h K ∘ f` and `H ·r g ∘ f` are ambiguous (note that this is not an issue with left whiskering).
- Homotopy type formation could have precedence over identity type formation,
as `f ~ g ＝ X` only ever makes sense as `(f ~ g) ＝ X` since functions cannot be homotopic to types.
- Products should have precedence over, or the same precedence, as coproducts.
e.g. `A × B + C` means `(A × B) + C` or is ambiguous.
- The operator `_∘_` may have precedence over `_,_`, as the alternative is always ill-typed.
e.g. `f , g ∘ h` means `f , (g ∘ h)`.

fredrik-bakke:
Resolves #479. See the document `MIXFIX-OPERATORS.md` for details.


fredrik-bakke:
I think we could define the path concatenation operations to associate to a particular side as well, but only if we define strict guidelines for when we allow parentheses to be omitted. For instance, when we are forming an equation involving paths, it is essential to have the expressions bracketed for readability, e.g.
```agda
  p ∙ (q ∙ r) =  (p ∙ q) ∙ r 
```
as opposed to
```agda
  p ∙ q ∙ r =  (p ∙ q) ∙ r 
```

However, when we are simply concatenating many paths to prove that two other things are equal, I think the parenthization is detrimental to the readability. Take for instance this random argument from `commuting-squares-of-identification` and compare:
```agda
  coherence-square-identifications-comp-horizontal p q =
    ( ( ( inv (assoc p-left p-bottom q-bottom) ∙
          ap-binary (_∙_) p (refl {x = q-bottom})) ∙
        assoc p-top middle q-bottom) ∙
      ap-binary (_∙_) (refl {x = p-top}) q) ∙
    inv (assoc p-top q-top q-right)
```
to
```agda
  coherence-square-identifications-comp-horizontal p q =
    inv (assoc p-left p-bottom q-bottom) ∙
    ap-binary (_∙_) p (refl {x = q-bottom}) ∙
    assoc p-top middle q-bottom ∙
    ap-binary (_∙_) (refl {x = p-top}) q ∙
    inv (assoc p-top q-top q-right)
```
Here it would be beneficial to have path concatenation have a defined association.

EgbertRijke:
I started to work with private infix definitions where I was also giving precedence levels to the privately defined operations. I would like to facilitate this, but then we should leave more space between "standard" precedence levels. I suggest that we go up in increments of 5, instead of increments of 1.

fredrik-bakke:
Precedence levels can take any signed fractional value, although I agree that integral values look better.

fredrik-bakke:
> In your example you chose an ill-formatted construction of an identification, to make the contrast bigger ;)

Please don't misinterpret me. I am not trying to convince you in any particular direction, I'm only exploring the different options and trying to lay out demonstrative arguments.

EgbertRijke:
I forgot to say that you took a great initiative here, to determine good conventions for precedence levels of infix notation. I definitely agree that this is important.

EgbertRijke:
I've been thinking about precedence levels in the past week, but my ideas are somewhat differently than the ones proposed here. I think we should clearly distinguish between precedence levels on (structured) types and (structured) elements, and give all the operations on structured types lower precedence than operations on structured elements.

For each of these domains we have the following classes: unary operations > multiplicative operations > additive operations > exponentiative operations > relations. Note that I'm not quite sure what the best practice is for precedence levels of exponentiative operations. In Agda, the arrow has lower precedence than nearly everything. So if we want to be consistent with precedence levels for exponentiative operations, we should give them low precedence. On the other hand, the PEMDAS gives exponentiation high precedence (between unary and multiplicative).

The reasoning syntax can still have the very lowest precedence level.

I've been using private notation in some files recently, which I started to like a lot because it doesn't take up names but it does make things very readable. So I propose to leave gaps of size 10 for clarity. I'm thinking that perhaps we should have a convention for global precedence levels to be 0-4 mod 10 and private/local precedence levels to be 5-9 mod 10.

|              | Relations | Exponentiative | Additive | Multiplicative | Unary |
|-----------------------------|-----------|----------------|----------|----------------|-------|
| Operations on Types         | 10-19     | 20-29          | 30-39    | 40-49          | 50-59 |
| Operations on Elements      | 60-69     | 70-79          | 80-89    | 90-99          | 100-109 |

Note that I would categorize equality as an operation on elements, and therefore give it precedence level 60. Not a low precedence level, as was suggested above. Would that be unreasonable?

EgbertRijke:
Another way of using the gaps would instead of distinguishing between global/private precedence levels, to distinguish within each range 0-9 mod 10 between nondomain specific operations and domain specific operations. For example, we could say that binary operations on domain specific structured types (eg groups, pointed types) must be in the range 5-9 mod 10.

EgbertRijke:
One thing I didn't think about is "default" precedence level. What is the purpose of the default precedence level precisely?

fredrik-bakke:
> One thing I didn't think about is "default" precedence level. What is the purpose of the default precedence level precisely?

The default precedence level is what Agda assigns to an operator when no precedence level is given.

EgbertRijke:
Ah, thanks! Is that level adjustable per library?

fredrik-bakke:
It doesn't seem like it. At the very least, [the documentation](https://agda.readthedocs.io/en/v2.6.3.20230805/language/mixfix-operators.html) does not mention any way to modify it.

fredrik-bakke:
Perhaps it is best if we interpret function types as relational operations rather than exponentiative?

EgbertRijke:
That's perhaps a good idea, because that actually resolves an inconsistency in my proposal: that agdas function types have even lower precedence than relations. And then we can be consistent with PEMDAS

fredrik-bakke:
So, let me try to write up a more concrete mapping. I'm struggling a bit to see the differences between your solution and mine.

EgbertRijke:
I didn't see a clear distinction between type operations and element operations in your proposal, but I think having that distinction would be clarifying. Also I would leave much more space between the "standard" precedence levels, so that we can also have conventions for private operations or for domain-specific operations between them.

fredrik-bakke:
So, one problem that came up when I first tried to define precedences is where to place `_,_`. It currently has a very low precedence and is used a lot together with both types and terms. I assume you would consider it an operation on structured elements and thus would propose a higher precedence level for it. That would force us to parenthesize many expressions differently than we do currently. For `_,_`, I think it makes sense to prioritize it below almost all other operations.

fredrik-bakke:
I do appreciate your more principled approach, and having larger gaps does sound better aesthetically speaking.

EgbertRijke:
That's a very good point!

EgbertRijke:
So the question is perhaps what an operation on (structured) types is, for the purpose of this discussion about precedence levels. Perhaps it is an operation parametrized in such a way that it _could_ take as one of its arguments something which we would consider a structured type. Then the identity type and the pairing operation would again fall under the umbrella of type operations. Is this too evil?:)

EgbertRijke:
Ah, is the point perhaps that we should distinguish between parametric and nonparametric operations? The identity type and the pairing operation are parametric, but for example multiplication on the natural numbers is not.

fredrik-bakke:
Another thing, I chose to give many different constructions the same precedence level to force contributors to parenthesize their expressions. Even though I'm introducing many precedence levels with this PR, I still think that much of the time, expressions should still be parenthesized, even if it is not strictly needed for the code to type check. By having things have the same precedence level, we are effectively forcing people to be more explicit. This will also lead to less confusion when reading code, or writing arithmetic expressions. Hence, much of the work that went into defining my scheme was figuring out when omitting parentheses should not cause confusion. This is also why I think we should have a strict convention for when code must still be parenthesized.

fredrik-bakke:
> Ah, is the point perhaps that we should distinguish between parametric and nonparametric operations? The identity type and the pairing operation are parametric, but for example multiplication on the natural numbers is not.

That is an interesting distinction, I think maybe you're right

EgbertRijke:
> Another thing, I chose to give many different constructions the same precedence level to force contributors to parenthesize their expressions. Even though I'm introducing many precedence levels with this PR, I still think that much of the time, expressions should still be parenthesized, even if it is not strictly needed for the code to type check. By having things have the same precedence level, we are effectively forcing people to be more explicit. This will also lead to less confusion when reading code, or writing arithmetic expressions. Hence, much of the work that went into defining my scheme was figuring out when omitting parentheses should not cause confusion. This is also why I think we should have a strict convention for when code must still be parenthesized.

This should certainly be mentioned in the part of the guide where we explain our choices for precedence levels.

fredrik-bakke:
Alright, so then the main differences between your scheme and mine seem to be the following:
For your scheme
- Most operations end up having higher precedence than the default one. Perhaps we should shift your levels a bit so that the default level ends up between parametric and non-parametric operations? Note that precedence levels can also be negative.
- Function composition has the same precedence as taking products of types, and, for instance, lower precedence than forming inequality on the natural numbers.
- There are more precedence levels for different operations on types.
   So, for instance, the expression `A ↔ B + C ↪ D + E × F` is not ill-formed.

EgbertRijke:
Shouldn't logical equivalence and embedding (and other kinds of arrows) all have the same precedence level (that fall under the precedence level of parametric relations)?

Instead of going for negative precedence levels, we could have a standard increment of 5 instead of 10? Then the parametric exponentiative operations (which perhaps we won't use very much) have the default precedence level, and unary parametric operations will bind stronger.

|                             | Relations | Additive | Multiplicative | Exponentiative | Unary |
|-----------------------------|-----------|----------|----------------|----------------|-------|
| Parametric Operations       | 5-9       | 10-14    | 15-19          | 20-24          | 25-29 |
| Nonparametric Operations    | 30-34     | 35-39    | 40-44          | 45-49          | 50-54 |

fredrik-bakke:
Ah, right, that's a thing. I wanted `→∗` to have lower precedence than `≃∗`, and `~∗`, so that something like
`A →∗ B ≃∗ (f ~∗ g)` would parse similarly to `A → B ≃ (f ~ g)`. This is why I gave the directed arrows a lower precedence than the bi-/undirected ones.

EgbertRijke:
Perhaps we could assign the range `0-4` to arrow-like parametric operations, and bump the reasoning ones down to `-5 - -1`

EgbertRijke:
Or perhaps it is fine to use level `5` for arrow-like parametric operations and levels `6-9` for other parametric relations. We are giving ourselves the space to do this, after all.

fredrik-bakke:
True, that does sound better

fredrik-bakke:
how about assigning ranges 3 - 7 for relations, 8 - 12 for parametric additive operations and so on, so that we have some wiggle room both up and down in each category?

EgbertRijke:
> how about assigning ranges 3 - 7 for relations, 8 - 12 for parametric additive operations and so on, so that we have some wiggle room both up and down in each category?

Don't you think that looks confusing?

fredrik-bakke:
Not if we phrase it as "5 ± 2" and so on

EgbertRijke:
I'm not convinced:)

fredrik-bakke:
So if you are defining an additive operator and want it to have a lower precedence than the other ones, do you propose to put it in the category of relational operators, since there is no level available in the category of additive operators that is lower than the default one, or do you propose setting the default value to 12 instead of 10?

fredrik-bakke:
Or what is your perspective?

EgbertRijke:
I'm inclined to being fine with not allowing it, but there's a chance that your situation will actually come up:)

EgbertRijke:
If this situation comes up, it is most likely that we have a compelling reason that our new operation should bind lower than one or a few other operations. This means that there is a compelling reason that those few other operations shouldn't have the lowest precedence level in their range, and we can adjust them.

fredrik-bakke:
Hmm. So how about instead of assigning ranges, we just assign default values, and don't define the rest of the spectrum. I.e.

|                             | Relations | Additive | Multiplicative | Default binary | Unary |
|-----------------------------|-----------|----------|----------------|----------------|-------|
| Parametric Operations       | 5      | 10    | 15          | 20          | 25 |
| Nonparametric Operations    | 30     | 35    | 40          | 45         | 50 |

fredrik-bakke:
I'm not entirely convinced by your argument. If I read it correctly, it seems to suggest that the default value shouldn't necessarily be the default value.

EgbertRijke:
Doesn't this raise questions about what our conventions are when someone sees a precedence level of 13? I think this makes our conventions less clear.

fredrik-bakke:
I agree. My primary issue is that I don't see why the conventions should give flexibility upwards, and not downwards.

fredrik-bakke:
I don't want to have the issue that precedence levels need to be juggled around in the future because of this.

EgbertRijke:
> I'm not entirely convinced by your argument. If I read it correctly, it seems to suggest that the default value shouldn't necessarily be the default value.

I think my point is that if we assign ranges of precedence levels for classes of operations, then we already build in some flexibility of how we assign precedence levels within each range. For example, with parametric relations we saw that we can give the arrow-like operations a precedence level of `5`, and other parametric relations a precedence level of higher than `5`. I can imagine something similar happens in other ranges, especially when clear reasons show up on how to organize different precedence levels within each range. For example, something might show up around the `_,_` and equality that makes us want to assign it a "special" precedence level within its range.

I also think that it will still be good that besides the table that gives the overview of the precedence level ranges, that it will still be good to give a more detailed and complete table like the one you gave in the PR description, that lists all the operations in the library that are given an explicit precedence level. This might help navigate precedence levels.

EgbertRijke:
> I don't want to have the issue that precedence levels need to be juggled around in the future because of this.

But I don't want the issue that precedence levels are "set in stone" once they are defined. We should be able to adjust precedence levels if we have good reason to do so.

fredrik-bakke:
Alright, let's just go with your approach. Sorry for dragging this out.

EgbertRijke:
No need to be sorry, these are good things to discuss! I actually started to be a bit more sympathetic to your idea, because your argument that we should be able to give ourselves room both upwards and downwards definitely has merit. I just also think that (and I could be wrong) this might occur less often and we would sacrifice some clarity in our conventions for it.

EgbertRijke:
Perhaps we could have a convention that in exceptional circumstances we can use precedence levels `-1 mod 5` to give some operations lower precedence than all other (nonexceptional) operations in its class

fredrik-bakke:
Hmm. Don't you think it's more confusing that when defining relational operators, the default precedence level will be 6 instead of 5?

EgbertRijke:
Won't there be about as many arrow-like relations ad nonarrow-like relations, so that we can hardly speak of a "default"?

src/foundation-core/equivalences.lagda.md
```diff
@@ -45,6 +45,7 @@ module _
   is-equiv : (A → B) → UU (l1 ⊔ l2)
   is-equiv f = section f × retraction f
 
+infix 10 _≃_
```
EgbertRijke, line 48: Isn't this one at level 6?

fredrik-bakke: Thanks for spotting that! It will have to be level 7, so we can insert `_,_` at level 6.

EgbertRijke: Lol, that's one odd parametric relation:)

EgbertRijke: Would it be unreasonable to stick `_,_` at level `3` or something? What is your reasoning for sticking it at level `6`?

fredrik-bakke: It's not unreasonable. My reason for putting it at `6` was so the expression `A → B , q` would parse the same way as `A →∗ B , q`, but now that I think of it, the first expression is always ill-formed.

EgbertRijke: If only we could make a precedence poset

fredrik-bakke: I know, right!

fredrik-bakke: Or at least be able to redefine the precedence level of `→`

fredrik-bakke: So should I change it to 3 or keep it at 6 do you think?

EgbertRijke: I think 3 is better, because then it is not confusing that it is in the precedence range of parametric relations

EgbertRijke: I tried hard to think of ways that `_,_` is a parametric relation btw. Perhaps it is a parametric relation in the sense that it is a binary operation into a type that _could_ be some type of structured types. It's a bit of a stretch:)

fredrik-bakke: I appreciate the effort!

EgbertRijke:
Could you also explain your choice of when something is associating to the left/to the right/unspecified?

fredrik-bakke:
I'm working on it. Writing `PRECEDENCE-LEVELS.md` at the moment.

EgbertRijke:
Awesome! I was going to ask you if you wanted to do that or if you wanted me to have a go at it.

EgbertRijke:
Could a better name for that file be `INFIX-OPERATIONS.md` or `MIXFIX-OPERATIONS`?

Since it is not only about precedence levels, but also about how to introduce them and how they associate?

fredrik-bakke:
I was thinking the same thing. 😄
Probably MIXFIX-OPERATIONS is best, given that it includes reasoning syntax and unary operators.

EgbertRijke:
Not to be annoying... (uh oh)

But I would associate + and - and * and / all to the left. My reasoning is as follows: If you see an expression like 6 + 17 + 9 +2 then you add them from left to right. At least I think most people do. This is also consistent with 6 + 17 - 9 - 2. It would be more unintuitive if it were assocated to the right.

Exponentiation on the other hand should be associated to the right (contrary to agda's arrow notation), because we often write 3^4^7 without parentheses to mean 3^(4^7) and we would always add parentheses if we meant (3^4)^7.

EgbertRijke:
So I would associate all additive and multiplicative operations to the left, exponentiative operations to the right in general. Of course, there could be exceptions when there is a compelling reason to do otherwise.

fredrik-bakke:
Ah, thanks, that was a mistake on my part in regards to the arithmetic operations. However, I disagree with your choice of associativity for multiplicative operations in general. I would like the term `a , b , c` to live in the type `A × B × C`, hence I think it should be right-associative.

EgbertRijke:
isn't `_,_` special down there at level 3?

fredrik-bakke:
Yes, I am not done with the file

EgbertRijke:
Also, if both pairing and cartesian products both associate to the left then your concern resolved isn't it?

fredrik-bakke:
Oh god no

EgbertRijke:
Wait a sec... maybe I agree with you. With dependent pair types it would be weird if (a,b,c) isn't interpreted as an element of Sigma (a:A), Sigma (b:B(a)), C(a,b).

Ok, I'm good with associating multiplicative operations to the right. Thanks for correcting me on this!

EgbertRijke:
I'll stop annoying you =)

fredrik-bakke:
When it comes to type formers, I think it is more common to think of it in terms of introduction rules from left to right, I.e. you want a single `inl` to map into `A + B + C` and so on. Therefore, I'm inclined to make them right associative

fredrik-bakke:
No no, I appreciate it! Sorry for my less verbose responses. I'm hard at work writing atm 😅

EgbertRijke:
> When it comes to type formers, I think it is more common to think of it in terms of introduction rules from left to right, I.e. you want a single `inl` to map into `A + B + C` and so on. Therefore, I'm inclined to make them right associative

This also agrees with the generalization to infinite sums.

(I'll really stop annoying you now)

EgbertRijke:
I'm also working hard to finish that cyclic PR

fredrik-bakke:
Alright, enjoy yourself!

MIXFIX-OPERATORS.md
```diff
@@ -0,0 +1,107 @@
+# Mixfix operators in agda-unimath
+
+In this document, we outline our conventions for assigning precedence levels and
+associativity to different
+[mixfix operators in Agda](https://agda.readthedocs.io/en/v2.6.3.20230805/language/mixfix-operators.html).
+
+## Mixfix operators in Agda
+
+Mixfix operators in Agda can each be assigned a
```
EgbertRijke, line 9: Before jumping into precedence levels I think the first sentence should be explaining what mixfix operations are. For instance:

"Infix operations in Agda are binary operations that take arguments on either side. For example, addition is commonly written in infix notation as `1 + 2`. Mixfix operations are operations with longer names potentially containing multiple components, where the arguments appear between the components. For example, the commutator `[a,b]` is a common mixfix operation. The purpose of introducing infix and mixfix operations in Agda is to make the code more readable by using commonly accepted notation for widely used operations.

fredrik-bakke:
I will have to finish this PR tomorrow

EgbertRijke:
No problem!

fredrik-bakke:
Alright, I think this PR is ready for review again now.

CODINGSTYLE.md
```diff
@@ -387,6 +387,12 @@ the story. Here's how we handle indentation and line breaks in the
   Sometimes, however, `equation-n` is a short proof term that fits on the same
   line as `by` within the 80 character limit. In that case it is ok to do so.
 
+- Expressions involving mixfix operators are appropriately parenthesized when
+  the particular association bears relevance, or if there is any chance of
+  confusion from omitting the parentheses. A reader of the code should not be
+  expected to know the precedence levels or associativity of particular
+  operators.
```
EgbertRijke, line 394: This is a very good principle!

MIXFIX-OPERATORS.md
```diff
@@ -0,0 +1,179 @@
...
+_non-parametric_ operators. The general rule is that non-parametric operator has
+higher precedence than parametric operators. Parametric operators are operators
+like the [cartesian product `_×_`](foundation-core.cartesian-product-types.md) ,
+the [identity type former `_＝_`](foundation-core.identity-types.md) and the
+[pairing operator `_,_`](foundation.dependent-pair-types.md). Examples of
+non-parametric operators are
+[difference of integers `_-ℤ_`](elementary-number-theory.difference-integers.md)
+[strict inequality on natural numbers `_<-ℕ_`](elementary-number-theory.strict-inequality-natural-numbers.md)
+and
+[multiplication of Eisenstein integers `_*ℤ[ω]_`](commutative-algebra.eisenstein-integers.md).
```
EgbertRijke, line 58: Perhaps we can leave one more hint about what parametric operations are, instead of only explaining it by example. We could say that a parametric operation is an operation that takes a universe level as one of its arguments. We consider an operation to be parametric even it its universe level is an implicit argument.

src/reflection/type-checking-monad.lagda.md
```diff
@@ -177,11 +177,11 @@ postulate
 ## Monad syntax
 
 ```agda
-infixl 3 _<|>_
+infixl 15 _<|>_
 _<|>_ : {l : Level} {A : UU l} → TC A → TC A → TC A
 _<|>_ = catchTC
 
-infixl 1 _>>=_ _>>_ _<&>_
+infixl 10 _>>=_ _>>_ _<&>_
```
EgbertRijke, line 184: Lol, and we said that the purpose of infix operations was to implement commonly used, widely accepted notation:)

EgbertRijke (approved):
Awesome! I had only one meaningful remark about the need for explaining what we mean by parametric operations. After that one is addressed I will merge this.

Thanks for this huge improvement to the library!

EgbertRijke (approved):
Awesome! I had only one meaningful remark about the need for explaining what we mean by parametric operations. After that one is addressed I will merge this.

Thanks for this huge improvement to the library!

MIXFIX-OPERATORS.md
fredrik-bakke: Thanks, I didn't quite know how to define it 😅

EgbertRijke: It's not a formal definition, but we don't need a formal definition for the purpose of our guide. We only need to say: if it has a universe level as an argument then we consider it to be in the parametric class.

EgbertRijke:
I'm excited to hit that green button:)

fredrik-bakke:
Thanks for all the assistance in developing this PR! I'm quite relieved it is merged now 😄

fredrik-bakke:
Huh, you didn't get any credit for this PR! :(

EgbertRijke:
No big deal. Only saying stuff doesn't count:)

PR #719, opened 2023-09-05
The statement of the flattening lemma

EgbertRijke:
This is a small pull request adding only the statement of the flattening lemma to the library.

### DIFF
--- a/src/foundation/functoriality-dependent-pair-types.lagda.md
+++ b/src/foundation/functoriality-dependent-pair-types.lagda.md
@@ -9,8 +9,11 @@ open import foundation-core.functoriality-dependent-pair-types public
 <details><summary>Imports</summary>
 
 ```agda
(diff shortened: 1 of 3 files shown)


src/foundation/functoriality-dependent-pair-types.lagda.md
```diff
@@ -192,6 +194,76 @@ module _
...
+
+#### Functoriality of `Σ` preserves commuting squares of maps
+
+```agda
+module _
+  {l1 l2 l3 l4 l5 l6 l7 l8 : Level}
+  {A : UU l1} {P : A → UU l2}
+  {B : UU l3} {Q : B → UU l4}
+  {C : UU l5} {R : C → UU l6}
+  {D : UU l7} (T : D → UU l8)
```
EgbertRijke, line 207: It looks like I forgot how the alphabet goes, lol

Issue #702, opened 2023-08-21
Formalize the Hopf Fibration


morphismz:
Construct the hopf fibration in such a way that the generator of $\mathbb{S}^3$ is sent to a 3-loop in $\mathbb{S}^2$ constructed using Eckmann-Hilton. 

A detailed plan for this may be too much to write out in a comment here. But I'll list a few components that will go into this.

Done or currently being worked on:

- [x] finish dependent universal property of suspensions
- [x] "Universal Property of Fibers of maps": build up infrastructure for working with the family of fibers of a map `f : A --> B` as the initial type family equipped with a section `(a : A) --> fib_f (f a)`. 
Open tasks
- [x] show EH is (more or less) the naturality condition of whiskering a 2-path on the left by a 1-path
- [x] "Eckmann-Hilton in the universe" idea: show that `tr` sends EH on identifications to EH on homotopies (almost done, missing a few coherences)

Open:

- [ ] descent for $\mathbb{S}^1$, $\mathbb{S}^2$, $\mathbb{S}^3$ 
-----  related to issue #1122 . This proof works best the the point and n-loop presentation of spheres, which have yet to be postulated.
- [ ] characterizing (higher) homotopies between (homotopies) functions with domain $\mathbb{S}^1$,
- [ ] characterizing homotopies between dependent functions functions with domain $\mathbb{S}^3$.

Quasi-references: 
- a talk presenting the current version of the proof https://morphismz.github.io/talks/2024-04-03-hottuf

PR #727, opened 2023-09-09
Refactor shift, pointed identity types are loop spaces

morphismz:
This pr does two related things:

(i) it shows that pointed identity types are equivalent to loop spaces. 

this will be useful in the Eckmann-Hilton and Hopf proof (issue #702), since we can construct an eckmann-hilton term of type $s \cdot s = s \cdot s$, where `s` is the generator of $\Omega^2 \mathbb{S}^2$`. Then, using this equivalence, we get a term of type $\Omega^3 \mathbb{S}^2$. Due to the way the equivalence is defined, we will get many nice functoriality properties. In particular, the equivalence preserves path concatination. This will be helpful in the proof. It will also later be helpful when applying sypllepsis.

(ii) the pr uses the above equivalenecs to replace `unshift` in the file `universal-property-suspensions-pointed-types`


src/synthetic-homotopy-theory/double-loop-spaces.lagda.md
```diff
@@ -118,3 +122,20 @@ interchange-concat-Ω² :
...
+
+
+```agda
+module _
+  {l : Level} {A : Pointed-Type l} {x : type-Pointed-Type A}
+  (p : point-Pointed-Type A ＝ x)
+  where
+
+  pointed-equiv-2-loop-pointed-identity :
+    Ω ((pair (point-Pointed-Type A ＝ x)  p)) ≃∗ Ω² A
```
EgbertRijke, line 137: I'm still puzzled why you like `(( .. ))` so much

morphismz: I'm so sorry about that, that looks like a typo that I didn't catch... Not sure why I keep making that typo :thinking:

VojtechStep: I think that's an interactive Agda thing, that if you have a hole in parantheses it (sometimes?) adds redundant ones when you give it a term (I'm not at my computer, but I suspect it can happen if you create a hole, load the file, add the parantheses, _don't_ load the file, and give a term).

EgbertRijke: Ah yes, the automatic agda-formatting when resolving a goal feels like some kind of trolling by our agda-overlords. Don't get me started!

morphismz: @VojtechStep ah, that would make sense. I"ll try that out

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -200,12 +182,14 @@ module _
...
-            Σ ( type-Pointed-Type X → (point-Pointed-Type Y) ＝ (pr1 t))
+            Σ ( type-Pointed-Type X →
+              ( point-Pointed-Type Y) ＝ (pr1 t))
               ( λ f → f (point-Pointed-Type X) ＝ (pr2 t))))) ∘e
       ( ( equiv-tot (λ y1 → equiv-left-swap-Σ)) ∘e
         ( ( associative-Σ
             ( type-Pointed-Type Y)
-            ( λ y1 → type-Pointed-Type X → (point-Pointed-Type Y) ＝ y1)
+            ( λ y1 → type-Pointed-Type X →
+              (point-Pointed-Type Y) ＝ y1)
```
fredrik-bakke, line 192: Always line break at the earliest possible branching point first. I.e. here
```suggestion
            ( λ y1 →
              type-Pointed-Type X → point-Pointed-Type Y ＝ y1)
```
although I don't see precisely why you want to break the line at all.

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -200,12 +182,14 @@ module _
           ( type-Pointed-Type Y)
           ( λ z → (point-Pointed-Type Y) ＝ z)
           ( λ t →
-            Σ ( type-Pointed-Type X → (point-Pointed-Type Y) ＝ (pr1 t))
+            Σ ( type-Pointed-Type X →
+              ( point-Pointed-Type Y) ＝ (pr1 t))
```
fredrik-bakke, line 186:
```suggestion
                ( point-Pointed-Type Y) ＝ (pr1 t))
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -215,7 +199,8 @@ module _
                             ( λ y1 →
                               type-Pointed-Type X →
                               point-Pointed-Type Y ＝ y1)) →
-                  is-contr-total-path ((pr2 z) (point-Pointed-Type X))))) ∘e
+                  is-contr-total-path
+                    ( (pr2 z) (point-Pointed-Type X))))) ∘e
```
fredrik-bakke, line 203: This part is ill-formatted starting on line 198. Does it not type check if you omit the type of `z`?

src/synthetic-homotopy-theory/functoriality-loop-spaces.lagda.md
```diff
@@ -95,3 +96,62 @@ module _
       ( pointed-map-faithful-pointed-map f)
       ( is-faithful-faithful-pointed-map f)
 ```
+
+### Pointed embeddings induce pointed equivalces on loop spaces
```
fredrik-bakke, line 100:
```suggestion
### Pointed embeddings induce pointed equivalences on loop spaces
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -67,56 +67,37 @@ module _
...
   where
 
-  shift :
-    (type-Ω (suspension-Pointed-Type X)) → (north-suspension ＝ south-suspension)
-  shift l = l ∙ (meridian-suspension (point-Pointed-Type X))
-
-  shift∗ :
-    Ω (suspension-Pointed-Type X) →∗
+  pointed-equiv-loop-pointed-identity-suspension :
     (pair
```
fredrik-bakke, line 71:
```suggestion
    ( pair
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -67,56 +67,37 @@ module _
...
-      (meridian-suspension (point-Pointed-Type X)))
-  pr1 shift∗ = shift
-  pr2 shift∗ = refl
-
-  unshift :
-    (north-suspension ＝ south-suspension) →
-    (type-Ω (suspension-Pointed-Type X))
-  unshift p = p ∙ inv (meridian-suspension (point-Pointed-Type X))
+      ( meridian-suspension (point-Pointed-Type X))) ≃∗
+    Ω (suspension-Pointed-Type X)
```
fredrik-bakke, line 74:
```suggestion
    ( Ω (suspension-Pointed-Type X))
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -67,56 +67,37 @@ module _
...
+    pointed-equiv-loop-pointed-identity
+      ( suspension-Pointed-Type X)
+      ( meridian-suspension (point-Pointed-Type X))
 
-  unshift∗ :
+  pointed-map-loop-pointed-identity-suspension :
     (pair
       ( north-suspension ＝ south-suspension)
       ( meridian-suspension (point-Pointed-Type X))) →∗
     Ω (suspension-Pointed-Type X)
```
fredrik-bakke, line 84:
```suggestion
    ( Ω (suspension-Pointed-Type X))
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -67,56 +67,37 @@ module _
...
+      ( meridian-suspension (point-Pointed-Type X))) ≃∗
+    Ω (suspension-Pointed-Type X)
+  pointed-equiv-loop-pointed-identity-suspension =
+    pointed-equiv-loop-pointed-identity
+      ( suspension-Pointed-Type X)
+      ( meridian-suspension (point-Pointed-Type X))
 
-  unshift∗ :
+  pointed-map-loop-pointed-identity-suspension :
     (pair
```
fredrik-bakke, line 81:
```suggestion
    ( pair
```

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
```diff
@@ -130,7 +111,8 @@ module _
         ( point-Pointed-Type X) ,
         ( id))
 
-  counit-susp-loop-adj∗ : ((suspension (type-Ω X)) , north-suspension) →∗ X
+  counit-susp-loop-adj∗ :
+    ( pair (suspension (type-Ω X)) north-suspension) →∗ X
   pr1 counit-susp-loop-adj∗ = counit-susp-loop-adj
   pr2 counit-susp-loop-adj∗ =
```
fredrik-bakke, line 117:
```suggestion
  pointed-map-counit-susp-loop-adj :
    ( pair (suspension (type-Ω X)) north-suspension) →∗ X
  pr1 pointed-map-counit-susp-loop-adj = counit-susp-loop-adj
  pr2 pointed-map-counit-susp-loop-adj =
```
I do kind of like the use of `∗`-notation here, although if we were to implement this, it must be consistent across the library. Also, in this case, we should distinguish between free and pointed loop spaces by writing `Ω∗`.

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
morphismz: I don't think it does. One thing is, this code will soon be replaced, since the goal of refactoring the suspension loop space adjunction is mainly to replace this equivalence with one whose underlying map is much nicer. Thus it may not make too much sense to spends lots of time trying to improve this particular chunk of code. Though I'll fix those formatting errors.

fredrik-bakke: Okay, that's reasonable. I won't object to keeping this for the moment

src/synthetic-homotopy-theory/universal-property-suspensions-of-pointed-types.lagda.md
morphismz: I agree, the `*`-notation doesn't fit it well, so I'll get rid of it. 

Is there a need to distinguish between free and pointed loop spaces? Usually `Ω` just refers to pointed loop spaces. I think free loops are given a name like `free-loop`. I'm fine with this convention, unless there is other need to change it.

PR #729, opened 2023-09-10
`mixfix operation` -> `mixfix operator`

fredrik-bakke:


EgbertRijke:
It seemed that you previously preferred the word "operation" and I also didn't have a problem with it. I'm not really sure what reason there could be that operator is a better word than operation. Could you explain your reasoning?

EgbertRijke:
Ah, I asked ChatGPT and it was able to explain me that "operator" stands for the entity that engages in "operations". So I agree, "operator" seems to be the right word in this context.

fredrik-bakke:
I didn't really make the choice of using "operation" consciously. Since we're talking about the operators specifically in this file, and not necessarily the operation that is performed, I find it a better choice :)

PR #730, opened 2023-09-10
Rename `inv-con` and `con-inv` to `left/right-transpose-eq-concat`

fredrik-bakke:
Similarly, `inv-con-htpy` and `con-inv-htpy` are renamed to `left/right-transpose-htpy-concat`.


EgbertRijke:
For groups, the tranposition operations are called `transpose-eq-mul-Group`. This is a fairly well-used naming scheme, so perhaps `transpose-eq-concat` would be a better name. At least we should be consistent with the naming schemes for transposition in groups as well.

PR #750, opened 2023-09-11
The categories of rings and commutative rings

EgbertRijke:
This small PR adds the categories of rings and commutative rings to the library.


fredrik-bakke (approved):
Excellent, looks great!

PR #752, opened 2023-09-11
`check-merge-conflict` after file generation

fredrik-bakke:


.pre-commit-config.yaml
```diff
@@ -12,8 +12,6 @@ repos:
       - id: check-yaml
       # - id: check-json # Doesn't accept json with comments
       # - id: requirements-txt-fixer # Can't handle comments
-      - id: check-case-conflict
-      - id: check-merge-conflict
```
EgbertRijke, line 16: Should it not check for merge conflicts before it does other things that can edit the file? The point of generating the indexing files before merge conflicts is simply that it could resolve some merge conflicts simply by regenerating the indexing files, but I feel like other edits should happen after it checked for merge conflicts. Or does the order not matter so much?

fredrik-bakke: It would be nice if it was possible define a dependency graph, but sadly pre-commit doesn't support this. To implement this, we would have to write a custom script

EgbertRijke: ChatGPT would have to do it:)

PR #753, opened 2023-09-11
adjusting loc

EgbertRijke:
With the approval of Vojtech, who has the approval of ChatGPT

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,10 +33,10 @@ METAFILES := \
 agdaFiles:
 	@rm -rf $@
 	@rm -rf ./src/everything.lagda.md
-	@git ls-files src | grep '\.lagda.md$' > $@
+	@git ls-files src | grep '\.lagda.md$$' > $@
 	@sort -o $@ $@
 	@wc -l $@
-	@echo "$(shell git ls-files src | grep '.lagda.md$' | xargs cat | wc -l) LOC"
+	@echo "$(shell (git ls-files src | grep '.lagda.md$$' | xargs cat) | wc -l) LOC"
 
 .PHONY: ./src/everything.lagda.md
 src/everything.lagda.md: agdaFiles
--- a/Makefile
+++ b/Makefile
@@ -33,10 +33,10 @@ METAFILES := \
 agdaFiles:
 	@rm -rf $@
 	@rm -rf ./src/everything.lagda.md
-	@find src -name temp -prune -o -type f \( -name "*.agda" -o -name "*.lagda" -o -name "*.lagda.md" \) -print > $@
+	@git ls-files src | grep '\.lagda.md$' > $@
 	@sort -o $@ $@
 	@wc -l $@
-	@echo "$(shell (find src -name '*.lagda.md' -print0 | xargs -0 cat ) | wc -l) LOC"
+	@echo "$(shell git ls-files src | grep '.lagda.md$' | xargs cat | wc -l) LOC"
 
 .PHONY: ./src/everything.lagda.md
 src/everything.lagda.md: agdaFiles


EgbertRijke:
But not with the approval of me. It has a bug

VojtechStep:
Just to protect my honor: I wasn't the one asking ChatGPT for approval 😶

fredrik-bakke:
After this PR, I occasionally get errors due to something that perhaps is cached files that don't type check. E.g. when I run `make check`:

```text
Typechecking library
cat: src/category-theory/isomorphisms-categories.lagda.md: No such file or directory
cat: src/category-theory/isomorphisms-large-precategories.lagda.md: No such file or directory
cat: src/category-theory/isomorphisms-precategories.lagda.md: No such file or directory
    1413 agdaFiles
  252824 LOC
time agda -v1 src/everything.lagda.md
Checking everything (/Users/frebak/Repositories/agda-unimath/src/everything.lagda.md).
/Users/frebak/Repositories/agda-unimath/src/everything.lagda.md:25,1-52
Failed to find source of module
category-theory.isomorphisms-categories in any of the following
locations:
  /Users/frebak/Repositories/agda-unimath/src/category-theory/isomorphisms-categories.agda
  /Users/frebak/Repositories/agda-unimath/src/category-theory/isomorphisms-categories.lagda
  /opt/homebrew/Cellar/agda/2.6.3/libexec/ghc-9.4.4/Agd-2.6.3-d4728884/share/lib/prim/category-theory/isomorphisms-categories.agda
  /opt/homebrew/Cellar/agda/2.6.3/libexec/ghc-9.4.4/Agd-2.6.3-d4728884/share/lib/prim/category-theory/isomorphisms-categories.lagda
when scope checking the declaration
  open import category-theory.isomorphisms-categories
        1.48 real         1.37 user         0.10 sys
make[1]: *** [check] Error 42
make: *** [pre-commit] Error 2
```

After I've renamed said files

VojtechStep:
Shoot, `git ls-files` probably lists deleted files in the output, which then end up in the Everything file. That could be handled by having git include the change information marker and pipe it through `sed`, deleting lines starting with `D`. To be honest this PR wasn't tested as much as it should have been 😬.

In the meantime, could you try staging the deletion of the files?

EgbertRijke:
I say we revert that merge.

EgbertRijke:
However, the correct workflow for using make pre-commit is anyway that you should stage your changes before you use it, because the indexing files it generates will otherwise be out of sync. Therefore the correct thing to do is perhaps not to revert https://github.com/UniMath/agda-unimath/pull/753 but to add a pre-commit hook that checks if there are unstaged new files or deleted files. If that check fails, make pre-commit should be aborted, because it should then not touch the indexing files.

EgbertRijke:
Perhaps more clearly: `make pre-commit` should first check if there are any unstaged changes (of any kind). It should exit immediately with an error and the suggestion to stage the changes before using `make pre-commit`. It should only be allowed to continue with the other checks and hooks if there are no unstaged changes.

EgbertRijke:
@VojtechStep does putting this into our makefile do the job?:

```
pre-commit:
	# Check for unstaged changes
	@if git diff --exit-code > /dev/null; then \
		echo "Running pre-commit checks..."; \
		pre-commit run --all-files; \
		echo; \
		echo "Typechecking library"; \
		make check; \
	else \
		echo "There are unstaged changes. Please commit or stash them before running make pre-commit."; \
		exit 1; \
	fi
```

VojtechStep:
I propose an alternative: Let pre-commit and typechecking run only on the staged content, and fail fast if there are no staged changes. In my experience this is how these pre-commit frameworks are usually used: their purpose is to check that whatever you are about to commit is validated. We currently invoke pre-commit with the `--all-files` flag, sidestepping this logic. I'd need to do some experimentation to see how pre-commit handles it, but I expect it to create a stash with unstaged changes, so that the working directory is in the exact state that would be visible if the staged changes were committed and pushed, then it runs the checks (so our hooks and typechecking would act on the "correct" repo content), and then pops the stash (returning all your unstaged changes).

VojtechStep:
I mis-refactored the second sentence when I was writing the comment, sorry 😅. What I meant is we should run hooks only on staged files (since these should be ensuring "local consistency" IIRC), and run the typechecking on the whole repository with only the staged changes added. That would require us to add a pre-commit hook for the typechecking, since we'd want to run it on the version of the working directory as seen by pre-commit; but that should be possible.

EgbertRijke:
Do you think the pre-commit hook for generating the indexing files could work correctly if it only sees the staged changes?

Shouldn't it see all the tracked files in addition to the staged changes (minus the deleted ones, I guess)?

VojtechStep:
It should see the version of the repository with the staged changes applied.

At any point when you work on a repository, your working directory consists of:
- the state of the repository as of the last commit you have checked out
- your local changes, which fall under:
  - ignored changes (as specified by `.gitignore`)
  - Tracked changes, which further divide into:
    - staged changes (you added/modified/deleted parts of files, or you added/deleted files); these are conceptually a diff w.r.t. the clean state of the repository at the last commit on the branch you have checked out
    - unstaged changes (essentially the same as staged, except they aren't marked as staged); these are conceptually a diff w.r.t. the staged changes applied on top of the clean state of the repository at the last commit on the branch you have checked out

Pre-commit should then first reset the working directory to the "staged changes applied on top of the clean state of the repository at the last commit on the branch you have checked out" version, saving your unstaged changes as a stash. So when the hooks run, they have access to the whole repository, _including_ the staged changes, but not just them. Then pre-commit runs the hooks: usually the hooks are passed _names of files which have staged changes_, but the hook can ignore that information completely and just run `make check`, which should in theory do what we want, and typecheck the whole library as the CI would if you committed and pushed the staged changes.

EgbertRijke:
I like your proposal! Thanks so much for breaking it down for me, it looks very sensible now.

EgbertRijke:
I'm still a bit worried that the formalization checks can succeed, even if it shouldn't. Take the following scenario: I have added files `foo` and `bar` to the library. `foo` is unstaged and `bar` is staged. However, `bar` imports `foo`. Now pre-commit will not index `foo`, but it will index `bar`. However, since `foo` is locally present, when we check the library it will succeed on `bar`. On the other hand, if we push it to a pull request, then `foo` will not be present and the check will fail.

EgbertRijke:
Perhaps this scenario is not so bad, because it is both locally workable and the fact that a file is missing gets detected in the PR.

VojtechStep:
In this case the changes in `foo` would be moved to the stash before typechecking, so pre-commit would fail locally; the point of doing the stash is exactly so that there are no "invisibile" local-only changes.

VojtechStep:
I'll try to play with it (eventually, I'd really like the flattening lemma finished first 😩) and see if there are any complications that I'm missing

EgbertRijke:
Lol, so long for not letting library maintenance be held up by shiny new features XD

Anyway, I'm just being silly. I agree it would be nice to have the flattening lemma done.

PR #757, opened 2023-09-12
The classification of cyclic rings

EgbertRijke:
The purpose of this pull request is to introduce and classify rings of which the underlying abelian group is cyclic. We show that such rings are quotients of the initial ring. Consequently, the category of cyclic rings is a large poset. Indeed, this is the large poset of orders of group elements, which is ordered by divisibility. This large poset is also the opposite of the large poset of subgroups of the integers.

In future work we plan to show that the poset reflection of the category of connected set bundles over the circle is the poset of cyclic rings.

In this PR we introduce the following concepts:

**Category theory:**
- Initial objects in large precategories
- Initial objects in large categories

**Commutative algebra:**
- Groups of units of commutative rings

**Elementary number theory:**
- Multiplicative units of the integers
- Multiplicative units of modular arithmetic
- Rings of modular arithmetic

**Group theory:**
- Exponents of elements of abelian groups
- Exponents of elements of groups
- Isomorphisms of monoids

**Ring theory:**
- Additive orders of elements of rings
- The category of cyclic rings
- Characteristics of rings
- Cyclic rings
- Free rings with one generator
- Generating elements of rings
- Groups of units of rings
- Homomorphisms of cyclic rings
- Initial rings
- Kernels of ring homomorphisms
- Multiplicative orders of units of rings
- The poset of cyclic rings


src/group-theory/invertible-elements-monoids.lagda.md
fredrik-bakke, line 1: Seems we both did the exact same work on this file 😐

EgbertRijke: I saw that, but I didn't want to hold you back.

src/group-theory/isomorphisms-monoids.lagda.md
```diff
@@ -0,0 +1,331 @@
...
+    is-iso-hom-Monoid (id-hom-Monoid {M = M})
+  pr1 is-iso-id-hom-Monoid = id-hom-Monoid
+  pr1 (pr2 is-iso-id-hom-Monoid) =
+    left-unit-law-comp-hom-Monoid (id-hom-Monoid)
+  pr2 (pr2 is-iso-id-hom-Monoid) =
+    left-unit-law-comp-hom-Monoid (id-hom-Monoid)
+
+  id-iso-Monoid : iso-Monoid M M
+  pr1 id-iso-Monoid = id-hom-Monoid
+  pr2 id-iso-Monoid = is-iso-id-hom-Monoid
```
fredrik-bakke, line 129: these should just be corollaries of the corresponding theorems in large precategories as well

EgbertRijke: I'm doing that right now

fredrik-bakke:
Perhaps there is an intermediate goal you can set for this PR instead? It already has many good additions that would be nice to have merged. By the way, just a heads-up. I'm doing a little work on natural transformations at the moment.

EgbertRijke:
Ok, I think that won't interfere with the small additions that I did in `category-theory`. I agree I should work towards a natural stopping point. At the moment I'm showing that the underlying additive group of a ring is cyclic if and only if `1` is a generator for that cyclic group. When that is done I will tie up the loose ends.

src/elementary-number-theory/relatively-prime-natural-numbers.lagda.md
```diff
@@ -1,18 +1,25 @@
 # Relatively prime natural numbers
 
 ```agda
+{-# OPTIONS --allow-unsolved-metas #-}
```
EgbertRijke, line 4: `--allow-unsolved-metas`

src/ring-theory/cyclic-rings.lagda.md
```diff
@@ -0,0 +1,433 @@
+# Cyclic rings
+
+```agda
+{-# OPTIONS --allow-unsolved-metas #-}
```
EgbertRijke, line 4: `--allow-unsolved-metas`

EgbertRijke:
This PR now has a clear target, i.e., updated description. It is still not done though.

EgbertRijke:
The classification of cyclic rings is done. Now I have one more open hole in `elementary-number-theory.relatively-prime-natural-numbers` to deal with.

EgbertRijke:
I wrapped it up by removing the open holes:P This PR is now ready for review.

fredrik-bakke:
Awesome! I'm a bit tired at the moment, but I'll review it first thing tomorrow

src/elementary-number-theory/rings-of-modular-arithmetic.lagda.md
```diff
@@ -0,0 +1,52 @@
...
+open import ring-theory.integer-multiples-of-elements-rings
+open import ring-theory.rings
+```
+
+</details>
+
+## Idea
+
+The **standard cyclic rings** `ℤ/n` are the [rings](ring-theory.rings.md) of
+[modular arithmetic](elementary-number-theory.modular-arithmetic.md).
```
EgbertRijke, line 34: The fact that these are cyclic rings is for another PR

fredrik-bakke: Have you left a comment about this in the file?

EgbertRijke: No, is that a good thing to do? What is your convention?

fredrik-bakke: I think it is for a few of reasons.
- It conveys extra information about the topic at hand
- It documents your intention and serves as an effective todo note
- If you end up not formalizing it, someone else can notice that more easily and follow up on it
- If someone is looking for that particular result, they know to stop searching

fredrik-bakke: I also recommend using a standard phrase like "This remains to be formalized." or "It remains to be defined." in the file, so it is easy to find such things by text search. The particular phrase can be discussed of course.

EgbertRijke: Perfect! I will do this

fredrik-bakke: Actually, maybe we should develop a more elaborate phrase. Something that calls more explicitly for contribution by eventual readers

fredrik-bakke: We don't have to do that now, but its something to keep in mind.

EgbertRijke:
I will merge this, because my student will need it today. I can still make changes that come up in the review in a separate pull request.

fredrik-bakke (commented):
Phew, this has got to be the most detailed review I have given so far. There is tons of absolutely excellent work in this PR! Also, it is nice to see some of my own contributions being used for something. It's good to know not all of my contributions are completely useless.

Anyway, I took the time to sit down and think about some of the issues we have encountered but not resolved in the past with this review and wrote down my thoughts in the comments below. I hope this takes us one step closer to finding a better solution to these points. Also, I don't expect you to resolve all of my comments, but there are a few things I think you _should_ resolve, even though this PR has already been merged.

src/ring-theory/category-of-cyclic-rings.lagda.md
```diff
@@ -0,0 +1,111 @@
...
+## Idea
+
+The **large category of cyclic rings** is the
+[large category](category-theory.large-categories.md) consisting of
+[cyclic rings](ring-theory.cyclic-rings.md) and
+[ring homomorphisms](ring-theory.homomorphisms-cyclic-rings.md).
+
+Note that we already showed that there is at most one ring homomorphism between
+any two cyclic rings, so it follows that the large category of cyclic rings is
+in fact a [large poset](order-theory.large-posets.md). The large poset of cyclic
```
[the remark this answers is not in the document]
EgbertRijke: The point here is that the category of cyclic rings is first constructed as a category, and then it is a theorem that this category is in fact a poset. So it makes sense to have a file for it being introduced as a category, and a later file in which it is introduced as a poset, using the theorem that this category is a poset.

I am not proposing that for every poset we should have a corresponding file where it is a category. Whether there is a file for a category of certain objects should be context dependent, and I would be opposed to a blanket rule that disallows us from having first a file where we introduce the category of a certain kind of objects, and later deduce that it is a poset.

The reason that there should not be such a rule, is that we cannot foresee situations where it is natural to introduce a category of a certain kind of objects, but it might be a very deep and hard theorem, requiring an elaborate formalization effort, to show that it is a poset. With your rule in mind, we will not be allowed to introduce the file with the category of these objects, naturally refactor the hypothetically complicated proof that it is a poset, and finally introduce the poset in the file about the poset of these objects.

We need to allow for context dependency in our considerations for the organization of the library, and also for the naming of entries. Global rules that are blind to any relevant context that can help organize the rule, such as "no categories file if it is in fact a poset" can be damaging to the natural organization of the library. If for a certain category it makes sense to first introduce it as a category, and later deduce that it is a poset, then we should be able to organize the library accordingly.

In general, I am in favor of factoring stuff out into more files, and I don't see how having a "category-of-cyclic-rings" will have any significant effect on our ability to maintain the library.

fredrik-bakke:
No worries! And I'm glad to hear it

src/ring-theory/category-of-cyclic-rings.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: I was not trying to propose a blanket ban, and in this particular case, you've already deduced that if it's a category, it is a poset, before ever defining the category. However, that's not the point. What worries me is if this pattern reduces maintainability in the long run. In either case, we don't have to do anything about this now, and I think we can continue the pattern. But if we notice deterioration in maintainability in the long run, I suggest having a second look if restraining how many files we create can help this.

src/ring-theory/localizations-rings.lagda.md
```diff
@@ -62,7 +62,7 @@ is-left-inverse-inv-inverts-element-hom-Ring :
       ( inv-inverts-element-hom-Ring R S x f H)
       ( map-hom-Ring R S f x))
     ( one-Ring S)
-is-left-inverse-inv-inverts-element-hom-Ring R S x f H = pr1 (pr2 H)
+is-left-inverse-inv-inverts-element-hom-Ring R S x f H = pr2 (pr2 H)
```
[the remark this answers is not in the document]
fredrik-bakke: On second thought, there are probably instances where naming every projection is overkill, but I'll let the comment stand as I don't know if this is the case here or not.

PR #761, opened 2023-09-13
Refactor structured monoids

fredrik-bakke:
Playing around a little with representations :)
Disclaimer: I don't know what I'm doing here.

The library is undecided on the terminology `coherent-H-space` vs `wild-unital-magma`. Which do you prefer?


fredrik-bakke:
wild representations should probably go in a new file, but should it go into `structured-types`? Maybe a new directory for `wild-algebra` or `incoherent-algebra` is in order?

fredrik-bakke:
Not that I will be doing particularly much work in this direction, I'm just exploring a little at the moment.

EgbertRijke:
They should simply be called H-spaces, and the noncoherent ones should get an adjective. See https://arxiv.org/abs/2301.02636

EgbertRijke:
> wild representations should probably go in a new file, but should it go into `structured-types`? Maybe a new directory for `wild-algebra` or `incoherent-algebra` is in order?

There was a time where the structured-types folder was called wild algebra. But then I decided that structured types was a better name for it. So the wild representations could go in there.

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
```diff
@@ -0,0 +1,68 @@
...
+
+### The monoid of endomorphisms on an object in a precategory
+
+```agda
+module _
+  {l1 l2 : Level} (C : Precategory l1 l2) (X : obj-Precategory C)
+  where
+
+  endo-Precategory : UU l2
+  endo-Precategory = type-hom-Precategory C X X
```
EgbertRijke, line 33: I noticed in a few places that we're currently inconsistent about when we the prefix `type-`

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
```diff
@@ -0,0 +1,68 @@
...
+
+  left-unit-law-comp-endo-Precategory :
+    (f : endo-Precategory) → comp-endo-Precategory id-endo-Precategory f ＝ f
+  left-unit-law-comp-endo-Precategory = left-unit-law-comp-hom-Precategory C
+
+  right-unit-law-comp-endo-Precategory :
+    (f : endo-Precategory) → comp-endo-Precategory f id-endo-Precategory ＝ f
+  right-unit-law-comp-endo-Precategory = right-unit-law-comp-hom-Precategory C
+
+  set-endo-Precategory : Set l2
```
EgbertRijke, line 55: I think this should be named `endo-set-Precategory`. Given a precategory in your context (among other things), you construct a set, so that sets `set-Precategory`, and  this set should be named because it is not just an "underlying set". It's name is `endo` and that goes in front as a prefix.

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
```diff
@@ -0,0 +1,68 @@
...
+  set-endo-Precategory = hom-Precategory C X X
+
+  semigroup-endo-Precategory : Semigroup l2
```
EgbertRijke, line 58: `endo-semigroup-Precategory`

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
```diff
@@ -0,0 +1,68 @@
...
+  pr1 semigroup-endo-Precategory = set-endo-Precategory
+  pr1 (pr2 semigroup-endo-Precategory) = comp-endo-Precategory
+  pr2 (pr2 semigroup-endo-Precategory) = associative-comp-endo-Precategory
+
+  monoid-endo-Precategory : Monoid l2
```
EgbertRijke, line 63: `endo-monoid-Precategory`

src/foundation/homotopy-induction.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+open import foundation-core.equivalences
+open import foundation-core.functoriality-dependent-pair-types
+open import foundation-core.homotopies
+open import foundation-core.identity-types
+open import foundation-core.sections
+```
+
+</details>
+
+## Statement
```
EgbertRijke, line 26: There should definitely be an `## Idea` section here.

Given the fair amount of refactoring in this PR, I also think that this PR should not be held up by new developments of representation theory. It should focus on getting these nice refactorings merged into the library, and representation theory can be done in a separate PR.

src/foundation/homotopy-induction.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+
+```agda
+ev-refl-htpy :
+  {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2}
+  (f : (x : A) → B x) (C : (g : (x : A) → B x) → f ~ g → UU l3) →
+  ((g : (x : A) → B x) (H : f ~ g) → C g H) → C f refl-htpy
+ev-refl-htpy f C φ = φ f refl-htpy
+
+IND-HTPY :
```
EgbertRijke, line 35: How about `homotopy-induction`? I think that matches the new `equivalence-induction` statement.

src/foundation/homotopy-induction.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+        ( is-contr-total-path' f)
+```
+
+### Homotopy induction is equivalent to function extensionality
+
+```agda
+abstract
+  IND-HTPY-FUNEXT :
+    {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2} (f : (x : A) → B x) →
+    FUNEXT f → IND-HTPY {l3 = l3} f
```
EgbertRijke, line 74: I guess in a similar vein we can rename `FUNEXT` to `function-extensionality`

src/foundation/homotopy-induction.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+  {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2}
+  (f : (x : A) → B x) (C : (g : (x : A) → B x) → f ~ g → UU l3) →
+  ((g : (x : A) → B x) (H : f ~ g) → C g H) → C f refl-htpy
+ev-refl-htpy f C φ = φ f refl-htpy
+
+IND-HTPY :
+  {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2}
+  (f : (x : A) → B x) → UU (l1 ⊔ l2 ⊔ lsuc l3)
+IND-HTPY {l1} {l2} {l3} {A} {B} f =
+  (C : (g : (x : A) → B x) → f ~ g → UU l3) → section (ev-refl-htpy f C)
```
EgbertRijke, line 39: Can this be expressed in terms of identity systems without causing a cyclic dependency?

src/group-theory/representations-monoids.lagda.md
```diff
@@ -11,66 +11,126 @@ open import category-theory.categories
...
+    id-endo-Category C obj-representation-category-Monoid
+  preserves-unit-action-representation-category-Monoid =
     preserves-unit-hom-Monoid M
-      ( monoid-endo-Category C obj-categorical-representation-Monoid)
-      ( hom-action-categorical-representation-Monoid)
+      ( monoid-endo-Category C obj-representation-category-Monoid)
+      ( hom-action-representation-category-Monoid)
+```
+
+### Wild representations of a monoid in types
```
EgbertRijke, line 87: I would suggest we put this into a separate file called `wild-representations-monoids`. We can still keep it in this folder, because it is about monoids.

EgbertRijke: Actually I suggest that you take this into a separate PR, because it is a small amount of code (easy to take out) that seems to be quite unfinished, and the other refactorings are of quite different nature.

src/structured-types/coherent-h-spaces.lagda.md
```diff
@@ -31,9 +34,9 @@ open import structured-types.pointed-types
 
 ## Idea
 
-A coherent H-space is a "wild unital magma", i.e., it is a pointed type equipped
-with a binary operation for which the base point is a unit, with a coherence law
-between the left and the right unit laws.
+A **coherent H-space** is a "wild unital magma", i.e., it is a pointed type
```
EgbertRijke, line 37: I think we should convert these to the standard definition of H-spaces and the current definition of H-spaces should be called noncoherent H-spaces. This doesn't have to be done in this PR, but it can be done here if you want.

src/structured-types/coherent-h-spaces.lagda.md
```diff
@@ -171,3 +189,16 @@ compute-pointed-section-ev-point-Pointed-Type (pair A a) =
       Σ ( (x : A) → pr1 μp x a ＝ x)
         ( λ H → H a ＝ ( ( ap (λ h → h a) (pr2 μp) ∙ refl) ∙ refl))))
 ```
+
+### Monoids are coherent H-spaces
+
+```agda
+coherent-h-space-Monoid : {l : Level} (M : Monoid l) → Coherent-H-Space l
```
EgbertRijke, line 196: This belongs to the file about monoids. 

Similarly, we also don't construct the multiplicative monoid of a ring in the file about monoids.

src/structured-types/dependent-products-coherent-h-spaces.lagda.md
```diff
@@ -0,0 +1,104 @@
...
+
+</details>
+
+## Idea
+
+Given a family of [coherent H-spaces](structured-types.coherent-h-spaces.md)
+`Mᵢ` indexed by `i : I`, the dependent product `Π(i : I), Mᵢ` is a coherent
+H-space consisting of dependent functions taking `i : I` to an element of the
+underlying type of `Mᵢ`. The multiplicative operation and the unit are given
+pointwise.
```
EgbertRijke, line 29: Nice!

src/structured-types/pointed-types.lagda.md
```diff
@@ -45,6 +45,15 @@ ev-point-Pointed-Type :
 ev-point-Pointed-Type A f = f (point-Pointed-Type A)
 ```
 
+### Dependent products of pointed types
```
EgbertRijke, line 48: This can be in a separate file

EgbertRijke (commented):
Very nice pull request! I made one recommendation to take out a small code block that seemed unfinished. The reason is that your PR contains a lot of useful refactorings that are independent and shouldn't be held up by trying to do something new (a mistake that I'm learning the hard way).

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
fredrik-bakke: The reasoning for using `set-endo-Precategory` is that it is the projection that gives the underlying `set` of `endo-Precategory`. Similarly, we could define `is-set-endo-Precategory` and so on. If we change the ordering, this will not be possible. The reason I switched the name `hom-Wild-Monoid` to `type-hom-Wild-Monoid` is because I expect `hom-Wild-Monoid` is perhaps a wild monoid itself. Similarly, we could reserve `endo-Precategory` for the associated precategory structure, but I think that is a less relevant construction.

fredrik-bakke: `[conclusion]-[context]`
`[semigroup]-[endo-Precategory]`

I thought this was comparable to
`Prop` -> `type-Prop` -> `is-prop-type-Prop`

fredrik-bakke: With my reasoning above, I guess `endo-Precategory` should in some sense refer to the "most structured" thing we can build.

EgbertRijke: I know that we're not entirely consistent with this in the library. I follow your reasoning for `set-endo-Precategory`, but I'd note that `endo-Precategory` is a type so it doesn't have an underlying set. If `endo-Precategory` would be the monoid, then `set-endo-Precategory` would make more sense.

For example, with the kernel of a group homomorphism we define the kernel to be the normal subgroup, and then we have the naming scheme `subgroup-kernel-hom-Group`, `set-kernel-hom-Group`, and so on.

I feel like the naming scheme `[name]-[target]-[source]` is also very common in our library, and `endo-set-Precategory` would fit in that scheme.

src/foundation/homotopy-induction.lagda.md
fredrik-bakke: I would be more inclined to name it `induction-principle-homotopy` and `induction-principle-equivalence`. This would match better with the rest of the library.

src/foundation/homotopy-induction.lagda.md
fredrik-bakke: `extensionality-Π` and `extensionality-function` you mean? These don't ring that well...

fredrik-bakke: Anyways, this refactoring can be part of a separate PR I think. I only wanted to do a little algebra for once...

src/foundation/homotopy-induction.lagda.md
EgbertRijke: Sure! Let's get rid of those all-caps

EgbertRijke: except that `induction-principle-homotopy` is blatantly ungrammatical.

fredrik-bakke: homotopies* then

EgbertRijke: I think it's fine to read `equivalence-induction` as "the statement of equivalence induction". What other parts of the library does this name have a conflict with?

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
fredrik-bakke: Personally, I want to keep it in the current order. This is how I have always named things in the library up until now, I believe. If it makes you happier, I'm fine with renaming `endo-Precategory` to `type-endo-Precategory`.

EgbertRijke: Sure, can you make `endo-Precategory` to be the monoid in that case?

src/foundation/homotopy-induction.lagda.md
fredrik-bakke: `universal-property-***` is pretty common. I had this discussion with @VojtechStep recently, and we concluded that `induction-principle` is the most consistent name.

fredrik-bakke: Sorry, I didn't mean to at you there, Vojtech

EgbertRijke: Yep, you're right!

Ok, let's go with `induction-principle-homotopies` and `induction-principle-equivalences`

src/category-theory/endomorphisms-of-objects-precategories.lagda.md
fredrik-bakke: I guess this is consistent with both of my proposals since precategories are monoid(-pre-?)oids, but perhaps it is better to just not define `endo-Precategory`, or define it as an on the dot `Precategory`?

EgbertRijke: I think it is fine to define it as a monoid in the existing sense of the word, and there could separately be a construction that turns a monoid into a single-object precategory.

fredrik-bakke: I think this is a bad choice for a couple of reasons
- We usually use `[name]-[Subuniverse]` to define constructions internal to the subuniverse. I.e. `endo-Precategory` should be the internal object of endomorphisms, so a `Precategory`.
- If I went to go look for the endomorphism monoid, I would probably start looking for `monoid-endo-Precategory`. I don't think it is immediately obvious to a user that this should have just the name `endo-Precategory`.

fredrik-bakke: > With my reasoning above, I guess `endo-Precategory` should in some sense refer to the "most structured" thing we can build.

So to backtrack, this seems to be a bad principle to follow to me.

EgbertRijke: That's a valid concern. I don't think we have ever dealt with internal objects in this library yet, but wouldn't we write that as `endo-monoid-object-Precategory-Precategory`?

I.e. it is the monoid object named `endo` in the precategory of precategories?

EgbertRijke: > > With my reasoning above, I guess `endo-Precategory` should in some sense refer to the "most structured" thing we can build.
> 
> So to backtrack, this seems to be a bad principle to follow to me.

So the next best thing is to follow my suggestion? 👀

EgbertRijke: [hides real fast]

fredrik-bakke: I don't see how your suggestion improves the situation.
I've always thought of indexed products and the like as "internal" constructions, but maybe I'm wrong. I also expected for instance `hom-Monoid` to be the monoid of homomorphisms of monoids by multiplying point-wise, but I see I'm wrong about that, so perhaps I've just misinterpreted things.

fredrik-bakke: I guess this would cause problems when one considers what would then be a double monoid of endomorphisms on a monoid.

EgbertRijke: It used your ordering of the name btw

src/foundation/homotopy-induction.lagda.md
EgbertRijke: > `extensionality-Π` and `extensionality-function` you mean? These don't ring that well...

No, I didn't mean these things. We have a fairly well-established naming scheme for `extensionality-***` to mean the _proof_ of extensionality of `***`, not the _statement_of extensionality. On the other hand, if you would insist on foregoing a natural name like `function-extensionality` we could make the naming scheme `extensionality-principle-***` and use `function-type` instead of `function`.

EgbertRijke: I do wonder how rigid we want naming schemes to be. We want to be fairly consistent, to be sure, but we also want to enable natural, commonly used names that concisely describe what is meant. On the other hand, I am also just wrong sometimes.

EgbertRijke: > Anyways, this refactoring can be part of a separate PR I think. I only wanted to do a little algebra for once...

Ah yes! Don't let me trap you into refactoring forever. Do some math =)

src/foundation/homotopy-induction.lagda.md
fredrik-bakke: I am going to leave this for someone else to do :)

EgbertRijke:
```suggestion
## Idea

The principle of **homotopy induction** asserts that homotopies form an [identity system](foundation.identity-systems.md) on dependent function types.

## Statement
```

There you go! Now there's something, and then someone can improve it later.

src/structured-types/coherent-h-spaces.lagda.md
fredrik-bakke: I can't really tell, is the file on "The involutive type of H-space structures on a pointed type" about noncoherent or coherent h-spaces?

EgbertRijke: That's about coherent H-spaces

fredrik-bakke: Thanks! 😁

src/foundation/homotopy-induction.lagda.md
fredrik-bakke: Thanks! The duo strikes again 😎

EgbertRijke: All you need is 💖

fredrik-bakke:
On second thought, maybe `endomorphisms-in-categories` and `endomorphisms-of-categories` is a more natural phrasing than `endomorphisms-of-objects-categories` and `endomorphisms-of-categories`. When it is written `endomorphisms-of-objects-categories`, it feels more like a secondary notion, although that's clearly not true.

fredrik-bakke:
(I figured I might as well rename all the category theory files in this PR, to make everything consistent immediately)

EgbertRijke:
Don't tell Vojtech

fredrik-bakke:
This is now officially a refactor PR 👨‍⚖️

EgbertRijke:
Agreed 👊

fredrik-bakke:
If the checks pass now, I think this PR is ready for review.

fredrik-bakke:
There is a slight discrepancy in the naming of files in `category-theory` currently. For instance, we have the files `functors-categories` and `equivalences-of-categories`. Also,  for file titles, the preposition "between" is used instead of "of". For instance, `# Functors between categories`.

EgbertRijke:
I think `equivalences-of-categories` is perfectly good

EgbertRijke:
Let's see who's faster, the checks or you. Because I just went over your changes and it looks ready to be merged:)

fredrik-bakke:
Alright, I'll leave things as they are.

PR #762, opened 2023-09-13
Revert "adjusting loc"

EgbertRijke:
Reverts UniMath/agda-unimath#753

It is causing problems.


EgbertRijke:
The alternative is that `make pre-commit` should return a warning if there are unstaged new files or deleted files

PR #763, opened 2023-09-13
rational commutative monoids

EgbertRijke:
This small PR factors out rational commutative monoids from #623.

### DIFF
--- a/src/group-theory.lagda.md
+++ b/src/group-theory.lagda.md
@@ -116,6 +116,7 @@ open import group-theory.orbits-concrete-group-actions public
 open import group-theory.orbits-group-actions public
 open import group-theory.orbits-monoid-actions public
 open import group-theory.orders-of-elements-groups public
+open import group-theory.powers-of-elements-commutative-monoids public
 open import group-theory.powers-of-elements-groups public
 open import group-theory.powers-of-elements-monoids public
 open import group-theory.precategory-of-abelian-groups public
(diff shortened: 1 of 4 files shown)


fredrik-bakke (approved):
Looks good. I have just one comment.

EgbertRijke:
done

fredrik-bakke:
Ah, now I see why you suggested the opposite naming scheme in my other PR!

EgbertRijke:
There is the old naming scheme that would give `is-rational-Commutative-Monoid-Prop`, but I started to really dislike this because we only do something odd like this for propositions, and it goes against the rest of the naming scheme. So I thought, ok we start with an object, construct a proposition, and that proposition has a name, so the naming scheme should be `[name]-prop-[Namespace]` where `Namespace` is the name of a category that we kind of treat like an artificial namespace. For example `Ring`, `Group`, or `Commutative-Monoid`.

To me it is clear that `[name]-prop-Commutative-Monoid` would take a context with a commutative monoid featuring prominently in it, and construct a proposition. That proposition then has name `[name]`. Idk... to me it makes sense and it fits well with the rest of the naming scheme.

fredrik-bakke:
Yep, it definitely makes more sense to me now. I fear the library is wildly inconsistent on this point though.

EgbertRijke:
I don't want to undertake this refactoring at the moment, or anybody else to do this refactoring... It is more important to finish refactoring the synthetic homotopy theory (which I do intend to work on).

fredrik-bakke:
Yep, very understandable. This is not important at all. Also, I'm not entirely convinced yet.

PR #765, opened 2023-09-13
Define representations of monoids

fredrik-bakke:
### Summary
- Define concrete monoids
- Define wild representations of monoids in types
- Define precategory representations of monoids
- Define (dependent) products of (pre)categories
- Isomorphisms in (dependent) product (pre)categories are fiberwise isos
- Characterize equality of cospans and spans with fixed end points
- A pinch of refactoring
- Move trivial higher groups and concrete groups to separate files


EgbertRijke:
Would it be sufficient for your purpose if they are precategories?

fredrik-bakke:
That would help, sure. Is it formalized?

EgbertRijke:
It's not, but I'll do it for you:)

EgbertRijke:
Too late:)

fredrik-bakke:
I find the name `G-Space` a bit awkward since the `G` refers to a group. Is it perhaps the same for `H`-spaces? Or do you have a better name for them? Something like `types-equipped-with-a-group-action`?

Btw, I still have no idea what I'm doing. Please don't yell at me if it's stupid

EgbertRijke:
We call it `Abstract-Group-Action` in the library, to distinguish it from `Concrete-Group-Action`. See

https://unimath.github.io/agda-unimath/group-theory.group-actions.html

EgbertRijke:
Are you redoing this:

https://unimath.github.io/agda-unimath/higher-group-theory.higher-group-actions.html

but assuming that the higher group is a concrete 1-group? Then I guess we can still call it a higher group action over a concrete group.

EgbertRijke:
That naming scheme is broken by the way, and it should be in line with the naming scheme for concrete groups and abstract groups, which are developed much further.

fredrik-bakke:
Ah, yep, looks like it. Thanks!

src/higher-group-theory/higher-groups.lagda.md
fredrik-bakke, line 1: I would expect a file with this name to contain the definition of `n`-groups as well. But I would also expect that to be in a different file from infinity-groups. Thus, I propose having files `n-groups` and `infinity-groups` instead.

EgbertRijke: Would it be reasonable to create a file `truncated-higher-groups` for the `n`-groups, and keep the file `higher-groups` for the untruncated case? 

The reason I would speak of truncated groups and not of `n`-groups is just that I don't like to refer to variable names in names of things.

EgbertRijke: Note that I wouldn't mind if the truncation levels of truncated higher groups are indexed by the natural numbers instead of the type of truncation levels, to keep some consistency with the literature.

src/foundation/spans.lagda.md
```diff
@@ -0,0 +1,132 @@
...
+    fundamental-theorem-id (is-contr-total-htpy-span c) (htpy-eq-span c)
+
+  extensionality-span :
+    (c d : span l A B) → (c ＝ d) ≃ (htpy-span c d)
+  pr1 (extensionality-span c d) = htpy-eq-span c d
+  pr2 (extensionality-span c d) = is-equiv-htpy-eq-span c d
+
+  eq-htpy-span : (c d : span l A B) → htpy-span c d → c ＝ d
+  eq-htpy-span c d = map-inv-equiv (extensionality-span c d)
+```
```
EgbertRijke, line 132: Perhaps it can be left as a good first issue to show that spans are equivalent to relations

fredrik-bakke: please go ahead!

src/foundation/spans.lagda.md
```diff
@@ -0,0 +1,132 @@
...
+  left-map-span = pr1 (pr2 c)
+
+  right-map-span : type-span → B
+  right-map-span = pr2 (pr2 c)
+```
+
+### Homomorphisms of cospans
+
+One notion of homomorphism of spans `c` and `d` with common domains is a map
+between their codomains so that the triangles on either side commute:
```
fredrik-bakke, line 60:
```suggestion
between their domains so that the triangles on either side commute:
```

src/foundation/spans.lagda.md
EgbertRijke: It is issue #767. Would you be so kind to record it here?

src/group-theory/concrete-monoids.lagda.md
```diff
@@ -0,0 +1,63 @@
...
+[category](category-theory.categories.md) whose type of objects is
+[pointed](structured-types.pointed-types.md) and
+[connected](foundation.connected-types.md).
+
+## Definition
+
+```agda
+is-concrete-monoid-Category : {l1 l2 : Level} → Category l1 l2 → UU l1
+is-concrete-monoid-Category C =
+  obj-Category C × is-0-connected (obj-Category C)
```
fredrik-bakke, line 48: I'm curious, is this definition nonsensical? 😅

fredrik-bakke: I'm for instance worried there's a coherence issue between the unit object and the family of identity morphisms

EgbertRijke: I don't think this definition has coherence issues. What issue are you worried about?

fredrik-bakke: I'm not entirely sure 😅 But then I will soldier ahead :)

fredrik-bakke:
Oh dear, I'm creating a hydra PR again

EgbertRijke:
Welcome to the club

src/higher-group-theory/higher-groups.lagda.md
fredrik-bakke: Not going to do this in this PR

EgbertRijke:
I'll review this tomorrow

fredrik-bakke:
Excellent! There is no rush :)

src/group-theory/wild-representations-monoids.lagda.md
```diff
@@ -0,0 +1,84 @@
...
+open import group-theory.monoids
+
+open import structured-types.morphisms-wild-monoids
+```
+
+</details>
+
+## Idea
+
+A coherent action of a [monoid](group-theory.monoids.md) on a type `X` requires
```
fredrik-bakke, line 25:
```suggestion
A coherent action of a [monoid](group-theory.monoids.md) `M` on a type `X` requires
```

src/group-theory/wild-representations-monoids.lagda.md
```diff
@@ -0,0 +1,84 @@
...
+an infinite hierarchy of explicit coherences. Instead, as a first order
+approximation, we can consider **wild representations** of `M` on `X`,
+consisting of a
+[wild monoid homomorphism](structured-types.morphisms-wild-monoids.md) from `M`
+to the [wild monoid](structured-types.wild-monoids.md) of
+[endomorphisms](category-theory.endomorphisms-in-categories.md) on `X`.
```
fredrik-bakke, line 31:
```suggestion
[endomorphisms](foundation.endomorphisms.md) on `X`.
```

src/higher-group-theory/higher-groups.lagda.md
```diff
@@ -24,8 +24,10 @@ open import synthetic-homotopy-theory.loop-spaces
 
 ## Idea
 
-An ∞-group is just a pointed connected type. Its underlying type is its loop
-space, and the classifying type is the pointed connected type itself.
+An **∞-group** is just a [pointed](structured-types.pointed-types.md)
+[connected](foundation.connected-types.md) type. Its underlying type is its
```
fredrik-bakke, line 28:
```suggestion
[connected](foundation.0-connected-types.md) type. Its underlying type is its
```

src/foundation/spans.lagda.md
```diff
@@ -0,0 +1,142 @@
...
+  left-map-span = pr1 (pr2 c)
+
+  right-map-span : domain-span → B
+  right-map-span = pr2 (pr2 c)
+```
+
+### Homomorphisms of cospans
+
+One notion of homomorphism of spans `c` and `d` with common domains is a map
+between their domains so that the triangles on either side commute:
```
fredrik-bakke, line 60:
```suggestion
### Homomorphisms between spans with fixed codomains

One notion of homomorphism of spans `c` and `d` with common codomains is a map
between their domains so that the triangles on either side commute:
```

src/foundation/cospans.lagda.md
```diff
@@ -8,23 +8,134 @@ module foundation.cospans where
...
+  codomain-cospan = pr1 c
+
+  left-map-cospan : A → codomain-cospan
+  left-map-cospan = pr1 (pr2 c)
+
+  right-map-cospan : B → codomain-cospan
+  right-map-cospan = pr2 (pr2 c)
+```
+
+### Homomorphisms of cospans
```
fredrik-bakke, line 57:
```suggestion
### Homomorphisms between cospans with fixed domains
```

src/category-theory/categories.lagda.md
```diff
@@ -101,9 +106,22 @@ module _
...
+      hom-Category
+      associative-composition-structure-Category
+  is-unital-composition-structure-Category =
+    is-unital-composition-structure-Precategory precategory-Category
+
   is-category-Category :
     is-category-Precategory precategory-Category
   is-category-Category = pr2 C
+
+  equiv-iso-eq-Category :
```
EgbertRijke, line 120: `extensionality-object-Category`

EgbertRijke: Or `extensionality-obj-Category`, because we're using `obj-Category` as the name for the type of objects

src/category-theory/dependent-products-of-categories.lagda.md
```diff
@@ -0,0 +1,112 @@
...
+
+  is-category-precategory-Π-Category :
+    is-category-Precategory precategory-Π-Category
+  is-category-precategory-Π-Category x y =
+    is-equiv-htpy-equiv
+      ( equiv-iso-Π-fiberwise-iso-Precategory I (precategory-Category ∘ C) ∘e
+        equiv-Π-equiv-family (λ i → equiv-iso-eq-Category (C i)) ∘e
+        equiv-funext)
+      ( λ {refl → refl})
+
```
EgbertRijke, line 52: You could add `extensionality-obj-Π-Category` in this file too.

src/category-theory/dependent-products-of-precategories.lagda.md
```diff
@@ -0,0 +1,227 @@
...
+  pr1 Π-Precategory = obj-Π-Precategory
+  pr1 (pr2 Π-Precategory) = hom-Π-Precategory
+  pr1 (pr2 (pr2 Π-Precategory)) =
+    associative-composition-structure-Π-Precategory
+  pr2 (pr2 (pr2 Π-Precategory)) = is-unital-Π-Precategory
+```
+
+## Properties
+
+### Isomorphisms in the dependent product precategory are fiberwise isomorphisms
```
EgbertRijke, line 105: These properties also transfer to dependent products of categories

src/category-theory/function-categories.lagda.md
```diff
@@ -0,0 +1,101 @@
...
+function type `I → C` is a category consisting of functions taking `i : I` to an
+object of `C`. Every component of the structure is given pointwise.
+
+## Definition
+
+```agda
+module _
+  {l1 l2 l3 : Level} (I : UU l1) (C : Category l2 l3)
+  where
+
```
EgbertRijke, line 33: add `extensionality-obj-function-Category` to this module.

The name "function category" does sound a bit wrong though. What are such categories called in CFWM?

src/category-theory/function-precategories.lagda.md
```diff
@@ -0,0 +1,93 @@
...
+  right-unit-law-comp-hom-function-Precategory =
+    right-unit-law-comp-hom-Precategory function-Precategory
+
+  is-unital-function-Precategory :
+    is-unital-composition-structure-Set
+      hom-function-Precategory
+      associative-composition-structure-function-Precategory
+  is-unital-function-Precategory =
+    is-unital-composition-structure-Precategory function-Precategory
+```
```
EgbertRijke, line 93: The characterization of isos as fiberwise isos could also be recorded here in this file.

src/category-theory/function-categories.lagda.md
```diff
@@ -0,0 +1,101 @@
...
+  right-unit-law-comp-hom-function-Category =
+    right-unit-law-comp-hom-Category function-Category
+
+  is-unital-function-Category :
+    is-unital-composition-structure-Set
+      hom-function-Category
+      associative-composition-structure-function-Category
+  is-unital-function-Category =
+    is-unital-composition-structure-Category function-Category
+```
```
EgbertRijke, line 101: The characterization of isos as fiberwise isos could also be recorded here in this file.

src/category-theory/one-object-precategories.lagda.md
```diff
@@ -0,0 +1,154 @@
...
+
+open import group-theory.monoids
+```
+
+</details>
+
+## Definition
+
+```agda
+is-one-object-Precategory : {l1 l2 : Level} → Precategory l1 l2 → UU l1
```
EgbertRijke, line 28: Perhaps we also want to record the proposition `is-one-object-prop-Precategory`

src/group-theory/concrete-groups.lagda.md
```diff
@@ -83,25 +85,26 @@ module _
   is-set-type-Concrete-Group = pr2 G
 
   set-Concrete-Group : Set l
-  set-Concrete-Group = pair type-Concrete-Group is-set-type-Concrete-Group
+  pr1 set-Concrete-Group = type-Concrete-Group
+  pr2 set-Concrete-Group = is-set-type-Concrete-Group
 
   is-1-type-classifying-type-Concrete-Group :
```
EgbertRijke, line 91: We should have a policy that definitions in which we use pattern matching lambda should always be abstract, because they don't compute nicely anyway.

src/group-theory/concrete-monoids.lagda.md
```diff
@@ -0,0 +1,76 @@
+# Concrete monoids
```
EgbertRijke, line 1: I like this definition a lot. Well done for coming up with it!

src/category-theory/function-categories.lagda.md
fredrik-bakke: I had a quick look, but doesn't seem to be in there. Other possible alternatives are `exponent-Category` and `function-type-Category`

src/higher-group-theory/higher-groups.lagda.md
```diff
@@ -117,7 +119,8 @@ module _
```
EgbertRijke, line 119: Actually this definition should be in a separate file called `trivial-higher-groups.lagda.md`

EgbertRijke:
Very nice PR! I had just a few small remarks, and I want to add that I really liked these new additions to the library. Great work!

src/higher-group-theory/higher-groups.lagda.md
fredrik-bakke: Would you define a trivial higher group to be one whose classifying type is contractible, or whose loop space is contractible? 😅

fredrik-bakke: of course both are equivalent

src/category-theory/function-categories.lagda.md
EgbertRijke: Perhaps `exponent-Category` is a good name. One risk of confusion is that exponential objects in a category takes two objects `X` and `Y` and returns an object `X^Y`. Then, the exponential category of two categories would be the functor category. But we already have a good name for that exponential category (namely the functor category), so perhaps we can get away with `exponent-Category`.

EgbertRijke: Or we can just leave it as it is. I'll let you decide.

src/higher-group-theory/higher-groups.lagda.md
EgbertRijke: Showing that both are equivalent would be a natural thing to prove in that file about trivial higher groups. I guess the most natural definition of a trivial group is that its underlying type (the loop space) is contractible, so that makes the other definition `has-contractible-classifying-type-oo-Group`.

fredrik-bakke:
I'm glad to hear it! Sometimes my PRs end up making way more definitions and proving way fewer things than I intended 😅

src/category-theory/function-categories.lagda.md
fredrik-bakke: I find them both about equally as confusing. Let's keep it like it is.

fredrik-bakke:
Alright, that should address all your comments. Note that I did not prove that the two definitions of triviality for higher groups are equivalent, however.

fredrik-bakke:
let me quickly trim down some more unused imports by the way

EgbertRijke:
> let me quickly trim down some more unused imports by the way

How did you do that so quickly? I ran the script overnight last nigth.

fredrik-bakke:
I introduced an optimization some months back making it process the most recently modified files first. So when you only want to find unused imports related to your current PR, it's pretty quick.

EgbertRijke:
What do you type to run the script only on the current PR?

fredrik-bakke:
Oh, I'll just run it as normal and throw out changes not related to the files I've edited.

fredrik-bakke:
I find it a little amusing that we always check if our proofs are valid on both Ubuntu and Mac.

PR #766, opened 2023-09-13
The one-object precategory of a monoid

EgbertRijke:


EgbertRijke:
There you go @fredrik-bakke

fredrik-bakke:
Haha, it really wasn't necessary, but thanks!

EgbertRijke:
I like big libraries, and I cannot lie

PR #772, opened 2023-09-15
update contributors, remove unused imports

EgbertRijke:


EgbertRijke:
Note that we're currently not including co-authors of commits on our contributors page. This seems to be an oversight.

CONTRIBUTORS.md
```diff
@@ -31,7 +32,6 @@ We are grateful to the following people for their contributions to the library.
 - [Matthias Hutzler](https://github.com/MatthiasHu)
 - [Nathan van Doorn](https://github.com/Taneb)
 - [Pierre Cagne](https://github.com/pierrecagne)
-- [Tom de Jong](https://github.com/tomdjong)
```
EgbertRijke, line 34: I don't know why Tom isn't recognized anymore

VojtechStep: Apparently the `/repo/.../contributors` API endpoint has a default limit of 30 results. For now we can bump it to 100, which is the maximum, but a proper fix would include a more robust fetching implementation which properly handles pagination

VojtechStep: To do the hotfix, change
```python
url = f'https://api.github.com/repos/{repo}/contributors'
```
to
```python
url = f'https://api.github.com/repos/{repo}/contributors?per_page=100'
```

in `scripts/update-contributors.py`

EgbertRijke: After running the script with this change I get:

```
egbertrijke@MacBook-Pro-3 agda-unimath % python3 scripts/update_contributors.py
Failed to retrieve information for user MatthiasHu. Status code: 403
Failed to retrieve information for user Taneb. Status code: 403
Failed to retrieve information for user pierrecagne. Status code: 403
Failed to retrieve information for user tomdjong. Status code: 403
Failed to retrieve information for user louismntnu. Status code: 403
```

EgbertRijke: Let me just add Tom back in by hand and fix this later.

VojtechStep: I did some digging in the morning, and apparently the way GitHub handles attribution, usernames and emails is more fucked up then I thought; I'd like to discuss it in the office, but if you want some background reading this shows that squash merges aren't great for keeping that information (GH rewrites author's info to whatever they have setup for the GH account during merging instead of the original commits): https://github.com/isaacs/github/issues/1368

EgbertRijke: I'm fine with not squashing. It was a concern of other developers that the history might be messy if we don't squash.

fredrik-bakke: I'm personally a fan of squashing. It cleans up the mess that I and others make of our commit histories before they reach master. But if there is an argument that a finer granularity on the history is better, then I can be convinced otherwise on this point. Another benefit of not squashing is that there won't be a need for rebasing if you want to merge a PR that branched off from a previous PR before it was itself merged, if I'm not mistaken. Anyways, that's just my input disregarding the issue at hand.

PR #782, opened 2023-09-17
Moving file about hexagons of identifications

EgbertRijke:
1. This PR moves the file to `foundation.commuting-hexagons-of-identifications.lagda.md`
2. Furthermore it moves the content of `foundation-core.commuting-cubes-of-maps.lagda.md` to `foundation`, and removes the `foundation-core` file.


src/foundation/commuting-cubes-of-maps.lagda.md
```diff
@@ -2,24 +2,413 @@
...
+      ( inv-htpy (rectangle-top-front-left-cube f g h k f' g' h' k' hA hB hC hD
+        top back-left back-right front-left front-right bottom))))
+    ( pair g' (pair (f ∘ hA)
+      ( inv-htpy
+        ( rectangle-back-right-bottom-cube f g h k f' g' h' k' hA hB hC hD
+          top back-left back-right front-left front-right bottom))))
+    ( refl-htpy' g')
+    ( inv-htpy back-left)
+coherence-htpy-square-rectangle-top-fl-rectangle-br-bot-cube = {!!}
+-}
```
fredrik-bakke, line 371: Do you have a stance on commented code like this? I think it should be removed.

fredrik-bakke (approved):

src/foundation/commuting-cubes-of-maps.lagda.md
EgbertRijke: Yes, let's keep it until someone proves this.

For instance, in the case of the flattening lemma, we had commented code in `26-descent` which was very helpful to us.

We shouldn't add new commented code though.

fredrik-bakke: Thanks! I agree we shouldn't add new commented code. The chance it will ever be finished is not the best, and it can be quite hard to build on someone's old commented code, to the point where it would be easier to write out the commented code from scratch again. In addition, we have the literate file format for people to elaborate on unfinished ideas instead. This also looks a lot better than the commented code. And if all you have is a type signature, like here, I think it is better to just make a definition for the type and explain in prose what has been left for future work.

EgbertRijke: Ok I'll try to crack it

EgbertRijke: The totient function will just have to wait longer :P

fredrik-bakke: Just to be clear, I was not at all trying to convince you to prove it. Clearly, it hasn't been very important to prove this particular statement...

EgbertRijke: I'm already on it, and I found some other ways to clean up that file.

fredrik-bakke: Alright, nice!

EgbertRijke: I did my refactorings and rewrote that commented code so that it type-checks again, but I moved it to the comments outside the code block because I don't want to waste a bunch of time on this at this moment.

EgbertRijke:
I'm merging this, so that @morphismz doesn't get stuck.

src/foundation/cones-over-cospans.lagda.md
```diff
@@ -100,8 +100,8 @@ cone-family {C = C} PX {f = f} {g} f' g' c PC =
 Next we characterize the identity type of the type of cones with a given vertex
 C. Note that in the definition of htpy-cone we do not use pattern matching on
 the cones c and c'. This is to ensure that the type htpy-cone f g c c' is a
-Σ-type for any c and c', not just for c and c' of the form (pair p (pair q H))
-and (pair p' (pair q' H')) respectively.
+Σ-type for any c and c', not just for c and c' of the form (p , q , H) and (p' ,
+q' , H') respectively.
```
fredrik-bakke, line 104: A lot of things here should be put in code guards

src/foundation/commuting-cubes-of-maps.lagda.md
```diff
@@ -222,248 +230,145 @@ module _
...
+    (c : coherence-cube-maps f g h k f' g' h' k' hA hB hC hD
+      top back-left back-right front-left front-right bottom) →
+    coherence-htpy-parallel-cone
+      ( inv-htpy front-right)
+      ( refl-htpy' h)
+      ( g' , hB ∘ f' , inv-htpy (rectangle-top-front-left-cube))
+      ( g' , f ∘ hA , inv-htpy (rectangle-back-right-bottom-cube))
+      ( refl-htpy' g')
+      ( inv-htpy back-left)
+```
```
fredrik-bakke, line 315: Good call, this is also an alternative I like better than comments, and the code won't be highlighted in your editor, so it should be easy enough to distinguish from type checked agda code.

fredrik-bakke:
Sorry for the delayed response, this looks.

src/foundation/cones-over-cospans.lagda.md
EgbertRijke: That will be for another PR:)

EgbertRijke:
> Sorry for the delayed response. This looks good.

No worries. Thank you for looking at it!

Issue #769, opened 2023-09-14
Category Theory


fredrik-bakke:
This issue outlines formalization goals for the `category-theory` module.

### Formalization targets
The following list outlines missing formalizations that should be part of any standard library on category theory:
- [ ] Extensionality of
  - [ ] Small precategories 
  - [ ] Large precategories
  - [ ] Small categories 
  - [ ] Large categories
- [x] (large) Functor categories
- [ ] Rezk completion
  - [ ] Large Rezk completion of small precategory
  - [ ] Small Rezk completion of small precategory
  - [ ] Large Rezk completion of large precategory
- [ ] The (large) (pre)groupoid core of a (large) (pre)category
  - [X] The pregroupoid core of a precategory
  - [x] The groupoid core of a category
  - [ ] The large pregroupoid core of a large precategory
  - [ ] The large groupoid core of a large category
  - [ ] the (large) (pre)groupoidal completion
  - [ ] their adjunctions
- [x] The Yoneda lemma for (large) (pre)categories
   - [ ] Naturality of the Yoneda lemma
   - [ ] The Yoneda embedding is an embedding
- [ ] Colimits and limits in (large) (pre)categories
      Equivalence of
   - [ ] Definition as adjoint to constant functor
   - [ ] Definition as terminal object in cone category
   - [ ] Object with universal property
- [x] Dependent products of (pre)categories
  - [x] Dependent products of large (pre)categories
- [ ] Fully-faithful+essentially surjective factorization
- [ ] (large) Sub(pre)categories
  - [x] Full (large) Sub(pre)categories
  - [ ] Wide (large) Sub(pre)categories  
- [ ] (large) Quotient (pre)categories
- [ ] (Co)cartesian families/fibrations
- [ ] The adjoint functor theorem
- [ ] Equivalence of transposing adjunctions and diagrammatic adjunctions
- [ ] Adjoint equivalences of (pre)categories (i.e. what is named "equivalences" in HoTT-UF)
- [x] Presheaf categories
  - [ ] Presheaf categories are free cocompletions
  - [ ] Representable presheaves

It would also be nice to have
- [ ] A basic treatment of wild categories
  - [ ] The homotopy precategory of a wild category
- [x] A basic treatment of strict categories
  - [ ] "Adjoint equivalences are equivalences" is equivalent to choice
- [ ] Left/right fibrations

EgbertRijke:
[...]

PR #783, opened 2023-09-17
Yoneda's lemma for categories

fredrik-bakke:
This is a shameless copy-and-paste of Emily's recent contribution (#632) on the Yoneda lemma for precategories (also generalizes one universe level).
Progresses #769.


fredrik-bakke:
This PR does not record how the theorem is natural in various variables. Also, I'm aware some of the names given are a little awkward, but that is only to keep them consistent across the distinction of pre-/categories. I would prefer if we could withhold discussion on those names until we have landed on a common naming scheme and are planning to refactor names for the whole of category-theory/agda-unimath.

EgbertRijke:
Note that there is also a tiny bit of category theory in #757, regarding isomorphisms and initial objects.

I'll review this later, but I would still like to be able to make requests for specific changes of names. You are taking away an essential way of working towards a _good_ naming scheme if you say that we can't make suggestions for names unless we propose a globally uniform scheme and everything has to be refactored immediately. That's one way of working, but another way that allows for more experimentation is to be able to request for name changes within the scope of the pull request.

EgbertRijke:
Either way, I like that you're adding new mathematics to the library, and I certainly don't want that you'd be refactoring most of the time. Adding new math is much more fun! So I won't request that you refactor all the category theory, and only make requests within the scope of this PR.

fredrik-bakke:
Apologies, you are right. I just don't want to be held up too much in this PR at the moment.

EgbertRijke:
Understandable. Let's agree that as a general rule of thumb for reviewing we try to keep our requests within the scope of the PR. Sweeping changes such as refactorings that could be related to a PR but not essential for its merging can come up, but we should be aware that their natural place is an issue. And we should also be aware that we have limited time, and it is always ok to reject requests for changes. It's nice if PR's are good, but they certainly don't have to be perfect.

EgbertRijke:
Given that you don't want to be held up with this PR, let me just merge it. The contents of this PR will improve with use.

fredrik-bakke:
I agree wholeheartedly. I think this library has come to a point where sweeping changes are somewhat less frequent, and also harder to implement generally. Therefore, it will very often be a project in and of itself to implement changes that come up in discussions.

EgbertRijke:
One sensible and purposeful way of refactoring is that if you are working on adding a new feature/definition/theorem, and you notice that something in the library needs to be improved for your purpose, then you can do it. You can spend a lifetime just refactoring for the sake of it, but it still won't make the library perfect, and it is better to do that refactoring in connection to something you already wanted to work on.

fredrik-bakke:
I believe this PR did that. In essence, that is all that it did.

EgbertRijke:
Also, a sensible and purposeful way of developing category theory is to just develop the parts that are needed in order to do something else. That way we don't get an isolated huge categories library, but a well integrated category library.

EgbertRijke:
Imagine if we had a large categories library doing small categories, and then to apply it it turned out to be much better to work with large categories. It would be such a shame:)

fredrik-bakke:
To be honest, I only I wonder what Amelia would have to say about that 🫠

EgbertRijke:
Large categories won again today, because I was able to state and prove the initiality of the ring of integers with respect to _all_ rings usinc the concept of initial object in a large category. If we'd only have small categories, then we'd have to lift the ring of integers or come up with some other way to construct _an_ initial ring of universe level `l`, but now we can just say that the integers are initial with respect to universe level `l`.

fredrik-bakke:
My main question is why stop at `ω`.

EgbertRijke:
Because the type constructors are only defined for universe levels of type Level

fredrik-bakke:
That was not a question directed at you. 😅

EgbertRijke:
Lol, yeah, it should indeed be a question for the implementers. The type constructors should just be defined irrespective of universe level.

PR #787, opened 2023-09-18
WIP Functor Categories

fredrik-bakke:
### Summary
- Defines the representing arrow category
- Composition of large natural transformations
- "Extensionality" of functors
- Refactor isomorphisms in small precategories and categories
- Develop files on natural isomorphisms
- Define the functor category

Lots of refactoring in this one.

#769


src/commutative-algebra/isomorphisms-commutative-rings.lagda.md
```diff
@@ -200,11 +200,11 @@ module _
       ( ring-Commutative-Ring A)
       ( ring-Commutative-Ring B)
 
-  is-iso-iso-Commutative-Ring :
+  is-iso-hom-iso-Commutative-Ring :
```
EgbertRijke, line 203: Here is a question whether we want our naming scheme to be [property]-[underlying-thing]-[thing] or [property]-[thing]. I have always deliberately chosen the latter, because it is shorter and it conveys all the information.

So I've been using `is-iso-iso`, `is-category-category`, `is-ideal-ideal`, and so on. And I prefer this option, because names are already long enough in agda-unimath.

EgbertRijke:
[Posting my previous comment here again, so that it is not going to be snowed under my my next few comments, because I think it is important]

There is a question whether we want our naming scheme to be [property]-[underlying-thing]-[thing] or [property]-[thing]. I have always deliberately chosen the latter, because it is shorter and it conveys all the information.

So I've been using `is-iso-iso`, `is-category-category`, `is-ideal-ideal`, and so on. And I prefer this option, because names are already long enough in agda-unimath.

src/category-theory/anafunctors-categories.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Anafunctors between categories
+# Anafunctors on categories
```
EgbertRijke, line 1: Why this change? Isn't it more natural to speak of functors between categories? The nlab does so too...

src/category-theory/anafunctors-precategories.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Anafunctors between precategories
+# Anafunctors on precategories
```
EgbertRijke, line 1: Why this change? Isn't it more natural to speak of functors between categories?

src/category-theory/functors-categories.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Functors between categories
+# Functors on categories
```
EgbertRijke, line 1: Why this change? Isn't it more natural to speak of functors between categories?

src/category-theory/functors-large-precategories.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Functors between large precategories
+# Functors on large precategories
```
EgbertRijke, line 1: Why this change? Isn't it more natural to speak of functors between categories?

src/category-theory/functors-precategories.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Functors between precategories
+# Functors on precategories
```
EgbertRijke, line 1: Why this change? Isn't it more natural to speak of functors between categories?

src/category-theory/category-of-functors.lagda.md
```diff
@@ -0,0 +1,68 @@
+# The category of functors and natural transformations between two fixed categories
```
EgbertRijke, line 1: In this title you use "between" correctly:)

src/category-theory/isomorphisms-in-large-precategories.lagda.md
```diff
@@ -92,9 +92,9 @@ module _
   hom-iso-Large-Precategory : type-hom-Large-Precategory C X Y
   hom-iso-Large-Precategory = pr1 f
 
-  is-iso-iso-Large-Precategory :
+  is-iso-hom-iso-Large-Precategory :
```
EgbertRijke, line 95: I prefer the shorter name.

src/category-theory/isomorphisms-in-large-precategories.lagda.md
```diff
@@ -428,12 +422,11 @@ module _
   where
 
   inv-iso-Large-Precategory :
-    iso-Large-Precategory C X Y →
-    iso-Large-Precategory C Y X
+    iso-Large-Precategory C X Y → iso-Large-Precategory C Y X
   pr1 (inv-iso-Large-Precategory f) = hom-inv-iso-Large-Precategory C f
   pr2 (inv-iso-Large-Precategory f) =
     is-iso-inv-is-iso-hom-Large-Precategory C
```
EgbertRijke, line 428: Here it also shows that your new longer names are creating an inconsistency in the naming scheme. According to your longer naming scheme we should be saying `is-iso-hom-inv-is-iso-hom-Large-Precategory`. I prefer the shorter version.

src/category-theory/precategories.lagda.md
```diff
@@ -93,13 +93,15 @@ module _
...
+    type-hom-Precategory y z →
+    type-hom-Precategory x y →
     type-hom-Precategory x z
   comp-hom-Precategory = pr1 associative-composition-structure-Precategory
 
   comp-hom-Precategory' :
     {x y z : obj-Precategory} →
-    type-hom-Precategory x y → type-hom-Precategory y z →
+    type-hom-Precategory x y →
+    type-hom-Precategory y z →
```
EgbertRijke, line 104: I've actually been dying to change `type-hom-Precategory` to `hom-Precategory` and `hom-Precategory` to `hom-set-Precategory` :)

src/category-theory/functors-precategories.lagda.md
```diff
@@ -43,42 +123,41 @@ module _
...
-              (f : type-hom-Precategory C x y) →
-              ( F₁ (comp-hom-Precategory C g f)) ＝
-              ( comp-hom-Precategory D (F₁ g) (F₁ f))) ×
-            ( (x : obj-Precategory C) →
-              F₁ (id-hom-Precategory C {x}) ＝ id-hom-Precategory D {F₀ x})))
-
-  obj-functor-Precategory :
+          ( λ F₁ → is-functor-map-Precategory C D (F₀ , F₁)))
+
+  map-obj-functor-Precategory :
```
EgbertRijke, line 131: The way I read `obj-functor-Precategory` as that it is the action on objects of a functor on precategories. So the short name looks totally fine with me.

I thought it was perhaps because you want `obj-functor-Precategory` to be the type of objects in the functor precategory, but this doesn't seem to be the case, because you call that precategory `functor-precategory-Precategory`.

src/category-theory/functors-precategories.lagda.md
```diff
@@ -43,42 +123,41 @@ module _
...
     functor-Precategory → obj-Precategory C → obj-Precategory D
-  obj-functor-Precategory = pr1
+  map-obj-functor-Precategory = pr1
 
-  hom-functor-Precategory :
+  map-hom-functor-Precategory :
```
EgbertRijke, line 135: Same here, the shorter name looks totally fine with me.

src/category-theory/functors-precategories.lagda.md
```diff
@@ -107,46 +186,42 @@ comp-functor-Precategory :
   (C : Precategory l1 l2) (D : Precategory l3 l4) (E : Precategory l5 l6) →
   functor-Precategory D E → functor-Precategory C D → functor-Precategory C E
 pr1 (comp-functor-Precategory C D E G F) =
-  obj-functor-Precategory D E G ∘ obj-functor-Precategory C D F
+  map-obj-functor-Precategory D E G ∘ map-obj-functor-Precategory C D F
```
EgbertRijke, line 189: Here it starts showing that you have to say `map-` everywhere, but it doesn't contain extra information.

src/category-theory/functors-precategories.lagda.md
```diff
@@ -32,6 +40,78 @@ A functor from a precategory `C` to a precategory `D` consists of:
 
 ## Definition
 
+### Maps between precategories
```
EgbertRijke, line 43: This could easily have its own file.

src/category-theory/functors-precategories.lagda.md
```diff
@@ -32,6 +40,78 @@ A functor from a precategory `C` to a precategory `D` consists of:
...
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  (C : Precategory l1 l2)
+  (D : Precategory l3 l4)
+  where
+
+  map-Precategory : UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  map-Precategory =
+    Σ ( obj-Precategory C → obj-Precategory D)
```
EgbertRijke, line 54: Could this be defined in terms of morphisms of relations?

EgbertRijke:
This is a great PR adding many useful new things to the categories formalization. But perhaps I would like to discuss naming, because it looks like we have somewhat different ideas about this.

src/category-theory/anafunctors-precategories.lagda.md
fredrik-bakke: I changed it so that "natural transformations between functors on categories" read more naturally, but I'm not convinced by this change myself either, so I'll revert it.

src/category-theory/precategories.lagda.md
fredrik-bakke: I totally agree with that change!

src/category-theory/functors-precategories.lagda.md
fredrik-bakke: yup, I think I changed that in anticipation for the objects of the functor category, but then realized there was a naming conflict for `functor-Precategory`

src/category-theory/functors-precategories.lagda.md
fredrik-bakke: Ah, thanks for that suggestion!

src/commutative-algebra/isomorphisms-commutative-rings.lagda.md
fredrik-bakke: We use `[property]-[underlying-thing]-[thing]` for very common things in the library like sets and propositions, and in the case of `Contr`, the name `is-contr-Contr` is already taken. It shouldn't be a problem to change that one, since I bet it's rarely used if ever, but it begs the question how do we name properties of the types of objects such as `iso-Precategory` when they can be interpreted as both properties of the type itself, but also as a property of its objects.

EgbertRijke: Have a look in #793 to see if there is more I could add there. I think it would be good to settle a clear naming convention

src/category-theory/functors-precategories.lagda.md
fredrik-bakke: Here it could be mistaken for the hom-functor, though

EgbertRijke: But that one is called `rep-functor-Precategory` (violating our convention to not abbreviate a name unless it is very common to do so)

src/commutative-algebra/isomorphisms-commutative-rings.lagda.md
fredrik-bakke: Another one is `(categorical law-)comp-hom-Category`. Either the `comp` or the `hom` is superfluous most of the time. E.g. `right-unit-law-comp-hom-Category` could be shortened to `right-unit-law-comp-Category`, or even `right-unit-law-Category`.

EgbertRijke: We have those laws for morphisms and for isomorphisms, hence the need to disambiguate hom, and we should disambiguate the operations because that way the laws can be stated uniformly for any setting.

fredrik-bakke: I'm arguing for consistency, not for longer names.

EgbertRijke: I think in the case of properties of underlying objects it is quite natural to omit reference to the operation of taking the underlying object in order to state that the object has a certain property. In other words, this permits us to use shorter names while still remaining consistent with our other naming scheme. 

It is also a question on how robust we want to be. Do we want a naming scheme that is so rigid that we can algorithmically generate names for entries, or do we permit some leniency that allows for natural shortenings in large classes of namings, which wouldn't be permitted by the hypothetical algorithm. I think a quite clear class of entries where a shorter name is permitted and can be consistent with the rest of the naming in agda-unimath is about properties of underlying objects. I would like our naming scheme to be a little bit flexible in that way.

fredrik-bakke: I agree with the sentiment of this. My main pushback is that having a clear algorithm for names is also useful, even at the expense of longer names. Mainly because that makes it possible to easily guess the names of things in the library so that people who work in it don't have to look them up or memorize them.

fredrik-bakke: I'm a bit confused though why you preferred having non-abbreviated names such as `natural-transformation` instead of `nat-trans` when that clearly wouldn't cause confusion, but now want to omit words that actually provide meaning to things for the sake of brevity.

fredrik-bakke: Maybe we can revert that change? I find it kind of odd that we spell out `natural-isomorphism`, but also use `iso`.

src/category-theory/functors-precategories.lagda.md
fredrik-bakke: On second inspection, I think they are different enough that it makes more sense to keep the definition separate. The base varies, and the objects are implicit arguments as opposed to the arguments of relation homomorphisms.

src/category-theory/precategories.lagda.md
fredrik-bakke: doing this renaming now

fredrik-bakke: Done :) Sorry for making the review process messier

fredrik-bakke:
oh no, I messed up my merge earlier

src/category-theory/precategories.lagda.md
EgbertRijke: Oh! You were actually doing that in this PR! Thank you so much!

PR #793, opened 2023-09-21
Expanding on the naming conventions

EgbertRijke:


NAMING.md
fredrik-bakke, line 1: I'm curious about your thinking behind moving this to a new file. It must be quite daunting for new contributors at this point to get a grasp on all the conventions and principles of the library, and having even more guides doesn't help that I think, even if the total number of lines of text doesn't increase. Soon, we'll need a guide to give viewers an overview of the guides. 😛
Jokes aside, at some point maybe we want a quick guide to not scare away new contributors with all of our documents. Perhaps that was/is part of the intention of the design principles file. But that document reads more like a statement of principles, although it is also somewhat summarizing.

EgbertRijke: I thought the coding style guide was already quite long, and that naming was also a separate topic, because most of the coding style guidelines are about formatting.

fredrik-bakke, line 1: I think a better name for this file is `NAMING-CONVENTIONS.md`

NAMING.md
```diff
@@ -0,0 +1,116 @@
+# Naming conventions
+
+A good naming convention is essential for being able to navigate and maintain
```
fredrik-bakke, line 3: Maybe add just a sentence about the scope of this file?

E.g.
```suggestion
# Naming conventions

This document provides general guidelines for naming definitions that apply anywhere in the library. A good naming convention is essential for being able to navigate and maintain
```

NAMING.md
```diff
@@ -0,0 +1,116 @@
...
+
+The library contains, for example, an entry `is-iso-hom-Ring` for the predicate
+that a ring homomorphism is an isomorphsim. The most significant aspect of this
+predicate is the assertion that something is an isomorphism. Furthermore, we
+make this assertion about ring homomorphisms. The name `is-iso-hom-Ring` is
+therefore a logical name for the predicate that a ring homomorphism is an
+isomorphism.
+
+In our naming scheme we strive for a direct correspondence between a
+constructions's name and its type. Take, for example, the proof that the
```
fredrik-bakke, line 16:
```suggestion
construction's name and its type. Take, for example, the proof that the
```

NAMING.md
```diff
@@ -0,0 +1,116 @@
+# Naming conventions
+
+A good naming convention is essential for being able to navigate and maintain
+the library, and for being able to make progress with your formalization
+project. Good names provide coincise descriptions of an entry's purpose, and
+help making the code in the library readable.
```
fredrik-bakke, line 6: I think it should be explained how good naming conventions also make it possible to refer to things without looking them up first.

fredrik-bakke (commented):
I would like to review this PR carefully before it is merged, but that will have to wait at the moment.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+### Naming conventions for mathematical laws
+
+Often, mathematical laws are recorded by first specifying in full generality
+what it means to satisfy that law. For example, the
+[unit laws](foundation.unital-binary-operations.md) for a binary operation are
+stated as `left-unit-law` and `right-unit-law`. The fact that
+[multiplication on the integers](elementary-number-theory.multiplication-integers.md)
+satisfies the unit laws is then stated as `left-unit-law-mul-ℤ` and
+`right-unit-law-mul-ℤ`. In the general naming scheme, this naming follows the
+pattern `[type]-[Namespace]`, because it states the type of which an element is
```
VojtechStep, line 132: I'd also say that this is `[type]-[name]`, or possibly `[type]-[name]-[Namespace]`, because otherwise aren't you saying that `left-unit-law-mul` is a type?

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+[interchange law](foundation.interchange-law.md). This interchange law requires
+two operations. When we want to prove an interchange law for two specific
+operations `mul1` and `mul2`, we name it `interchange-law-mul1-mul2`. We use
+this naming scheme, even if the two operations are the same. In fact two
+_associative_ operations that satisfy the interchange law will necessarily be
+the same. For example, in the [integers](elementary-number-theory.integers.md)
+there is an interchange law for addition over addition, which states that
+`(a + b) + (c + d) ＝ (a + c) + (b + d)`. This law is called
+`interchange-law-add-add-ℤ`. Again, this naming follows the pattern
+`[type]-[Namespace]`, because it states the type of which an element is
```
VojtechStep, line 145: Again I'd say this is `[type]-[name]-[Namespace]`

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+  mathematical practice.
+
+- Just as we do with abbreviations, we use special symbols sparingly in names.
+
+- If a symbol is not available, we describe the concept concisely in words.
+
+- We prioritize the readability of the code and avoid subtly different
+  variations of the same symbol. An important exception is the use of the
+  [full width equals sign](https://codepoints.net/U+ff1d) for the identity type,
+  as the standard equals sign is a reserved symbol in Agda.
```
VojtechStep, line 192: I'd also mention the homotopy composition and whiskering situation

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+
+If you are unsure about what to name your entry, ask us on the Univalent Agda
+discord server. One of the most commonly asked questions is what to name a
```
VojtechStep, line 195: Is this advice intentionally mentioned twice?

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+  and omit the symbolic notation for anything but the operation itself. For
+  example, the commutativity of cartesian products is called `commutative-prod`
+  and not `commutative-×`, and the unit of a modality is called `unit-modality`
+  and not `unit-○`.
+
+- To enhance conceptual clarity, we suggest names of constructions avoid
+  referring to variable names. This makes code more understandable, even at a
+  glance, and easier to work with in subsequent code.
+
+- Using the projection functions `pr1` and `pr2`, particularly their
```
VojtechStep, line 219: I'd keep this paragraph where it was, I don't really see how it pertains to naming things - it tells you to add some definitions, not necessarily how to name them

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+  provide more informative responses when jumping to the definition.
+  Furthermore, it makes it easier to change the definition later on.
+
+- Lastly, we recommend not naming constructions after infix notation of
+  operations included in them. Preferring primary prefix notation over infix
+  notation can help keep our code consistent. For example, it's preferred to use
+  `commutative-prod` instead of `commutative-×` for denoting the commutativity
+  of cartesian products.
+
+- Names never reference variables.
```
VojtechStep, line 234: This is a restatement of the two paragraphs above

VojtechStep (commented):
Thank you for putting it all together. I do feel like there are some inconsistencies in the description of some of the examples, could you elaborate a bit on the comments?

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+The general naming pattern breaks down as follows:
+
+- **[name]:** This part is used to give a custom descriptive name for the entry.
+- **[type]:** This part of the name refers to the output type of the entry.
+- **[hypotheses]:** This part of the name consists of descriptors for the
+  hypotheses used in the type specification of the entry.
+- **[Namespace]:** This part of the name describes what important concept or
+  general category the entry is about.
+
+Given Agda's current lack of a namespace mechanism, we've incorporated what
```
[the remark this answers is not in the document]
EgbertRijke: Agda has a module system. New entries cannot be added to a module once it is closed. A namespace mechanism would allow one to add things to a namespace across several files.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+There is also a common class of entries where we don't use the `[name]` part in
+the name of an entry: underlying objects. For example, the underlying set of a
+group is called `set-Group`, which uses the pattern `[type]-[Namespace]`. The
+construction of the underlying set of a group returns for each group a set,
+which is an element of type `Set`. Similarly, we have `type-Group]`,
+`semigroup-Group`, `type-Ring`, `set-Ring`, and so on. Another instance where
+this happens is in `hom-iso-Group`, which is the construction that returns the
+underlying group homomorphism of an isomorphism of group. The fact that a group
+isomorphism is an isomorphsim is called `is-iso-iso-Group`, which also uses the
+pattern `[type]-[Namespace]`. One could also consider calling it
```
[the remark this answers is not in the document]
EgbertRijke: Things that go under [type] are the **output types** of the specification of an entry. In the case of `is-iso-hom-Ring` the output type of its specification is `UU`. We don't put this in the name. Instead `is-iso` is a descriptor of the specification, falling under `[name]` in the general pattern.

NAMING-CONVENTIONS.md
EgbertRijke: The type of the output of this entry is `interchange-law add-ℤ add-ℤ`, so the name here merely describes the type, and the namespace for the integers. In the case `is-equiv succ-ℤ` this would also fall under `[type]-[Namespace]`.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+follow the patterns `[type]-[Namespace]` and `[type]-[hypotheses]-[Namespace]`.
+
+Another class of entries where the naming scheme needs some explanation, is
+where we define a structured object from another structured object. For
+instance, the [kernel](group-theory.kernels.md) of a
+[group homomorphism](group-theory.homomorphisms-groups.md) is defined to be the
+[normal subgroup](group-theory.normal-subgroups.md) `kernel-hom-Group`. This
+name follows the scheme `[name]-[hypotheses]-[Namespace]`. When we want to
+define the underlying structure of the kernel of a group homomorphism, we follow
+the scheme `[type]-[hypotheses]-[Namespace]`. For instance, the underlying group
```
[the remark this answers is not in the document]
EgbertRijke: Good point!

NAMING-CONVENTIONS.md
EgbertRijke: Probably it's an oversight. Thanks!

NAMING-CONVENTIONS.md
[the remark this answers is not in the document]
VojtechStep: Alright, that makes sense, although it wasn't immediately clear to me - could you maybe rephrase it as "Given Agda's current lack of a namespace mechanism, we can't have named logical collection of definitions that span multiple files. Instead, we've incorporated [..]", or something like that?

NAMING-CONVENTIONS.md
VojtechStep: Then I don't see where the `[Namespace]` part comes in? Since the `-Z` suffix is already in the `[type]` part.

EgbertRijke: But if it were counted in the type part, then we'd have to say `interchange-law-add-ℤ-add-ℤ`. Instead, here the proposal is to consider `ℤ` in the namespace part. The name of the type we're landing in of course also mentions the namespace, but in a namespace you don't repeat the name of the namespace for every entry in it that you use. Instead, _within the namespace_ the type would be `left-unit-law mul`, which would appear in the name of an entry as `left-unit-law-mul`, and we append the namespace.

EgbertRijke: BTW, all of these are very good questions. I will try to clear it up in the text too.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,234 @@
...
+construction that constructs the underlying homomorphism of an isomorphisms of
+rings. This name follows the pattern `[type]-[hypotheses]-[Namespace]`.
+
+There is also a common class of entries where we don't use the `[name]` part in
+the name of an entry: underlying objects. For example, the underlying set of a
+group is called `set-Group`, which uses the pattern `[type]-[Namespace]`. The
+construction of the underlying set of a group returns for each group a set,
+which is an element of type `Set`. Similarly, we have `type-Group]`,
+`semigroup-Group`, `type-Ring`, `set-Ring`, and so on. Another instance where
+this happens is in `hom-iso-Group`, which is the construction that returns the
```
VojtechStep, line 102: `hom-iso-Group` has the approximate type `iso -> hom`, shouldn't it then be `[type]-[hypotheses]-[Namespace]`?

EgbertRijke: That looks right

NAMING-CONVENTIONS.md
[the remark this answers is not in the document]
VojtechStep: I don't think I follow. What I'm reading is that
- `is-iso-iso-Group : (f : iso) -> is-iso f` follows `[type]-[Namespace]`, but
- `is-ideal-ideal-Ring : (I : ideal) -> is-ideal I` follows `[type]-[hypotheses]-[Namespace]`

Is that correct?

NAMING-CONVENTIONS.md
VojtechStep: Oh, that does make it way clearer, thanks!

NAMING-CONVENTIONS.md
VojtechStep: The explanation on `left-unit-law-mul-ℤ` makes this clear now 👌

NAMING-CONVENTIONS.md
```diff
@@ -61,64 +63,71 @@ naming scheme aims for clarity and predictability.
...
+  entry.
+- **[input-types]:** This part of the name consists of references to the input
+  types used in the type specification of the entry.
 - **[Namespace]:** This part of the name describes what important concept or
   general category the entry is about.
 
-Given Agda's current lack of a namespace mechanism, we've incorporated what
-would typically be a namespace directly into the name. This convention is
+Given Agda's current lack of a namespace mechanism, we can't have named logical
+collection of definitions that span multiple files. Instead, we've incorporated
```
VojtechStep, line 84:
```suggestion
collections of definitions that span multiple files. Instead, we've incorporated
```
Sorry I had a typo in my original suggestion.

NAMING-CONVENTIONS.md
```diff
@@ -61,64 +63,71 @@ naming scheme aims for clarity and predictability.
...
+what would typically be a namespace directly into the name. This convention is
 particularly applied to key mathematical concepts, such as groups, rings,
-categories, graphs, and trees.
+categories, graphs, and trees, and so on.
```
VojtechStep, line 87:
```suggestion
categories, graphs, trees, and so on.
```

NAMING-CONVENTIONS.md
EgbertRijke: Perfect, I was going to ask you to use github suggestions. Since you are helping out a lot with this PR, you should be a coauthor after it is merged, and using github's suggestion system is one way to do that.

VojtechStep: I was using my phone to write the other comments, and it's a lot less convenient to add GH suggestions from the app 😅

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,228 @@
...
+anymore, or sometimes we gain a better understanding of what an entry is about
+and update its name accordingly. We should therefore remember that neither the
+naming scheme nor the names we use in the library are set in stone. If you find
+a particular entry where the naming seems off, for any reason, we would love to
+know about it. Maintaining and improving it is part of the work of maintaining
+the agda-unimath library.
+
+## Examples
+
+For example, the library has an entry named `is-iso-hom-Ring` for the predicate
```
fredrik-bakke, line 34: The transition here is a little abrupt. How about

```suggestion
## Examples

Before we derive a general scheme, let us first get a feel for the structure of names in agda-unimath by considering some examples.
The library has an entry named `is-iso-hom-Ring` for the predicate
```

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,228 @@
...
+construction's name and its type. Take, for example, the proof that the
+successor function on integers is an equivalence. It has the type
+`is-equiv succ-ℤ`, so we name it `is-equiv-succ-ℤ`.
+
+We also consider the type of hypotheses when naming a construction. Typically,
+when including hypotheses in the name, we mention them after the type of the
+main construction. Let's take the entry `is-equiv-is-contr-map` as an example.
+In this entry, we show that any
+[contractible map](foundation.contractible-maps.md) is an
+[equivalence](foundation.equivalences.md). The type of this entry is therefore
```
fredrik-bakke, line 51: I tend to link to core, as that is usually where the relevant definition is usually located

```suggestion
[equivalence](foundation-core.equivalences.md). The type of this entry is therefore
```

EgbertRijke: The reason I usually don't link to core is because except in some special cases I don't remember (and don't want to rememeber) if it has a base file in core. But you're right, linking to core is linking to the actual definitions, and I agree with you.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,229 @@
...
+- Names are unique; we steer clear of namespace overloading.
+
+- Names should accurately convey the concept of its construction.
+
+- We use US English spelling of words in names.
+
+- When an entry is predominantly about objects of an important concept, the
+  names of these concepts can be capitalized. Usually, these are categories like
+  `Prop`, `Set`, `Semigroup`, `Monoid`, `Group`, `Preorder`, `Poset`,
+  `Precategory`, `Category`, `Directed-Graph`, `Undirected-Graph`, and so on.
```
fredrik-bakke, line 181: I think it is better if the meaning of category is loosened here.
```suggestion
  names of these concepts can be capitalized. Usually, these are categories in one way or another, like
  `Prop`, `Set`, `Semigroup`, `Monoid`, `Group`, `Preorder`, `Poset`,
  `Precategory`, `Category`, `Directed-Graph`, `Undirected-Graph`, `Truncated-Type`, `Spectrum` and so on.
```
and maybe some more exotic example to pique

NAMING-CONVENTIONS.md
fredrik-bakke: Other examples are the asterisk vs asterisk operator, and big vee vs small vee. Maybe rephrase to "try to avoid subtly different"

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+construction that returns the underlying group homomorphism of an isomorphism of
+group. The fact that a group isomorphism is an isomorphsim is called
+`is-iso-iso-Group`, which also uses the pattern `[output-type]-[Namespace]`. One
+could also consider calling it `is-iso-hom-iso-Group`, to emphasize that the
+underlying group homomorphism of the isomorphism is an isomorphism. However,
+this name does not fit our patterns in any way, and the addition of `hom` to the
+name adds no extra useful information. This situation is common in instances
+where we omit the `[descriptor]` part of a name. For instance
+`is-category-Category` and `is-ideal-ideal-Ring` follow the patterns
+`[output-type]-[Namespace]` and `[output-type]-[input-types]-[Namespace]`,
```
VojtechStep, line 119: On second reading I'm not sure why
- `is-category-Category : (C : Category) -> is-category C` and
- `is-ideal-ideal-Ring : (I : ideal-Ring) -> is-ideal I`
are different? Even the modules where they are defined seem to follow the exact same structure.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+  [descriptor]-[output-type]-[input-types]-[Namespace]
+```
+
+In this general pattern, all the components are optional. However, their order
+is fixed and should be maintained for consistency.
+
+The general naming pattern breaks down as follows:
+
+- **[descriptor]:** This part is used to give a custom descriptive name for the
+  entry.
```
fredrik-bakke, line 76:
```suggestion
- **[descriptor]:** This part is used to give a custom descriptive name for the
  entry, or refer to the custom descriptive name of another entry.
```
If I didn't misunderstand, this is how it is used in the below example `is-iso-prop-hom-Ring`.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+of rings. This name follows the pattern
+`[output-type]-[input-types]-[Namespace]`.
+
+There is also a common class of entries where we don't use the `[descriptor]`
+part in the name of an entry: underlying objects. For example, the underlying
+set of a group is called `set-Group`, which uses the pattern
+`[output-type]-[Namespace]`. The construction of the underlying set of a group
+returns for each group a set, which is an element of type `Set`. Similarly, we
+have `type-Group`, `semigroup-Group`, `type-Ring`, `set-Ring`, and so on.
+Another instance where this happens is in `hom-iso-Group`, which is the
```
fredrik-bakke, line 109: Isn't `hom-iso-Group` an instance of `[output-type]-[input-types]-[Namespace]`?

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+construction that returns the underlying group homomorphism of an isomorphism of
+group. The fact that a group isomorphism is an isomorphsim is called
+`is-iso-iso-Group`, which also uses the pattern `[output-type]-[Namespace]`. One
```
fredrik-bakke, line 112: Isn't this also an instance of `[output-type]-[input-types]-[Namespace]`?

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+could also consider calling it `is-iso-hom-iso-Group`, to emphasize that the
+underlying group homomorphism of the isomorphism is an isomorphism. However,
+this name does not fit our patterns in any way, and the addition of `hom` to the
```
fredrik-bakke, line 115: It's not clear to me how this does not fit our patterns better than the proposed one. Also, the name would be more consistent with `is-iso-hom-inv-iso-Group`

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,230 @@
...
+name adds no extra useful information. This situation is common in instances
```
fredrik-bakke, line 116: It seems to me the added `hom` is disambiguative and does add extra information. To take an example where I more explicitly disagree with this convention, consider your recent change renaming `is-unital-semigroup-Monoid` to `is-unital-Monoid`, but there, the added `semigroup` _did_ add extra useful information. It disambiguated from for instance `is-unital-magma-Monoid` and `is-unital-h-space-Monoid`. Of course, you could argue that semigroup is the correct assumption when the information is omitted, but a reader may not immediately know to conclude that. They shouldn't have to know which way a monoid is defined, as a unital magma that is associative, or a semigroup that is unital.

NAMING-CONVENTIONS.md
fredrik-bakke, line 1: I wonder if it would be a good idea to also provide a table of commonly used descriptors. Some that come to mind are
- `compute`
- `htpy`
- `extensionality`
- `is-property`

Perhaps that section is also a good occasion to comment on the use of `'` for variants of the same thing, like when we have a covariant and a contravariant definition, or one that takes implicit and one that takes explicit arguments. Sometimes, I also think it is used when the thing without the `'` fits more naturally as a construction in the given namespace, but the definition with the `'` has weaker assumptions.

NAMING-CONVENTIONS.md
fredrik-bakke: Also, let me paraphrase from a discussion yesterday:

We use `[property]-[underlying-thing]-[thing]` for very common things in the library like sets and propositions. E.g. `is-set-type-Set`. For the case of `Contr`, the name `is-contr-Contr` is already taken. This begs the question: how do we name properties of the types of objects when they can be interpreted as both properties of the type itself, but also as a property of its objects?

NAMING-CONVENTIONS.md
fredrik-bakke: Another one is `coherence`

EgbertRijke: Yep, we can definitely write something about them

NAMING-CONVENTIONS.md
EgbertRijke: I think your argument that the reader shouldn't have to know what is asserted to be unital is very nice. Our naming scheme should be somewhat intuitive, even for readers who are not well-versed with our naming scheme. I will add something like that to the text. 

On the other hand, the term `is-unital-Monoid` *is* intuitively clear? What does it assert to be unital: the monoid! It extracts in a direct way as possible the unitality of the monoid. That is what the name conveys, and I think the short name conveys this better than `is-unital-semigroup-Monoid`.

I could also see where adding the underlying object could be *more* confusing. For example, `is-category-Category` extracts in a direct way as possible that a category is a category. On the other hand, we could write `is-category-precategory-Category`. This to me reads more like the entry proving that the precategory of categories is a category. 

Your point about the important exceptions, where the disambiguation is necessary, also brings up an important thing that needs to be clarified. If a disambiguation is necessary, for whatever reason, then we disambiguate. But generally we don't adjust our entire naming scheme for that. We work with the nicest possible scheme, that gives concise descriptive names that are well aligned with standard mathematical practice, and if there are exceptions we make exceptions. *Even if the exceptions are in very important cases.*

It reminds me a bit of the early choice I had to make about whether a concepts-first approach to library-design was possible. _Of course_ there were going to be exceptions. Right at the heart of type theory: Identity types, equivalences, dependent pair types, propositions, sets, and so on all had to be split up because it was not strictly possible to maintain a concepts-first design free of cyclic dependencies. But I found it very important to _not let the library design be dictated by a few exceptions_, however prominent those exceptions were. If I would want a super consistent library design, it coudn't be concepts-first and it would probably look more like the standard library.

So, your point about the important exceptions of `Set` and `Contr` is well taken, but we have to carefully decide what to do with those exceptions. Both options are valid choices, but neither are going to be free of ambiguity. And since the longer names are not necessarily more clarifying, I would still have a preference for shorter names, which will work well most of the time, but sometimes need to be disambiguated.

EgbertRijke: (The disadvantage of the standard library design is of course that you have to know how something is formalized (i.e., as a instance of what general construction or concept) in order to find it. So it may be more consistent, but this consistency comes at the price that it straining the user.

EgbertRijke: And another point is that if we choose `is-unital-semigroup-Monoid`, then we force the user to know that the monoids are formalised as semigroups in order to use its unitality. If we would like to not require this of the user, then `is-unital-Monoid` would be a better name, because it gives the user access to the unitality of the monoid without them having to know what the name of its underlying structure is.

fredrik-bakke: Ah, haha, I see I failed to observe how the statements coincide. My apologies, then I am on board.

fredrik-bakke: Thank you for spelling this out for me, it is now much clearer where the principles are coming from.

NAMING-CONVENTIONS.md
EgbertRijke: I just added a list of common descriptors to the file. I omitted `htpy` because it is directly tied to homotopies, but I could be convinced to put it in there. With such lists, it probably doesn't hurt to put in too much information. I will resolve this conversation here, but have a look if more common descriptors could come up in the list. We can start a new conversation near the source code of the list.

fredrik-bakke: Excellent! I think this helps contextualize the discussion on the general scheme, and gets readers started on forming their own names.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,247 @@
...
+[conjugation](synthetic-homotopy-theory.conjugation-loops.md). The name for
+transport along identifications is `tr` and the name for loop spaces in the
+library is `Ω`. Therefore, we record the function that transport computes to as
+`tr-Ω` and we record the [homotopy](foundation.homotopies.md) that transport is
+pointwise equal to `tr-Ω` as `compute-tr-Ω`.
+
+## List of common descriptors
+
+In the table below we list some common descriptors that are not directly tied to
+a type with that name.
```
fredrik-bakke, line 173:
```suggestion
To give a sense of the kind of general descriptors we use, we list in the table below some common descriptors that are not directly tied to
a type with that name.
```

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,247 @@
...
+
+| Descriptor       | Purpose                                                                                                        |
+| ---------------- | -------------------------------------------------------------------------------------------------------------- |
+| `coh`            | Used for proofs of coherence                                                                                   |
+| `coherence`      | Used to assert coherence                                                                                       |
+| `compute`        | Used for proofs of computation, which may be an identification or an element of any other identity system      |
+| `Eq`             | Used for identity systems on types of a particular form                                                        |
```
EgbertRijke, line 180: I wonder if we shouldn't use `identity-system` where we currently use `Eq` (but I will not make such a change in the current PR)

fredrik-bakke: Yeah, I thought `Eq` was deprecated. I've been using `htpy` for this so far 😬

NAMING-CONVENTIONS.md
EgbertRijke: Can you edit your suggestion with a period directly after "some common descriptors"? I think it is perhaps better to not restrict ourselves by adding that last part of the sentence.

NAMING-CONVENTIONS.md
EgbertRijke: It's not deprecated. For many types it makes sense to use `htpy` to say that the inhabitants are a structured tuple of homotopies. Similarly, for many types it makes sense to use `equiv` for the descriptor of its identity system. But for some types neither is possible, and then we go with `Eq`. Many number systems are of this kind, and they are certainly not deprecated:)

EgbertRijke: Actually, `identity-system-X` should return an element of the type of identity systems of `X`, while `Eq` returns the underlying family of types. There is a difference.

fredrik-bakke: I see I've just been making assumptions about the general case from my special cases. 😅 In either case, the capitalized `Eq` is a bit of an eye-sore.

EgbertRijke: I agree

scripts/contributors_data.toml
```diff
@@ -233,3 +233,8 @@ github = "andrejbauer"
 displayName = "Matej Petković"
 usernames = [ "Matej Petković" ]
 github = "Petkomat"
+
+[[contributors]]
```
VojtechStep, line 237: I'm adding Gregor as part of #795, since it's more in scope there and it required adding one more `username`.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,268 @@
...
+| `eq`             | Used as a descriptor for the identity type                                                                                                                                                        |
+| `equiv`          | Used for equivalences, and also for names of identity systems of universe-like types                                                                                                              |
+| `extensionality` | Used for computations of identity types                                                                                                                                                           |
+| `htpy`           | Used for constructions of homotopies, and also for names of identity systems of function-like types                                                                                               |
+| `is-property`    | Used when `is-prop` is unavailable                                                                                                                                                                |
```
fredrik-bakke, line 203: I don't remember if I already mentioned this, but did you know we actually define `is-property` in `foundation-core.subtypes`? Although IIRC it is not referred to anywhere.

NAMING-CONVENTIONS.md
```diff
@@ -0,0 +1,268 @@
...
+library is `Ω`. Therefore, we record the function that transport computes to as
+`tr-Ω` and we record the [homotopy](foundation.homotopies.md) that transport is
+pointwise equal to `tr-Ω` as `compute-tr-Ω`.
+
+## List of common descriptors
+
+To give a sense of the kind of general descriptors we use, we list some common
+descriptors in the table below.
+
+| Descriptor       | Purpose                                                                                                                                                                                           |
```
fredrik-bakke, line 193: Add `statement`

fredrik-bakke, line 193: I think it would be very good to have an example column for this table as well. That's what I did for the sHoTT project

fredrik-bakke, line 193: Another descriptor pair we use a lot is `is`/`has` :)

EgbertRijke: Why do you think of them as descriptors? That seems grammatically very weird.

fredrik-bakke: They are modifiers signifying that the type is a property/structure, and then we omit them for the associated total types. It just seems consistent with the scheme.

EgbertRijke: I'm willing to come up with our own naming convention, but I'm not willing to come up with a new theory that the verbs to be and to have are modifiers. That idea seems out of this world to me.

PR #794, opened 2023-09-21
Maps of categories, functor categories, and small subprecategories

fredrik-bakke:
### Summary

- Define the (pre)category of maps of (pre)categories
  - Define maps of (pre)categories 
  - Define natural transformations/isomorphisms of maps between (pre)categories
  - Characterize equality of maps between precategories
- Refactor isomorphisms in small precategories and categories
- Develop files on natural isomorphisms
- Define the functor category
  - Characterize equality of functors between precategories
- Define (small) subprecategories and their precategory structures
- Define faithful functors (using embeddings, and show that is equivalent to being injective)
- Define the representing arrow category
- Composition of large natural transformations

Lots of refactoring in this one.

#769


fredrik-bakke:
I messed up #787, so this one has to take over.

src/category-theory/maps-precategories.lagda.md
```diff
@@ -0,0 +1,208 @@
...
+  is-contr-total-htpy-map-Precategory :
+    (f : map-Precategory C D) →
+    is-contr (Σ (map-Precategory C D) (htpy-map-Precategory f))
+  is-contr-total-htpy-map-Precategory f =
+    is-contr-total-Eq-structure _
+      ( is-contr-total-htpy (obj-map-Precategory C D f))
+      ( obj-map-Precategory C D f , refl-htpy)
+      ( is-contr-total-Eq-Π _
+        ( λ x →
+          is-contr-total-Eq-Π _
```
fredrik-bakke, line 175: It seems I am at a bit of an impasse here currently. I want to apply this theorem, so endpoints have to be given explicitly to `hom-map-Precategory`. This also means that functors have to take endpoints explicitly for their action on homs. Alternatively, it seems I have to reprove `is-contr-total-Eq-Π` for functions taking an implicit argument. I feel like I'm missing something, but I gotta ask, what should I do here?

EgbertRijke: I think that's not a bad idea. It can be done using the equivalence from dependent function types with explicit argument to the dependent function type with implicit argument. We have it in the library, but it is not nicely factored out. It is constructed in the proof of `is-prop-Π'` in `foundation.propositions`. It should not be too hard to prove `is-contr-total-Eq-Π'` using `is-contr-total-Eq-Π` and `is-contr-equiv`, and it would be nice if you can factor out that equivalence from `foundation.propositions` into the file about dependent function types (it's just a copy-paste).

fredrik-bakke: Alright, then I'll do that, thanks! It feels like there should be an agda command to make it unify implicit and explicit arguments though

fredrik-bakke: ah, yes, of course, I missed the part where it's a contractibility proof 😅

EgbertRijke: The implicit arguments should really just be syntactic sugar, and not be a separate type. Another thing I would fix if I would delve down into the agda source code.

fredrik-bakke: If only we were born as proof-assistant developers 🙏

EgbertRijke: It's quite surprising to me that I've never faced this situation before, that I needed to apply `is-contr-total-Π` but with implicit arguments.

fredrik-bakke: Same here

fredrik-bakke: Wasn't sure where precisely to place it, so let me know if you want me to move it

fredrik-bakke:
So, I've realized that maps of categories should probably form a category by the last extensionality principle I have to formalize. Inside this category, we recover functor categories as full subcategories. Thus natural transformations are useful more generally than for just functors. Moreover, I don't know if perhaps this mapping category is useful when we go to displayed categories. I haven't really looked at any references for that yet, except browsing 1lab a little. In either case, what do you think about defining natural transformations more generally for maps of precategories?

fredrik-bakke:
Nahh, probably won't need this

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -85,34 +98,39 @@ module _
...
 
   map-inv-equiv-is-adjoint-pair-Large-Precategory :
     (H : is-adjoint-pair-Large-Precategory) {l1 l2 : Level}
-    (X : obj-Large-Precategory C l1) (Y : obj-Large-Precategory D l2) →
-    ( type-hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y) →
-    ( type-hom-Large-Precategory C X (obj-functor-Large-Precategory G Y))
+    (X : obj-Large-Precategory C l1)
+    (Y : obj-Large-Precategory D l2) →
+    ( hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y) →
+    ( hom-Large-Precategory C X (obj-functor-Large-Precategory G Y))
```
EgbertRijke, line 121:
```suggestion
    hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y →
    hom-Large-Precategory C X (obj-functor-Large-Precategory G Y)
```

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -121,7 +139,9 @@ module _
           ( hom-functor-Large-Precategory F f))
       ( λ h →
         comp-hom-Large-Precategory C
-          ( comp-hom-Large-Precategory C (hom-functor-Large-Precategory G g) h)
```
EgbertRijke, line 124: It's fine to put line breaks here, but wasn't it also fine to have it all on one line?

fredrik-bakke: probably wasn't while they had the `map-hom-...` prefix

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -140,173 +162,237 @@ module _
...
-      level-left-adjoint-Adjunction :
+      level-left-adjoint-Adjunction-Large-Precategory :
         Level → Level
-      left-adjoint-Adjunction :
-        functor-Large-Precategory C D level-left-adjoint-Adjunction
-      level-right-adjoint-Adjunction :
+      left-adjoint-Adjunction-Large-Precategory :
+        functor-Large-Precategory C D
+          level-left-adjoint-Adjunction-Large-Precategory
+      level-right-adjoint-Adjunction-Large-Precategory :
```
EgbertRijke, line 204: I would prefer it if the level is a parameter, just like for any other concept surrounding large categories. Why would we treat adjunctions differently?

fredrik-bakke: I haven't changed any definitions in this file, only done a little bit of formatting. So that has been that way

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -322,85 +408,93 @@ module _
...
       ( id-hom-Large-Precategory D)
   coherence-square-natural-transformation-Large-Precategory
-    ( unit-Adjunction FG) {X = X} {Y} f =
-    ( inv
-      ( left-unit-law-comp-hom-Large-Precategory C
-        ( comp-hom-Large-Precategory C (η Y) f))) ∙
-      ( ( ap
+    ( unit-Adjunction-Large-Precategory FG) {X = X} {Y} f =
+    inv
+      ( ( inv
```
fredrik-bakke, line 427: oh, by the way, that reminds me that if you agree with how I defined naturality squares for large categories now, then I will go through and distribute `inv` over this and the next proof

EgbertRijke: I don't know what you mean, but if you can simplify the proof then of course you are welcome to do it.

src/category-theory/adjunctions-large-precategories.lagda.md
EgbertRijke: I see. Then you can leave it, but keep in mind that I might change this definition in the near future:)

src/category-theory/adjunctions-large-precategories.lagda.md
fredrik-bakke: I changed the direction of the coherence in naturality squares, so I had to update this proof by prepending an `inv` to the whole thing.

src/category-theory/subprecategories.lagda.md
```diff
@@ -0,0 +1,385 @@
...
+    is-in-subtype (P₁ y z) g →
+    is-in-subtype (P₁ x y) f →
+    is-in-subtype (P₁ x z) (comp-hom-Precategory C g f)
+
+  is-prop-contains-comp-subtype-Precategory :
+    is-prop contains-comp-subtype-Precategory
+  is-prop-contains-comp-subtype-Precategory =
+    is-prop-Π¹⁰
+      ( λ x y z g f _ _ _ _ _ →
+        is-prop-is-in-subtype (P₁ x z) (comp-hom-Precategory C g f))
```
fredrik-bakke, line 82: What do you think about this trick by the way? :)

src/category-theory/commuting-squares-of-morphisms-in-precategories.lagda.md
```diff
@@ -0,0 +1,46 @@
...
+```text
+  x ------> y
+  |         |
+  |         |
+  |         |
+  V         V
+  z ------> w
+```
+
+is said to **commute** in if there is an
```
EgbertRijke, line 31:
```suggestion
 in a [precategory](category-theory.precategories.md) `C` is said to **commute** in if there is an
```

It should be said that this square of morphisms is in a precategory.

src/category-theory/commuting-squares-of-morphisms-in-precategories.lagda.md
```diff
@@ -0,0 +1,46 @@
+# Commuting squares of morphisms in precategories
```
EgbertRijke, line 1: I like this addition! We should probably also have such files for categories, large precategories, and large categories.

fredrik-bakke: Defined as needed, of course 😬

src/category-theory/subprecategories.lagda.md
EgbertRijke: You mean all the underscores? They're fine.

Or the exponential naming? I don't think it fits very well in our naming scheme to be honest. Eg, to anyone new to this notation it is not intuitively clear what this could mean, and our proposed naming conventions state that we aspire to make our namings intuitively clear to the reader:P

src/category-theory/subprecategories.lagda.md
```diff
@@ -0,0 +1,385 @@
...
+    is-prop-Π²
+      ( λ x _ → is-prop-is-in-subtype (P₁ x x) (id-hom-Precategory C))
+
+  contains-id-prop-subtype-Precategory : Prop (l1 ⊔ l3 ⊔ l4)
+  pr1 contains-id-prop-subtype-Precategory =
+    contains-id-subtype-Precategory
+  pr2 contains-id-prop-subtype-Precategory =
+    is-prop-contains-id-subtype-Precategory
+
+  contains-comp-subtype-Precategory : UU (l1 ⊔ l2 ⊔ l3 ⊔ l4)
```
EgbertRijke, line 65: This should be called `is-closed-under-composition-subtype-Precategory`, to follow the same naming scheme for `large-subprecategories` and the already established `is-closed-under-multiplication` for subgroups.

src/category-theory/subprecategories.lagda.md
fredrik-bakke: I meant as opposed to the only previous alternative:

```suggestion
  is-prop-contains-comp-subtype-Precategory :
    is-prop contains-comp-subtype-Precategory
  is-prop-contains-comp-subtype-Precategory =
    is-prop-Π
      ( λ x →
        is-prop-Π
          ( λ y →
            is-prop-Π
              ( λ z →
                is-prop-Π
                  ( λ g →
                    is-prop-Π
                      ( λ f →
                        is-prop-Π
                          ( λ _ →
                            is-prop-Π
                              ( λ _ →
                                is-prop-Π
                                  ( λ _ →
                                    is-prop-Π
                                      ( λ _ →
                                        is-prop-Π
                                          ( λ _ →
                                            is-prop-is-in-subtype
                                              ( P₁ x z)
                                              ( comp-hom-Precategory
                                                ( C)
                                                ( g)
                                                ( f))))))))))))
```

We can discuss how to enumerate them, but they must be enumerated somehow.

src/foundation-core/propositions.lagda.md
```diff
@@ -271,6 +271,232 @@ pr1 (Π-Prop' A P) = type-Π-Prop' A P
...
+    is-prop (A10 x1 x2 x3 x4 x5 x6 x7 x8 x9)) →
+  is-prop
+    ( (x1 : A1) (x2 : A2 x1) (x3 : A3 x1 x2) (x4 : A4 x1 x2 x3)
+      (x5 : A5 x1 x2 x3 x4) (x6 : A6 x1 x2 x3 x4 x5)
+      (x7 : A7 x1 x2 x3 x4 x5 x6) (x8 : A8 x1 x2 x3 x4 x5 x6 x7)
+      (x9 : A9 x1 x2 x3 x4 x5 x6 x7 x8) →
+      A10 x1 x2 x3 x4 x5 x6 x7 x8 x9)
+is-prop-Π⁹ H = is-prop-Π (is-prop-Π⁸ ∘ H)
+
+is-prop-Π¹⁰ :
```
EgbertRijke, line 452: Oh, lol, now I see what it means:)

(Remember that I can't click through definitions in a github PR)

Ok, it is pretty obvious.

fredrik-bakke: I put all of that code in `<details>` tags, so it's hidden by default on the web page 👌

EgbertRijke: There should be a better way to define these though. I'm pretty sure that we can make a uniform definition `is-prop-iterated-Π` that ranges natural numbers, and then plug in the number 10. That would be much nicer, because this is kind of ugly.

fredrik-bakke: Oh, yeah, that would be nice. Without using Sigma types though?

EgbertRijke: > I put all of that code in `<details>` tags, so it's hidden by default on the web page 👌

No, I think we shouldn't hide ugly code. That's very bad practice. We should show our code, and if it is ugly then we should improve on it.

fredrik-bakke: fair enough

EgbertRijke: > Oh, yeah, that would be nice. Without using Sigma types though?

Yeah, you can define towers of type families of arbitrary height by coinduction.

fredrik-bakke: Please do elaborate!

EgbertRijke: My suggestion for now is: remove the <details> wrapper and keep it as it is. To define towers of type familes and starting to use it to improve some code is beyond the scope of this PR (please pretty please).

fredrik-bakke: alright alright

EgbertRijke: > Please do elaborate!

Can we do that outside of this PR (and perhaps on another moment)?

fredrik-bakke: Sure, no rush!

fredrik-bakke: For what it's worth, despite what GitHub is saying about how many files I've edited, this PR is more focused than my average PR.

EgbertRijke:
If you get over 10k lines of code, will you promise that you will be wrapping up the PR and not add new material anymore? 😄

fredrik-bakke:
Lol! I think at 10k lines, I will finally be able to prove the thing I wanted to begin with 😬

src/foundation-core/propositions.lagda.md
EgbertRijke: :) I believe you, but still do try to wrap this up

fredrik-bakke: Yessir! I will also review this myself before I officially hand it over to you ^^

src/category-theory/category-of-functors.lagda.md
```diff
@@ -0,0 +1,63 @@
...
+**category of functors**.
+
+## Definition
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} (C : Category l1 l2) (D : Category l3 l4)
+  where
+
+  functor-category-Precategory : Precategory (l1 ⊔ l2 ⊔ l3 ⊔ l4) (l1 ⊔ l2 ⊔ l4)
```
EgbertRijke, line 35: Given that this definition works on categories, I would suggest that it would be in the `Category` name space. For example, we wouldn't call the multiplicative monoid of a ring `multiplicative-ring-Monoid`. 

So here I would suggest the name `functor-precategory-Category`. This name indicates that it works on categories, it produces a precategory, and this precategory is the functor precategory.

src/foundation-core/propositions.lagda.md
EgbertRijke: Since this PR is so big that I won't be able to review it in one sitting (or two or three) anyway, I am already starting to review it.

fredrik-bakke: No worries! There is a lot of boilerplate code in it, so hopefully it's not _toooo_ bad. But since I haven't reviewed it myself yet, I won't make any guarantees on the quality :)

EgbertRijke: I always like to say that our library contains _no_ boilerplate code:)

fredrik-bakke: Well then I am curious to know what you think is boilerplate

EgbertRijke: Actually, that's not true. But the boilerplate code is not in our apis or infrastructure. The boilerplate code is rather in the longer proofs that _should_ be using general theory but aren't, and therefore contain a lot of unnecessary conversions and in-proof deriviations. Those are the boilerplate code, in my opinion. I can start a rant on this, lol

src/category-theory/equivalences-of-precategories.lagda.md
```diff
@@ -20,14 +20,19 @@ open import foundation.universe-levels
...
-`G : D → C` such that:
+A [functor](category-theory.functors-precategories.md) `F : C → D` is an
+**equivalence** of [precategories](category-theory.precategories.md) if there is
 
-- `G ∘ F` is naturally isomorphic to the identity functor on `C`,
-- `F ∘ G` is naturally isomorphic to the identity functor on `D`.
+1. a functor `G : D → C` such that `G ∘ F` is
+   [naturally isomorphic](category-theory.natural-isomorphisms-precategories.md)
+   to the identity functor on `C`,
+2. a functor `H : D → C` such that `F ∘ H` is naturally isomorphic to the
```
EgbertRijke, line 29: _Thank you_ for making this change! 🙏

fredrik-bakke: Btw, I assume what is named "equivalences" in the HoTT-UF book we will call adjoint equivalences?

EgbertRijke: I suppose so. We also have to have a name for the essentially surjective full and faithful functors, but maybe we can just call them `essentially-surjective-full-and-faithful-functor`. 

(for categories, they will just be surjective, i.e., the essentially prefix doesn't add anything, but for precategories it does).

EgbertRijke: btw don't add all the variants of equivalences of categories in this PR. That's also a separate PR:P

fredrik-bakke: I won't 😆

src/category-theory/adjunctions-large-precategories.lagda.md
EgbertRijke: Ah yes, is it now consistent with the direction of all our other commuting squares?

fredrik-bakke: God damn it.

fredrik-bakke: No 🤦‍♂️

fredrik-bakke: Then it is natural transformations for small categories that is wrong

fredrik-bakke: I'll swap it

fredrik-bakke: or, I guess this depends on what you define to be left right top and bottom

EgbertRijke: first argument: top
second argument: left
third argument: right
fourth argument: bottom

Also: use the definition of commuting squares of morphisms in precategories/categories/large-precategories/large-categories. Then you can't go wrong.

src/category-theory/subprecategories.lagda.md
EgbertRijke: Just to resolve this conversation: We agreed to keep it as it is for the current pull request, but eventually `is-prop-Π¹⁰` should become `is-prop-iterated-Π 10`

src/category-theory/adjunctions-large-precategories.lagda.md
fredrik-bakke: No, I mean, do you want

```agda
coherence-square-hom-Precategory D
  ( hom-map-Precategory C D F f)
  ( γ x)
  ( γ y)
  ( hom-map-Precategory C D G f)
```
or
```agda
coherence-square-hom-Precategory D
  ( γ x)
  ( hom-map-Precategory C D F f)
  ( hom-map-Precategory C D G f)
  ( γ y)
```

EgbertRijke: Ah, yes, didn't we say that it should be the same direction as the coherence of a morphism from `hom-map-Precategory C D F f` to `hom-map-Precategory C D G f`? This would suggest that it should be implemented as an instance of morphisms in the arrow category. But we may not have this yet... and in that case I suggest that you make the most sensible choice.

fredrik-bakke: Okay, I declare it should be the way it is currently implemented. 👨‍⚖️ court dismissed

src/category-theory/category-of-functors.lagda.md
fredrik-bakke: I agree this name was wrong. But the issue is I want to define the category of functors from a precategory to a category first. So actually, I think this should be called `precategory-functor-category-Category` in anticipation of `functor-category-Category` (and `functor-precategory-Category` for functors from a precategory to a category)

EgbertRijke: I think we shouldn't have a separate definition for functors from a precategory to a category. We also don't have a separate definition for a homomorphism from a ring to a commutative ring.

In my opinion it is best to have a definition for functors from a precategory to a precategory, and a definition for functors from a category to a category, and if you need to define a functor from a precategory to a category, then convert the category to a precategory.

So, with this suggestion we would have `functor-precategory-Precategory` for the precategory of functors between precategories. My reading of this name is that it operates on precategories, hence the namespace `Precategory`. It outputs a precategory, hence `precategory-Precategory`, and the name of this precategory is `functor`, so the full name of this entry is `functor-precategory-Precategory`.

I would prefer not to name the object before the name, as in your suggestion `precategory-functor`. I would have the name first, and then the kind of object it is, eg as in `id-equiv`. We don't call it `equiv-id` either.

Similarly, we could have `functor-precategory-Category` for the precategory of functors between categories, and `functor-category-Category` for the category of functors between categories.

fredrik-bakke: I want to record somewhere the result that when the codomain is a category the functor precategory is a category.

fredrik-bakke: Should I just define `is-category-functor-precategory-Precategory` and avoid giving the Sigma term then?

EgbertRijke: `is-category-functor-category-Category` if the input is a category, and `is-category-functor-precategory-is-category-Precategory` if the input is a precategory and the assumption that this precategory is a category.

EgbertRijke: The naming scheme takes into account when you are working with the hypothesis that something is a category, as opposed when something is given to you as a category.

A similar situation is with equivalences or isomorphisms: we use `is-equiv-map-inv-is-equiv` when the input is the assumtion that a map is an equivalence to conclude that its inverse is an equivalence; on the other hand, we use `is-equiv-map-inv-equiv` if the input is an equivalence.

EgbertRijke: I updated my previous comment with some more clarifications

EgbertRijke: Maybe even sharper would be `functor-category-Precategory` instead of my longer suggestion `functor-category-is-category-Precategory`, and not naming the assumption that the codomain precategory is a category. This shorter name is available, and it makes sense from its specification.

fredrik-bakke: Thanks for the clarifications! I guess I wanted the slickness of statements like the one for `Π-Prop`, which mirrors how one would in informal math write "assume `C` is a precategory and `D` is a category", compared to the slightly more cumbersome "assume `C` and `D` are precategories, where `D` is a category."

EgbertRijke: I see. Let’s come up with a solution then:)

Perhaps `functor-category-Precategory-Category`?

It’s not perfect…

fredrik-bakke: Let's sleep on it. At the moment, it's getting more and more tempting to start writing `Functor-Precategory` for functors between precategories, and `functor-Precategory` for the functor precategory, and it feels like sin.

fredrik-bakke: or the other way around

fredrik-bakke: Either way, I think it is important for our naming scheme to lead us to making good definitions.

EgbertRijke: On the other hand, if we would go with the short name functor-category-Precategory, then we could also be content with using

`functor-category-Precategory C (precategory-Category D) (is-category-Category D)`

which is also slightly inelegant, but at least it is not super long.

fredrik-bakke: yeah, that can work

EgbertRijke: > Either way, I think it is important for our naming scheme to lead us to making good definitions.

I fully agree. We shouldn’t avoid a definition because our naming scheme wouldn’t have a good name for it. Our naming scheme should be helpful, it should not be in the way.

fredrik-bakke: For now, I am content with defining `is-category-functor-precategory-Precategory` I think

src/category-theory/adjunctions-large-precategories.lagda.md
fredrik-bakke: I think I will leave inverting this proof to remove one tiny `inv` as future work, as it seems too much of an effort at the moment. Especially if it turns out we want to change the direction back.

fredrik-bakke:
Alright, this PR is officially ready for a review now :)

src/category-theory/subprecategories.lagda.md
```diff
@@ -0,0 +1,420 @@
...
+open import foundation.sets
+open import foundation.subtypes
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+A **subprecategory** of a [precategory](category-theory.large-precategories.md)
```
fredrik-bakke, line 28:
```suggestion
A **subprecategory** of a [precategory](category-theory.precategories.md)
```

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -22,10 +22,13 @@ open import foundation.universe-levels
...
+[large precategories](category-theory.large-precategories.md). Two
+[functors](category-theory.functors-large-precategories.md) `F : C → D` and
+`G : D → C` constitute an **adjoint pair** if
 
-- for each pair of objects `X` in `C` and `Y` in `D`, there is an equivalence
+- for each pair of objects `X` in `C` and `Y` in `D`, there is an
+  [equivalence](foundation-core.equivalences.md)
   `ϕ X Y : hom X (G Y) ≃ hom (F X) Y` such that
 - for every pair of morhpisms `f : X₂ → X₁` and `g : Y₁ → Y₂` the following
   square commutes:
```
EgbertRijke, line 34: Should this not reference `foundation.commuting-squares-of-maps`?

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -85,34 +98,39 @@ module _
...
-    (H : is-adjoint-pair-Large-Precategory) {l1 l2 : Level}
-    (X : obj-Large-Precategory C l1) (Y : obj-Large-Precategory D l2) →
-    ( type-hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y) ≃
-    ( type-hom-Large-Precategory C X (obj-functor-Large-Precategory G Y))
+    (H : is-adjoint-pair-Large-Precategory)
+    {l1 l2 : Level}
+    (X : obj-Large-Precategory C l1)
+    (Y : obj-Large-Precategory D l2) →
+    ( hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y) ≃
+    ( hom-Large-Precategory C X (obj-functor-Large-Precategory G Y))
```
EgbertRijke, line 112:
```suggestion
    hom-Large-Precategory D (obj-functor-Large-Precategory F X) Y ≃
    hom-Large-Precategory C X (obj-functor-Large-Precategory G Y)
```

src/category-theory/adjunctions-large-precategories.lagda.md
fredrik-bakke: Hmm, isn't it a commuting square of morphisms? If so it should link to a file `commuting-squares-of-morphisms-large-precategories`, but that doesn't currently exist.

fredrik-bakke: oh, wait, it's a square of hom-sets

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -140,173 +160,237 @@ module _
...
+    hom-functor-Large-Precategory
+      ( left-adjoint-Adjunction-Large-Precategory FG)
+
+  preserves-id-left-adjoint-Adjunction-Large-Precategory :
+    (FG : Adjunction-Large-Precategory)
+    {l1 : Level}
+    (X : obj-Large-Precategory C l1) →
+    ( hom-left-adjoint-Adjunction-Large-Precategory FG
+      ( id-hom-Large-Precategory C {X = X})) ＝
     ( id-hom-Large-Precategory D)
```
EgbertRijke, line 242:
```suggestion
    hom-left-adjoint-Adjunction-Large-Precategory FG
      ( id-hom-Large-Precategory C {X = X}) ＝
    id-hom-Large-Precategory D
```

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -140,173 +160,237 @@ module _
...
+    hom-functor-Large-Precategory
+      ( right-adjoint-Adjunction-Large-Precategory FG)
+
+  preserves-id-right-adjoint-Adjunction-Large-Precategory :
+    (FG : Adjunction-Large-Precategory)
+    {l : Level}
+    (Y : obj-Large-Precategory D l) →
+    ( hom-right-adjoint-Adjunction-Large-Precategory FG
+      ( id-hom-Large-Precategory D {X = Y})) ＝
     ( id-hom-Large-Precategory C)
```
EgbertRijke, line 276:
```suggestion
    hom-right-adjoint-Adjunction-Large-Precategory FG
      ( id-hom-Large-Precategory D {X = Y}) ＝
    id-hom-Large-Precategory C
```

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -140,173 +160,237 @@ module _
...
-    inv-equiv (equiv-is-adjoint-pair-Adjunction FG X Y)
-
-  map-inv-equiv-is-adjoint-pair-Adjunction :
-    (FG : Adjunction) {l1 l2 : Level} (X : obj-Large-Precategory C l1)
+    ( hom-Large-Precategory D
+      ( obj-left-adjoint-Adjunction-Large-Precategory FG X)
+      ( Y)) ≃
+    ( hom-Large-Precategory C
+      ( X)
+      ( obj-right-adjoint-Adjunction-Large-Precategory FG Y))
```
EgbertRijke, line 349:
```suggestion
    hom-Large-Precategory D
      ( obj-left-adjoint-Adjunction-Large-Precategory FG X)
      ( Y) ≃
    hom-Large-Precategory C
      ( X)
      ( obj-right-adjoint-Adjunction-Large-Precategory FG Y)
```

src/category-theory/category-of-functors.lagda.md
```diff
@@ -0,0 +1,129 @@
...
+  (D : Precategory l3 l4)
+  (is-category-D : is-category-Precategory D)
+  where
+
+  equiv-natural-isomorphism-htpy-map-functor-is-category-Precategory :
+    (F G : functor-Precategory C D) →
+    ( htpy-map-Precategory C D
+      ( map-functor-Precategory C D F)
+      ( map-functor-Precategory C D G)) ≃
+    ( natural-isomorphism-Precategory C D F G)
```
EgbertRijke, line 55:
```suggestion
    htpy-map-Precategory C D
      ( map-functor-Precategory C D F)
      ( map-functor-Precategory C D G) ≃
    natural-isomorphism-Precategory C D F G
```

src/category-theory/category-of-functors.lagda.md
```diff
@@ -0,0 +1,129 @@
...
+      is-category-Precategory (functor-precategory-Precategory C D)
+    is-category-functor-precategory-Precategory F G =
+      is-equiv-htpy-equiv
+        ( ( equiv-iso-functor-natural-isomorphism-Precategory C D F G) ∘e
+          ( extensionality-functor-is-category-Precategory
+              C D is-category-D F G))
+        ( λ
+          { refl →
+            compute-iso-functor-natural-isomorphism-eq-Precategory
+              C D F G refl})
```
EgbertRijke, line 92:
```suggestion
        ( λ { refl →
              compute-iso-functor-natural-isomorphism-eq-Precategory
                C D F G refl})
```

This is not necessarily a suggestion for a change, but the way I have written the pattern matching lambdas untilnow is as suggested above, with the curly braces on the same line. However, with my convention the indentation level increases by four spaces. Still, I find that it looks better. It would probably be good if we write in the coding style guide how we format pattern matching lambdas.

src/category-theory/category-of-maps-of-categories.lagda.md
```diff
@@ -0,0 +1,159 @@
...
+## Lemmas
+
+### Extensionality of maps of precategories when the codomain is a category
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  (C : Precategory l1 l2)
+  (D : Precategory l3 l4)
+  (is-category-D : is-category-Precategory D)
```
EgbertRijke, line 56: So, we're going with long variable names now?

fredrik-bakke: I don't know. Are we? I've always preferred variable names with this form

src/category-theory/category-of-maps-of-categories.lagda.md
```diff
@@ -0,0 +1,159 @@
...
+                      ( equiv-eq
+                        ( ap-binary
+                          ( λ p q →
+                            coherence-square-hom-Precategory D
+                              ( hom-map-Precategory C D f a)
+                              ( p)
+                              ( q)
+                              ( hom-map-Precategory C D g a))
+                          ( compute-hom-iso-eq-Precategory D (K x))
+                          ( compute-hom-iso-eq-Precategory D (K y))))))))
```
EgbertRijke, line 89:
```suggestion
      ( equiv-right-swap-Σ) ∘e
      ( equiv-Σ
        ( is-natural-transformation-map-Precategory C D f g ∘ pr1)
        ( ( distributive-Π-Σ) ∘e
          ( equiv-Π-equiv-family
            ( λ x →
              extensionality-obj-Category (D , is-category-D)
                ( obj-map-Precategory C D f x)
                ( obj-map-Precategory C D g x))))
        ( λ K →
          equiv-implicit-Π-equiv-family
            ( λ x →
              equiv-implicit-Π-equiv-family
                ( λ y →
                  equiv-Π-equiv-family
                    ( λ a →
                      ( equiv-eq
                        ( ap-binary
                          ( λ p q →
                            coherence-square-hom-Precategory D
                              ( hom-map-Precategory C D f a)
                              ( p)
                              ( q)
                              ( hom-map-Precategory C D g a))
                          ( compute-hom-iso-eq-Precategory D (K x))
                          ( compute-hom-iso-eq-Precategory D (K y)))))))))
```

I think in this case adding the parentheses benefits the readability, because in this way the parentheses are places more predictably, and therefore the tree structure of the proof is a bit clearer.

src/category-theory/category-of-maps-of-categories.lagda.md
```diff
@@ -0,0 +1,159 @@
...
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  (C : Precategory l1 l2)
+  (D : Precategory l3 l4)
+  (is-category-D : is-category-Precategory D)
+  where
+
+  abstract
```
EgbertRijke, line 110: Nice! Have I already mentioned to you that I would like to propose a convention that if we use pattern matching lambdas in a definition, then that definition should be abstract?

fredrik-bakke: You did so last week 😄

src/category-theory/category-of-functors.lagda.md
VojtechStep: There are also occurrences of the style
```agda
λ {
(pat x) ->
  expr ;
(pat y) ->
  expr }
```
which I quite like, and it doesn't increase the indentation as much

EgbertRijke: I think it is quite strange that this style proposes code to be within the scope of `{ }` but it is indented more to the left than the opening brace `{`.

VojtechStep: Which isn't all that weird if you consider e.g. the record value construction syntax
```agda
record {
  fieldname1 = term1 ;
  fieldname2 = term2 ;
  ...
}
```

VojtechStep: Oh, and in the examples I found the lambda was in parantheses, so it would IMO be pretty consistent with function application:
```agda
( λ {
  (pat x) →
    expr ;
  (pat y) →
    expr })
```
and
```agda
λ {
  (pat x) →
    expr ;
  (pat y) →
    expr }
```

VojtechStep: Searching for `λ\s+\{\n` gives me
- `src/elementary-number-theory/powers-of-two.lagda.md`, line 65
- `src/finite-group-theory/orbits-permutations.lagda.md`, line 1129
- `src/set-theory/countable-sets.lagda.md`, lines 154, 157, 345, 368
- `src/synthetic-homotopy-theory/coforks.lagda.md`, line 279
- `src/synthetic-homotopy-theory/dependent-coforks.lagda.md`, line 353
🥹

EgbertRijke: Thanks! These definitely need fixing. I'll be on it.

EgbertRijke:
If you update this branch, then I'll merge it. I don't have time to look over all of it, but this PR contains many changes that shouldn't be held up too long, and most of it looks good anyway.

PR #796, opened 2023-09-24
The Legendre symbol

wrest64:
A few things to point out:

1) The function here `legendre` does not actually care whether or not `p` is prime, it just generally characterizes whether or not a number is a square in `Zp` for any natural `p`. I'm not sure whether or not this is confusing. Most interesting/important facts about the legendre symbol leverage the fact that `p` is prime, but I felt that fact should be invoked in those theorems, not in the definition of the function itself, seeing as there is no type of prime numbers.

2) Most of this file is definitions and theorems about squareness in `N`, `Z`, and `Zp`. It might be better if these are in separate files, but I wasn't sure where they should go. These didn't exist before as far as I could tell, and the basic justification for their existence is that the legendre symbol needs to decide whether or not a number is a square in` Zp`, which for `p=0` (see above) requires deciding squareness in `Z`, which in turn requires deciding squareness in `N`. At least, that was the most straightforward way I could think of for defining and proving these things. Maybe at least the functions `square-Z` and `square-Z-Mod` should be in the files where `mul-Z` and `mul-Z-Mod` are, since `square-N` was already defined and is where `mul-N` is.

3) There's a pretty long proof of an elementary algebraic identity: `(n+2)^2 = n(n+2) + (n+2) + (n+2)`. I didn't feel this was a general enough fact to warrant being in the top level, so it's in a `where` clause. There also might be a way to prove the theorem it's used for without it, not sure.

### DIFF
--- a/src/elementary-number-theory/legendre-symbol.lagda.md
+++ b/src/elementary-number-theory/legendre-symbol.lagda.md
@@ -1,0 +1,286 @@
+# Legendre Symbol
+
+```agda
+module elementary-number-theory.legendre-symbol where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import elementary-number-theory.absolute-value-integers
+open import elementary-number-theory.addition-natural-numbers
+open import elementary-number-theory.decidable-types
+open import elementary-number-theory.equality-natural-numbers
+open import elementary-number-theory.inequality-natural-numbers
+open import elementary-number-theory.integers
+open import elementary-number-theory.modular-arithmetic
+open import elementary-number-theory.multiplication-integers
+open import elementary-number-theory.multiplication-natural-numbers
+open import elementary-number-theory.natural-numbers
+open import elementary-number-theory.prime-numbers
+
+open import foundation.action-on-identifications-functions
+open import foundation.coproduct-types
+open import foundation.decidable-equality
+open import foundation.decidable-types
+open import foundation.dependent-pair-types
+open import foundation.identity-types
+open import foundation.logical-equivalences
+open import foundation.negation
+open import foundation.unit-type
+open import foundation.universe-levels
+
+open import foundation-core.cartesian-product-types
+open import foundation-core.transport-along-identifications
+
+open import univalent-combinatorics.fibers-of-maps
+```
+
+</details>
+
+## Idea
+
+The **Legendre Symbol** is a function which determines whether or not an integer
+is a perfect square in `ℤₚ` (i.e. whether or not it is a quadratic residue).
+Specifically, let `p : ℕ` be a natural and `a : ℤ` be an integer. If `a` is a
+square in `ℤₚ` then `legendre(p,a) = 1`. Similarly if `a` is not a square in
+`ℤₚ` then `legendre(p,a) = -1`. Finally if `a` is congruent to `0` in `ℤₚ` then
+`legendre(p,a) = 0`. Typically `p` is taken to be prime.
+
+## Natural Numbers
+
+### Definition for squareness in ℕ
+
+```agda
+is-square-ℕ : ℕ → UU lzero
+is-square-ℕ n = Σ ℕ (λ x → n ＝ square-ℕ x)
+```
+
+### n > √n for n > 1
+
+The idea is to assume `n = m + 2 ≤ sqrt(m + 2)` for some `m : ℕ` and derive a
+contradiction by squaring both sides of the inequality
+
+```agda
+greater-than-square-root-ℕ :
+  (n root : ℕ) → ¬ ((n +ℕ 2 ≤-ℕ root) × (n +ℕ 2 ＝ square-ℕ root))
+greater-than-square-root-ℕ n root (pf-leq , pf-eq) =
+  reflects-order-add-ℕ
+    ( square-ℕ root)
+    ( n *ℕ (n +ℕ 2) +ℕ n +ℕ 2)
+    0
+    ( tr
+      ( leq-ℕ (n *ℕ (n +ℕ 2) +ℕ n +ℕ 2 +ℕ square-ℕ root))
+      ( inv (left-unit-law-add-ℕ (square-ℕ root)))
+      ( concatenate-eq-leq-ℕ
+        ( square-ℕ root)
+        ( inv
+          ( helper ∙
+            ( ap
+              ( add-ℕ (n *ℕ (n +ℕ 2) +ℕ n +ℕ 2))
+              pf-eq)))
+        ( preserves-leq-mul-ℕ (n +ℕ 2) root (n +ℕ 2) root pf-leq pf-leq)))
+  where
+  helper : square-ℕ (n +ℕ 2) ＝ n *ℕ (n +ℕ 2) +ℕ n +ℕ 2 +ℕ n +ℕ 2
+  helper =
+    equational-reasoning
+    square-ℕ (n +ℕ 2)
+    ＝ n *ℕ (n +ℕ 2) +ℕ (2 *ℕ (n +ℕ 2))
+      by (right-distributive-mul-add-ℕ n 2 (n +ℕ 2))
+    ＝ n *ℕ (n +ℕ 2) +ℕ 2 *ℕ n +ℕ 4
+      by
+        ( ap
+          ( add-ℕ (n *ℕ (n +ℕ 2)))
+          ( left-distributive-mul-add-ℕ 2 n 2))
+    ＝ n *ℕ (n +ℕ 2) +ℕ (n +ℕ 2 +ℕ (n +ℕ 2))
+      by
+        ( ap
+          ( add-ℕ (n *ℕ (n +ℕ 2)))
+          ( equational-reasoning
+            2 *ℕ n +ℕ 4
+            ＝ 4 +ℕ 2 *ℕ n
+              by (commutative-add-ℕ (2 *ℕ n) 4)
+            ＝ 2 +ℕ 2 +ℕ (n +ℕ n)
+              by
+                ( ap
+                  ( add-ℕ 4)
+                  ( left-two-law-mul-ℕ n))
+            ＝ (2 +ℕ 2 +ℕ n) +ℕ n
+              by
+                ( inv
+                  ( associative-add-ℕ 4 n n))
+            ＝ n +ℕ (2 +ℕ 2 +ℕ n)
+              by (commutative-add-ℕ (2 +ℕ 2 +ℕ n) n)
+            ＝ n +ℕ (2 +ℕ (2 +ℕ n))
+              by
+                ( ap
+                  ( add-ℕ n)
+                  ( associative-add-ℕ 2 2 n))
+            ＝ n +ℕ (2 +ℕ (n +ℕ 2))
+              by
+                ( ap-add-ℕ {n} {2 +ℕ (2 +ℕ n)}
+                  refl
+                  ( ap
+                    ( add-ℕ 2)
+                    ( commutative-add-ℕ 2 n)))
+            ＝ n +ℕ 2 +ℕ (n +ℕ 2)
+              by
+                ( inv
+                  ( associative-add-ℕ n 2 (n +ℕ 2)))))
+    ＝ n *ℕ (n +ℕ 2) +ℕ n +ℕ 2 +ℕ (n +ℕ 2)
+      by
+        ( inv
+          ( associative-add-ℕ
+            ( n *ℕ (n +ℕ 2))
+            ( n +ℕ 2)
+            ( n +ℕ 2)))
+```
+
+### Squareness in ℕ is decidable
+
+```agda
+is-decidable-big-root :
+  (n : ℕ) → is-decidable (Σ ℕ (λ root → (n ≤-ℕ root) × (n ＝ square-ℕ root)))
+is-decidable-big-root 0 = inl (0 , star , refl)
+is-decidable-big-root 1 = inl (1 , star , refl)
+is-decidable-big-root (succ-ℕ (succ-ℕ n)) =
+  inr (λ H → greater-than-square-root-ℕ n (pr1 H) (pr2 H))
+
+is-decidable-square-ℕ : (n : ℕ) → is-decidable (is-square-ℕ n)
+is-decidable-square-ℕ n =
+  is-decidable-Σ-ℕ n
+    ( λ x → n ＝ square-ℕ x)
+    ( λ x → has-decidable-equality-ℕ n (square-ℕ x))
+    ( is-decidable-big-root n)
+```
+
+## Integers
+
+### Definitions for squareness in ℤ
+
+```agda
+square-ℤ : ℤ → ℤ
+square-ℤ a = mul-ℤ a a
+
+is-square-ℤ : ℤ → UU lzero
+is-square-ℤ a = Σ ℤ (λ x → a ＝ square-ℤ x)
+```
+
+### Squares in ℤ are nonnegative
+
+```agda
+is-decidable-nonnegative-square-ℤ :
+  (a : ℤ) →
+  (is-nonnegative-ℤ a) + (is-nonnegative-ℤ (neg-ℤ a)) →
+  is-nonnegative-ℤ (square-ℤ a)
+is-decidable-nonnegative-square-ℤ _ (inl x) = is-nonnegative-mul-ℤ x x
+is-decidable-nonnegative-square-ℤ a (inr x) =
+  tr
+    is-nonnegative-ℤ
+    ( double-negative-law-mul-ℤ a a)
+    ( is-nonnegative-mul-ℤ x x)
+
+is-nonnegative-square-ℤ : (a : ℤ) → is-nonnegative-ℤ (square-ℤ a)
+is-nonnegative-square-ℤ a =
+  is-decidable-nonnegative-square-ℤ a decide-is-nonnegative-ℤ
+```
+
+### The squares in ℤ are exactly the squares in ℕ
+
+```agda
+is-square-int-if-is-square-nat : {n : ℕ} → is-square-ℕ n → is-square-ℤ (int-ℕ n)
+is-square-int-if-is-square-nat (root , pf-square) =
+  ( pair
+    ( int-ℕ root)
+    ( ( ap int-ℕ pf-square) ∙
+      ( inv (mul-int-ℕ root root))))
+
+is-square-nat-if-is-square-int : {a : ℤ} → is-square-ℤ a → is-square-ℕ (abs-ℤ a)
+is-square-nat-if-is-square-int (root , pf-square) =
+  ( pair
+    ( abs-ℤ root)
+    ( ( ap abs-ℤ pf-square) ∙
+      ( multiplicative-abs-ℤ root root)))
+
+is-nat-square-iff-is-int-square :
+  (n : ℕ) → is-square-ℕ n ↔ is-square-ℤ (int-ℕ n)
+pr1 (is-nat-square-iff-is-int-square n) = is-square-int-if-is-square-nat
+pr2 (is-nat-square-iff-is-int-square n) H =
+  tr is-square-ℕ (abs-int-ℕ n) (is-square-nat-if-is-square-int H)
+
+is-int-square-iff-nonneg-nat-square :
+  (a : ℤ) → is-square-ℤ a ↔ is-nonnegative-ℤ a × is-square-ℕ (abs-ℤ a)
+pr1 (is-int-square-iff-nonneg-nat-square a) (root , pf-square) =
+  ( pair
+    ( tr is-nonnegative-ℤ (inv pf-square) (is-nonnegative-square-ℤ root))
+    ( is-square-nat-if-is-square-int (root , pf-square)))
+pr2 (is-int-square-iff-nonneg-nat-square a) (pf-nonneg , (root , pf-square)) =
+  ( pair
+    ( int-ℕ root)
+    ( ( inv (int-abs-is-nonnegative-ℤ a pf-nonneg)) ∙
+      ( pr2 (is-square-int-if-is-square-nat (root , pf-square)))))
+```
+
+### Squareness in ℤ is decidable
+
+```agda
+is-decidable-square-ℤ : (a : ℤ) → is-decidable (is-square-ℤ a)
+is-decidable-square-ℤ (inl n) =
+  inr (map-neg (pr1 (is-int-square-iff-nonneg-nat-square (inl n))) pr1)
+is-decidable-square-ℤ (inr (inl n)) =
+  inl (zero-ℤ , refl)
+is-decidable-square-ℤ (inr (inr n)) =
+  is-decidable-iff
+    is-square-int-if-is-square-nat
+    is-square-nat-if-is-square-int
+    ( is-decidable-square-ℕ (succ-ℕ n))
+```
+
+## ℤₚ
+
+### Definitions for squareness in ℤₚ
+
+```agda
+square-ℤ-Mod : (p : ℕ) → ℤ-Mod p → ℤ-Mod p
+square-ℤ-Mod 0 a = mul-ℤ a a
+square-ℤ-Mod (succ-ℕ p) a = mul-ℤ-Mod (succ-ℕ p) a a
+
+is-square-ℤ-Mod : (p : ℕ) → ℤ-Mod p → UU lzero
+is-square-ℤ-Mod 0 k = is-square-ℤ k
+is-square-ℤ-Mod (succ-ℕ p) k =
+  Σ (ℤ-Mod (succ-ℕ p)) (λ x → square-ℤ-Mod (succ-ℕ p) x ＝ k)
+```
+
+### Squareness in ℤₚ is decidable
+
+```agda
+is-decidable-square-ℤ-Mod :
+  (p : ℕ) (k : ℤ-Mod p) → is-decidable (is-square-ℤ-Mod p k)
+is-decidable-square-ℤ-Mod 0 k = is-decidable-square-ℤ k
+is-decidable-square-ℤ-Mod (succ-ℕ p) k =
+  is-decidable-fiber-Fin {succ-ℕ p} {succ-ℕ p} (square-ℤ-Mod (succ-ℕ p)) k
+```
+
+## Legendre Symbol
+
+### Definitions for the Legendre Symbol
+
+```agda
+int-is-square-ℤ-Mod :
+  {p : ℕ} {k : ℤ-Mod p} →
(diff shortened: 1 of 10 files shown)


src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+</details>
+
+## Idea
+
+The **Legendre Symbol** is a function which determines whether or not an integer
+is a perfect square in `ℤₚ` (i.e. whether or not it is a quadratic residue).
+Specifically, let `p : ℕ` be a natural and `a : ℤ` be an integer. If `a` is a
+square in `ℤₚ` then `legendre(p,a) = 1`. Similarly if `a` is not a square in
+`ℤₚ` then `legendre(p,a) = -1`. Finally if `a` is congruent to `0` in `ℤₚ` then
+`legendre(p,a) = 0`. Typically `p` is taken to be prime.
```
EgbertRijke, line 48:
```suggestion
The **Legendre symbol** is a function which determines whether or not an
[integer](elementary-number-theory.integers.md)
is a perfect square in the ring `ℤₚ` of
[integers modulo `p`](elementary-number-theory.modular-arithmetic.md)
(i.e. whether or not it is a quadratic residue).
Specifically, let `p : ℕ` be a
[natural number](elementary-number-theory.natural-numbers.md)
and `a : ℤ` be an integer. If `a` is a
square in `ℤₚ` then `legendre(p,a) = 1`. Similarly if `a` is not a square in
`ℤₚ` then `legendre(p,a) = -1`. Finally if `a` is congruent to `0` in `ℤₚ` then
`legendre(p,a) = 0`. Typically `p` is taken to be prime.
```

We like to put hyperlinks in our comments, which make it easier to navigate the website. In the above suggestion, I have added some. However, if you commit it, then likely `make pre-commit` will fail on the markdown formatting. In that case you'll have to do a `git pull` from your machine and run `make pre-commit` again to let it automatically fix the markdown formatting.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Natural Numbers
```
EgbertRijke, line 50: I would put the contents of this section into a separate file called `squares-natural-numbers.lagda.md`. Such a file would then be a natural place to record other facts about squares of natural numbers, that would not fit naturally in a file about the Legendre symbol.

Maybe it looks slightly funny to put "squares" in plural in the file name, but I have two reasons for suggesting it in this way: The file can be read to be about "squares of natural numbers" and we usually omit words like "of" and "in" in filenames unless we have a specific reason to include them, and also it is to keep consistent with the filename `squares-modular-arithmetic.lagda.md` that I'm suggesting below, where it would make less sense to put the word square in singular.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+
+is-decidable-square-ℕ : (n : ℕ) → is-decidable (is-square-ℕ n)
+is-decidable-square-ℕ n =
+  is-decidable-Σ-ℕ n
+    ( λ x → n ＝ square-ℕ x)
+    ( λ x → has-decidable-equality-ℕ n (square-ℕ x))
+    ( is-decidable-big-root n)
+```
+
+## Integers
```
EgbertRijke, line 157: I would put the contents of this section into a separate file called `squares-integers.lagda.md`. Such a file would then be a natural place to record other facts about squares of integers, that would not fit naturally in a file about the Legendre symbol.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+is-decidable-square-ℤ (inr (inl n)) =
+  inl (zero-ℤ , refl)
+is-decidable-square-ℤ (inr (inr n)) =
+  is-decidable-iff
+    is-square-int-if-is-square-nat
+    is-square-nat-if-is-square-int
+    ( is-decidable-square-ℕ (succ-ℕ n))
+```
+
+## ℤₚ
```
EgbertRijke, line 239: I would put the contents of this section into a separate file called `squares-modular-arithmetic.lagda.md`. Such a file would then be a natural place to record other facts about squares in modular arithmetic, that would not fit naturally in a file about the Legendre symbol.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+
+```agda
+is-decidable-square-ℤ-Mod :
+  (p : ℕ) (k : ℤ-Mod p) → is-decidable (is-square-ℤ-Mod p k)
+is-decidable-square-ℤ-Mod 0 k = is-decidable-square-ℤ k
+is-decidable-square-ℤ-Mod (succ-ℕ p) k =
+  is-decidable-fiber-Fin {succ-ℕ p} {succ-ℕ p} (square-ℤ-Mod (succ-ℕ p)) k
+```
+
+## Legendre Symbol
```
EgbertRijke, line 264: So if all the previous generalities about squares go into their separate files, the contents of this file will begin here. In that case you can follow the usual structure of our files and call this section `## Definition`.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,286 @@
...
+int-is-square-ℤ-Mod (inr _) (inr _) = neg-one-ℤ
+
+legendre : ℕ → ℤ → ℤ
+legendre p a =
+  int-is-square-ℤ-Mod
+    ( has-decidable-equality-ℤ-Mod p arg-mod-p (zero-ℤ-Mod p))
+    ( is-decidable-square-ℤ-Mod p arg-mod-p)
+  where
+  arg-mod-p : ℤ-Mod p
+  arg-mod-p = mod-ℤ p a
```
EgbertRijke, line 285: Very nice! Just a tiny nit-pick is that you don't really need a where clause here, so I would just replace `arg-mod-p` above with `mod-ℤ p a`.

One can also argue whether this is the natural extension of the Legendre symbol to all natural numbers, or whether the natural extension of the Legendre symbol to all natural numbers should be the Jacobi symbol. You noted in the pull request description that there is no type of all primes. However, we do have the type of all primes defined here:

https://unimath.github.io/agda-unimath/elementary-number-theory.prime-numbers.html#1598

So it would be an option to restrict the Legendre symbols to the primes. Perhaps that is better, so that we keep in line with the standard convention of what the Legendre symbol is.

EgbertRijke (approved):
Very nice pull request, thank you for doing this! I have actually been interested in trying to see if there is anything that can be said about the Legendre symbol from a univalent point of view. So I am very happy that you are adding the Legendre symbol to the library.

My biggest comments were that I would factor out the generalities about squares into separate files. The rest were just some small remarks. Would you mind doing this, or would you prefer that we do this after we merge your PR?

wrest64:
> I have actually been interested in trying to see if there is anything that can be said about the Legendre symbol from a univalent point of view.

I'd be very interested to hear your thoughts about this at some point.

> My biggest comments were that I would factor out the generalities about squares into separate files. The rest were just some small remarks. Would you mind doing this, or would you prefer that we do this after we merge your PR?

Yes I can do this, and also make the specific changes laid out in your comments. When the changes are made, should I create a new pull request?

EgbertRijke:
> Yes I can do this, and also make the specific changes laid out in your comments. When the changes are made, should I create a new pull request?

No, thankfully that is not necessary on github. You simply commit and push your changes, and they should automatically show up here.

Make sure to run `make pre-commit` again every time before you push. And also, since you will be adding links, it is good to run `make website` after you run `make pre-commit`. This will build the website locally, mainly to test if everything works as expected, and you'll get warnings if there are any broken links.

EgbertRijke:
> I'd be very interested to hear your thoughts about this at some point.

André Joyal once pointed me to Cartier's article "Sur une généralisation des symboles the Legendre-Jacobi".

src/elementary-number-theory/legendre-symbol.lagda.md
wrest64: > Very nice! Just a tiny nit-pick is that you don't really need a where clause here, so I would just replace `arg-mod-p` above with `mod-ℤ p a`. 

Yeah that `where` was mostly habit from traditional programming in not wanting to re-compute the same thing.

> One can also argue whether this is the natural extension of the Legendre symbol to all natural numbers, or whether the natural extension of the Legendre symbol to all natural numbers should be the Jacobi symbol. You noted in the pull request description that there is no type of all primes. However, we do have the type of all primes defined here:
> 
> https://unimath.github.io/agda-unimath/elementary-number-theory.prime-numbers.html#1598
> 
> So it would be an option to restrict the Legendre symbols to the primes. Perhaps that is better, so that we keep in line with the standard convention of what the Legendre symbol is.

Thanks for the call out to the type of primes! I agree it would be more consistent to have the legendre symbol be defined for primes, and so can make that change.

I forgot about the Jacobi symbol, but yes I think that makes sense. With Legendre defined here and some prime decomposition stuff given in the page for the fundamental theorem of arithmetic, I can get the Jacobi symbol defined. But I won't do it here as part of this PR.

EgbertRijke: Perfect! I agree that the definition of the Jacobi symbol would be better in a separate PR

src/elementary-number-theory/legendre-symbol.lagda.md
wrest64: This all sounds good. However unlike `Z` and `Zp`, the `square-N` function is already defined in the page for multiplication on `N`. There is also the definition for `cube-N` there. I feel the `square-N` function should be in `squares-natural-numbers` for the sake of consistency.

EgbertRijke: Good call! You can move it over to the new file.

wrest64: I see from grepping around that there are a couple pages which use `square-N` (pythagorean-triples, square-free-natural-numbers, and multiplication-natural-numbers itself). I can make the import changes to those files by hand. Going forward, is there a mechanism for bulk changing import statements when files are renamed or definitions moved around?

EgbertRijke: The way I do it is that first I generate the `evertything` file with `make check`. Then I open the `everything` file and start running `agda` on it. This will jump me to the files of which the imports need modification, and then I do that by hand. I don't know if there's a better way, but in the end the `everything` file needs to compile anyway so I think it doesn't loose much time.

EgbertRijke: All the things in `multiplication-natural-numbers` that reference `square-N` can move to the new file about squares of natural numbers.

wrest64: This seems less prudent/could go etiher way, but perhaps the page square-free-natural-numbers should be converted into a section within squares-natural-numbers? At the moment there's nothing on that page except for the definition of being square-free in `N`. I think it might make sense for facts about square-freeness to be in the relevant squares page, but not sure.

EgbertRijke: I like to have it in a separate file, because the concept of square-free natural number is it’s own concept, with eventually its own theorems and lemmas about it. We separate literally everything by concept:)

src/elementary-number-theory/legendre-symbol.lagda.md
EgbertRijke: Note that the decision whether something should be in its own file is made on a conceptual basis, not on the basis of how much has been formalized about it.

EgbertRijke: The above comment is misplaced because I’m using GitHub on my phone. My apologies.

wrest64: no problem!

src/elementary-number-theory/legendre-symbol.lagda.md
wrest64: gotcha make sense :+1:

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,51 @@
...
+int-is-square-ℤ-Mod :
+  {p : ℕ} {k : ℤ-Mod p} →
+  is-decidable (is-zero-ℤ-Mod p k) →
+  is-decidable (is-square-ℤ-Mod p k) →
+  ℤ
+int-is-square-ℤ-Mod (inl _) _ = zero-ℤ
+int-is-square-ℤ-Mod (inr _) (inl _) = one-ℤ
+int-is-square-ℤ-Mod (inr _) (inr _) = neg-one-ℤ
+
+legendre : Prime-ℕ → ℤ → ℤ
```
EgbertRijke, line 46: Perhaps a better name for this entry would be `legendre-symbol`, as there might be more things due to Legendre.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -0,0 +1,51 @@
+# The Legendre symbol
```
EgbertRijke, line 1: Very good! Now this file is sharply focused on the legendre symbol.

src/elementary-number-theory/legendre-symbol.lagda.md
wrest64: Ah yes right, I was meaning to make that change thanks for bringing that up.

src/elementary-number-theory/squares-natural-numbers.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+      ( concatenate-eq-leq-ℕ
+        ( square-ℕ root)
+        ( inv
+          ( helper ∙
+            ( ap
+              ( add-ℕ (n *ℕ (n +ℕ 2) +ℕ n +ℕ 2))
+              pf-eq)))
+        ( preserves-leq-mul-ℕ (n +ℕ 2) root (n +ℕ 2) root pf-leq pf-leq)))
+  where
+  helper : square-ℕ (n +ℕ 2) ＝ n *ℕ (n +ℕ 2) +ℕ n +ℕ 2 +ℕ n +ℕ 2
```
EgbertRijke, line 86: Now I think it makes sense to take the helper out into a separate definition called `square-succ-succ-ℕ`. Also note that we have a lemma `square-succ-ℕ` about `square-ℕ (succ-ℕ n)`. Could this lemma be used to shorten your derivation?

EgbertRijke:
This PR is shaping up very nicely. I will make a few more formatting suggestions, and then we can merge.

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+is-decidable-nonnegative-square-ℤ :
+  (a : ℤ) →
+  (is-nonnegative-ℤ a) + (is-nonnegative-ℤ (neg-ℤ a)) →
+  is-nonnegative-ℤ (square-ℤ a)
+is-decidable-nonnegative-square-ℤ _ (inl x) = is-nonnegative-mul-ℤ x x
+is-decidable-nonnegative-square-ℤ a (inr x) =
+  tr
+    is-nonnegative-ℤ
+    ( double-negative-law-mul-ℤ a a)
+    ( is-nonnegative-mul-ℤ x x)
```
EgbertRijke, line 61:
```suggestion
  tr
    ( is-nonnegative-ℤ)
    ( double-negative-law-mul-ℤ a a)
    ( is-nonnegative-mul-ℤ x x)
```

Even though the type family doesn't require to be wrapped in parentheses, we like to put them anyway, to treat all three arguments of `tr` the same.

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+
+### The squares in ℤ are exactly the squares in ℕ
+
+```agda
+is-square-int-if-is-square-nat : {n : ℕ} → is-square-ℕ n → is-square-ℤ (int-ℕ n)
+is-square-int-if-is-square-nat (root , pf-square) =
+  ( pair
+    ( int-ℕ root)
+    ( ( ap int-ℕ pf-square) ∙
+      ( inv (mul-int-ℕ root root))))
```
EgbertRijke, line 76:
```suggestion
  ( ( int-ℕ root) ,
    ( ( ap int-ℕ pf-square) ∙
      ( inv (mul-int-ℕ root root))))
```

Some old files still use `pair` but we're mostly moving away from that. Now we favor `_,_`.

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+
+is-square-nat-if-is-square-int : {a : ℤ} → is-square-ℤ a → is-square-ℕ (abs-ℤ a)
+is-square-nat-if-is-square-int (root , pf-square) =
+  ( pair
+    ( abs-ℤ root)
+    ( ( ap abs-ℤ pf-square) ∙
+      ( multiplicative-abs-ℤ root root)))
```
EgbertRijke, line 83:
```suggestion
  ( ( abs-ℤ root) ,
    ( ( ap abs-ℤ pf-square) ∙
      ( multiplicative-abs-ℤ root root)))
```

src/elementary-number-theory/squares-natural-numbers.lagda.md
wrest64: > note that we have a lemma `square-succ-ℕ` about `square-ℕ (succ-ℕ n)`. Could this lemma be used to shorten your derivation?

I see what you're saying that's a good call, I'll definitely look at that. 

> Now I think it makes sense to take the helper out into a separate definition called `square-succ-succ-ℕ`

In general I wasn't sure about `square-succ-succ-N` as a lemma because aren't there many different identities which could serve as that lemma? I have the same question about `square-succ-N` but I didn't write that of course. In fact the particular identity I use is not what I would choose as the one for the lemma if I had to choose.

What I mean is, `(n+2)² = n² + 4n + 4` and `(n+2)² = n(n+2) + (n+2) + (n+2)` are true. The former is what I initially think about, while the second is the one I need for my proof.

src/elementary-number-theory/squares-integers.lagda.md
wrest64: Okay interesting. Maybe it's habit from lisp `cons` but the word `pair` feels more natural for anything of substantial length. But agreement matters more than anything for stuff like this, so I'm glad at least a decision has been made :+1:

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,117 @@
+# Squares in the Integers
```
fredrik-bakke, line 1: We usually don't use title case for our headers
```suggestion
# Squares in the integers
```

src/elementary-number-theory/squares-natural-numbers.lagda.md
```diff
@@ -0,0 +1,157 @@
+# Squares in the Natural Numbers
```
fredrik-bakke, line 1: We usually don't use title case for our headers.
```suggestion
# Squares in the natural numbers
```

src/elementary-number-theory/squares-natural-numbers.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+```agda
+square-succ-ℕ :
+  (k : ℕ) →
+  square-ℕ (succ-ℕ k) ＝ succ-ℕ ((succ-ℕ (succ-ℕ k)) *ℕ k)
+square-succ-ℕ k =
+  ( right-successor-law-mul-ℕ (succ-ℕ k) k) ∙
+  ( commutative-add-ℕ (succ-ℕ k) ((succ-ℕ k) *ℕ k))
+```
+
+### n > √n for n > 1
```
fredrik-bakke, line 61: We place equations in guarded blocks even in headers
```suggestion
### `n > √n` for `n > 1`
```

fredrik-bakke (approved):
Sorry to double-team on you, I just had some minor corrections on formatting. The PR is looking very good!

EgbertRijke:
> Sorry to double-team on you, I just had some minor corrections on formatting. The PR is looking very good!

Thanks for taking over! I am in a train through Austria and it goes through a lot of tunnels, so I don't have good connection. I'll let you take it from here.

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+
+is-nonnegative-square-ℤ : (a : ℤ) → is-nonnegative-ℤ (square-ℤ a)
+is-nonnegative-square-ℤ a =
+  is-decidable-nonnegative-square-ℤ a decide-is-nonnegative-ℤ
+```
+
+### The squares in ℤ are exactly the squares in ℕ
+
+```agda
+is-square-int-if-is-square-nat : {n : ℕ} → is-square-ℕ n → is-square-ℤ (int-ℕ n)
```
VojtechStep, line 71: Unless the `-if-` in the name stands for something else than implication that I'm missing, the more conventional name would be `is-square-int-is-square-nat`

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+    ( ( ap int-ℕ pf-square) ∙
+      ( inv (mul-int-ℕ root root))))
+
+is-square-nat-if-is-square-int : {a : ℤ} → is-square-ℤ a → is-square-ℕ (abs-ℤ a)
+is-square-nat-if-is-square-int (root , pf-square) =
+  ( ( abs-ℤ root) ,
+    ( ( ap abs-ℤ pf-square) ∙
+      ( multiplicative-abs-ℤ root root)))
+
+is-nat-square-iff-is-int-square :
```
VojtechStep, line 83: When proving a term of type `A <-> B`, we prefer to name it with prefix `iff-B-A` (because the "underlying map" has type `A -> B`), so the conventional name for this definition would be `iff-is-square-int-is-square-nat`

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+### The squares in ℤ are exactly the squares in ℕ
+
+```agda
+is-square-int-if-is-square-nat : {n : ℕ} → is-square-ℕ n → is-square-ℤ (int-ℕ n)
+is-square-int-if-is-square-nat (root , pf-square) =
+  ( ( int-ℕ root) ,
+    ( ( ap int-ℕ pf-square) ∙
+      ( inv (mul-int-ℕ root root))))
+
+is-square-nat-if-is-square-int : {a : ℤ} → is-square-ℤ a → is-square-ℕ (abs-ℤ a)
```
VojtechStep, line 77: Similarly -> `is-square-nat-is-square-int`

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+    ( ( ap abs-ℤ pf-square) ∙
+      ( multiplicative-abs-ℤ root root)))
+
+is-nat-square-iff-is-int-square :
+  (n : ℕ) → is-square-ℕ n ↔ is-square-ℤ (int-ℕ n)
+pr1 (is-nat-square-iff-is-int-square n) = is-square-int-if-is-square-nat
+pr2 (is-nat-square-iff-is-int-square n) H =
+  tr is-square-ℕ (abs-int-ℕ n) (is-square-nat-if-is-square-int H)
+
+is-int-square-iff-nonneg-nat-square :
```
VojtechStep, line 89: Similarly -> `iff-is-nonneg-nat-square-is-int-square`

src/elementary-number-theory/squares-integers.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+pr2 (is-int-square-iff-nonneg-nat-square a) (pf-nonneg , (root , pf-square)) =
+  ( ( int-ℕ root) ,
+    ( ( inv (int-abs-is-nonnegative-ℤ a pf-nonneg)) ∙
+      ( pr2 (is-square-int-if-is-square-nat (root , pf-square)))))
+```
+
+### Squareness in ℤ is decidable
+
+```agda
+is-decidable-square-ℤ : (a : ℤ) → is-decidable (is-square-ℤ a)
```
VojtechStep, line 103: I'd consider naming this `is-decidable-is-square-Z`, because that's its type. It doesn't decide squares, it decides the property of an integer being a square

src/elementary-number-theory/squares-modular-arithmetic.lagda.md
```diff
@@ -0,0 +1,54 @@
...
+square-root-ℤ-Mod {0} _ (root , _) = root
+square-root-ℤ-Mod {succ-ℕ p} _ (root , _) = root
+```
+
+## Properties
+
+### Squareness in ℤₚ is decidable
+
+```agda
+is-decidable-square-ℤ-Mod :
```
VojtechStep, line 48: Similarly -> `is-decidable-is-square-Z-Mod`

src/elementary-number-theory/squares-natural-numbers.lagda.md
```diff
@@ -0,0 +1,156 @@
...
+
+```agda
+is-decidable-big-root :
+  (n : ℕ) → is-decidable (Σ ℕ (λ root → (n ≤-ℕ root) × (n ＝ square-ℕ root)))
+is-decidable-big-root 0 = inl (0 , star , refl)
+is-decidable-big-root 1 = inl (1 , star , refl)
+is-decidable-big-root (succ-ℕ (succ-ℕ n)) =
+  inr (λ H → greater-than-square-root-ℕ n (pr1 H) (pr2 H))
+
+is-decidable-square-ℕ : (n : ℕ) → is-decidable (is-square-ℕ n)
```
VojtechStep, line 150: Similarly -> `is-decidable-is-square-N`

VojtechStep (commented):
Nice PR 👌. I left a few suggestions for names that would be more in line with our conventions.

fredrik-bakke:
By the way! If any call out to you, you can add number sequences derived from the Legendre Symbol to our [sequences of the online encyclopedia of integer sequences](https://unimath.github.io/agda-unimath/online-encyclopedia-of-integer-sequences.oeis.html#sequences-of-the-online-encyclopedia-of-integer-sequences) 😄

fredrik-bakke:
Alright, sorry for the slight delay. This PR is ready to merge now, so I will do that. Thank you for the nice contribution!

PR #798, opened 2023-09-26
The inverse Eckmann-Hilton argument

morphismz:
This pr formalizes the inverse Eckmann-Hilton construction. This construction uses whiskering on the right, instead of whiskering on the left. This allows us to construct an inverse to the previous Eckmann-Hilton term.


src/foundation/path-algebra.lagda.md
```diff
@@ -341,6 +341,10 @@ module _
 
 ### The whiskering operations allow us to commute higher identifications
 
+There are two natural ways to commute higher identifications: using whiskering
+on the left versus using whiskering on the right. Thes two ways "undo" each
```
EgbertRijke, line 345: Thes -> These

src/synthetic-homotopy-theory/eckmann-hilton-argument.lagda.md
```diff
@@ -31,9 +33,10 @@ open import synthetic-homotopy-theory.triple-loop-spaces
 
 There are two classical statements of the Eckmann-Hilton argument. The first
 states that a group object in the category of groups is abelian. The second
-states that `π₂ (X)` is abelian, for any space `X`.. The first statement can be
-thought of as a more algebraic phrasing, while the second can be thought of as a
-more homotopy theoretic phrasing.
+states that `π₂ (X)` is abelian, for any space `X`. The former is an algebraic
+statement, while the later is a homotopy theoretic statment. As it turns out,
```
EgbertRijke, line 37: while the later -> while the latter

src/synthetic-homotopy-theory/eckmann-hilton-argument.lagda.md
```diff
@@ -176,19 +180,163 @@ module _
...
       ( identification-left-whisk refl α)
     by ( path-swap-nat-identification-left-whisk α β)
     ＝ β ∙ α
     by ( horizontal-concat-Id²
-      ( right-unit-law-identification-left-whisk-Ω² β)
+      ( right-unit-law-identification-right-whisk-Ω² β)
       ( left-unit-law-identification-left-whisk-Ω² α))
 ```
 
+There is another natural construction of an Eckmann-Hilton term along these
```
EgbertRijke, line 193: There is always a debate over how to use the word "term" in type theory. "Term" refers to a syntactic object, so one could argue that what is written down in a proof assistant is always a term. However, those terms refer to elements of types, which can be elements, numbers, functions, identifications, and so on, but not terms unless those types specifically denote syntax. So, when you speak of the "EH term", I think it would be more accurate to speak of the "EH identification".

morphismz: That makes sense to me, I'll make that change

fredrik-bakke: Apologies, using "term" was originally my suggestion.

PR #799, opened 2023-09-26
Define the Jacobi symbol and prove the Legendre symbol is periodic

wrest64:
Figured I would add the Jacobi symbol since the Legendre symbol is now defined.

I wasn't sure which "squareness" page to link to when mentioning it in the "Idea" section of the Jacobi symbol, so I just decided to link to the squareness in `Zp` page since that's the one most relevant for the Legendre (and herce, Jacobi) symbol. But perhaps there should be no link there at all.

Here there's also some minor reorganization of the Legendre symbol definition and a proof of its periodicity (in its top argument).

Also in here are some very minor formatting tweaks to some other entries. Do let me know if those kinds of changes should be in separate pull requests from more contentful ones.


src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -43,9 +45,26 @@ int-is-square-ℤ-Mod (inl _) _ = zero-ℤ
...
+  legendre-symbol p a ＝ legendre-symbol p b
+is-legendre-symbol-periodic p a b H =
+  equational-reasoning
+  legendre-symbol p a
+  ＝ legendre-symbol' p (mod-ℤ (pr1 p) a)
+    by refl
+  ＝ legendre-symbol' p (mod-ℤ (pr1 p) b)
+    by ap (legendre-symbol' p) H
+  ＝ legendre-symbol p b
+    by refl
```
EgbertRijke, line 69:
```suggestion
  ap (legendre-symbol' p) H
```

We usually don't record unfolding the definition in equational reasonings.

wrest64: oh whoop thanks I think I got a bit of tunnel vision there, since I started writing the proof first and then rewrote the definition for legendre-symbol.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -43,9 +45,26 @@ int-is-square-ℤ-Mod (inl _) _ = zero-ℤ
...
   int-is-square-ℤ-Mod
-    ( has-decidable-equality-ℤ-Mod p (mod-ℤ p a) (zero-ℤ-Mod p))
-    ( is-decidable-is-square-ℤ-Mod p (mod-ℤ p a))
+    ( has-decidable-equality-ℤ-Mod p k (zero-ℤ-Mod p))
+    ( is-decidable-is-square-ℤ-Mod p k)
+
+legendre-symbol : Prime-ℕ → ℤ → ℤ
+legendre-symbol p a = legendre-symbol' p (mod-ℤ (pr1 p) a)
+
+is-legendre-symbol-periodic :
```
fredrik-bakke, line 57: `is-periodic-legendre-symbol`*

Notice how with an appropriate definition of the predicate `is-periodic`, the type signature of `is-periodic-legendre-symbol` would have a conclusion of type `is-periodic legendre-symbol`.

src/elementary-number-theory/legendre-symbol.lagda.md
```diff
@@ -43,9 +45,26 @@ int-is-square-ℤ-Mod (inl _) _ = zero-ℤ
 int-is-square-ℤ-Mod (inr _) (inl _) = one-ℤ
 int-is-square-ℤ-Mod (inr _) (inr _) = neg-one-ℤ
 
-legendre-symbol : Prime-ℕ → ℤ → ℤ
-legendre-symbol (p , _) a =
+legendre-symbol' : ((p , _) : Prime-ℕ) → ℤ-Mod p → ℤ
```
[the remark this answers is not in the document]
fredrik-bakke: Also, pattern matching in type signatures is sadly bad practice, due to how Agda handles them with let-bindings. If you've ever seen `.patternInTele0` in your Agda pane, this is due to a let-binding. Instead, I suggest using the named projection `nat-Prime-ℕ` to get  the underlying natural number.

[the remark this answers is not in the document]
wrest64: interesting, good to know. thanks for pointing out `nat-Prime-N` I'll make that change.

src/elementary-number-theory/legendre-symbol.lagda.md
wrest64: That sounds good for consistency, I'll make the name change. Where would be the best place to define periodicity? It seems like it would depend a bit on how general it should be. One option is to have it work for any function `f : A -> B` and any binary operation on `A`. But there could also just be a few periodicity definitions for each of the common number types.

fredrik-bakke: Hmm, I'm not sure what is a good general definition for periodicity. If you want, you can define it for integers or abelian groups, but I don't think it is necessary to actually define `is-periodic` here. What your statement is really saying is that the map somehow descends to a quotient group, and I'm not sure that is generally considered "periodicity". In this context, periodicity is just a special case of a property this map is exhibiting then. Still, using the word "periodic" in your definition is demonstrative, so it can stay. My original point was mainly that this formulation was consistent with other names where the predicate does have its own definition.

wrest64: That seems worth doing (though yes not in this PR). If it idoes get done though, wouldn't it be confusing calling this specific fact `is-periodic-legendre-symbol` if its type is not `is-periodic (legendre-symbol)`? I suppose it's still about periodicity in the sense that `legendre(p,a) = legendre(p,a+np)` for natural `n`, so `p` is the fundamental period for `legendre(p)`.

EgbertRijke: Sorry, I marked this conversation as resolved prematurely. Fredrik suggested correctly that this function should be renamed to

`is-periodic-legendre-symbol`

fredrik-bakke: So, first of all, the entry has to get _some_ name, and I think the prefix `is-periodic` does a good job of capturing the essence of what the entry proves. Having it prepended instead of your current proposal is just our naming convention. By having a rigid naming convention, we get the benefit that we can usually guess the names of entries so we don't have to look them up all the time. Also, prefixing conclusions makes it quick to infer the type of some term when it appears in a proof.

Moreover, regarding the prefix `is-periodic`, this entry should capture the fact that this function is periodic in such a way that _if_ we were to define the `is-periodic` predicate, then this entry would produce a term of that type, thus not lead to confusion. Note that we do use similar prefixes in the library without defining their predicate. For instance `is-section` and `is-retraction` are examples of such prefixes.

wrest64: I see now yes I agree that makes sense.

src/elementary-number-theory/jacobi-symbol.lagda.md
```diff
@@ -0,0 +1,51 @@
...
+The **Jacobi symbol** is a function which encodes information about the
+[squareness](elementary-number-theory.squares-modular-arithmetic.md) of an
+[integer](elementary-number-theory.integers.md) within certain
+[rings of integers modulo `p`](elementary-number-theory.modular-arithmetic.md),
+for [prime](elementary-number-theory.prime-numbers.md) `p`. Specifically,
+`jacobi-symbol(a,n)` for an integer `a : ℤ` and natural number `n : ℕ` is the
+product of the [legendre symbols](elementary-number-theory.legendre-symbol.md)
+`legendre-symbol(p₁,a),legendre-symbol(p₂,a),...,legendre-symbol(pₖ,a)`, where
+`p₁,...,pₖ` are the prime factors of `n`, not necessarily distinct (i.e. it is
+possible that`pᵢ = pⱼ`).
```
fredrik-bakke, line 36:
```suggestion
possible that `pᵢ = pⱼ`).
```

fredrik-bakke:
After this last fix goes through, I'll merge this PR. Thank you for another nice contribution! :)

wrest64:
> After this last fix goes through, I'll merge this PR. Thank you for another nice contribution! :)

fun as always :+1:

PR #797, opened 2023-09-24
Iterated type families

EgbertRijke:
This is an experimental attempt to get some sort of uniform treatment of iterated type families going.

Work in progress, and ideas are very welcome.

PR #802, opened 2023-09-28
Subobject classifiers

maybemabeline:
I plan to implement subobject classifiers, construct them in the category of presheaves and prove some adjacent results.
To start off I realized I need pullbacks in large precategories, so I tried enlarging the file. The object w' in line 123 is forced to be at level l1 by the following functions though and I can't seem to tell why. I also added the type of all monomorphisms in a large precategory.

### DIFF
--- a/src/category-theory/monomorphisms-in-large-precategories.lagda.md
+++ b/src/category-theory/monomorphisms-in-large-precategories.lagda.md
@@ -11,6 +11,7 @@ open import category-theory.isomorphisms-in-large-precategories
 open import category-theory.large-precategories
 
 open import foundation.action-on-identifications-functions
+open import foundation.dependent-pair-types
 open import foundation.embeddings
 open import foundation.equivalences
 open import foundation.identity-types
@@ -49,6 +50,33 @@ module _
   is-prop-is-mono-Large-Precategory : is-prop is-mono-Large-Precategory
   is-prop-is-mono-Large-Precategory =
     is-prop-type-Prop is-mono-prop-Large-Precategory
+```
+
+## The type of monomorphisms in a large precategory
+
+```agda
+module _
+  {α : Level → Level} {β : Level → Level → Level}
+  (C : Large-Precategory α β) {l1 l2 : Level} (l3 : Level)
+  (X : obj-Large-Precategory C l1) (Y : obj-Large-Precategory C l2)
+  where
+
+  mono-Large-Precategory : UU (α l3 ⊔ β l3 l1 ⊔ β l3 l2 ⊔ β l1 l2)
+  mono-Large-Precategory =
+    Σ
+      ( hom-Large-Precategory C X Y)
+      ( is-mono-Large-Precategory C l3 X Y)
+
+  module _
+    (f : mono-Large-Precategory)
+    where
+
+    hom-mono-Large-Precategory : hom-Large-Precategory C X Y
+    hom-mono-Large-Precategory = pr1 f
+
+    is-mono-mono-Large-Precategory :
+      is-mono-Large-Precategory C l3 X Y hom-mono-Large-Precategory
+    is-mono-mono-Large-Precategory = pr2 f
 ```
 
 ## Properties
--- a/src/category-theory/monomorphisms-in-large-precategories.lagda.md
+++ b/src/category-theory/monomorphisms-in-large-precategories.lagda.md
@@ -67,16 +67,14 @@ module _
       ( hom-Large-Precategory C X Y)
       ( is-mono-Large-Precategory C l3 X Y)
 
-  module _
-    (f : mono-Large-Precategory)
-    where
+  hom-mono-Large-Precategory :
+    (f : mono-Large-Precategory) → hom-Large-Precategory C X Y
+  hom-mono-Large-Precategory f = pr1 f
 
-    hom-mono-Large-Precategory : hom-Large-Precategory C X Y
-    hom-mono-Large-Precategory = pr1 f
-
-    is-mono-mono-Large-Precategory :
-      is-mono-Large-Precategory C l3 X Y hom-mono-Large-Precategory
-    is-mono-mono-Large-Precategory = pr2 f
+  is-mono-mono-Large-Precategory :
+    (f : mono-Large-Precategory) →
+    is-mono-Large-Precategory C l3 X Y (hom-mono-Large-Precategory f)
+  is-mono-mono-Large-Precategory f = pr2 f
 ```
 
 ## Properties
--- a/src/category-theory/monomorphisms-in-large-precategories.lagda.md
+++ b/src/category-theory/monomorphisms-in-large-precategories.lagda.md
@@ -96,34 +96,34 @@ module _
       ( λ P →
         ( inv
           ( left-unit-law-comp-hom-Large-Precategory C g)) ∙
-          ( ( ap
+          ( ap
             ( λ h' → comp-hom-Large-Precategory C h' g)
             ( inv (is-retraction-hom-inv-iso-Large-Precategory C f))) ∙
-            ( ( associative-comp-hom-Large-Precategory C
+          ( associative-comp-hom-Large-Precategory C
+            ( hom-inv-iso-Large-Precategory C f)
+            ( hom-iso-Large-Precategory C f)
+            ( g)) ∙
+          ( ap
+            ( comp-hom-Large-Precategory C
+              ( hom-inv-iso-Large-Precategory C f))
+            ( P)) ∙
+          ( inv
+            ( associative-comp-hom-Large-Precategory C
               ( hom-inv-iso-Large-Precategory C f)
               ( hom-iso-Large-Precategory C f)
-              ( g)) ∙
-              ( ( ap
-                ( comp-hom-Large-Precategory C
-                  ( hom-inv-iso-Large-Precategory C f))
-                ( P)) ∙
-                ( ( inv
-                  ( associative-comp-hom-Large-Precategory C
-                    ( hom-inv-iso-Large-Precategory C f)
-                    ( hom-iso-Large-Precategory C f)
-                    ( h))) ∙
-                  ( ( ap
-                    ( λ h' → comp-hom-Large-Precategory C h' h)
-                    ( is-retraction-hom-inv-iso-Large-Precategory C f)) ∙
-                    ( left-unit-law-comp-hom-Large-Precategory C h)))))))
+              ( h))) ∙
+          ( ap
+            ( λ h' → comp-hom-Large-Precategory C h' h)
+            ( is-retraction-hom-inv-iso-Large-Precategory C f)) ∙
+          ( left-unit-law-comp-hom-Large-Precategory C h))
       ( λ p →
-        eq-is-prop
+        ( eq-is-prop
           ( is-set-hom-Large-Precategory C Z Y
             ( comp-hom-Large-Precategory C
               ( hom-iso-Large-Precategory C f)
               ( g))
             ( comp-hom-Large-Precategory C
               ( hom-iso-Large-Precategory C f)
-              ( h))))
+              ( h)))))
       ( λ p → eq-is-prop (is-set-hom-Large-Precategory C Z X g h))
 ```
--- a/src/category-theory/pullbacks-in-large-precategories.lagda.md
+++ b/src/category-theory/pullbacks-in-large-precategories.lagda.md
@@ -27,12 +27,12 @@ A pullback of two morphisms `f : hom y x` and `g : hom z x` in a category `C`
 consists of:
 
 - an object `w`
-- morphisms `p₁ : hom w y` and `p₂ : hom w z` such that
-- `f ∘ p₁ = g ∘ p₂` together with the universal property that for every object
-  `w'` and pair of morphisms `p₁' : hom w' y` and `p₂' : hom w' z` such that
-  `f ∘ p₁' = g ∘ p₂'` there exists a unique morphism `h : hom w' w` such that
-- `p₁ ∘ h = p₁'`
-- `p₂ ∘ h = p₂'`.
+- morphisms `p : hom w y` and `q : hom w z` such that
+- `f ∘ p = g ∘ q` together with the universal property that for every object
+  `w'` and pair of morphisms `p' : hom w' y` and `q' : hom w' z` such that
+  `f ∘ p' = g ∘ q'` there exists a unique morphism `h : hom w' w` such that
+- `p ∘ h = p'`
+- `q ∘ h = q'`.
 
 We say that `C` has all pullbacks if there is a choice of a pullback for each
 object `x` and pair of morphisms into `x` in `C`.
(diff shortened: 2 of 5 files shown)


src/category-theory/monomorphisms-in-large-precategories.lagda.md
```diff
@@ -51,6 +52,33 @@ module _
...
+
+  mono-Large-Precategory : UU (α l3 ⊔ β l3 l1 ⊔ β l3 l2 ⊔ β l1 l2)
+  mono-Large-Precategory =
+    Σ
+      ( hom-Large-Precategory C X Y)
+      ( is-mono-Large-Precategory C l3 X Y)
+
+  module _
+    (f : mono-Large-Precategory)
+    where
```
fredrik-bakke, line 72: We avoid nested modules. Here I would probably just copy the hypothesis of `f` to each definition, but you can also close the previous module and open a completely new one.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+A pullback of two morphisms `f : hom y x` and `g : hom z x` in a category `C`
+consists of:
+
+- an object `w`
+- morphisms `p₁ : hom w y` and `p₂ : hom w z` such that
+- `f ∘ p₁ = g ∘ p₂` together with the universal property that for every object
+  `w'` and pair of morphisms `p₁' : hom w' y` and `p₂' : hom w' z` such that
+  `f ∘ p₁' = g ∘ p₂'` there exists a unique morphism `h : hom w' w` such that
+- `p₁ ∘ h = p₁'`
+- `p₂ ∘ h = p₂'`.
```
fredrik-bakke, line 35: I suggest using different letters for `p₁` and `p₂` instead of indices. E.g. `p` and `q`. Indices usually make statements harder to read and remember. Our only exception to this rule as far as I can recall are universe levels.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+open import foundation.propositions
+open import foundation.unique-existence
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+A pullback of two morphisms `f : hom y x` and `g : hom z x` in a category `C`
```
fredrik-bakke, line 26:
```suggestion
A **pullback** of two morphisms `f : hom y x` and `g : hom z x` in a [category](category-theory.categories.md) `C`
```
We write terms in bold-face when defining them, and add links when mentioning terms for the first time in a file that are defined elsewhere.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+
+- an object `w`
+- morphisms `p₁ : hom w y` and `p₂ : hom w z` such that
+- `f ∘ p₁ = g ∘ p₂` together with the universal property that for every object
+  `w'` and pair of morphisms `p₁' : hom w' y` and `p₂' : hom w' z` such that
+  `f ∘ p₁' = g ∘ p₂'` there exists a unique morphism `h : hom w' w` such that
+- `p₁ ∘ h = p₁'`
+- `p₂ ∘ h = p₂'`.
+
+We say that `C` has all pullbacks if there is a choice of a pullback for each
```
fredrik-bakke, line 37:
```suggestion
We say that `C` **has all pullbacks** if there is a choice of a pullback for each
```

src/category-theory/pullbacks-in-large-precategories.lagda.md
maybemabeline: I would tend to agree, but I copied these from the existing file on pullbacks in precategories, so I thought I would leave it as is. I can change it both files.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+
+module _
+  {α : Level → Level} {β : Level → Level → Level}
+  {l1 l2 l3 l4 : Level} (C : Large-Precategory α β)
+  (x : obj-Large-Precategory C l1)
+  (y : obj-Large-Precategory C l2)
+  (z : obj-Large-Precategory C l3)
+  (f : hom-Large-Precategory C y x)
+  (g : hom-Large-Precategory C z x)
+  (w : obj-Large-Precategory C l4)
```
fredrik-bakke, line 169: You probably want `x` `y` `z` and `w` to be implicit arguments here, unless Agda is unable to infer them generally

src/category-theory/pullbacks-in-large-precategories.lagda.md
fredrik-bakke: Excellent! Some of the files in this module suffer from being a little outdated currently :)

src/category-theory/pullbacks-in-large-precategories.lagda.md
maybemabeline: Good point! I'll change this in the other file too.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+  (C : Large-Precategory α β) {l1 l2 l3 l4 : Level}
+  where
+
+  is-pullback-Large-Precategory :
+    (x : obj-Large-Precategory C l1) →
+    (y : obj-Large-Precategory C l2) →
+    (z : obj-Large-Precategory C l3) →
+    (f : hom-Large-Precategory C y x) →
+    (g : hom-Large-Precategory C z x) →
+    (w : obj-Large-Precategory C l4) →
```
fredrik-bakke, line 54:
```suggestion
    {x : obj-Large-Precategory C l1}
    {y : obj-Large-Precategory C l2}
    {z : obj-Large-Precategory C l3}
    {w : obj-Large-Precategory C l4}
    (f : hom-Large-Precategory C y x)
    (g : hom-Large-Precategory C z x) →
```

src/category-theory/pullbacks-in-large-precategories.lagda.md
fredrik-bakke: Also, when they are implicit, I would place `w` right after `z` instead

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+    hom-Large-Precategory C z x →
+    UU (α l1 ⊔ α l4 ⊔ β l1 l1 ⊔ β l1 l2 ⊔ β l1 l3 ⊔
+        β l1 l4 ⊔ β l4 l1 ⊔ β l4 l2 ⊔ β l4 l3)
+  pullback-Large-Precategory x y z f g =
+    Σ (obj-Large-Precategory C l4) λ w →
+    Σ (hom-Large-Precategory C w y) λ p₁ →
+    Σ (hom-Large-Precategory C w z) λ p₂ →
+    Σ (comp-hom-Large-Precategory C f p₁
+    ＝ comp-hom-Large-Precategory C g p₂) λ α →
+      is-pullback-Large-Precategory x y z f g w p₁ p₂ α
```
fredrik-bakke, line 84: The proper formatting for this is

```suggestion
    Σ ( obj-Large-Precategory C l4)
      ( λ w →
        Σ ( hom-Large-Precategory C w y)
          ( λ p →
            Σ ( hom-Large-Precategory C w z)
              ( λ q →
                Σ ( comp-hom-Large-Precategory C f p ＝
                    comp-hom-Large-Precategory C g q)
                  ( is-pullback-Large-Precategory x y z f g w p q))))
```
(if that stays within the 80-character limit)

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+    (p₁' : hom-Large-Precategory C w' y)
+    (p₂' : hom-Large-Precategory C w' z)
+    (α : comp-hom-Large-Precategory C f p₁'
+      ＝ comp-hom-Large-Precategory C g p₂')
+    where
+
+    morphism-into-pullback-Large-Precategory :
+      hom-Large-Precategory C w' object-pullback-Large-Precategory
+    morphism-into-pullback-Large-Precategory =
+      pr1 (pr1 (pr2 (pr2 (pr2 (pr2 (t x y z f g)))) w' p₁' p₂' α))
```
fredrik-bakke, line 133: You will want to define named projections for things like this instead of using cryptic composites of `pr1` and `pr2`. Maybe a good reference for how to do this is to look at how we define projections of (small) precategories.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+module _
+  {α : Level → Level} {β : Level → Level → Level}
+  (C : Large-Precategory α β) {l1 l2 : Level}
+  (t : has-all-pullbacks-Large-Precategory C)
+  (x y z : obj-Large-Precategory C l1)
+  (f : hom-Large-Precategory C y x)
+  (g : hom-Large-Precategory C z x)
+  where
+
+  object-pullback-Large-Precategory : obj-Large-Precategory C l2
```
fredrik-bakke, line 106: `obj-pullback-...`

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+  object-pullback-Large-Precategory = pr1 (t x y z f g)
+
+  pr1-pullback-Large-Precategory :
```
fredrik-bakke, line 109: Usually, I would want to say that you should take a look at `foundation.pullbacks` for how to name these. Sadly, that file is in need of refactoring as well. But you can have a look at `foundation.cones-over-cospans` for how to name these.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+    {z : obj-Large-Precategory C l3}
+    {w : obj-Large-Precategory C l4}
+    (f : hom-Large-Precategory C y x)
+    (g : hom-Large-Precategory C z x) →
+    (p₁ : hom-Large-Precategory C w y) →
+    (p₂ : hom-Large-Precategory C w z) →
+    comp-hom-Large-Precategory C f p₁ ＝ comp-hom-Large-Precategory C g p₂ →
+    UU (α l1 ⊔ β l1 l1 ⊔ β l1 l2 ⊔ β l1 l3 ⊔ β l1 l4)
+  is-pullback-Large-Precategory x y z f g w p₁ p₂ _ =
+    (w' : obj-Large-Precategory C l1) →
```
fredrik-bakke, line 60: `w'` must be able to live at any universe level

fredrik-bakke: this should mirror the definition of `universal-property-pullback` in foundation-core.universal-property-pullbacks

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+    (p₁' : hom-Large-Precategory C w' y) →
+    (p₂' : hom-Large-Precategory C w' z) →
+    comp-hom-Large-Precategory C f p₁' ＝ comp-hom-Large-Precategory C g p₂' →
+    ∃!
```
fredrik-bakke, line 64: Instead of using unique existence, we prefer to define universal properties in terms of certain evaluation maps being equivalences. Here, that unfolds to requiring that the map that sends a morphism `w' -> w`, where `w` is the pullback, to the cone constructed by postcomposing with `f` and `g` being an iso in the span category with fixed codomains `y` and `z`. Maybe Egbert has a different opinion on this, but that is what I would say this definition should be at the moment.

src/category-theory/pullbacks-in-large-precategories.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+  (w : obj-Large-Precategory C l4)
+  (p₁ : hom-Large-Precategory C w y)
+  (p₂ : hom-Large-Precategory C w z)
+  (α₁ : comp-hom-Large-Precategory C f p₁ ＝ comp-hom-Large-Precategory C g p₂)
+  where
+
+  is-prop-is-pullback-Large-Precategory :
+    is-prop (is-pullback-Large-Precategory C x y z f g w p₁ p₂ α₁)
+  is-prop-is-pullback-Large-Precategory =
+    is-prop-Π³ (λ w' p₁' p₂' → is-prop-function-type is-property-is-contr)
```
fredrik-bakke, line 178: Note that we are changing `is-prop-Π³` to a general `is-prop-iterated-Π 3` with #797 😁

src/category-theory/pullbacks-in-large-precategories.lagda.md
fredrik-bakke: Just to confirm your question earlier, using the above reference, x y z and w should indeed all be able to have different universe levels.

src/category-theory/pullbacks-in-large-precategories.lagda.md
fredrik-bakke: Maybe to expand a little on the above remark. Notice how this definition makes itself more readily available for use with univalence down the line.

fredrik-bakke:
Do let me know if you have any questions about my comments!

src/category-theory/pullbacks-in-large-precategories.lagda.md
maybemabeline: I'm not sure I understand how this would be an isomorphism in the category of spans. Do you mean that postcomposition with `f` and `g` forms equivalence between the slice category `C/w` and the category of spans with codomains `y` and `z`?

fredrik-bakke: Ah, yes, my bad, you're right!

fredrik-bakke: Also, I shouldn't have said span category. It's the cone category over the cospan of `f` and `g`. And the cone is constructed by postcomposing with `p` and `q`, not `f` and `g`.

fredrik-bakke:
If this helps, they are currently working on formalizing elementary topoi over at 1lab. See in particular here for the subobject classifier: <https://github.com/plt-amy/1lab/blob/e2c7a58fcf534c0611dd8bdea7d054f1eb80bd28/src/Cat/Diagram/Subobject.lagda.md>, and note that they take a fibrational approach. Are you planning on taking a similar path? If so, that would be super neat! I was hoping to do some basic work on fibrations of categories at some point, but I don't currently know when/if I will take the time to do that.

EgbertRijke:
Hi @maybemabeline! How is this PR going?

fredrik-bakke:
That's too bad! Do get in touch if you want to coordinate on working towards formalizing this in the future, @maybemabeline. A little bit of topos theory wouldn't hurt the library :) I also did refactor pullbacks in types a bit in one of my recent PRs (#839), so that should give a better starting point now.

fredrik-bakke:
No pressure though, there's an endless supply of cool things to formalize. 😄

PR #809, opened 2023-09-29
Refactor library to use `λ where`

fredrik-bakke:
Also adds a new rule removing blank lines right before `where` keywords.


src/category-theory/category-of-functors.lagda.md
```diff
@@ -82,9 +82,10 @@ module _
         ( ( equiv-iso-functor-natural-isomorphism-Precategory C D F G) ∘e
           ( extensionality-functor-is-category-Precategory
               C D is-category-D F G))
-        ( λ { refl →
-              compute-iso-functor-natural-isomorphism-eq-Precategory
-                C D F G refl})
+        ( λ where
+          refl →
+            compute-iso-functor-natural-isomorphism-eq-Precategory
+              C D F G refl)
```
fredrik-bakke, line 88: Okay, so this is a little annoying. The `λ where` isn't strictly compatible with our indentation rules. Consider the code here, if I try to reduce the indentation level of the last two lines by one, or move `refl →` to the end of the line before, then the code does not type check anymore.

fredrik-bakke: Or if we indent `refl →` by one more level, it also doesn't type check.

fredrik-bakke:
I also remembered that we don't actually have a script focused on best practices for agda code, and I won't bother making one now just for this convention.

src/category-theory/category-of-functors.lagda.md
fredrik-bakke: Yeah, I get the comparison, but it's strange to me it behaves this way when normal lambdas don't.

fredrik-bakke: > It kind of makes sense if you think about it like defining an anonymous function with an arrow instead of `=`. If it had a name, we would write it as
> 
> ```agda
> f refl =
>   compute...
> ```
> 
> but I agree that it feels more like rationalizing a defect than an intuitive consequence of our style guide. Do you consider this a blocker for merging the PR?

Actually, I don't think this comparison is quite accurate. I think it would correspond more closely with

```agda
f refl =
    compute...
```
(second line is doubly indented)
or
```agda
f
refl =
  compute...
```

EgbertRijke: Would it be possible to format it as follows?:

```suggestion
        ( λ where
            refl →
            compute-iso-functor-natural-isomorphism-eq-Precategory
              C D F G refl)
```

fredrik-bakke: Nope

EgbertRijke: Perhaps we could just work towards merging this PR but leave the question about a formatting guideline open until we find a best practice by actually putting lambda-where into practice the best we can. At the moment I also don't know what the guideline would be.

EgbertRijke: Given that lambda-where is not without trouble with respect to formatting, we should perhaps still permit pattern-matching lambda as well.

fredrik-bakke: Okay, I will try and format it as I think looks best, and then flag this PR as ready for review

fredrik-bakke:
This one's ready now. At some point in the future, when there are fewer pull requests in the pipeline, I'd like to go through the repository to change most instances of `pair` with `_,_` using replace patterns like
- ``(?<!-)pair (?!of)([^`()\s]+)( |$)`` -> `$1 ,$2`

EgbertRijke:
> This one's ready now. At some point in the future, when there are fewer pull requests in the pipeline, I'd like to go through the repository to change most instances of `pair` with `_,_` using replace patterns like
> 
> * `` (?<!-)pair (?!of)([^`()\s]+)( |$) `` -> `$1 ,$2`

Aren't there some instances where `pair` is nicer than `_,_`?

fredrik-bakke:
I agree, when both arguments have to be placed on their own line, I think `pair` is nicer than `_,_`. But I thought I saw you recently say that we always preferred the latter? Either way, the pattern I presented only matches instances where `pair` and its first argument are on the same line. In this case I think `_,_` will generally be nicer. I don't feel too strongly about cases where we have
```agda
 x , 
 y
   z
   w
```
So if you want, I can restrict to only replacing instances where both arguments of `pair`  appear on the same line.

src/foundation/0-connected-types.lagda.md
```diff
@@ -117,7 +117,7 @@ is-trunc-map-ev-point-is-connected k {A} {B} a H K =
       ( universal-property-contr-is-contr star is-contr-unit B))
     ( is-trunc-map-precomp-Π-is-surjective k
       ( is-surjective-point-is-0-connected a H)
-      ( λ _ → pair B K))
+      ( λ _ → B , K))
```
fredrik-bakke, line 120:
```suggestion
      ( λ _ → (B , K)))
```

src/group-theory/generating-elements-groups.lagda.md
```diff
@@ -465,8 +467,9 @@ module _
         ( U x)
         ( U y)
         ( Id-Prop (set-Group G) (mul-Group G x y) (mul-Group G y x))
-        ( λ { (k , refl) (l , refl) →
-              commute-integer-powers-Group G k l refl})
+        ( λ where
+          (k , refl) (l , refl) →
+              commute-integer-powers-Group G k l refl)
```
fredrik-bakke, line 472:
```suggestion
            commute-integer-powers-Group G k l refl)
```

EgbertRijke:
I feel like we should still include some guidelines about what options we have to type-set lambdas. Eg we could mention that we always wrap the lambda-expression in parentheses, even if it is the last argument of a function and therefore not strictly required. Then we could mention that there are several lambda-styles available:
1. regular lambda
2. pattern-matching lambda without `{ }` as in `lambda (x , y) -> ...`
3. pattern-matching lambda with `{ }`
4. lambda-where
and we could mention that whenever lambdas 2-4 appear in an entry we mark it as abstract. We could also mention that while these fancy lambdas are permitted in our code, in many common situations it might be preferable to factor out that lambda-expression into a separate definition.

EgbertRijke:
@VojtechStep also mentioned some lambda-styles that we don't use in our library. I already forgot about them, but we could include those too in our style guide and say that we don't use them.

src/set-theory/cumulative-hierarchy.lagda.md
```diff
@@ -729,29 +730,31 @@ needed.
...
-                    ( H))))))})
+                            ( unit-trunc-Prop (a , refl))) ,
+                          ( refl))
+                      ( ∈-cumulative-hierarchy-mere-preimage H))
+                  ( λ H →
+                    mere-preimage-∈-cumulative-hierarchy
+                      ( map-universal-property-trunc-Prop
+                        ( ∃-Prop A _)
+                        ( λ where
+                          (z , K , refl) →
```
fredrik-bakke:
```suggestion
                          ( z , K , refl) →
```

scripts/blank_line_conventions.py
```diff
@@ -23,6 +22,10 @@
         output = re.sub(r'[ \t]+$', '', inputText, flags=re.MULTILINE)
         output = re.sub(r'\n(\s*\n){2,}', '\n\n', output)
 
+        # Remove blank lines before a `where`
```
EgbertRijke, line 25: Why should blank lines before `where` not be permitted?

fredrik-bakke: It seemed every instance in the library was a misformatting: https://github.com/UniMath/agda-unimath/pull/809/commits/f7cf33b976db6cb18d74ea93110c504f39e764e1

I can disable it if you want.

src/foundation/identity-types.lagda.md
```diff
@@ -103,8 +103,8 @@ module _
   pr2 equiv-concat-equiv =
     is-equiv-is-invertible
       equiv-concat
-      (λ { refl → refl})
-      (λ e → eq-htpy (λ y → eq-htpy-equiv (λ { refl → right-unit})))
+      (λ where refl → refl)
+      (λ e → eq-htpy (λ y → eq-htpy-equiv (λ where refl → right-unit)))
```
EgbertRijke, line 107: Do you want formatting suggestions in this PR?

```suggestion
      ( λ where refl → refl)
      ( λ e → eq-htpy (λ y → eq-htpy-equiv (λ where refl → right-unit)))
```

fredrik-bakke: Sure, go ahead!

VojtechStep:
> 2\. pattern-matching lambda without `{ }` as in `lambda (x , y) -> ...`

I don't think this is technically a pattern-matching lambda - it's only allowed for record types with eta equality (which, to be fair, covers many of our use cases, except path induction).

scripts/blank_line_conventions.py
EgbertRijke: I think it is ok. I saw that indeed it caught some unnecessary blank lines. Only the one in russell's paradox seemed reasonable, but it would also be reasonable without that blank line:)

EgbertRijke: So no need to disable.

src/foundation/identity-types.lagda.md
EgbertRijke: I saw some unnecessary parentheses somewhere before, but now I already forgot where they were:) So I don't have more formatting suggestions

EgbertRijke (approved):

EgbertRijke:
The conventions you use for `lambda-where` look very reasonable. They may not exactly fit uniformly in our indentation conventions, but I think we can stick with them if we explain it in the guidelines.

fredrik-bakke:
Am I correct that @VojtechStep intended to update the guidelines regarding pattern-matching lambdas, or did you want me to do it here?

EgbertRijke:
I don't know if Vojtech was planning to do that, but since this PR introduces lambda-where globally to the library, the natural PR to include some formatting guidelines seems to be this one.

fredrik-bakke:
Okay, that makes sense. And then I guess he will write the guidelines for the other proposed instance where he wanted to use pattern-matching lambdas.

fredrik-bakke:
> I feel like we should still include some guidelines about what options we have to type-set lambdas. Eg we could mention that we always wrap the lambda-expression in parentheses, even if it is the last argument of a function and therefore not strictly required. Then we could mention that there are several lambda-styles available:
> 
> 1. regular lambda
> 2. pattern-matching lambda without `{ }` as in `lambda (x , y) -> ...`
> 3. pattern-matching lambda with `{ }`
> 4. lambda-where
> 
> and we could mention that whenever lambdas 2-4 appear in an entry we mark it as abstract. We could also mention that while these fancy lambdas are permitted in our code, in many common situations it might be preferable to factor out that lambda-expression into a separate definition. And then we could also say that the formatting guidelines for these lambdas are not quite cleared up yet, just to be upfront.

I don't think we should have a blanket rule that any of these entail marking definitions as abstract. For instance, we don't mark identification concatenation as abstract even though it is defined by pattern matching. Also, it seems particularly counter-productive to have such a rule in the case of item 2.

VojtechStep:
Oh yeah, please do include the styleguide changes for this usage of lambdas here, I'll eventually open an analogous PR for lambdas replacing copattern matching where appropriate.

EgbertRijke:
> > I feel like we should still include some guidelines about what options we have to type-set lambdas. Eg we could mention that we always wrap the lambda-expression in parentheses, even if it is the last argument of a function and therefore not strictly required. Then we could mention that there are several lambda-styles available:
> > 
> > 1. regular lambda
> > 2. pattern-matching lambda without `{ }` as in `lambda (x , y) -> ...`
> > 3. pattern-matching lambda with `{ }`
> > 4. lambda-where
> > 
> > and we could mention that whenever lambdas 2-4 appear in an entry we mark it as abstract. We could also mention that while these fancy lambdas are permitted in our code, in many common situations it might be preferable to factor out that lambda-expression into a separate definition. And then we could also say that the formatting guidelines for these lambdas are not quite cleared up yet, just to be upfront.
> 
> I don't think we should have a blanket rule that any of these entail marking definitions as abstract. For instance, we don't mark identification concatenation as abstract even though it is defined by pattern matching. Also, it seems particularly counter-productive to have such a rule in the case of item 2.

I don't understand your example. We're discussing pattern-matching lambdas/lambda-wheres inside definitions. Concatenation of identification has neither of these.

EgbertRijke:
Note that this discussion is about when specific syntax is used, not about what they might be judgmentally equal to after the proof assistant compiles it.

fredrik-bakke:
> Note that this discussion is about when specific syntax is used, not about what they might be judgmentally equal to after the proof assistant compiles it.

Alright, whatever, I don't really want to fight you on this right now. Either way, we will soon enough be able to control unfolding much better with `opaque`. I'll implement the convention for items 3 and 4, but it should not apply to item 2.

EgbertRijke:
> > Note that this discussion is about when specific syntax is used, not about what they might be judgmentally equal to after the proof assistant compiles it.
> 
> Alright, whatever, I don't really want to fight you on this right now. Either way, we will soon enough be able to control unfolding much better with `opaque`. I'll implement the convention for items 3 and 4, but it should not apply to item 2.

I'm sorry, I didn't mean to argue. I was just really puzzled by your example, because it didn't involve a lambda expression.

fredrik-bakke:
That's fine. I see now how the phrasing of your convention restricts the cases where it applies. However, it still seems a bit artificial to me that if the formalizer is able to define their argument in terms of other definitions that use pattern matching, instead of writing a pattern match locally, then it can be declared nonabstract.

EgbertRijke:
I see now.

The reason why I would suggest making them abstract is that if a lambda-where appears inside a definition as one of the arguments of another function, then we don't have good control over its computational behavior. To me, this means that we shouldn't want to compute with these entries, and therefore we make them abstract because that makes the library faster. 

If computation is necessary for entries that now have a lambda-where in them, then I would suggest factoring out that lambda-where subexpression.

EgbertRijke:
On the other hand, it the pattern-matching happens at the top-level, then of course you should be able to compute with it. The point is that these lambda-where's appear at the level of a proper subexpression, and not at the top level because then we prefer regular pattern-matching.

fredrik-bakke:
Yep, your convention is making more sense to me as I'm thinking about it.

src/foundation-core/empty-types.lagda.md
```diff
@@ -101,6 +101,9 @@ abstract
 empty-Prop : Prop lzero
 pr1 empty-Prop = empty
 pr2 empty-Prop = is-prop-empty
+
+is-prop-is-empty : {l : Level} {A : UU l} → is-empty A → is-prop A
```
EgbertRijke, line 105: This is not really the right PR to do this sort of work in ;)

fredrik-bakke: I thought I needed this for a proof that previously computed since it was marked abstract. However, I got around that another way. Hence why it is there.

src/commutative-algebra/products-ideals-commutative-rings.lagda.md
```diff
@@ -170,37 +170,39 @@ module _
...
-            ( x))
-          ( λ where
-            ( s , t , refl) →
-              contains-product-product-ideal-Commutative-Ring A
-                ( ideal-subset-Commutative-Ring A S)
-                ( ideal-subset-Commutative-Ring A T)
-                ( pr1 s)
-                ( pr1 t)
-                ( contains-subset-ideal-subset-Commutative-Ring A S
+  abstract
```
EgbertRijke, line 173: Nice! Thank you for doing this! I thought I'd go through the library to make them abstract but I see that you found a way to quickly do it. Thanks so much!

src/foundation-core/empty-types.lagda.md
EgbertRijke: No worries, it can still be here. I just thought that I'd say something before you start fixing everything you see that needs fixing:) Because then this will become a long pull request again:)

fredrik-bakke: Thanks for keeping me on track ;)

EgbertRijke: This PR should be ready to merge once we have the formatting guideline:)

EgbertRijke:
Btw my github keeps warning me about the lambda in the title of this PR. Is it a problem to have unicode in PR titles and in commit messages? (I'm fine with it, but I don't know if it might lead to some technical issues.)

fredrik-bakke:
No idea, I changed it now.

EgbertRijke:
It looks like it didn't make a difference.
Screenshot 2023-10-09 at 00 46 32

Anyway, it is not important.

fredrik-bakke:
Oh, yeah, that's because my branch name uses a lambda

src/organic-chemistry/ethane.lagda.md
```diff
@@ -226,72 +226,69 @@ module _
       (inr star) (inr star) =
       star
 
-  abstract
```
EgbertRijke: I made an issue #821 for this, because actually this definition needs to be broken down into smaller pieces. That doesn't have to happen in this PR.

fredrik-bakke: Thanks!

src/finite-group-theory/orbits-permutations.lagda.md
```diff
@@ -463,12 +463,11 @@ module _
...
-                ( unit-trunc-Prop
-                  ( pair
-                    ( k2 +ℕ k1)
-                    ( (iterate-add-ℕ k2 k1 (map-equiv f) a) ∙
-                      ( ap (iterate k2 (map-equiv f)) p ∙ q))))}))
+          ( λ (k2 , q) →
+            ( unit-trunc-Prop
+              ( ( k2 +ℕ k1) ,
+                ( (iterate-add-ℕ k2 k1 (map-equiv f) a) ∙
+                  ( ap (iterate k2 (map-equiv f)) p ∙ q))))))
```
EgbertRijke, line 470:
```suggestion
            unit-trunc-Prop
              ( ( k2 +ℕ k1) ,
                ( ( iterate-add-ℕ k2 k1 (map-equiv f) a) ∙
                  ( ap (iterate k2 (map-equiv f)) p ∙ q)))))
```

fredrik-bakke:
Man, type-checking the whole library is getting quite slow

fredrik-bakke:
I changed the title back because I don't think it should cause trouble. If it does, then we need to address that another way in any case I think.

src/foundation/identity-types.lagda.md
```diff
@@ -97,14 +97,15 @@ module _
...
 
-  equiv-concat-equiv :
-    {x x' : A} → ((y : A) → (x ＝ y) ≃ (x' ＝ y)) ≃ (x' ＝ x)
-  pr1 (equiv-concat-equiv {x}) e = map-equiv (e x) refl
-  pr2 equiv-concat-equiv =
-    is-equiv-is-invertible
-      equiv-concat
-      (λ { refl → refl})
-      (λ e → eq-htpy (λ y → eq-htpy-equiv (λ { refl → right-unit})))
+  abstract
```
EgbertRijke, line 100: This entry shouldn't be abstract. 

Usually we would write four definitions for `inv`, `is-section-inv`, `is-retraction-inv` and `is-equiv-...` and then use those definitions as arguments for `is-equiv-is-invertible`. If we would have done that, then we wouldn't have any need for lambda-where or to make anything abstract.

fredrik-bakke: Nicely spotted! I'll fix it

src/foundation/isolated-points.lagda.md
```diff
@@ -240,20 +240,21 @@ is-set-isolated-point :
...
-    ( inclusion-isolated-point A)
-    ( const unit (isolated-point A) a)
-    ( is-emb-inclusion-isolated-point A)
-    ( is-emb-is-injective
-      ( is-set-isolated-point A)
-      ( λ { {star} {star} p → refl}))
-pr2 (pr2 (decidable-emb-isolated-point {l1} {A} a)) x =
-  is-decidable-prod is-decidable-unit (pr2 a x)
+abstract
+  decidable-emb-isolated-point :
```
EgbertRijke, line 244: This entry should also not be abstract. Instead, it should be broken up into separate definitions for each of its projections. The entry corresponding to `pr1 (pr2 ...)` can then be marked abstract.

src/foundation/type-arithmetic-unit-type.lagda.md
```diff
@@ -166,7 +166,7 @@ module _
 
   is-retraction-map-inv-left-unit-law-Π :
     ( map-inv-left-unit-law-Π ∘ map-left-unit-law-Π) ~ id
-  is-retraction-map-inv-left-unit-law-Π f = eq-htpy (λ { star → refl})
+  is-retraction-map-inv-left-unit-law-Π f = eq-htpy (λ star → refl)
```
EgbertRijke, line 169: I thought @VojtechStep said that it only worked with records, but here we see a pattern-matching lambda without `{ }` for a data type:)

fredrik-bakke:
```agda
record unit : UU lzero where
  instance constructor star
```

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -255,188 +255,190 @@ inv-precomp-vector-set-quotient (succ-ℕ n) A R X f (qa0 , qa) =
...
-          ( inv-precomp-set-quotient-prod-set-quotient
-            ( R (inr star))
-            ( all-sim-Equivalence-Relation n
-              ( tail-functional-vec n A)
-              ( λ x → R (inl x))) X f)
-          ( eq-pair-Σ refl
-            ( map-equiv-equiv-set-quotient-vector-quotient-map n _ _ a)) ∙
-        ( htpy-eq
+abstract
+  is-section-inv-precomp-vector-set-quotient :
```
EgbertRijke, line 259: This one also seems a bit strange to be marked as abstract, because again it defines an element of an iterated Sigma type and only one of its components is defined using lambda-where. So it feels like the definition should be split up. 

However, I don't know if there are natural names for the components in this case. This might make it more difficult to break this definition down. So I suggest we keep it abstract as it is, and if it ever gives us trouble we can unmark it abstract.

src/foundation/vectors-set-quotients.lagda.md
```diff
@@ -255,188 +255,190 @@ inv-precomp-vector-set-quotient (succ-ℕ n) A R X f (qa0 , qa) =
...
-      ( reflecting-map-quotient-vector-map n A R)
-      ( X)))
-pr1 (is-retraction-inv-precomp-vector-set-quotient n A R X) =
-  inv-precomp-vector-set-quotient n A R X
-pr2 (is-retraction-inv-precomp-vector-set-quotient zero-ℕ A R X) f =
-  eq-htpy (λ {(map-raise star) → refl})
-pr2 (is-retraction-inv-precomp-vector-set-quotient (succ-ℕ n) A R X) f =
-  ap (_∘ set-quotient-vector-prod-set-quotient)
-    is-inv-map-inv-equiv-f ∙ lemma-f
-  where
```
EgbertRijke, line 326: Holy moly there's a lot going on in this definition. It's best that I don't look at this right now:)

fredrik-bakke: Probably best not to look at the file on pi finite types too then 🤐

src/foundation/identity-types.lagda.md
fredrik-bakke: do you want what I defined as `map-concat-equiv` to be called `inv-equiv-concat` instead?

src/univalent-combinatorics/pi-finite-types.lagda.md
```diff
@@ -588,51 +587,46 @@ has-finite-connected-components-Σ-is-0-connected {A = A} {B} C H K =
...
-          ε : Id a x' → is-decidable (mere-eq (pair x y) (pair x' y'))
+          ε : Id a x' → is-decidable (mere-eq (x , y) (x' , y'))
           ε refl =
             is-decidable-equiv e
               ( is-decidable-type-trunc-Prop-is-finite
                 ( is-finite-Σ
                   ( pr2 H a a)
                   ( λ ω → is-finite-is-decidable-Prop (P ω) (d ω))))
-
             where
```
EgbertRijke, line 630: Yikes they're nested 6 deep!

I wrote this code before the agda-unimath project was started:)

src/foundation/type-arithmetic-unit-type.lagda.md
EgbertRijke: Oh! I'm wrong:)

src/foundation/isolated-points.lagda.md
fredrik-bakke: I'm factoring this out, but actually, I don't think `is-emb-inclusion-isolated-point` should be marked as abstract, since it is pattern matching on the record type `unit`.

fredrik-bakke: wait, that name is already taken. What do we generally call maps from the unit type representing terms?

EgbertRijke:
> Man, type-checking the whole library is getting quite slow

Strange! Do you know why it is getting slow?

fredrik-bakke:
IDK, my guess is a large code base. Maybe it would be a good idea to keep a log of type-checking time for the whole library after each PR to keep a watch on it

src/foundation/identity-types.lagda.md
EgbertRijke: Let's see...

Since the name of the equivalence is `equiv-concat-equiv`, that makes the underlying map `map-equiv-concat-equiv`, its inverse `map-inv-equiv-concat-equiv`, its section `is-section-map-inv-equiv-concat-equiv`, its retraction `is-retraction-map-inv-equiv-concat-equiv`, and the proof that it is an equivalence `is-equiv-map-equiv-concat-equiv`.

This is the uninspired approach to naming... I don't know if it is the best, but it will do and it is systematic.

EgbertRijke:
> IDK, my guess is a large code base. Maybe it would be a good idea to keep a log of type-checking time for the whole library after each PR to keep a watch on it

Do you mean that it is generally taking a long time to compile the library, or that something in this PR happened that makes it take longer to compile?

fredrik-bakke:
Hard for me to say if it's gotten slower with this PR. It may just be that I'm type-checking a lot while working on this PR, hence noticing the slow

src/foundation/isolated-points.lagda.md
EgbertRijke: `point`

See here:
https://unimath.github.io/agda-unimath/foundation.unit-type.html#points-as-maps-out-of-the-unit-type

EgbertRijke: > I'm factoring this out, but actually, I don't think `is-emb-inclusion-isolated-point` should be marked as abstract, since it is pattern matching on the record type `unit`.

Proofs that a map is an embedding sound very reasonable candidates to be marked as abstract to me. Why should we allow Agda to try to compute with those gnarly proofs?

fredrik-bakke: Thanks! Then presumably `isolated-point` should be renamed `isolated-term` to avoid a naming conflict. Currently, the name `point-isolated-point` is already taken

fredrik-bakke: or `isolated-element`

EgbertRijke: `isolated-element` would be better.

We should only speak of terms when we are speaking about syntax.

src/foundation/vectors-set-quotients.lagda.md
fredrik-bakke: fixed it

fredrik-bakke:
Let me know what you think of the new convention

CODINGSTYLE.md
```diff
@@ -128,6 +128,47 @@ strategic endeavour to ensure the longevity, vitality, and success of the
     a = construction-of-a
   ```
 
+- **Lambda expressions**: We always wrap lambda expressions in parentheses, even
+  if it is the last argument of a function and thus isn't strictly required to
+  be parenthesized.
```
EgbertRijke, line 133: I think we might have some cases where the lambda-expression is at top level, and then we don't parenthesize it.

```suggestion
- **Lambda expressions**: When a lambda expression appears as the argument of a function, we always wrap it in parentheses. We do this even if it is the last argument of a function, and thus isn't strictly required to
  be parenthesized. Note that in some rare cases, a lambda expression might appear at the top level. In this case we don't require the lambda expression to be parenthesized.
```

EgbertRijke:
> Let me know what you think of the new convention

Beautiful!

src/foundation/isolated-elements.lagda.md
```diff
@@ -0,0 +1,392 @@
...
+  where
+
+  element-isolated-element : X
+  element-isolated-element = pr1 x
+
+  is-isolated-isolated-element : is-isolated element-isolated-element
+  is-isolated-isolated-element = pr2 x
+```
+
+### Complements of isolated points
```
EgbertRijke, line 72:
```suggestion
### Complements of isolated elements
```

src/foundation/isolated-elements.lagda.md
```diff
@@ -0,0 +1,392 @@
...
+```agda
+complement-isolated-element :
+  {l1 : Level} (X : UU l1) → isolated-element X → UU l1
+complement-isolated-element X x =
+  Σ X (λ y → ¬ (element-isolated-element x ＝ y))
+```
+
+## Properties
+
+### A element is decidable if and only if the constant map pointing at it is decidable
```
EgbertRijke, line 83:
```suggestion
### An element is isolated if and only if the constant map pointing at it is decidable
```

src/foundation/identity-types.lagda.md
fredrik-bakke: doesn't it make sense to call `map-equiv-concat-equiv` just `map-concat-equiv` though? especially in light of the new naming conventions

EgbertRijke: I don't think so in this case, because `concat-equiv` is a completely different equivalence.

EgbertRijke: The name `equiv-concat-equiv` is kind of inaccurate, but it's what we have at the moment and I don't have a better suggestion. So it's unfortunate that we have both `equiv-concat-equiv` and `concat-equiv`, but given that we have both of them we need to disambiguate `equiv-concat-equiv` throughout.

fredrik-bakke: Ah, right

fredrik-bakke: Okay, another stupid question, since `equiv-concat` is the inverse, can't we call the section proof `is-section-equiv-concat` and so on?

EgbertRijke: We also have `equiv-concat`?

Lol

I feel like the equivalence `equiv-concat-equiv` should be called something like `compute-dependent-product-of-equivalences-identity-types` of something along those lines

EgbertRijke: > Okay, another stupid question, since `equiv-concat` is the inverse, can't we call the section proof `is-section-equiv-concat` and so on?

Yeah, I think you're right

fredrik-bakke: Lol! I'll just use `is-section-equiv-concat` for now, and then we can think about names for `equiv-concat-equiv` at a time when the sun is up.

PR #813, opened 2023-10-06
Torsion-free groups

EgbertRijke:
This PR adds the notion of torsion element of a group and and the notion of torsion-free group to the library. Furthermore, it adds a short discussion on the notion of elements of finite order, which looks too strong from a constructive point of view.


src/group-theory/elements-of-finite-order-groups.lagda.md
```diff
@@ -0,0 +1,59 @@
...
+An element `x` of a [group](group-theory.groups.md) `G` is said to be of
+**finite order** if the [subgroup](group-theory.subgroups.md) `order x` of `ℤ`
+is [generated by](group-theory.subgroups-generated-by-elements-groups.md) a
+[nonzero integer](elementary-number-theory.nonzero-integers.md).
+
+Note that the condition of being of finite order is slightly stronger than the
+condition of being a [torsion element](group-theory.torsion-elements-groups.md).
+The latter condition merely asserts that there
+[exists](foundation.existential-quantification.md) a nonzero integer in the
+ideal `order x`.
```
fredrik-bakke, line 36: Depending on what is most appropriate, either
- link to ideal
- change wording to subgroup

src/group-theory/torsion-elements-groups.lagda.md
```diff
@@ -0,0 +1,104 @@
...
+```text
+  ∃ (k : nonzero-ℤ), xᵏ ＝ 1.
+```
+
+Note that the condition of being a torsion element is slightly weaker than the
+condition of being of
+[finite order](group-theory.elements-of-finite-order-groups.md). The condition
+of being a torsion element holds
+[if and only if](foundation.logical-equivalences.md) the
+[subgroup](group-theory.subgroups.md) `order x` of `ℤ` contains a nonzero
```
fredrik-bakke, line 40:
```suggestion
[subgroup](group-theory.subgroups.md) `order x` of `ℤ` contains a [nonzero](elementary-number-theory.nonzero-integers.md)
```

src/group-theory/torsion-free-groups.lagda.md
```diff
@@ -0,0 +1,130 @@
...
+element of finite [order](group-theory.orders-of-elements-groups.md) is the
+identity element. In other words, torsion-free groups are groups in which the
+condition
+
+```text
+  ∀ (k : nonzero-ℤ), xᵏ ＝ 1 → x ＝ 1
+```
+
+holds for all elements `x : G`. This condition can be formulated in several
+equivalent ways:
```
fredrik-bakke, line 42:
```suggestion
[equivalent](foundation.logical-equivalences.md) ways:
```

fredrik-bakke (approved):
Looks good! Since you explicitly mention it I assume it is thought out why `is-nonzero-N` is defined elsewhere.

EgbertRijke:
> Looks good! Since you explicitly mention it I assume it is thought out why `is-nonzero-N` is defined elsewhere.

I think it is because some very elementary stuff about natural numbers is done with the notion of nonzero natural number. My other reason was that I didn't really feel like testing it out to refactor it, so I thought I'd just mention it.

src/group-theory/elements-of-finite-order-groups.lagda.md
EgbertRijke: I am changing the wording.

fredrik-bakke:
That's okay. Given the new dependency of telescopes on natural numbers, I've actually been itching to try and move some of the dependencies in `natural-numbers`, but that's for another PR.

EgbertRijke:
> That's okay. Given the new dependency of telescopes on natural numbers, I've actually been itching to try and move some of the dependencies in `natural-numbers`, but that's for another PR.

Do you mean to put some stuff from other files into `natural-numbers`?

fredrik-bakke:
No, I mean the other way around. When I was playing around with telescopes, I frequently encountered cyclic dependencies because of all the things that are imported in `elementary-number-theory.natural-numbers`

EgbertRijke:
> No, I mean the other way around. When I was playing around with telescopes, I frequently encountered cyclic dependencies because of all the things that are imported in `elementary-number-theory.natural-numbers`

That sounds good! Cyclic dependencies are so annoying!

EgbertRijke:
TBH I find the cyclic dependencies quite dumb. A proof assistant should just test for _entries_ whether they create cyclic dependencies, not for files. Then we would truly be able to get rid of foundation-core.

EgbertRijke:
I addressed all your comments, so it is ready

PR #701, opened 2023-08-21
Modal type theory

fredrik-bakke:
### Additions
- Module on modal type theory, including
  - Define the flat modality
  - Define the sharp modality
  - Start work on their adjunction
  - Some basic definitions about codiscrete types and the flat modal operator
- Some additions on higher modalities
  - Introduce (strong) subuniverse induction as a relaxation of modal subuniverse induction
  - Show we can induct on the identity types with a higher modality without appealing to univalence using strong subuniverse induction
  - Start on functorial properties of higher modalities
  - Multivariable sections is a definition of sections of multivariable maps that make it possible to circumvent applying function extensionality in many cases.
  
(If we at some point in the future want a module on simplicial homotopy type theory, then it would be fruitful to develop the module on modal type theory further :))
 
### A regression
It has become clear that the current formulation of higher modalities is not the most practical one, as it needs to assume everything is sufficiently contained within a single universe. In addition, the small relaxation to locally small types seems somewhat contentless and makes the formalizations more cumbersome. I have started a new line of formalization to fix these problems in #890.

PR #829, opened 2023-10-10
Clean up the commutators of elements file

EgbertRijke:


EgbertRijke:
I made some changes that require review:
- I changed the filename
- I made a new section `## External links` because the file contained an external link that I didn't just want to remove.

src/group-theory/commutators-of-elements-groups.lagda.md
```diff
@@ -68,3 +66,7 @@ module _
   inv-commutator-Group x y =
     inv-right-div-Group G (mul-Group G x y) (mul-Group G y x)
 ```
+
+## External links
```
VojtechStep, line 70: We have a `## References` header in other files, could you elaborate why this is different?

EgbertRijke: References are for citations of articles. When you cite an article, then you do that with [n] in the text where you cite it, and then you record what [n] refers to under references. Then we refer to it by the title, author, journal, publisher, year, and all that stuff. Just like in an article. 

This is just an external link, and a citation in the above style seems inappropriate to me. As far as I could see, it is just a pointer to another online source on the same topic. To me, references carry more weight than just some external links. But of course this is up for debate, and maybe @fredrik-bakke also thinks it should just be called references.

fredrik-bakke: I don't think our common patterns should restrict what we can call our sections, so I'm completely fine with `External links`

fredrik-bakke: But references also don't have to be extremely formal. In #701, I use that section to link to formalizations I used as a starting point for instance

src/group-theory/commutators-of-elements-groups.lagda.md
```diff
@@ -18,16 +18,14 @@ open import group-theory.groups
 
 ## Idea
 
-A commutator gives an indication of the extent to which a group multiplication
-fails to be commutative.
-
-The commutator of two elements, g and h, of a group G, is the element
+The **commutator** of two elements `g` and `h` of a
+[group](group-theory.groups.md) `G` is defined to be the element
 `[g, h] = (gh)(hg)⁻¹`.
```
fredrik-bakke, line 24: I think the informal motivation could stay in a little rephrased form. Something like

```
The commutator can be seen as a correction factor that measures how much `g` and `h` fail to commute.
```

src/group-theory/commutators-of-elements-groups.lagda.md
```diff
@@ -68,3 +66,7 @@ module _
   inv-commutator-Group x y =
     inv-right-div-Group G (mul-Group G x y) (mul-Group G y x)
 ```
+
+## External links
+
+- [Wikipedia](https://en.wikipedia.org/wiki/Commutator#Group_theory)
```
fredrik-bakke, line 72: Should we include the title of the wikipedia article? It seems kind of nondescript in this form

fredrik-bakke (approved):

src/group-theory/commutators-of-elements-groups.lagda.md
```diff
@@ -1,7 +1,7 @@
 # Commutators of elements in groups
```
fredrik-bakke, line 1: Also, the file name is good. Don't want this to be confused with the commutator groups

src/group-theory/commutators-of-elements-groups.lagda.md
EgbertRijke: Yep, I will do that

src/group-theory/commutators-of-elements-groups.lagda.md
EgbertRijke: Note that Wikipedia also distinguishes between references, external links, and further reading.

fredrik-bakke:
Looks good now. Merge when you're ready :)

PR #830, opened 2023-10-11
Participating in the MaGE project

EgbertRijke:
@andrejbauer, @naussicaa and @katjabercic are working on a unified metadatabase for mathematical concepts that gathers links to implementations and expositions of mathematical concepts across the web. By adding links to the MaGE project with wikidata identifiers we will be participating in this project, which I think is a good idea.

In this short PR I started making a few External links sections, in which those and other links are recorded.

This is perhaps not yet the most efficient or cleanest way to go about this, but it is an initial attempt. @VojtechStep do you have any ideas from a website implementers perspective what is a decent way to go about this?

### DIFF
--- a/src/graph-theory/acyclic-undirected-graphs.lagda.md
+++ b/src/graph-theory/acyclic-undirected-graphs.lagda.md
@@ -18,21 +18,24 @@ open import graph-theory.undirected-graphs
 
 ## Idea
 
-An **acyclic** undirected graph is an undirected graph of which the geometric
-realization is contractible.
+An **acyclic** undirected graph is an
+[undirected graph](graph-theory.undirected-graphs.md) of which the
+[geometric realization](graph-theory.geometric-realizations-undirected-graphs.md)
+is [contractible](foundation-core.contractible-types.md).
 
 The notion of acyclic graphs is a generalization of the notion of
 [undirected trees](trees.undirected-trees.md). Note that in this library, an
-undirected tree is an undirected graph in which the type of trails between any
-two points is contractible. The type of nodes of such undirected trees
-consequently has decidable equality. On the other hand, there are acyclic
-undirected graphs that are not undirected trees in this sense. One way to obtain
-them is via [acyclic types](synthetic-homotopy-theory.acyclic-types.md), which
-are types of which the
-[suspension](synthetic-homotopy-theory.suspensions-of-types.md) is contractible.
-The undirected suspension diagram of such types is an acyclic graph.
-Furthermore, any [directed tree](trees.directed-trees.md) induces an acyclic
-undirected graph by forgetting the directions of the edges.
+undirected tree is an undirected graph in which the type of
+[trails](graph-theory.trails-undirected-graphs.md) between any two points is
+contractible. The type of nodes of such undirected trees consequently has
+[decidable equality](foundation.decidable-equality.md). On the other hand, there
+are acyclic undirected graphs that are not undirected trees in this sense. One
+way to obtain them is via
+[acyclic types](synthetic-homotopy-theory.acyclic-types.md), which are types of
+which the [suspension](synthetic-homotopy-theory.suspensions-of-types.md) is
+contractible. The undirected suspension diagram of such types is an acyclic
+graph. Furthermore, any [directed tree](trees.directed-trees.md) induces an
+acyclic undirected graph by forgetting the directions of the edges.
 
 ## Definition
 
@@ -57,3 +60,13 @@ is-acyclic-Undirected-Graph l G =
 ### Table of files related to cyclic types, groups, and rings
 
 {{#include tables/cyclic-types.md}}
+
+## External links
+
+- <a href="http://ul-fmf.github.io/MaGE/Q3115453" id="acyclic-undirected-graph">Acyclic
+  undirected graphs</a> at the Math Grand Exchange.
+- <a href="https://ncatlab.org/nlab/show/tree">Trees</a> at the nlab.
+- <a href="https://en.wikipedia.org/wiki/Tree_(graph_theory)#Forest">Forests</a>
+  at Wikipedia
+- <a href="https://mathworld.wolfram.com/AcyclicGraph.html">Acyclic graphs</a>
+  at Wolfram Mathworld.
--- a/src/graph-theory/circuits-undirected-graphs.lagda.md
+++ b/src/graph-theory/circuits-undirected-graphs.lagda.md
@@ -21,9 +21,12 @@ open import graph-theory.undirected-graphs
 
 ## Idea
 
-A circuit in an undirected graph `G` consists of a `k`-gon `H` equipped with a
-totally faithful morphism of graphs from `H` to `G`. In other words, a circuit
-is a closed walk with no repeated edges.
+A **circuit** in an [undirected graph](graph-theory.undirected-graphs.md) `G`
+consists of a [`k`-gon](graph-theory.polygons.md) `H` equipped with a
+[totally faithful](graph-theory.totally-faithful-morphisms-undirected-graphs.md)
+[morphism](graph-theory.morphisms-undirected-graphs.md) of undirected graphs
+from `H` to `G`. In other words, a circuit is a closed walk with no repeated
+edges.
 
 ## Definition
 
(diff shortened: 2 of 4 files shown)


EgbertRijke:
They’re still building the website… 🙂

fredrik-bakke:
Do they provide some sort of api so we can fetch and update these links automatically?

VojtechStep:
If the links are stable, then we could write a pretty simple preprocessor to automatically add them, based on some markup in the source files.

EgbertRijke:
I don't think there's any way that we can fetch the wikidata identifiers automatically, because we have a much narrower and more disambiguated notion of concept than wikipedia.

Wikidata has some "notability" constraint, which is a strange constraint to have from a database-of-mathematical-concepts perspective. I hope that the MaGE project will just accept mathematical concepts without judgment of whether they are notable in any sense.

Once we link all the pages, it should be fairly easy to update all the links at once. @katjabercic promised some automated tool for that. She also promised to explain a little bit more about the project here:)

EgbertRijke:
For example we already ran into a little problem because "circuits" and "cycles" are distinct concepts in graph theory, but wikidata treats them as the same.

EgbertRijke:
> If the links are stable, then we could write a pretty simple preprocessor to automatically add them, based on some markup in the source files.

What is your idea of automatically linking our concepts to the analogous concepts in Wikipedia/Mathworld/nlab? Are you sure this can be automated?

VojtechStep:
By "based on some markup in the file" I meant we could have something like `{{#wikidata ID}}` in the file; nothing more sophisticated than that

EgbertRijke:
That's a very nice idea! So the proposal is to tag all our files with `{{#wikidata ID}}` and possibly also with `{{#mage ID}}`?

andrejbauer:
Disclaimer, I am just a spiritual leader here, all the work is being done by @katjabercic and her posse.

andrejbauer:
You should not tag files but rather specific entries within files, and if possible your scripts should generate anchors `<a name='...'>` which point to them. This is important because various harvesting tools will work much better if the information provided by the tags has good granularity.

EgbertRijke:
> You should not tag files but rather specific entries within files, and if possible your scripts should generate anchors `<a name='...'>` which point to them. This is important because various harvesting tools will work much better if the information provided by the tags has good granularity.

Would it be a good idea to do this for *all* our entries, with the anchor id being the entry identifier? That could easily be automated I think.

src/graph-theory/acyclic-undirected-graphs.lagda.md
```diff
@@ -57,3 +60,13 @@ is-acyclic-Undirected-Graph l G =
 ### Table of files related to cyclic types, groups, and rings
 
 {{#include tables/cyclic-types.md}}
+
+## External links
+
+- <a href="http://ul-fmf.github.io/MaGE/Q3115453" id="acyclic-undirected-graph">Acyclic
+  undirected graphs</a> at the Math Grand Exchange.
+- <a href="https://ncatlab.org/nlab/show/tree">Trees</a> at the nlab.
```
andrejbauer, line 68: The whole point of linking to WikiData is that manua links to nlab etc. become unecessary. Once you have the `Qxxxxx` link, you can ask MaGe to give you external links to nlab etc., or don't even bother because clicking on the MaGE link will lead to the "Grand Exchange" anyhow.

EgbertRijke:
> Maybe I'm misunderstanding something, but from the form of this PR I thought the idea was to link from agda-unimath to MaGE - do we also want to add anchors so that people can link to agda-unimath (which I support, I just wasn't sure if it's in scope of this effort)?

I think so! 

This PR was just an initial attempt, but now it seems that anchors are the way to go.

andrejbauer:
> Maybe I'm misunderstanding something, but from the form of this PR I thought the idea was to link from agda-unimath to MaGE - do we also want to add anchors so that people can link to agda-unimath (which I support, I just wasn't sure if it's in scope of this effort)?

Yes, you want MaGE to be able to link back to you, and when you become "notable" also WikiData.

andrejbauer:
> > Disclaimer, I am just a spiritual leader here, all the work is being done by @katjabercic and her posse.
> 
> Perhaps I can convince the spiritual leader that the MaGE IDs should not have the arbitrary "notability" constraint that wikidata has. On literally the first try to tag some of our files with wikidata IDs we noticed that wikidata conflates some distinct concepts that should really be separated out.

Well, there was a "committee" in Dagstuhl which looked at a bunch of things and discovered that WikiData was the best current solution. Perhaps we can enrich WikiData somehow? For instance, MaGE could have IDs for stuff that does not exist in WikiData (yet).

src/graph-theory/acyclic-undirected-graphs.lagda.md
andrejbauer: On second thought, if you do provide external links, that *helps* harversters associate those external links with tags. One of the main problem at present is that there are very few annotations between sources.

EgbertRijke:
> > > Disclaimer, I am just a spiritual leader here, all the work is being done by @katjabercic and her posse.
> > 
> > 
> > Perhaps I can convince the spiritual leader that the MaGE IDs should not have the arbitrary "notability" constraint that wikidata has. On literally the first try to tag some of our files with wikidata IDs we noticed that wikidata conflates some distinct concepts that should really be separated out.
> 
> Well, there was a "committee" in Dagstuhl which looked at a bunch of things and discovered that WikiData was the best current solution. Perhaps we can enrich WikiData somehow? For instance, MaGE could have IDs for stuff that does not exist in WikiData (yet).

My proposal would be that MaGE would work with its own identifiers, link them to wikidata where appropriate, but it should be much easier to add concepts to MaGE than it is to add concepts to wikidata.

Wikidata may currently be "the best" solution but it is very tailored to wikipedia's purpose, but consequently it will put variations of the same concept under the same identifier. This is understandable from the point of view of writing wikipedia articles, which puts those different variations into context, but it is wrong as an indexing mechanism for mathematical concepts. As in my example: cycles of graphs are not the same as circuits of graphs. And a thing that worries me more is that in mathematics based on type theory, we will necessarily work with different concepts that don't even exist in set theoretic mathematics. Therefore I hope it will be easier to disamibiguate MaGE's notion of concept than it is with Wikidata's notion of concept.

src/graph-theory/acyclic-undirected-graphs.lagda.md
EgbertRijke: One reason to put in manually selected external links is that several of our pages link to one "concept" in MaGE/Wikidata because those databases don't disambiguate sufficiently for our purposes. The same might happen for other sources. So in principle MaGE will be a many-to-many multivariable correspondence, and providing a manually selected link to another source can be helpful to the reader.

EgbertRijke:
There might also be several equivalent variations of a concept, which we distinguish in agda-unimath. For example, a cyclic set might be equivalently described as a directed polygon, a torsor for a cyclic group, a connected set bundle over the circle, and so on... All of these are "the same" concept, but talked about differently. We even have two prominent notions of group in our library that are equivalent, but it is a somewhat deep theorem that they are equivalent: abstract groups and concrete groups.

All of such nuances are swept under the rug in Wikidata, which is why I am quite instantly unhappy about how Wikidata organizes its concepts. Wikidata identifiers work for Wikipedia's purpose, but as far as I understand the MaGE project these identifiers can't possibly be good enough for MaGE's purpose (and that is still consistent with Wikidata being "the best" alternative).

katjabercic:
An attempt to resolve some of the things that came up. The MaGE prototype will be presented at a cyberinfrastructure for math workshop in December (hence the rush), as an attempt to get more people and libraries involved.

The plan for the most basic MaGE prototype is to:
* host a page for each math concept, with information harvested from WikiData and agda-unimath (including linking to the agda-unimath docs),
* have a basic concept search field (from which users can then make their way to, say, the agda-unimath docs).

The latter will involve harvesting the annotated definitions from agda-unimath and maintaining an index (which can be rebuilt as necessary).

It is very likely that we will end up with MaGE identifiers. Long term, these would be based on the currently non-existent mathematics ontology. My current best guess is that they are not needed for the minimal viable prototype, and I would like to avoid going down that way until it is necessary, as it adds further complications (see also the next paragraph, as it's related). The idea is to use WikiData identifiers as a proxy.

It would be ***extremely unwise*** to start building a mathematics ontology with MaGE that would accurately encompass the agda-unimath library while stretching to other libraries. Many projects have tried and failed; it is a well known problem. We are aiming to get something feasible despite the lack of ontology.

Issues with WikiData:
* notability constraint (on concepts and linked libraries),
* expectation of stable links,
* we do not have any control over what they do, which is not great while we're figuring out what we want,
* not enough concepts (Egbert estimates that about 60% are missing).
I will get in touch with people that have more experience with WikiData, in particular about contributing back.

EgbertRijke:
The problems arise exactly when you are trying to be very careful about what concepts to index. Let the mathematical communities decide what the useful mathematical concepts are! I would suggest a very liberal approach.

katjabercic:
The reason for Mage is that Wikidata will not link back to your library, as I mentioned before. It will also not link back to other libraries.

naussicaa:
One could add provenance metadata and link back to the libraries of
integrated items. The data model doesn't prevent this AFAIK

Le mer. 11 oct. 2023, 17:28, Katja Berčič ***@***.***> a
écrit :

PR #832, opened 2023-10-12
Creating internal and external links in Graph Theory

EgbertRijke:
This PR creates internal links in the `## Idea` sections and new `## External links` sections to all files in the `graph-theory` folder.

In particular, it already collects links to `wikidata`, which we will eventually need for the MaGE project, but it does not introduce nonexisting links to the MaGE project website.

### DIFF
--- a/src/graph-theory/acyclic-undirected-graphs.lagda.md
+++ b/src/graph-theory/acyclic-undirected-graphs.lagda.md
@@ -63,8 +63,6 @@ is-acyclic-Undirected-Graph l G =
 
 ## External links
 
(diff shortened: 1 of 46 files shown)


fredrik-bakke:
Nice work! I don't have time to review this at the moment, but from a quick view it looks good, so if you're confident with the PR I think it can be merged right away.

EgbertRijke:
Ok, it just adds some links and otherwise it is quite a trivial pull request. So I'll merge it and then we can fix it later if we discover faulty links or other issues.

EgbertRijke:
Ay... that came in a split second too late. I'll fix it. Let me know if you find anything else

PR #833, opened 2023-10-12
Define suspension prespectra and maps of prespectra

fredrik-bakke:
Just a small PR giving some definitions.


src/synthetic-homotopy-theory/functoriality-loop-spaces.lagda.md
```diff
@@ -102,27 +104,24 @@ module _
 ```agda
 module _
   {l1 l2 : Level} {A : Pointed-Type l1} {B : Pointed-Type l2}
-  (f : A →∗ B) (t : is-emb (map-pointed-map f))
+  (f : A →∗ B) (is-emb-f : is-emb (map-pointed-map f))
```
EgbertRijke, line 107: I'm still not a big fan of long variable names, hehe, but I won't ask you to revert this change.

More seriously, the structure of these assumptions suggests that we should really have a notion of pointed embedding in the library

src/synthetic-homotopy-theory/iterated-suspensions-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,37 @@
...
+## Definitions
+
+### The iterated suspension of a pointed type
+
+```agda
+iterated-suspension-Pointed-Type :
+  {l : Level} (n : ℕ) → Pointed-Type l → Pointed-Type l
+iterated-suspension-Pointed-Type 0 X = X
+iterated-suspension-Pointed-Type (succ-ℕ n) X =
+  suspension-Pointed-Type (iterated-suspension-Pointed-Type n X)
```
EgbertRijke, line 36: It would be nicer if we define this in terms of the official definition of iterating functions, because then it points the the file that has lemmas about iterations:

https://unimath.github.io/agda-unimath/foundation.iterating-functions.html

src/synthetic-homotopy-theory/iterated-suspensions-of-pointed-types.lagda.md
```diff
@@ -0,0 +1,37 @@
...
+open import synthetic-homotopy-theory.suspensions-of-pointed-types
+```
+
+</details>
+
+## Idea
+
+Given a [pointed type](structured-types.pointed-types.md) `X` and a
+[natural number](elementary-number-theory.natural-numbers.md) `n`, we can form
+the `n`-**iterated suspension** of `X`.
```
EgbertRijke, line 25: This should be ``**`n`-th iterated suspension``, in particular with the entire phrase in `**`.

src/synthetic-homotopy-theory/loop-spaces.lagda.md
```diff
@@ -23,9 +23,11 @@ open import structured-types.wild-quasigroups
 
 ## Idea
 
-The loop space of a pointed type `A` is the pointed type of self-identifications
-of the base point of `A`. The loop space comes equipped with a group structure
-induced by the groupoidal structure on identifications.
+The **loop space** of a [pointed type](structured-types.pointed-types.md) `A` is
+the pointed type of self-[identifications](foundation-core.identity-types.md) of
+the base point of `A`. The loop space comes equipped with a
+[group](group-theory.groups.md) structure induced by the
```
EgbertRijke, line 29: This is somewhat inaccurate, because the loop space is not necessarily a set. Perhaps it would be better to say "group-like structure"

src/synthetic-homotopy-theory/maps-of-prespectra.lagda.md
```diff
@@ -0,0 +1,81 @@
...
+open import synthetic-homotopy-theory.functoriality-loop-spaces
+open import synthetic-homotopy-theory.loop-spaces
+open import synthetic-homotopy-theory.prespectra
+```
+
+</details>
+
+## Idea
+
+A **map of prespectra** `f : A → B` is a [sequence](foundation.sequences.md) of
```
EgbertRijke, line 32: Hm... perhaps the link to sequences here is somewhat misleading, because a map of prespectra is really a dependent sequence, i.e., it is not a sequence of the kind being linked to.

src/synthetic-homotopy-theory/maps-of-prespectra.lagda.md
```diff
@@ -0,0 +1,81 @@
...
+  UU (l1 ⊔ l2)
+map-Prespectrum A B =
+  Σ ( (n : ℕ) →
+      pointed-type-Prespectrum A n →∗ pointed-type-Prespectrum B n)
+    ( λ f → (n : ℕ) → coherence-map-Prespectrum n A B f)
+```
+
+## References
+
+- J. P. May, _A Concise Course in Algebraic Topology_, 1999
```
EgbertRijke, line 80: Should there be a comma between the year and the link to the pdf?

src/synthetic-homotopy-theory/spectra.lagda.md
```diff
@@ -10,25 +10,105 @@ module synthetic-homotopy-theory.spectra where
...
 ```
 
 </details>
 
 ## Idea
 
-A spectrum is an infinite sequence of pointed types `A` such that
-`A n ≃∗ Ω (A (n+1))` for each `n : ℕ`.
+A **spectrum** is a [sequence](foundation.sequences.md) of
+[pointed types](structured-types.pointed-types.md) `A` such that
```
EgbertRijke, line 32: Instead of "such that" wouldn't it be better to say "equipped with an equivalence"?

src/synthetic-homotopy-theory/sphere-spectrum.lagda.md
```diff
@@ -0,0 +1,46 @@
...
+
+</details>
+
+## Idea
+
+The [spheres](synthetic-homotopy-theory.spheres.md) `Sⁿ` define a
+[spectrum](synthetic-homotopy-theory.spectra.md)
+
+```text
+  Sⁿ ≃∗ ΩSⁿ⁺¹
```
EgbertRijke, line 27: Here you make it look like that the loop space of the `n+1`-st sphere is the `n`-sphere, which is not true.

src/synthetic-homotopy-theory/loop-spaces.lagda.md
fredrik-bakke: Ah, thanks, I should have caught that.

src/synthetic-homotopy-theory/maps-of-prespectra.lagda.md
fredrik-bakke: ah, yes, thanks

src/synthetic-homotopy-theory/maps-of-prespectra.lagda.md
fredrik-bakke: AFAIK we still don't have a rigid format for references

fredrik-bakke: note that the link is parenthesized

src/synthetic-homotopy-theory/maps-of-prespectra.lagda.md
fredrik-bakke: I resolved this by defining dependent sequences and pointing to them instead.

EgbertRijke: Excellent!

src/synthetic-homotopy-theory/sphere-spectrum.lagda.md
fredrik-bakke: Ah, yeah, this can't be true since $S^0$ is not the integers. But now I am confused about what the correct statement is.

EgbertRijke: The sphere spectrum is the suspension spectrum of the 0-sphere, aka the spectrification of the sphere-prespectrum, which does consist of the spheres.

fredrik-bakke: Oh, I thought there was a more straightforward statement that could be made. Then I am still confused about spectra. Then does the final argument in the file even make sense?

fredrik-bakke: Thanks for the help by the way!

fredrik-bakke: :( Alright, I will just remove the file I think. Thanks again for your help!

EgbertRijke: Too bad, I would be totally ok with having the file here, even if it didn't contain any formalization. It explained what the sphere spectrum was, which was still worthwhile. Furthermore, it would be the right place to say something informative like "the sphere-prespectrum is not a spectrum because..." which is something you learned during this PR and this file could be a place to record that information for future learners.

fredrik-bakke: Alright, you have me convinced, I will add it back in and add an explanation. This way I can convince myself that I have learned something as well

EgbertRijke: Awesome!

fredrik-bakke: Yup, got it now. I disambiguated a couple of things that were conflated before :)

PR #837, opened 2023-10-13
Fix links to wikidata to the recommended links; add concept tags at end of file for testing purposes

EgbertRijke:
The recommended link to wikidata is of the form `http://www.wikidata.org/entity/ID` as recommended here: https://www.wikidata.org/wiki/Wikidata:Identifiers

I also added two `<concept id="" />` tags at the end of two files, so that Andrej and Katja can test their scripts. These tags will not be visible on our website.


fredrik-bakke (approved):
Looks good to me

PR #838, opened 2023-10-13
Abelianization

EgbertRijke:
In this pull request I will define the abelianization functor and show that it is left adjoint to the inclusion of abelian groups into groups.

### DIFF
--- a/src/group-theory/abelianization-groups.lagda.md
+++ b/src/group-theory/abelianization-groups.lagda.md
@@ -1,0 +1,41 @@
+# Abelianization of abstract groups
+
+```agda
+module group-theory.abelianization-groups where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+
+```
+
+</details>
+
+## Idea
+
+Consider a [group homomoprhism](group-theory.homomorphisms-groups.md) `f : G → A` from a [group](group-theory.groups.md) `G` into an [abelian group](group-theory.abelian-groups.md) `A`. We say that `f` **is an abelianization** of `G` if the precomposition function
+
+```text
+  - ∘ f : hom A B → hom G B
+```
+
+is an [equivalence](foundation-core.equivalences.md) for any abelian group `B`.
+
+The **abelianization** `Gᵃᵇ` of a group `G` always exists, and can be constructed as the [quotient group](group-theory.quotient-groups.md) `G/[G,G]` of `G` modulo its [commutator subgroup](group-theory.commutator-subgroups.md). Therefore we obtain an [adjunction](category-theory.adjunctions-large-categories.md)
+
+```text
+  hom Gᵃᵇ A ≃ hom G A,
+```
+
+i.e., the abelianization is left adjoint to the inclusion functor from abelian groups into groups.
+
+## External links
+
+- [Abelianization](https://groupprops.subwiki.org/wiki/Abelianization) at Groupprops
+- [Abelianization](https://ncatlab.org/nlab/show/abelianization) at nlab
+- [Abelianization](https://en.wikipedia.org/wiki/Commutator_subgroup#Abelianization) at Wikipedia
+- [Abelianization](https://mathworld.wolfram.com/Abelianization.html) at Wolfram Mathworld
+- [Commutator subgroup](https://www.wikidata.org/entity/Q522216) at Wikidata
+
+<content id="https://www.wikidata.org/entity/Q522216" />
--- a/src/category-theory/adjunctions-large-categories.lagda.md
+++ b/src/category-theory/adjunctions-large-categories.lagda.md
@@ -1,0 +1,40 @@
+# Adjunctions between large categories
+
+```agda
+module category-theory.adjunctions-large-categories where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+
+```
+
+</details>
+
+## Idea
+
+Let `C` and `D` be two
+[large categories](category-theory.large-categories.md). Two
+[functors](category-theory.functors-large-categories.md) `F : C → D` and
+`G : D → C` constitute an **adjoint pair** if
+
+- for each pair of objects `X` in `C` and `Y` in `D`, there is an
+  [equivalence](foundation-core.equivalences.md)
+  `ϕ X Y : hom X (G Y) ≃ hom (F X) Y` such that
+- for every pair of morhpisms `f : X₂ → X₁` and `g : Y₁ → Y₂` the following
+  square commutes:
+
+```text
+                        ϕ X₁ Y₁
+         hom X₁ (G Y₁) --------> hom (F X₁) Y₁
+              |                        |
+  G g ∘ _ ∘ f |                        | g ∘ _ ∘ F f
+              |                        |
+              v                        v
+         hom X₂ (G Y₂) --------> hom (F X₂) Y₂
+                        ϕ X₂ Y₂
+```
+
+In this case we say that `F` is **left adjoint** to `G` and `G` is **right
+adjoint** to `F` and write this as `F ⊣ G`.


src/category-theory/functors-large-precategories.lagda.md
```diff
@@ -72,52 +72,97 @@ module _
...
-    ( hom-functor-Large-Precategory G)
-    ( preserves-comp-functor-Large-Precategory F g f)) ∙
-  ( preserves-comp-functor-Large-Precategory G
-    ( hom-functor-Large-Precategory F g)
-    ( hom-functor-Large-Precategory F f))
-preserves-id-functor-Large-Precategory (comp-functor-Large-Precategory G F) =
-  ( ap
-    ( hom-functor-Large-Precategory G)
-    ( preserves-id-functor-Large-Precategory F)) ∙
-  ( preserves-id-functor-Large-Precategory G)
```
fredrik-bakke, line 122: This was not my definition

src/category-theory/representable-functors-large-precategories.lagda.md
```diff
@@ -38,20 +41,60 @@ associativity and the left unit law for the precategory `C`.
...
-obj-functor-Large-Precategory (representable-functor-Large-Precategory C c) =
-  hom-set-Large-Precategory C c
-hom-functor-Large-Precategory (representable-functor-Large-Precategory C c) g =
-  postcomp-hom-Large-Precategory C c g
-preserves-comp-functor-Large-Precategory
-  ( representable-functor-Large-Precategory C c) h g =
-  eq-htpy (associative-comp-hom-Large-Precategory C h g)
-preserves-id-functor-Large-Precategory
-  ( representable-functor-Large-Precategory C c) =
-  eq-htpy (left-unit-law-comp-hom-Large-Precategory C)
```
fredrik-bakke, line 54: I mean, come on, do you consider this an involved definition?

fredrik-bakke (commented):
I looked through the files you've touched to look for those definitions we talked about. These were the only two I could find. I will confess though that I sinned when defining `comp-natural-transformation-Large-Precategory`.

fredrik-bakke:
BTW, I'm thinking of doing small subcategories today. I know you had some formalization of large sub(pre)categories lying around, but am I correct you haven't done small subcategories?

src/category-theory/adjunctions-large-precategories.lagda.md
```diff
@@ -197,15 +214,20 @@ module _
       level-left-adjoint-Adjunction-Large-Precategory :
         Level → Level
```
fredrik-bakke, line 215: Note: with the new `--level-universe` flag, we literally can't have fields whose codomain is `Level`

EgbertRijke:
I'm marking this PR as ready for review, since it contains a lot of good stuff already, even if it doesn't formalize the abelianization adjunction yet.

src/foundation-core/functoriality-dependent-pair-types.lagda.md
```diff
@@ -330,24 +330,22 @@ module _
...
-        ( map-Σ-map-base f D)
-        ( tot g)
-        ( triangle-map-Σ D f g)
+        ( map-Σ D _ _)
+        ( map-Σ-map-base _ D)
+        ( tot _)
+        ( triangle-map-Σ D _ _)
         ( is-equiv-tot-is-fiberwise-equiv is-fiberwise-equiv-g)
-        ( is-equiv-map-Σ-map-base f D is-equiv-f)
+        ( is-equiv-map-Σ-map-base _ D is-equiv-f)
```
fredrik-bakke, line 342: Don't you think this definition looks weird now with all the inferred arguments?

src/group-theory/abelianization-groups.lagda.md
```diff
@@ -0,0 +1,233 @@
+# Abelianization of abstract groups
+
+```agda
+{-# OPTIONS --allow-unsolved-metas #-}
+
```
fredrik-bakke, line 5:
```suggestion
```

src/group-theory/abelianization-groups.lagda.md
```diff
@@ -0,0 +1,233 @@
...
+open import group-theory.homomorphisms-groups
+open import group-theory.normal-subgroups
+open import group-theory.quotient-groups
+```
+
+</details>
+
+## Idea
+
+Consider a [group homomoprhism](group-theory.homomorphisms-groups.md)
```
fredrik-bakke, line 40:
```suggestion
Consider a [group homomorphism](group-theory.homomorphisms-groups.md)
```

src/group-theory/abelianization-groups.lagda.md
```diff
@@ -0,0 +1,233 @@
...
+  abelianization-hom-Group =
+    hom-quotient-Group G H
+      ( commutator-normal-subgroup-Group G)
+      ( commutator-normal-subgroup-Group H)
+      ( f)
+      {!!}
+
+abelianization-functor-Group :
+  functor-Large-Category id Group-Large-Category Ab-Large-Category
+abelianization-functor-Group = {!!}
```
fredrik-bakke, line 201: Comment out?

src/group-theory/abelianization-groups.lagda.md
```diff
@@ -0,0 +1,233 @@
...
+    ( λ l → l)
+    ( Group-Large-Category)
+    ( Ab-Large-Category)
+left-adjoint-Adjunction-Large-Precategory
+  abelianization-adjunction-Group =
+  abelianization-functor-Group
+right-adjoint-Adjunction-Large-Precategory abelianization-adjunction-Group =
+  {!!}
+is-adjoint-pair-Adjunction-Large-Precategory abelianization-adjunction-Group =
+  {!!}
```
fredrik-bakke, line 219: Comment out

src/group-theory/abelianization-groups.lagda.md
```diff
@@ -0,0 +1,233 @@
...
+- [Abelianization](https://groupprops.subwiki.org/wiki/Abelianization) at
+  Groupprops
+- [Abelianization](https://ncatlab.org/nlab/show/abelianization) at nlab
+- [Abelianization](https://en.wikipedia.org/wiki/Commutator_subgroup#Abelianization)
+  at Wikipedia
+- [Abelianization](https://mathworld.wolfram.com/Abelianization.html) at Wolfram
+  Mathworld
+- [Commutator subgroup](https://www.wikidata.org/entity/Q522216) at Wikidata
+
+<content id="https://www.wikidata.org/entity/Q522216" />
```
fredrik-bakke, line 233: We should at least automate the connection between these two, so that both don't have to be maintained. I vote that the presence of a link to wikidata dictate that an anchor should be appended.

src/group-theory/commutator-subgroups.lagda.md
```diff
@@ -134,6 +143,16 @@ module _
   pr2 commutator-normal-subgroup-Group = is-normal-commutator-subgroup-Group
 ```
 
+## Properties
+
+### Every group homomorphism `f : G → H` maps `[G,G]` to `[H,H]`
+
```
fredrik-bakke, line 149:
```suggestion
### Every group homomorphism `f : G → H` maps `[G,G]` to `[H,H]`

This remains to be formalized.

```

src/group-theory/commuting-squares-of-group-homomorphisms.lagda.md
fredrik-bakke, line 1: Isn't this file a bit over the top? We already have files for commuting squares in categories, and commuting squares in types. Or is there something in particular you had in mind for commuting squares of group homomorphisms?

src/group-theory/homomorphisms-groups-equipped-with-normal-subgroups.lagda.md
```diff
@@ -0,0 +1,186 @@
+# Homomorphisms of groups equipped with normal subgroups
```
fredrik-bakke, line 1: We could also call this "Homomorphisms between inclusions of normal subgroups"

src/group-theory/nullifying-group-homomorphisms.lagda.md
```diff
@@ -0,0 +1,178 @@
...
+open import group-theory.normal-subgroups
+```
+
+</details>
+
+## Idea
+
+Consider a [group homomorphism](group-theory.homomorphisms-groups.md)
+`f : G → H` and a [normal subgroup](group-theory.normal-subgroups.md) `N` of the
+[group](group-theory.groups.md) `G`. We say that `f` \*\*nullifies `N` if it
```
fredrik-bakke, line 32:
```suggestion
[group](group-theory.groups.md) `G`. We say that `f` **nullifies** `N` if it
```

src/group-theory/quotient-groups.lagda.md
```diff
@@ -226,111 +205,117 @@ module _
...
+            ( compute-mul-quotient-Group x y)) ∙
+          ( ( compute-mul-quotient-Group (mul-Group G x y) z) ∙
+            ( ( ap
+                ( map-quotient-hom-Group)
+                ( associative-mul-Group G x y z)) ∙
+              ( ( inv
+                  ( compute-mul-quotient-Group x (mul-Group G y z))) ∙
+                ( ap
+                  ( mul-quotient-Group (map-quotient-hom-Group x))
+                  ( inv (compute-mul-quotient-Group y z)))))))
```
fredrik-bakke, line 276: Can this be simplified? Or is the particular association important?

src/group-theory/quotient-groups.lagda.md
```diff
@@ -226,111 +205,117 @@ module _
...
+            ( set-quotient-Group)
+            ( mul-quotient-Group (inv-quotient-Group y) y)
+            ( unit-quotient-Group))
+        ( λ x →
+          ( ap
+            ( mul-quotient-Group' (map-quotient-hom-Group x))
+            ( compute-inv-quotient-Group x)) ∙
+          ( ( compute-mul-quotient-Group (inv-Group G x) x) ∙
+            ( ap map-quotient-hom-Group
+              ( left-inverse-law-mul-Group G x))))
```
fredrik-bakke, line 297: Is the particular association important?

src/group-theory/quotient-groups.lagda.md
```diff
@@ -226,111 +205,117 @@ module _
...
+            ( set-quotient-Group)
+            ( mul-quotient-Group y (inv-quotient-Group y))
+            ( unit-quotient-Group))
+        ( λ x →
+          ( ap
+            ( mul-quotient-Group (map-quotient-hom-Group x))
+            ( compute-inv-quotient-Group x)) ∙
+          ( ( compute-mul-quotient-Group x (inv-Group G x)) ∙
+            ( ap map-quotient-hom-Group
+              ( right-inverse-law-mul-Group G x))))
```
fredrik-bakke, line 318: Is the particular association important?

fredrik-bakke (approved):
I couldn't sleep, so here I am reviewing your PR instead. This is well-written as always, despite some incomplete sections. Good work ⭐️

src/group-theory/homomorphisms-groups-equipped-with-normal-subgroups.lagda.md
EgbertRijke: My rule of thumb is that if you have to try really hard to come up with a better name, and nothing is really convincing, then it is better to stick with the plainest description of what you've got. I think "group equipped with normal subgroup" is easier to parse than "inclusion of normal subgroups" because the latter could be interpreted in many ways, one of which is a subcategory of the arrow category of Group, but another is that it is a subcategory of a slice category.

We don't need the category of groups equipped with normal subgroups for very many things, so I think it is ok to go with a straightforward but ugly name.

fredrik-bakke: Yep, I think you're right. Was just trying out some ideas :)

src/group-theory/abelianization-groups.lagda.md
fredrik-bakke: Actually, the new process is to mark the whole code block as `text` instead of agda code!

EgbertRijke:
I removed the `abelianization` file from this pull request, and will include it in a separate pull request.

src/foundation-core/functoriality-dependent-pair-types.lagda.md
EgbertRijke: Now I think these things look so different when you formalize them yourself as opposed to when it is someone elses code. I didn't think much of all these underscores, but you're right, it looks much better with `f` and `g` filled in.

VojtechStep: I'd rather keep the explicit arguments in, since they aren't distracting from the idea of the proof by being complex, and having them inferred requires Agda to do more work.

src/group-theory/abelianization-groups.lagda.md
EgbertRijke: Sure! I have no idea how to do that tho

src/group-theory/quotient-groups.lagda.md
EgbertRijke: Sure! My mind wasn't really on reformatting existing code in this pull request, but I'll do it

src/group-theory/commuting-squares-of-group-homomorphisms.lagda.md
EgbertRijke: I set it up so that it takes in groups and group homomorphisms, but outputs a homotopy between the underlying maps. That's the way it is nice to work with. For instance when proving that quotienting is functorial, you don't have to convert a homotopy of functions to an equality of group homomorphisms.

src/group-theory/quotient-groups.lagda.md
fredrik-bakke: Super!

EgbertRijke:
All comments are addressed now

src/group-theory/commuting-squares-of-group-homomorphisms.lagda.md
fredrik-bakke: Yeah, this makes sense. I'm still worried having so many files will catch up to us at some point. Perhaps we should have a second look at using instance arguments, it could help in cases like this. By marking the category of groups as an instance of a category, we could reference commuting squares of categories without explicitly passing the category every time.

fredrik-bakke: oh, wait, that doesn't address the function extensionality part

fredrik-bakke: But in _other_ cases, I guess

fredrik-bakke:
Awesome, I will set the doomsday clock ⏳

src/group-theory/abelianization-groups.lagda.md
VojtechStep: I don't think we should be adding these now - the `<concept />` tag (I assume `content` is a typo) is some ad-hoc syntax Andrej came up with which we don't have the infrastructure to handle. I'm not sure if these end up propagating to the website as-is, or if they get thrown out by the mdbook renderer, but neither case is desirable - we either have a technically "broken" website (not HTML5 compliant, because there is no `concept` tag, it's completely made up), or we're just duplicating information which has zero effect on the result.

src/foundation-core/functoriality-dependent-pair-types.lagda.md
EgbertRijke: Would you be willing to try this setup with implicit arguments?

Btw is it always the case that agda has to do more work when it has to infer implicit arguments? I thought that when the user provides arguments, then Agda has to do more: it has to parse the users input, and type check it. But parsing the users input if only a unique input is available is the same amount of work as inferring it...

Anyway, I don't really know how agda works:P

fredrik-bakke:
Is it okay to merge? @VojtechStep

VojtechStep:
Yup, LGTM, we can continue the discussion about `<concept />`s on the Discord

EgbertRijke:
Maybe we will never know 😄

PR #826, opened 2023-10-09
The codiagonal of a map

EgbertRijke:
This small PR adds the codiagonal of a map, and does some minor refactoring around the cogap map.

PR #839, opened 2023-10-14
Sequential limits

fredrik-bakke:
### Summary
- Defines towers of types, dependent towers of types, maps of towers, equivalences of towers, cones over towers, sequential limits, standard sequential limits, transfinite cocomposition
- Do some adjacent refactoring
- Characterize equality of towers, maps of towers, cones of towers, sequential limits
- Characterize equality in standard sequential limits


VojtechStep:
I'm coincidentally working on the dual notion of sequential colimits

VojtechStep:
Btw at some point I had a discussion with @EgbertRijke about (co)towers, and he suggested the names `sequential-diagram` for `cotower` and `contravariant-sequential-diagram` for `tower`; I'm using `sequential-diagram` at the moment, but I'm not opposed to switching to `cotower`.

fredrik-bakke:
Oh! I was planning on doing cotowers next, by dualizing this PR

fredrik-bakke:
I discussed the terminology with Egbert a moment ago as well. He was okay with `Cotower`, but we agreed on `sequential-limit` for the limit

EgbertRijke:
Oh yeah I forgot that I suggested "sequential diagram" as terminology. So when I was discussing with Fredrik I forgot, and he won because I was only talking about sequences.

How about sequential diagrams, Fredrik?:)

EgbertRijke:
> I'm coincidentally working on the dual notion of sequential colimits

Make a draft PR, so that we can see that you're working on it

EgbertRijke:
> Oh! I was planning on doing cotowers next, by dualizing this PR

Actually I think this PR will be better if it only does towers and sequential limits. Don't try to do everything in a PR.

scripts/generate_main_index_file.py
```diff
@@ -67,7 +67,7 @@ def generate_index(root, header):
     namespaces = sorted(set(utils.get_subdirectories_recursive(root)))
 
     for namespace in namespaces:
-        if namespace == 'temp':
+        if namespace == 'temp' or 'MAlonzo' in namespace:
```
EgbertRijke, line 70: What is this, and what does it have to do with towers?:)

scripts/generate_namespace_index_modules.py
```diff
@@ -45,7 +45,7 @@ def generate_agda_block(root, namespace):
     root = 'src'
 
     for namespace in utils.get_subdirectories_recursive(root):
-        if namespace == 'temp':
+        if namespace == 'temp' or 'MAlonzo' in namespace:
```
EgbertRijke, line 48: This too. Would it be better to make these changes in a separate PR, in which you would explain what these changes are?

EgbertRijke: I see now that they are explained in the summary. I still think such changes should be separate from formalization efforts.

scripts/generate_main_index_file.py
fredrik-bakke: Oh that thing. It's actually indispensable in the formalization of towers! Otherwise, I would have to go into my file system _again_ to find that pesky MAlonzo folder and delete it or the library won't type check :)

src/foundation/cones-over-towers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+open import foundation-core.homotopies
+open import foundation-core.identity-types
+open import foundation-core.whiskering-homotopies
+```
+
+</details>
+
+## Idea
+
+A **cone** over a [tower](foundation.towers-of-types.md) `A` with domain `X` is
```
VojtechStep: For my future reference, we now have definitions of what a "**cone** over a tower" and a "**cone** over a cospan" are, but the bold parts are the same. So far I took the bold parts to represent the name we want to define, but that's now overloaded. I'm in the same situation with "cocones under a span" and "cocones under a cotower", but I chose to make the whole "**cocone under a cotower**" bold. What should be the preferred approach?

src/foundation/cones-over-towers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+a [sequence](foundation.dependent-sequences.md) of functions from `X` into the
```
VojtechStep: There are now multiple definitions which link to the file on sequences, but don't actually use sequences in the definition itself; they don't even import that file (e.g. this and the recent (pre)spectra stuff). Is that intentional?

src/foundation/cones-over-towers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+## Definitions
+
+### Cones on towers
+
+```agda
+module _
+  {l1 : Level} (A : Tower l1)
+  where
+
+  cone-Tower : {l2 : Level} → UU l2 → UU (l1 ⊔ l2)
```
VojtechStep, line 56: Do we want to treat `Tower` as a namespace? I don't think we've done this with any other kind of diagrams

VojtechStep (commented):
This is not an actual review, I just had some questions to ensure the development of sequential limits and colimits is coherent.

src/foundation.lagda.md
VojtechStep, line 1: Is there a particular reason why put this stuff into `foundation` as opposed to `synthetic-homotopy-theory`?

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: Yes. This should be acceptable as long as the definitions agree, and I find using `dependent-sequence` would unnecessarily obscure things. Another good example of this is when we link to `foundation.structure`.

src/foundation.lagda.md
fredrik-bakke: Because they're not about homotopy theoretic constructions, as opposed to sequential colimits. Just like pullbacks are also in foundation.

src/foundation/cones-over-towers.lagda.md
EgbertRijke: I disagree. We should be accurate. If something is a dependent sequence, then *we* should not obscure things by not calling them so.

However, it is not a requirement that only imported files can be linked to. It should be possible to link to a file that is in the transitive closure of all the files being imported, i.e., to link to a file that the current file depends on via some of its imports. We also give forward references to files that are importing the current file, or sideways references to files that are related to the current file.

fredrik-bakke: You state that you disagree, but nothing in your comment entails that you have a different opinion than me.

src/foundation.lagda.md
VojtechStep: I guess I don't see how a homotopy limit is less homotopy theoretical then a homotopy colimit

EgbertRijke: Well, I wouldn't say that they're not about homotopy theoretic constructions. They're diagrams of spaces, so they are about homotopy theoretic constructions. Foundation is full of basics about homotopy theoretic constructions.

However, there is a big difference between limits and colimits. Limits exist just by function extensionality. That's why pullbacks and sequential limits are in foundation. Homotopy colimits, on the other hand are a new way to construct spaces. Synthetic homotopy theory begins with the introduction of homotopy pushouts to type theory. And all the other homotopy colimits that we have are built out of them.

One may note that the truncations are also higher inductive types. They're in foundation. That's just the way it currently is:) But perhaps one day we can put them in `orthogonal-factorization-systems`.

fredrik-bakke: Right, my bad, I should have said synthetic homotopy theoretic constructions

src/foundation/cones-over-towers.lagda.md
EgbertRijke: I thought you were arguing for linking to `sequences` when things were actually `dependent-sequences`, in order to not obscure things. Maybe I read your comment incorrectly. In that case I apologize for disagreeing, and don't disagree anymore:)

fredrik-bakke: I see, I was unclear. I argued for not forcing my definitions to use `dependent-sequence`

src/foundation.lagda.md
VojtechStep: Oh ok, if synthetic homotopy theory begins with homotopy pushouts then this separation makes sense 👌

EgbertRijke: Perhaps this should be said in `synthetic-homotopy-theory.lagda.md`

src/foundation/cones-over-towers.lagda.md
EgbertRijke: Why wouldn't you use dependent sequences if they are dependent sequences?

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: I'm curious if you have a better proposal for a naming scheme. I don't feel strongly about using it as a namespace, but it made sense to me in the moment, because it gives an easy way to discern limits of any shape of diagram we want.

src/foundation/cones-over-towers.lagda.md
EgbertRijke: The whole point of this library is that we refactor in this way, so that we can use infrastructure for one thing in the definition of another. Otherwise we might have to prove things/do some constructions multiple times.

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: An issue of course arises when we consider some other oo-category of types, like pointed types. So an alternative to my current approach is probably preferable

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: I mean, at some point I think we should accept that definitions are simple enough that we don't need to refactor them into smaller parts. Another example of this is proofs that maps are sections.

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: I guess a simple solution is to write `tower` in lower case.

src/foundation/cones-over-towers.lagda.md
EgbertRijke: I am absolutely certain that you're going to need operations on sequences such as shifting in the theory of (pre)spectra, and then it would be an abomination if you would redefine those operation specifically for spectra. It is better to use the definition of (dependent) sequences immediately.

src/foundation/cones-over-towers.lagda.md
EgbertRijke: fine with me

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: We can still use the infrastructure of sequences as the definitions agree.

fredrik-bakke:
> > Oh! I was planning on doing cotowers next, by dualizing this PR
> 
> Actually I think this PR will be better if it only does towers and sequential limits. Don't try to do everything in a PR.

That was the original plan.

fredrik-bakke:
> Oh yeah I forgot that I suggested "sequential diagram" as terminology. So when I was discussing with Fredrik I forgot, and he won because I was only talking about sequences.
> 
> How about sequential diagrams, Fredrik?:)

I still prefer towers...

fredrik-bakke:
What would you name the dual?
- cosequential diagrams?
- sequential codiagrams?
- dual sequential diagrams?
- opposite sequential diagrams?

src/foundation/cones-over-towers.lagda.md
VojtechStep: Sorry I didn't finish my thought - I meant to suggest exactly to write `tower` instead of `Tower`

fredrik-bakke: Excellent! Then we all agree on this :)

EgbertRijke:
> What would you name the dual?
> 
> * cosequential diagrams?
> * sequential codiagrams?
> * dual sequential diagrams?
> * opposite sequential diagrams?

IIRC we were going for contravariant sequential diagram

VojtechStep:
The original suggestion was `sequential-diagram` for `cotower`, where the maps are covariant with `succ`, and `contravariant-sequential-diagram` for `tower`. But honestly I'd rather go with `cotower` and `tower`

EgbertRijke:
> The original suggestion was `sequential-diagram` for `cotower`, where the maps are covariant with `succ`, and `contravariant-sequential-diagram` for `tower`. But honestly I'd rather go with `cotower` and `tower`

My argument for sequential diagram was that it is intuitive terminology for people who are not well-versed in homotopy theory. I.e., it is intuitive terminology no matter what your background is. "cotower" is a word only a category/homotopy theorist would utter.

fredrik-bakke:
I'm going to finish up this PR by characterizing equality of maps of towers, and then I am done

fredrik-bakke:
Doing dependent universal properties will be future work. This is not done for pullbacks either.

EgbertRijke:
What do you mean by the dependent universal property for pullbacks?

fredrik-bakke:
If that's a nonsensical statement then nevermind :)

src/foundation/cones-over-towers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+### Identifications of cones over towers of types
+
+```agda
+module _
+  {l1 l2 : Level} (A : tower l1) {X : UU l2}
+  where
+
+  coherence-htpy-cone-tower :
```
VojtechStep, line 82: Maybe we could try the convention of prepending `statement-` to this kind of definitons, as done in https://github.com/UniMath/agda-unimath/pull/826/files#r1351154625?

fredrik-bakke: Right, I thought about that, but it will require some refactoring work and I'm not yet convinced by it. For instance, `coherence-square-maps` read better without the `statement-` prefix, I think. Maybe it should be called `square-coherence-maps`? 🤔

EgbertRijke: Yeah, we should think about this some more, and base our discussion on the different ways the library already deals with this.

`square-coherence-maps` doesn't sound very good to me.

I also don't like prepending stuff with `statement` too much, even though I am guilty of doing that.

fredrik-bakke: One place I think we should use `statement-` where we don't do currently is for the axioms though. Like `statement-function-extensionality`, `statement-univalence`, `statement-axiom-K`, `statement-law-of-excluded-middle`...

fredrik-bakke: On a slightly related note. Do you prefer `coherence` for the commutation property of maps of towers, or `naturality`?

EgbertRijke: `naturality`

I've been changing that for natural transformations too.

fredrik-bakke: alright, awesome!

VojtechStep:
> > I'm coincidentally working on the dual notion of sequential colimits
> 
> Make a draft PR, so that we can see that you're working on it

#841

src/foundation/cones-over-towers.lagda.md
EgbertRijke: > One place I think we should/could* use `statement-` where we don't do currently is for the axioms though. Like `statement-function-extensionality`, `statement-univalence`, `statement-axiom-K`, `statement-law-of-excluded-middle`...

If you want to mark all of these entries with `statement`, then I start to be curious about when an entry would qualify in your opinion to be marked as `statement`. There are lots of statements in the library which we wouldn't mark as such, for instance the statement that something is an equivalence or is contractible. Is there a clear line?

The reason why I don't like statement so much, and why I would prefer to not have entries marked `statement`, is that I don't have a clear way to decide when to mark something as a statement. The reason that I used it last time was more out of laziness, i.e., I couldn't come up with something better in that moment. But that's not a very good reason, so I'm not compelled to mark more things by `statement`. I'd be happier to mark less things as `statement`.

fredrik-bakke: > If you want to mark all of these entries with statement, then I start to be curious about when an entry would qualify in your opinion to be marked as statement

How about something like: when inhabitants of a type are far more important than the type itself, and the literature does not give a distinct name to the statement?

EgbertRijke: I don't know how to make decisions based on such a vague criterion:P

fredrik-bakke: I believe in you :)

fredrik-bakke: jk

EgbertRijke: This is something we can definitely continue to discuss. Perhaps a clear criterion does come to the surface. I just don't know it yet.

fredrik-bakke:
There we go. This PR is ready for review now

src/foundation/sequential-limits.lagda.md
fredrik-bakke, line 1: Should we rename this file to `standard-sequential-limits`, and similarly `pullbacks` to `standard-pullbacks`?

src/foundation-core/commuting-triangles-of-maps.lagda.md
```diff
@@ -92,3 +97,17 @@ module _
       ( precomp right W)
   precomp-coherence-triangle-maps' H W = htpy-precomp H W
 ```
+
+### Coherences of commuting triangles of maps with fixed vertices
+
+```agda
+coherence-coherence-triangle-maps :
```
EgbertRijke, line 104: This name is probably not going to hold up. Here's how I would come up with a name for such a thing: We're looking at commuting triangles of maps with fixed vertices. This could be called `triangle-maps`. I.e., a triangle of maps with vertices `A`, `B`, and `X` consists of maps `f`, `g`, and `h`, and a homotopy witnessing that the triangle commutes. Then the projection maps would naturally be called `left-triangle-maps`, `right-triangle-maps`, `top-triangle-maps`, and `coh-triangle-maps`. Now we may wish to characterize the identity type of this type, i.e., we would make a type `htpy-triangle-maps` with four projections: `left-htpy-triangle-maps`, `right-htpy-triangle-maps`, `top-htpy-triangle-maps`, and `coh-htpy-triangle-maps`. The type you wrote here is the type of `coh-htpy-triangle-maps`, so it could be called `coherence-htpy-triangle-maps`, and then it looks like a name that fits in a scheme. Just repeating `coherence` twice doesn't look like it fits in a scheme.

src/foundation/sequential-limits.lagda.md
EgbertRijke: For pullbacks we have both `pullbacks` and `standard-pullbacks`. In `pullbacks` we define what pullbacks are, and in `standard-pullbacks` we define the standard one. We could do the same for sequential limits.

EgbertRijke: But actually it is more common to have `universal-property` files. Perhaps we should stick with that scheme. I can't decide at the moment:) Maybe tomorrow I will have a better idea:)

src/foundation-core/commuting-triangles-of-maps.lagda.md
fredrik-bakke: Thanks! I wrote it down as I was trying to compute what it should be. In the special case I wanted to use it, however, one of the maps was an identity, so I ended up not using this definition after all, and then I kind of forgot about it.

src/foundation-core/equality-dependent-pair-types.lagda.md
```diff
@@ -51,13 +52,26 @@ module _
...
+    {x y : A} {s : B x} (p : x ＝ y) → (x , s) ＝ (y , tr B p s)
+  eq-pair-Σ-eq-pr1 refl = refl
+
+  eq-pair-Σ-eq-pr1' :
+    {x y : A} {t : B y} (p : x ＝ y) → (x , tr B (inv p) t) ＝ (y , t)
+  eq-pair-Σ-eq-pr1' refl = refl
+
+  eq-pair-Σ-eq-pr2 :
+    {x : A} {s t : B x} → s ＝ t → (x , s) ＝ (x , t)
+  eq-pair-Σ-eq-pr2 {x} = ap {B = Σ A B} (x ,_)
```
EgbertRijke, line 73: This would read better as `ap (pair x)`

(plus or minus the implicit argument)

src/foundation-core/commuting-triangles-of-maps.lagda.md
EgbertRijke: Could you perhaps leave a short informal note here that this is the coherence in the characterization of identifications of commuting triangles of maps with fixed end points? Then someone can do that characterization at some point, so that this name doesn't come out of the blue.

src/foundation/sequential-limits.lagda.md
fredrik-bakke: We don't have a file called `standard-pullbacks`. We have `pullbacks`, which is about standard pullbacks, and we have `universal-property-pullbacks`, which is about the universal properties of pullbacks.

EgbertRijke: Oh I see, I was completely wrong. Nevermind then.

src/foundation-core/equality-dependent-pair-types.lagda.md
fredrik-bakke: I don't really agree given the type signature, but I've implemented your suggestion.

src/foundation/dependent-towers.lagda.md
```diff
@@ -0,0 +1,160 @@
...
+A **section of a dependent tower** `B` over `A` is a commuting diagram of the
+form
+
+```text
+  ⋯ ----> Aₙ₊₁ ----> Aₙ ----> ⋯ ----> A₁ ----> A₀
+           |         |               |        |
+  ⋯        |         |       ⋯       |        |
+           v         v               v        v
+  ⋯ ----> Bₙ₊₁ ----> Bₙ ----> ⋯ ----> B₁ ----> B₀.
+```
```
EgbertRijke, line 104: Ehm... would you say that a section of a type family `B` over `A` is a diagram `A ---> B`?

fredrik-bakke: Yeah, I agree, it's a bit awkward. I'll find a better formulation

src/foundation/dependent-towers.lagda.md
```diff
@@ -0,0 +1,160 @@
...
+  {l1 l2 : Level} {A : tower l1} (B : dependent-tower l2 A) →
+  section-dependent-tower A B →
+  (n : ℕ) (x : type-tower A n) → family-dependent-tower B n x
+map-section-dependent-tower B = pr1
+
+naturality-map-section-dependent-tower :
+  {l1 l2 : Level} {A : tower l1} (B : dependent-tower l2 A)
+  (f : section-dependent-tower A B) (n : ℕ) →
+  naturality-section-dependent-tower A B (map-section-dependent-tower B f) n
+naturality-map-section-dependent-tower B = pr2
```
EgbertRijke, line 130:
```suggestion
module _
  {l1 l2 : Level} (A : tower l1) (B : dependent-tower l2 A)
  where
  
  naturality-section-dependent-tower :
    (h : (n : ℕ) (x : type-tower A n) → family-dependent-tower B n x) (n : ℕ) →
    UU (l1 ⊔ l2)
  naturality-section-dependent-tower A B h n =
    h n ∘ map-tower A n ~ map-dependent-tower B n _ ∘ h (succ-ℕ n)

  section-dependent-tower :
    {l1 l2 : Level} (A : tower l1) (B : dependent-tower l2 A) → UU (l1 ⊔ l2)
  section-dependent-tower A B =
    Σ ( (n : ℕ) (x : type-tower A n) → family-dependent-tower B n x)
      ( λ h → (n : ℕ) → naturality-section-dependent-tower A B h n)

  map-section-dependent-tower :
    {l1 l2 : Level} {A : tower l1} (B : dependent-tower l2 A) →
    section-dependent-tower A B →
    (n : ℕ) (x : type-tower A n) → family-dependent-tower B n x
  map-section-dependent-tower B = pr1

  naturality-map-section-dependent-tower :
    {l1 l2 : Level} {A : tower l1} (B : dependent-tower l2 A)
    (f : section-dependent-tower A B) (n : ℕ) →
    naturality-section-dependent-tower A B (map-section-dependent-tower B f) n
  naturality-map-section-dependent-tower B = pr2
```

src/foundation/dependent-towers.lagda.md
```diff
@@ -0,0 +1,160 @@
...
+  ⋯ ----> Bₙ₊₁ ----> Bₙ ----> ⋯ ----> B₁ ----> B₀.
+```
+
+```agda
+naturality-section-dependent-tower :
+  {l1 l2 : Level} (A : tower l1) (B : dependent-tower l2 A)
+  (h : (n : ℕ) (x : type-tower A n) → family-dependent-tower B n x) (n : ℕ) →
+  UU (l1 ⊔ l2)
+naturality-section-dependent-tower A B h n =
+  h n ∘ map-tower A n ~ map-dependent-tower B n _ ∘ h (succ-ℕ n)
```
EgbertRijke, line 112: This should use `coherence-square-maps`. One purpose of `coherence-square-maps` is that it is a guarantee that all commuting squares are implemented in a uniform way, so that we don't unnecessarily have to invert commuting squares.

src/foundation/dependent-towers.lagda.md
```diff
@@ -0,0 +1,160 @@
...
+  (f : section-dependent-tower A B) (n : ℕ) →
+  naturality-section-dependent-tower A B (map-section-dependent-tower B f) n
+naturality-map-section-dependent-tower B = pr2
+```
+
+## Operations
+
+### Right shifting a dependent tower
+
+We can **right shift** a dependent tower of types by forgetting the first terms.
```
EgbertRijke, line 137: Do you really want to custom-define shift for every object of sequential nature in this library? All of these shifts will never be judgmentally equal to each other.

src/foundation/dependent-towers.lagda.md
fredrik-bakke: sadly, it does not conform to `coherence-square-maps`, as it is a dependent diagram. At least I think that is what blocks it

src/foundation/dependent-towers.lagda.md
fredrik-bakke: Do you suggest applying `distributive-Pi-Sigma` and use a common definition for shifting on sequences instead?

src/foundation/equivalences-towers.lagda.md
```diff
@@ -0,0 +1,100 @@
...
+pr1 (hom-equiv-tower A B e) n = pr1 (pr1 e n)
+pr2 (hom-equiv-tower A B e) = pr2 e
+```
+
+## Properties
+
+### Characterizing equality of towers
+
+```agda
+refl-equiv-tower :
```
EgbertRijke, line 69: `id-equiv-tower`

src/foundation/dependent-towers.lagda.md
EgbertRijke: Ah, it is for dependent maps. My bad!

src/foundation/dependent-towers.lagda.md
EgbertRijke: I don't know exactly what I want, so nevermind. It's ok like it is.

src/foundation/equivalences-towers.lagda.md
EgbertRijke: Other instances of `refl-equiv` in the library should also be `id-equiv`, if there are any.

fredrik-bakke: Yep, thanks! There is one in `27-sequences`, as that it where I stole this from :)

EgbertRijke: Can 27-sequences be removed with this PR?

src/foundation/maps-towers.lagda.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Composition of map of towers
+
+```agda
+comp-hom-tower :
+  {l : Level} (A B C : tower l) → hom-tower B C → hom-tower A B → hom-tower A C
+pr1 (comp-hom-tower A B C g f) n = map-hom-tower B C g n ∘ map-hom-tower A B f n
+pr2 (comp-hom-tower A B C g f) n x =
+  ( ap (map-hom-tower B C g n) (naturality-map-hom-tower A B f n x)) ∙
+  ( naturality-map-hom-tower B C g n (map-hom-tower A B f (succ-ℕ n) x))
```
EgbertRijke, line 93: Here, for instance, if you every find yourself writing `map-hom-tower comp-hom-tower ...` then that's an indication that it would have been better to define this in the refactored way, by first defining `map-comp-hom-tower` and `naturality-comp-hom-tower` and only then `comp-hom-tower`.

src/foundation/maps-towers.lagda.md
```diff
@@ -0,0 +1,150 @@
...
+## Properties
+
+### Characterization of equality of maps of towers
+
+```agda
+module _
+  {l1 l2 : Level} (A : tower l1) (B : tower l2)
+  where
+
+  coherence-hom-tower :
```
EgbertRijke, line 105: Should this be called `coherence-htpy-hom-tower`?

src/foundation/equivalences-towers.lagda.md
VojtechStep: 27-sequences defines cotowers and cocones under cotowers, so I should probably remove it in my PR on sequential colimits

src/foundation/pullbacks.lagda.md
```diff
@@ -285,33 +284,29 @@ module _
...
-        ( gap f g (triple p q H))
-        ( map-equiv-canonical-pullback-htpy Hf Hg)
-        ( gap f' g' (triple p' q' H'))
+        ( gap f g (p , q , H))
+        ( map-equiv-standard-pullback-htpy Hf Hg)
+        ( gap f' g' (p' , q' , H'))
         ( triangle-is-pullback-htpy
-          {c = triple p q H} {triple p' q' H'} (triple Hp Hq HH))
-        ( is-equiv-map-equiv-canonical-pullback-htpy Hf Hg)
+          {c = p , q , H} {p' , q' , H'} (Hp , Hq , HH))
```
EgbertRijke, line 308: We can probable remove `triple` if it is not useful anymore.

src/foundation/equivalences-towers.lagda.md
fredrik-bakke: Yep. There is no overlap between what is added in this PR and what's in `27-sequences`

src/foundation/pullbacks.lagda.md
fredrik-bakke: triple can still be nice for multiline definitions. It is still used some places in this file

src/group-theory/torsion-free-groups.lagda.md
```diff
@@ -119,14 +118,14 @@ module _
   {l1 : Level} (G : Group l1)
   where
 
-  is-equiv-first-projection-pullback-subgroup-prop-prop-Group :
-    Prop (lsuc l1)
+  is-equiv-first-projection-pullback-subgroup-prop-prop-Group : Prop (lsuc l1)
```
EgbertRijke, line 121: This should now be renamed to `is-equiv-vertical-map-standard-pullback-subgroup-prop-prop-Group`.

src/foundation/sequential-limits.lagda.md
```diff
@@ -0,0 +1,284 @@
...
+open import foundation-core.contractible-types
+open import foundation-core.function-types
+open import foundation-core.homotopies
+open import foundation-core.identity-types
+open import foundation-core.propositions
+```
+
+</details>
+
+## Definitions
```
EgbertRijke, line 35: Missing `## Idea` section

src/foundation/maps-towers.lagda.md
```diff
@@ -0,0 +1,150 @@
+# Maps of towers of types
```
EgbertRijke, line 1: Probably it's better to call these `morphisms` so that they will appear close to the other files called `morphisms`.

src/foundation/sequential-limits.lagda.md
```diff
@@ -0,0 +1,284 @@
...
+  is-property-is-sequential-limit :
+    (c : cone-tower A X) → is-prop (is-sequential-limit c)
+  is-property-is-sequential-limit c = is-property-is-equiv (gap-tower A c)
+
+  is-sequential-limit-Prop : (c : cone-tower A X) → Prop (l1 ⊔ l2)
+  pr1 (is-sequential-limit-Prop c) = is-sequential-limit c
+  pr2 (is-sequential-limit-Prop c) = is-property-is-sequential-limit c
+```
+
+### Functoriality of standard sequential limits
```
EgbertRijke, line 118: Given that we usually factor out functoriality files, at least in `foundation`, we should probably do that in this case too.

src/foundation/towers.lagda.md
```diff
@@ -0,0 +1,96 @@
...
+sequence-map-tower : {l : Level} → (ℕ → UU l) → UU l
+sequence-map-tower A = (n : ℕ) → A (succ-ℕ n) → A n
+
+tower : (l : Level) → UU (lsuc l)
+tower l = Σ (ℕ → UU l) (sequence-map-tower)
+
+type-tower : {l : Level} → tower l → ℕ → UU l
+type-tower = pr1
+
+map-tower :
```
EgbertRijke, line 53: Also, given that you call this `map-tower`, the maps of towers should really be called morphisms of towers:)

src/foundation/towers.lagda.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+### Composites in towers
+
+```agda
+comp-map-tower :
+  {l : Level} (A : tower l) (n r : ℕ) → type-tower A (n +ℕ r) → type-tower A n
+comp-map-tower A n zero-ℕ = id
+comp-map-tower A n (succ-ℕ r) = comp-map-tower A n r ∘ map-tower A (n +ℕ r)
+```
```
EgbertRijke, line 65: Could this stuff about iterated compostion go into its own file? I imagine that at some point we would need a bunch of lemmas about them, and it would just clutter this file up.

For a filename I suggest `iterated-composition-of-maps-towers`.

EgbertRijke (approved):
Awesome pull request! Well done!

src/foundation/towers.lagda.md
fredrik-bakke: Alright, that can be done. I think it should be called something like `composites-of-maps-in-towers` though. First, because iteration is not core to the definition, that is just the mechanism used to define them. We could also have considered a weak definition of towers with arrows between every two vertices when  `n < k` and coherences between the maps. That would also give a notion of composites of maps, but without the iteration. Second, to disambiguate maps in towers from maps between towers.

fredrik-bakke:
> Awesome pull request! Well done!

Thanks! I'm really happy with how pain-free this one has been to formalize! Some of your earlier formalization work is paying off in this one :)

src/foundation/maps-towers.lagda.md
fredrik-bakke: Thanks for pointing it out to me

src/foundation/towers.lagda.md
EgbertRijke: But under that philosophy, shouldn't `map-tower` be what you currently call `comp-map-tower`, and then `comp-map-tower` should compose `map-tower`s fro `k < l` and `n < k`? I think you're mixing two things now.

fredrik-bakke: You could argue that. It's a different definition as I said.

EgbertRijke: I think that "weaker" definition will eventually be useful, by the way.

fredrik-bakke: Awesome!

src/foundation/maps-towers.lagda.md
fredrik-bakke: Maybe the other files about morphisms of various types should be called maps of them instead. Or would we call equivalences of towers for isomorphisms of towers?

src/foundation/sequential-limits.lagda.md
fredrik-bakke: have you given this some more thought? I've always found the name of the file `pullbacks` a bit misleading. Perhaps it is good practice to add a table of links in these files

src/foundation/cones-over-towers.lagda.md
fredrik-bakke: I don't really know, maybe we can pick up this discussion again in a different place? For sure, having the entire phrase in bold will always be good if you want to be safe, but sometimes having the whole phrase together requires reformulating sentences, which can become a bit of work if we want to require that everywhere in the library.

src/foundation/maps-towers.lagda.md
EgbertRijke: Like you said in the other thread, whether we want to call equivalences isomorphisms is something we can discuss another day, but for now for consistency it is at least better to make this filename `morphisms-of-towers`.

EgbertRijke: There is actually an argument for calling equivalences isomorphisms. The argument is that it is Andre Joyal's preferred terminology:)

fredrik-bakke: That is pretty convincing to me!

EgbertRijke: You would break our current conventions in the library, so please don't do that in this pull request:)

EgbertRijke: By the way, your argument seems to be that if we call something a morphism, then we can't call some of the morphisms equivalences. But we do that even for groups: we have group homomorphisms, equivalences of groups *and* isomorphisms of groups. So we can have all three.

fredrik-bakke: Ah, because of function extensionality? I was going to say: for uncomplicated, concrete mathematical objects, the terminology "maps" feels more appropriate to me, while morphisms fit better for abstract objects, or when they are maps of concrete objects, but in some less trivial sense.

fredrik-bakke: for categories we also have different notions of "map" and "functor". Here, equivalences are not just invertible maps, so clearly there is some nuance to these concepts

fredrik-bakke: Anyways, feel free to ignore this discussion. Sorry for eating up even more of your time

src/foundation/sequential-limits.lagda.md
EgbertRijke: I don't know yet what to do, but I think it would be bad if the library contains a bunch of stuff about pullbacks but there is no file called `pullbacks`. Perhaps leave it as it is now, because this PR is about something else, and then we can discuss it on discord?

src/foundation/cones-over-cospans.lagda.md
```diff
@@ -157,21 +148,20 @@ module _
   pr1 (extensionality-cone c c') = htpy-eq-cone c c'
   pr2 (extensionality-cone c c') = is-equiv-htpy-eq-cone c c'
 
-  eq-htpy-cone : (c c' : cone f g C) → htpy-cone c c' → (c ＝ c')
+  eq-htpy-cone : (c c' : cone f g C) → htpy-cone c c' → c ＝ c'
   eq-htpy-cone c c' = map-inv-equiv (extensionality-cone c c')
 ```
 
-### Precomposing cones
+### Precomposition cones over cospans
```
VojtechStep, line 155: It this supposed to be read "(precomposition cones) over cospans"?

EgbertRijke: Good question! What was wrong with the old section header?

src/foundation/sequential-limits.lagda.md
```diff
@@ -0,0 +1,288 @@
...
+module _
+  {l1 l2 : Level} (A : tower l1) {X : UU l2}
+  where
+
+  gap-tower : cone-tower A X → X → standard-sequential-limit A
+  pr1 (gap-tower c x) n = map-cone-tower A c n x
+  pr2 (gap-tower c x) n = coherence-cone-tower A c n x
+```
+
+### The `is-sequential-limit` property
```
VojtechStep, line 121: I think "The property of being a sequential limit" would be more descriptive

src/foundation/sequential-limits.lagda.md
fredrik-bakke: I will add a table of links, and then we can see if that is satisfactory

src/foundation/sequential-limits.lagda.md
fredrik-bakke: Thanks!

src/foundation/cones-over-cospans.lagda.md
fredrik-bakke: I don't really find either header satisfactory. I edited it and the forgot about it. The issue I have with the first one is that we're not literally precomposing cones, we're composing cones with a map

EgbertRijke: Then how about `## Precomposing cones with maps`?

EgbertRijke:
Would you also be willing to start adding external links to the new files you added in this PR?

fredrik-bakke:
Manually? I thought we were going to get automation for that. To be honest I was hoping I would never have to start doing that

EgbertRijke:
I was actually hoping that we would start doing that by hand. The mathswitch project will give a comprehensive list of all files of all sources that relate to the thing that wikidata thinks is a concept. This means that for many concepts it will happen that many files in any one source will link to the same concept. This is invariably going to mean that the user will not know what the best link is to get more useful information.

On the other hand, if we give some hand-selected external links, then we point the user to what we think is useful. For example, links to towers in various sources would be useful.

Note that Wikipedia also has sections for external links, besides references. So they do it too.

EgbertRijke:
Anyway, if you are not willing to do that, then that's ok too. I'll do it at some point.

EgbertRijke:
But note that collecting the wikidata identifiers will not be automated, so we will anyway have to do this by hand.

fredrik-bakke:
This is a considerable amount of recurring work to add to the process of submitting PRs to agda-unimath for what I would consider to be very little gain. If I think a resource is truly useful to the readers of a file, I will link it as a reference, or as another file in agda-unimath.

EgbertRijke:
No worries. I will do it.

fredrik-bakke:
> No worries. I will do it.

I won't stop you, but I'd like to discuss with you at some point how useful such a thing is, to what extent it should be done, and when it is okay to remove/omit external links.

VojtechStep:
Are you about to start adding various unrelated tables to this PR? 👀

fredrik-bakke:
> Are you about to start adding various unrelated tables to this PR? 👀

I am also adding a table about sequential limits. I wrote the table of fibers together with pullbacks, and then concluded they didn't fit together. I added the table as a  separate one instead rather than deleting it, as I am not a fan of deleting work that is useful.

src/univalent-combinatorics/fibers-of-maps-finite-types.lagda.md
```diff
@@ -1,7 +1,7 @@
 # Fibers of maps between finite types
 
 ```agda
-module univalent-combinatorics.fibers-of-maps where
+module univalent-combinatorics.fibers-of-maps-finite-types where
```
EgbertRijke, line 4: A lot of files in `univalent-combinatorics` are just named after their concept, and the understanding is that they are about finite types. Do you not like that convention?

fredrik-bakke: I see, my bad. I guess combinatorics of infinite types is just called group theory

fredrik-bakke:
Is there a convention for how to order the reference tables yet? I find that the alphabetic ordering may not be the best one. Let me know if you have objections to the current order I wrote them in.

fredrik-bakke:
otherwise I think this PR is mostly done now

src/foundation/universal-property-sequential-limits.lagda.md
```diff
@@ -0,0 +1,248 @@
...
+      is-contr (Σ (Y → X) (λ h → htpy-cone-tower A (cone-map-tower A c h) c'))
+    uniqueness-universal-property-sequential-limit up Y c' =
+      is-contr-equiv'
+        ( Σ (Y → X) (λ h → cone-map-tower A c h ＝ c'))
+        ( equiv-tot
+          ( λ h → extensionality-cone-tower A (cone-map-tower A c h) c'))
+        ( is-contr-map-is-equiv (up Y) c')
+```
+
+### The universal property of sequential limits is a property
```
VojtechStep, line 160: This phrasing seems a little misleading to me, since as mentioned elsewhere, this universal property is a large property, and this part of the file formalizes the fact that the universal property is level-wise a property, so I'd mention that explicitly: "is levelwise a property" or "is fiberwise a property"

src/foundation/universal-property-sequential-limits.lagda.md
```diff
@@ -0,0 +1,248 @@
...
+  {l1 l2 l3 : Level} (A : tower l1) {X : UU l2} (c : cone-tower A X)
+  where
+
+  abstract
+    is-prop-universal-property-sequential-limit :
+      is-prop (universal-property-sequential-limit l3 A c)
+    is-prop-universal-property-sequential-limit =
+      is-prop-Π (λ Y → is-property-is-equiv (cone-map-tower A c))
+
+  map-universal-property-sequential-limit :
```
VojtechStep, line 173: I don't think `map-universal-property-sequential-limit` and `eq-map-universal-property-sequential-limit` belong in this anonymous module - I'd put the first under `## Definitions` and the second into its own section under `## Properties`

EgbertRijke:
> Is there a convention for how to order the reference tables yet? I find that the alphabetic ordering may not be the best one. Let me know if you have objections to the current order I wrote them in.

The convention is the alphabetic ordering. Why is it not the best convention?

fredrik-bakke:
It seems to me like unrelated entries end up together, making the table somewhat harder to navigate. Maybe you can have a look at the diff of the commit  `reorder table entries` and see if you dis-/agree with me?

tables/sequential-limits.md
```diff
@@ -0,0 +1,12 @@
...
+| Composite maps in towers                    | [`foundation.composite-maps-in-towers`](foundation.composite-maps-in-towers.md)                         |
+| Morphisms of towers                         | [`foundation.morphisms-towers`](foundation.morphisms-towers.md)                                         |
+| Equivalences of towers                      | [`foundation.equivalences-towers`](foundation.equivalences-towers.md)                                   |
+| Cones over towers                           | [`foundation.cones-over-towers`](foundation.cones-over-towers.md)                                       |
+| The universal property of sequential limits | [`foundation.universal-property-sequential-limits`](foundation.universal-property-sequential-limits.md) |
+| Sequential limits                           | [`foundation.sequential-limits`](foundation.sequential-limits.md)                                       |
+| Functoriality sequential limits             | [`foundation.functoriality-sequential-limits`](foundation.functoriality-sequential-limits.md)           |
```
EgbertRijke, line 11:
```suggestion
| Functoriality of sequential limits             | [`foundation.functoriality-sequential-limits`](foundation.functoriality-sequential-limits.md)           |
```

EgbertRijke:
> It seems to me like unrelated entries end up together, making the table somewhat harder to navigate. Maybe you can have a look at the diff of the commit `reorder table entries` and see if you dis-/agree with me?

Your current ordering is fine with me 👍

src/foundation-core/fibers-of-maps.lagda.md
```diff
@@ -403,3 +403,10 @@ reduce-Π-fiber :
   (C : B → UU l3) → ((y : B) → fiber f y → C y) ≃ ((x : A) → C (f x))
 reduce-Π-fiber f C = reduce-Π-fiber' f (λ y z → C y)
 ```
+
+## Table of files about fibers of maps
```
EgbertRijke, line 407: This is a fine way of including the table in a file. The other option, which is also used, is to put the table under the `## See also` section. I think either way is good.

src/foundation/universal-property-sequential-limits.lagda.md
fredrik-bakke: How about `The universal property of sequential limits at a universe level is a property`?

VojtechStep: Sounds good to me 👌

src/foundation/universal-property-sequential-limits.lagda.md
fredrik-bakke: Thanks for spotting this. However, are you okay with having both under `Definitions`, since they're both just unfolding the definition of the universal property?

VojtechStep: Could we then reflect the fact that `eq-map-universal-property-sequential-limit` is a sort of computation rule in its name?  Then I'd be happy having it under Definitions

src/foundation/towers.lagda.md
```diff
@@ -0,0 +1,91 @@
...
+open import foundation.iterating-functions
+open import foundation.unit-type
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+A **tower of types** `f` is a [sequence](foundation.sequences.md) of types
```
VojtechStep, line 22: Here you define a tower as `f`, but in `cones-over-towers` you write "over a tower `A`", and in `dependent-towers` you define "sections over `(A, f)`". I think it's ok to use `A` for `(A, f)`, but I don't think we should use `f` for `(A, f)`.

src/foundation/cones-over-cospans.lagda.md
```diff
@@ -27,9 +27,9 @@ open import foundation-core.whiskering-homotopies
 
 ## Idea
 
-A **cone** on a [cospan](foundation.cospans.md) `A --f--> X <--g-- B` with
-domain `C` is a triple `(p,q,H)` consisting of a map `p : C → A`, a map
-`q : C → B`, and a homotopy `H` witnessing that the square
+A **cone on a [cospan](foundation.cospans.md)** `A -f-> X <-g- B` with domain
```
VojtechStep, line 30: While you're at it, could you change "cone on a cospan" to "cone over a cospan" to keep it consistent?

src/foundation/towers.lagda.md
fredrik-bakke: I think all three are fine, but I'll change it for you ;)

EgbertRijke:
Merged 🚀

PR #843, opened 2023-10-15
Perfect groups

EgbertRijke:
This pull request adds perfect groups to the library

### DIFF
--- a/src/group-theory/perfect-groups.lagda.md
+++ b/src/group-theory/perfect-groups.lagda.md
@@ -1,0 +1,50 @@
+# Perfect groups
+
+```agda
+module group-theory.perfect-groups where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.propositions
+open import foundation.universe-levels
+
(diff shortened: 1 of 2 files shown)


src/group-theory/perfect-groups.lagda.md
```diff
@@ -0,0 +1,50 @@
...
+open import group-theory.groups
+```
+
+</details>
+
+## Idea
+
+A [group](group-theory.groups.md) `G` is said to be **perfect** if its
+[commutator subgroup](group-theory.commutator-subgroups.md) is a
+[full](group-theory.full-subgroups.md) [subgroup](group-theory.subgroups.md).
```
fredrik-bakke, line 24: @VojtechStep Just to get back to your earlier question. Here, the text wouldn't read as well if we were to require that we write `**perfect group**`, I think, although there may be other concepts we define using the adjective `**perfect**`.

fredrik-bakke (approved):

fredrik-bakke:
If you're done with this PR it can be merged right away

PR #844, opened 2023-10-15
The decidable total order on a standard finite type

fredrik-bakke:
Defines the decidable total order on a standard finite type. Also does a littlebit of refactoring of infix binary comparison operators.


fredrik-bakke:
Rest assured, I am not planning on adding anything else to this PR.

fredrik-bakke:
I will revert the change in notation for infix binary comparison operators.

src/category-theory/faithful-functors-precategories.lagda.md
```diff
@@ -28,8 +28,8 @@ A [functor](category-theory.functors-precategories.md) between
 its an [embedding](foundation-core.embeddings.md) on hom-sets.
 
 Note that embeddings on [sets](foundation-core.sets.md) happen to coincide with
-[injections](foundation.injective-maps.md), but we define it in terms of the
-stronger notion, as this is the notion that generalizes.
+[injections](foundation.injective-maps.md), however, we define it in terms of
+the stronger notion as this is the notion that generalizes.
```
EgbertRijke, line 32: I think it's not a grammatical sentence anymore with this change. Maybe:

```suggestion
[injections](foundation.injective-maps.md). However, we define faithful functors in terms of
embeddings because this is the notion that generalizes.
```

src/elementary-number-theory/inequality-standard-finite-types.lagda.md
```diff
@@ -39,6 +40,10 @@ leq-Fin (succ-ℕ k) x (inr y) = unit
 leq-Fin (succ-ℕ k) (inl x) (inl y) = leq-Fin k x y
 leq-Fin (succ-ℕ k) (inr x) (inl y) = empty
 
+infix 30 _≤-Fin_
+_≤-Fin_ : {k : ℕ} → Fin k → Fin k → UU lzero
+_≤-Fin_ {k} = leq-Fin k
```
EgbertRijke, line 45: Have you tested that the natural number `k` can be implicit in practice?

fredrik-bakke: This notation should only be used if it can, but I haven't tested it, no

EgbertRijke: I have tested it in the past and it can't be implicit, so I think this notation isn't going to be useful.

A lot of our stuff about standard finite types has the natural number explicit, because if you make it implicit then you find yourself using the cursy braces all the time. So I think this notation should be removed again, because it will put users on the wrong track.

fredrik-bakke: Alright, thanks for pointing that out to me. I will remove it tomorrow.

fredrik-bakke: Done :)

EgbertRijke:
Cool! I'll merge when the checks are green

PR #845, opened 2023-10-15
The simplex precategory

fredrik-bakke:
Define the simplex precategory and precategories of various posets.


fredrik-bakke:
I'm not particularly happy with the current definition of the simplex category, so any advice on that is much appreciated. Also, I'd like to define simplicial sets for the purpose of defining quasi-categories, but `category-theory` does not seem like the proper place for that. Should a `homotopy-theory` module be made?

src/category-theory/augmented-simplex-category.lagda.md
```diff
@@ -0,0 +1,129 @@
...
+open import order-theory.order-preserving-maps-posets
+```
+
+</details>
+
+## Idea
+
+**The augmented simplex category** is the category consisting of finite total
+orders and
+[order-preserving maps](order-theory.order-preserving-maps-posets.md).
```
EgbertRijke, line 29: We have notions of finite poset and total orders in the library, so if you are defining the augmented simplex category to consist of finite total orders shouldn't at least one of your formalized definitions be literally that it is a finite total order?

src/category-theory/simplex-category.lagda.md
```diff
@@ -0,0 +1,131 @@
...
+open import foundation.universe-levels
+
+open import order-theory.order-preserving-maps-posets
+```
+
+</details>
+
+## Idea
+
+**The simplex category** is the category consisting of finite ordinals and
```
EgbertRijke, line 27: I think this definition should mention that these are _inhabited_ total finite posets. Note that the empty ordinal is an ordinal.

src/category-theory/augmented-simplex-category.lagda.md
fredrik-bakke: I realized after going to bed that these definitions are just not the right ones. I will change it tomorrow. Should the simplex category just be finite inhabited total orders?

EgbertRijke: I think we should keep the definition in which the type of objects is the type of natural numbers. I just saw that you also made a file about finite total orders. So one possibility is that you turn that into a category over at `order-theory`, and make a custom definition of the (augmented) simplicial category here in `category-theory`. Then we prove here too that these two categories are the same.

That's a bit more work, but in this PR you can just say that it remains to be shown that the (augmented) simplicial category is equivalent to the category of (inhabited) finite posets. (The parentheses don't match in that statement.)

src/category-theory/augmented-simplex-category.lagda.md
```diff
@@ -0,0 +1,129 @@
...
+open import foundation.universe-levels
+
+open import order-theory.order-preserving-maps-posets
+```
+
+</details>
+
+## Idea
+
+**The augmented simplex category** is the category consisting of finite total
```
EgbertRijke, line 27: link to the file about finite total orders

EgbertRijke:
Awesome pull request!

EgbertRijke:
> I'm not particularly happy with the current definition of the simplex category, so any advice on that is much appreciated. Also, I'd like to define simplicial sets for the purpose of defining quasi-categories, but `category-theory` does not seem like the proper place for that. Should a `homotopy-theory` module be made?

Yes, that could be. I would call that folder `simplicial-homotopy-theory`, because a `homotopy-theory` folder would also contain Quillen model categories and so on, but it would probably be too much to have that all under one umbrella.

fredrik-bakke:
Alright, I consider this a good stopping point for the PR. I'm aware that the file on inhabited finite total orders should be refactored a little. Ideally, one would perhaps want to have files all the way down to inhabited preorders, but that will be left for someone with more patience. Also, the files on different subprecategories of posets should probably just become files about sub_categories_ once we have the proper theory established.

EgbertRijke:
Excellent! I will review this in the evening

src/order-theory/precategory-of-posets.lagda.md
```diff
@@ -0,0 +1,59 @@
...
+    .obj-Large-Precategory l → Poset (α l) (β l)
+    .hom-set-Large-Precategory → hom-set-Poset
+    .comp-hom-Large-Precategory {X = X} {Y} {Z} → comp-hom-Poset X Y Z
+    .id-hom-Large-Precategory {X = X} → id-hom-Poset X
+    .associative-comp-hom-Large-Precategory {X = X} {Y} {Z} {W} →
+      associative-comp-hom-Poset X Y Z W
+    .left-unit-law-comp-hom-Large-Precategory {X = X} {Y} →
+      left-unit-law-comp-hom-Poset X Y
+    .right-unit-law-comp-hom-Large-Precategory {X = X} {Y} →
+      right-unit-law-comp-hom-Poset X Y
```
EgbertRijke, line 48: Oof I'm not sure if this is a style of definitions I want to promote too much. Maybe it's just me, but I would prefer the old style with copattern matching.

EgbertRijke:
Beautiful PR! I will merge it when you update it.

src/order-theory/precategory-of-posets.lagda.md
fredrik-bakke: lol. I do kinda agree, it looks better in my editor though, as my highlighter highlights expressions starting with a `.` differently :)

fredrik-bakke:
I see. I wouldn't have minded to be honest, but now that we assign credit on the website, maybe we should be more strict about this? It's kinda sad that this restricts when we should use `suggestion`s though. It really helps the review process.

EgbertRijke:
I think it's ok for suggestions.

PR #851, opened 2023-10-16
Make pre-commit hooks talk to git

### LINKED ISSUE #664: Pre-commit hooks should ignore all junk files
Some of the pre-commit hooks, like `generate_main_index_file.py`, don't properly behave when the `src` folder is dirty. For instance when a MAlonzo folder is present. To fix this, we can filter out patterns from the `.gitignore` file.

### LINKED ISSUE #848: Generate index files based only on tracked files
Apparently Ulrik had an untracked file for hypoabelian groups in his folder, but `make pre-commit` still added this line to the indexing file. I've never had this problem before.

_Originally posted by @EgbertRijke in https://github.com/UniMath/agda-unimath/pull/847#discussion_r1360772603_

fredrik-bakke:
Resolves #664
Resolves #848


scripts/spaces_conventions_simple.py
```diff
@@ -60,6 +68,10 @@ def no_whitespace_before_closing_curly_brace(line):
                     line = space_before_special_symbols(line)
                     line = no_whitespace_before_closing_parenthesis(line)
                     line = no_whitespace_before_closing_curly_brace(line)
+                    # line = no_whitespace_after_opening_parenthesis_inside_line(
+                    #     line)
+                    # line = no_whitespace_after_opening_curly_brace_inside_line(
+                    #     line)
```
fredrik-bakke, line 74: I've added two new spacing rules that we can choose to enable when the repository is in a better position for refactoring.

EgbertRijke: This occurs commonly with Sigma.

fredrik-bakke: Right, sorry, it implements an exception for that as well :)

EgbertRijke: Awesome!

fredrik-bakke: I had the rule written down for a while, so I forgot that I added the exception for that

scripts/generate_namespace_index_modules.py
```diff
@@ -34,24 +33,29 @@ def agda_file_filter(f): return utils.is_agda_file(
...
+def generate_agda_block(root, namespace, git_tracked_files):
+    return agda_block_template.format(namespace=namespace, imports=generate_imports(root, namespace, git_tracked_files))
 
 
 if __name__ == '__main__':
-    CHANGES_WERE_MADE_FLAG = 1
-    MISPLACED_TITLE_FLAG = 2
+    STATUS_FLAG_GIT_ERROR = 1
+    CHANGES_WERE_MADE_FLAG = 2
+    MISPLACED_TITLE_FLAG = 4
```
VojtechStep, line 43: It doesn't particularly matter here, but it's conventional to add new errors to the end of the list, not the front

scripts/generate_main_index_file.py
```diff
@@ -6,25 +6,36 @@
...
-    module_titles = tuple(map(utils.get_lagda_md_file_title, lagda_file_paths))
+    try:
+        git_tracked_files = utils.get_git_tracked_files()
+    except subprocess.CalledProcessError:
+        print('Failed to get Git-tracked files', file=sys.stderr)
+        sys.exit(STATUS_FLAG_GIT_ERROR)
 
-    module_mdfiles = tuple(
-        map(lambda m: utils.get_module_mdfile(namespace, m), modules))
+    root_path = pathlib.Path(root)
```
VojtechStep, line 27: How does this function even have access to `root`? 😃
Could you pass it as a parameter?

VojtechStep (commented):
I think the implementation will require a bit more fiddling to get right, because we traded one class of bugs (including untracked files) for another (including unstaged deleted/renamed files).

scripts/generate_main_index_file.py
fredrik-bakke: haha, yes :)

fredrik-bakke:
Thanks for the comments, and sorry for my messy python code. For some reason, my patience grows really thin whenever I have to touch it.

scripts/generate_namespace_index_modules.py
fredrik-bakke: I usually order them by in which order they can appear (when this makes sense), so if you get an error with a lower number than another, then that doesn't rule out that the later error can occur if the first one is fixed.

scripts/utils/__init__.py
```diff
@@ -206,3 +211,10 @@ def is_agda_opening_or_closing_tag(line):
     """
     tag_match = agda_block_tag_regex.match(line)
     return bool(tag_match), tag_match and bool(tag_match.group(1))
+
+
+def get_git_tracked_files():
+    # Get list of Git-tracked files
+    git_output = subprocess.check_output(['git', 'ls-files'], text=True)
```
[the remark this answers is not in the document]
fredrik-bakke: I've added some remarks to the `CONTRIBUTING.md` file. Let me know if you had something else/more in mind.

fredrik-bakke:
@VojtechStep If I recall correctly, you argued somewhere that this change in some sense trades one bug for another, and although both versions have their drawbacks, this PR at least makes the hooks consistent with each other, which I think is an overall improvement.

fredrik-bakke:
Ah, never mind, this one's still unfinished.

scripts/generate_main_index_file.py
```diff
@@ -6,25 +6,36 @@
...
+
+    # Filter out the relevant files in the given namespace
+    relevant_files = (
+        f for f in git_tracked_files if namespace_path in f.parents)
+
+    lagda_file_paths = (f for f in relevant_files if utils.is_agda_file(f))
+    modules = (p.name for p in lagda_file_paths)
+    module_titles = (utils.get_lagda_md_file_title(f)
+                     for f in lagda_file_paths)
+    module_mdfiles = (utils.get_module_mdfile(namespace, m) for m in modules)
```
VojtechStep, line 38: The generator expressions don't work here, because they are consume-once: for example iterating over `modules` consumes the underlying `lagda_file_paths`, which then leaves `module_titles` empty. I'm pretty sure that with the current implementation, only `relevant_files` can be left as a generator

scripts/generate_main_index_file.py
```diff
@@ -67,9 +78,7 @@ def generate_index(root, header):
     namespaces = sorted(set(utils.get_subdirectories_recursive(root)))
```
VojtechStep, line 78: `get_subdirectories_recursive` now assumes that its argument is a `pathlib.Path`, which `root` isn't, so this list is currently always empty (apart from the hardcoded template I get an empty SUMMARY.md locally). Root is converted into `root_path` elsewhere, so it makes sense to have it be a Path already in `main`.

VojtechStep: Then `root_path` can be removed, and `os.path.join(root, namespace)` can be replaced by `str(namespace_path)` on line 67, but I can't add a comment there because that part isn't changed in the PR

CONTRIBUTING.md
```diff
@@ -6,9 +6,9 @@ through their usage and help ensure your contributions meet our standards.
 ## The `make pre-commit` tool
 
 The `make pre-commit` tool checks library conventions and performs basic tasks.
-Before you run it, ensure your new files have been added to the `git`
-repository. Installation instructions for the `make pre-commit` tool can be
-found [here](HOWTO-INSTALL.md#after-the-setup).
+Before you run it, ensure that your files are properly tracked by git by staging
+any relevant file additions or deletions. Installation instructions for the
+`make pre-commit` tool can be found [here](HOWTO-INSTALL.md#after-the-setup).
```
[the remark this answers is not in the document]
fredrik-bakke: Ah, thanks for spotting that

fredrik-bakke:
I think I've adequately addressed all of your comments now, @VojtechStep. Thank you for your invaluable guidance, and let me know if there is anything more you think should be changed.
By the way, I think you deserve co-authorship over this PR, if you are comfortable with having your name on it.

PR #852, opened 2023-10-17
Classifying invertible maps

maybemabeline:
I showed that the type of invertible maps is equivalent to the type of equivalences, together with a loop at the underlying map, as discussed with Egbert. Input on the name "looped equivalence" is welcome, as I'm not sure if I like it very much.


src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+A looped equivalence is an equivalence together with a loop at the underlying
+map
+
+```agda
+module _
+  {l1 l2 : Level} (A : UU l1) (B : UU l2)
+  where
+
+  looped-equivalence : UU (l1 ⊔ l2)
+  looped-equivalence = Σ (A ≃ B) (λ f → map-equiv f ＝ map-equiv f)
```
EgbertRijke, line 314: I understand your concern that "looped equivalence" sounds a little weird. One way to get around such a name is to leave this type unnamed, and use the type `free-loop (A ≃ B)` instead. This might require you to fiddle a little bit with cyclic dependencies, which are usually a source of headaches so I don't recommend it.

On the other hand, this suggests alternative terminology: it is a free loop in the type of equivalences. We could call the type `free-loop-equiv` and then the two projections would be called `equiv-free-loop-equiv` and `loop-free-loop-equiv`.

That's one suggestion. Have a look whether you like it better, and perhaps @fredrik-bakke can weigh in too.

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+```
+
+#### The type of invertible equivalences is equivalent to the type of invertible maps
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2}
+  where
+
+  tot-is-equiv-is-invertible : (invertible-map A B) → Prop (l1 ⊔ l2)
```
EgbertRijke, line 334: There is no `tot` in either the specification or the definition, so it is a little confusing to see it in the name. The name `is-equiv-prop-invertible-map` is available, so we could call this proposition that. This name is constructed as follows: this entry asserts the proposition that something is an equivalence, so that's the first part of the name; the input of this entry is an invertible map, so `invertible-map` is mentioned as input in the second part of the name. Alternatively, you can read it as: given an invertible map as input, we will return a proposition, so this builds `prop-invertible-map`; the name of this proposition is `is-equiv`, so that builds `is-equiv-prop-invertible-map`.

On the other hand, given our naming scheme this name could easily be confused with the function that takes an invertible map as input, and returns the fact that it is an equivalence.

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+              ( map-equiv f)
+              ( section-is-equiv (is-equiv-map-equiv f)) ·r
+            ( map-equiv f)))
+          ( map-equiv f)))))
+```
+
+#### The equivalence
+
+```agda
+  equiv-looped-equivalence-invertible-map :
```
EgbertRijke, line 394: We started moving to a new naming scheme where we would call such equivalences `compute-...`. This naming scheme is not entirely crystallized yet, but you could call this `compute-free-loop-equiv` or `compute-invertible-map` according to taste:)

Perhaps @fredrik-bakke has a preference.

src/foundation/invertible-maps.lagda.md
EgbertRijke: I see, you're right. That's a subtle difference.

But that raises a question for me. For your result that the total space of all invertible maps in a universe is the type of 2-spheres in the universe, won't you need free loops in the equivalences?

It should be an easy conversion to actually get from these looped equivalences to the free loops in the equivalences, because equality of equivalences is the same as equality of their underlying functions.

maybemabeline: That's also true. In that case I agree it'd be best to work with free loops and include another step in the construction that lifts the loop on the function to the entire equivalence. 

> because equality of equivalences is the same as equality of their underlying functions

What you're talking about is the subtype identity principle, right?

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+    ( equiv-inclusion-is-full-subtype
+      ( tot-is-equiv-is-invertible)
+      ( is-full-subtype-tot-is-equiv-is-invertible)) ∘e
+    ( equiv-right-swap-Σ)
+```
+
+#### The type of looped equivalences is equivalent to the type of invertible equivalences
+
+```agda
+  equiv-is-retraction-section-is-invertible :
```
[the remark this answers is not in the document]
EgbertRijke: Well spotted!

src/foundation/invertible-maps.lagda.md
VojtechStep: We call it equivalence extensionality, but it's all the same https://unimath.github.io/agda-unimath/foundation.equivalence-extensionality.html#1247

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+    looped-equivalence A B ≃ Σ (A ≃ B) (is-invertible ∘ map-equiv)
+  equiv-looped-equivalence-invertible-equivalence =
+    ( equiv-tot
+      ( equiv-is-retraction-section-is-invertible ∘ map-equiv)) ∘e
+    ( equiv-tot
+      ( λ f →
+        ( inv-left-unit-law-Σ-is-contr
+          ( is-contr-section-is-equiv (is-equiv-map-equiv f))
+          ( section-is-equiv (is-equiv-map-equiv f))) ∘e
+        ( equiv-funext ∘e
```
EgbertRijke, line 373: So, if you would go with free loops in the type of equivalences, then at this step you would be applying equivalence extensionality, as @VojtechStep pointed out. That would skip the subtype identity principle.

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +298,106 @@ abstract
...
+    Σ (section f) (λ g → (map-section f g) ∘ f ~ id) ≃ is-invertible f
+  equiv-is-retraction-section-is-invertible f =
+    associative-Σ
+      ( B → A)
+      ( λ g → f ∘ g ~ id)
+      ( λ g → (map-section f g) ∘ f ~ id)
+
+  equiv-looped-equivalence-invertible-equivalence :
+    looped-equivalence A B ≃ Σ (A ≃ B) (is-invertible ∘ map-equiv)
+  equiv-looped-equivalence-invertible-equivalence =
```
EgbertRijke, line 365: Perhaps it would be good to also write down informally what this proof does, above this code block. E.g., you could write

```text
**Proof:** Your proof that looped equivalences are equivalently described as invertible equivalences
```

src/foundation/invertible-maps.lagda.md
fredrik-bakke: `compute-` is a nice prefix, but `compute-invertible-map` seems hard to interpret. Isn't the current name fine?

src/foundation/invertible-maps.lagda.md
fredrik-bakke: Here's my two cents: It seems like this definition should be done in terms of `htpy-equiv`, i.e. that a "looped equivalence" is an equivalence equipped with a loop-homotopy, not a loop-identification. Moreover, for the name, we have this naming scheme where we say a type equipped with an automorphism is a `Type-With-Automorphism`. This scheme could transfer to this setting, in which case it would be called something like `equiv-with-loop`. I'm not super fond of that naming scheme, but it's at least one possible source of inspiration.

fredrik-bakke:
This is a cool result by the way. Thanks for formalizing it! :)
We can use it to prove the instructive results that
- The type of invertibility proofs is an `n`-type when the codomain is an `n+1`-type.
- By extension, the type of invertibility proofs is equivalent to the type of equivalences when the codomain is a set.
- Not every type of invertibility proofs is an `n`-type when the codomain is not an `n+1`-type.

EgbertRijke:
> This is a cool result by the way. Thanks for formalizing it! :) We can use it to prove the instructive results that
> 
> * The type of invertibility proofs is an `n`-type when the codomain is an `n+1`-type.
> * By extension, the type of invertible maps is equivalent to the type of equivalences when the codomain is a set.
> * Not every type of invertibility proofs is an `n`-type when the codomain is not an `n+1`-type.

This current result shows that the type of invertible maps from `A` to `B` is equivalent to the type of maps from the circle into `A ≃ B`, by the universal property of the circle. Mabel will extend this result to showing that the type of all invertible maps in a universe `UU` is equivalent to the type of maps from the 2-sphere into the universe `UU`. This is also why she used identifications and not homotopies of equivalences.

fredrik-bakke:
Gotcha, very cool!

src/foundation/invertible-maps.lagda.md
maybemabeline: I was also planning to expand things a bit using equivalence reasoning, instead of just using `\circ e` everywhere.

src/foundation/invertible-maps.lagda.md
maybemabeline: Maybe `subtype-is-equiv-invertible-map`, since we're considering this prop as a subtype? I think `is-equiv-prop-invertible-map` should also be fine, a brief misinterpretation of the name of this auxiliary function wouldn't have any real consequences.

src/foundation/invertible-maps.lagda.md
EgbertRijke: Oh I quite like proofs without equivalence reasoning. They’re a lot shorter

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +301,98 @@ abstract
       ( is-equiv-map-associative-Σ _ _ _)
 ```
 
+### The type of invertible maps is equivalent to the type of free loops on equivalences
+
+#### The type of invertible equivalences is equivalent to the type of invertible maps
+
+**Proof:** Since every invertible map is an equivalence, the Sigma type of
```
EgbertRijke, line 308: Probably in informal text it is better to write `` `Σ`-type `` rather than `Sigma type`. Would you be willing to make that change throughout this file?

EgbertRijke: Also: That's super awesome that you added informal proofs to all the formal proofs. Great work!

src/foundation/invertible-maps.lagda.md
```diff
@@ -294,6 +301,98 @@ abstract
...
+#### The type of free loops on equivalences is equivalent to the type of invertible equivalences
+
+**Proof:** Firstly, associating the order of Sigma types shows that a function
+being invertible is equivalent to it having a section, such that this section is
+also its retraction. Now, since equivalences have a contractible type of
+sections, a proof of invertibility of the underlying map `f` of an equivalence
+contracts to just a single homotopy `g ∘ f ~ id`, showing that a section `g` of
+`f` is also its retraction. As `g` is a section, composing on the left with `f`
+and canceling `f ∘ g` yields a loop `f ~ f`. By equivalence extensionality, this
+loop may be lifted to a loop on the entire equivalence.
```
EgbertRijke, line 345: Well done!

EgbertRijke:
Great! If you make that one change about Sigma types in the informal text, and update this branch with master, than this PR will be ready to merge. This is a very nice pull request!

PR #857, opened 2023-10-18
If a map is an epimorphism, then its codagional is an equivalence.

tomdjong:
I tried to stick to the `agda-unimath` conventions, but I'm new to it, so any feedback is appreciated!

I will also formalise the converse, but wanted to get comments on this first.


tomdjong:
I just noticed that `pre-commit` messed up my diagram. What's the preferred way to include diagrams?

VojtechStep:
You put them in a text code block. I'm not sure how to show it here, since GitHub will interpret the Markdown. It's as if you were writing an Agda code block, but instead of `agda` in the header, you use `text`

tomdjong:
Thanks @VojtechStep! Fixed now.

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
   is-epimorphism : (A → B) → UUω
-  is-epimorphism f = {l : Level} (X : UU l) → is-emb (precomp f X)
+  is-epimorphism f = ∀ {l : Level} → is-epimorphism-level f l
+```
+
+## Properties
+
+### Epimorphisms, pushouts and codiagonals
+
+```text
```
VojtechStep, line 63: We only put the diagram inside the `text` code block

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} (f : A → B) (X : UU l3)
+  where
+
+  equiv-fibers-of-precomp-cocone :
+    Σ (B → X) (λ g → fiber (precomp f X) (g ∘ f)) ≃ cocone f f X
+  equiv-fibers-of-precomp-cocone =
+    equiv-comp
+      ( equiv-Σ _
+          ( id-equiv)
```
VojtechStep, line 85: These arguments should be less indented, so that the opening paranthesis aligns with `e`

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+          ( λ g → equiv-Σ _ id-equiv (λ h → equiv-funext)))
+      ( equiv-Σ _
+          ( id-equiv)
+          ( λ g → inv-equiv
```
VojtechStep, line 89: If we need to break a lambda, we do so right after the arrow, so `inv-equiv` goes on the next line, horizontally aligned with the lamda

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
-  {l1 l2 : Level} (l : Level) {A : UU l1} {B : UU l2}
+  {l1 l2 : Level} {A : UU l1} {B : UU l2}
   where
 
+  is-epimorphism-level : (A → B) → (l : Level) → UU (l1 ⊔ l2 ⊔ lsuc l)
+  is-epimorphism-level f l = (X : UU l) → is-emb (precomp f X)
+
   is-epimorphism : (A → B) → UUω
-  is-epimorphism f = {l : Level} (X : UU l) → is-emb (precomp f X)
+  is-epimorphism f = ∀ {l : Level} → is-epimorphism-level f l
```
VojtechStep, line 56: We don't use foralls

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+          ( λ g → equiv-Σ _ id-equiv (λ h → equiv-funext)))
+      ( equiv-Σ _
+          ( id-equiv)
+          ( λ g → inv-equiv
+                    ( inv-equiv
+                        ( equiv-fiber (precomp f X) (g ∘ f)))))
+
+  diagonal-into-fibers-of-precomp :
+    (B → X) → Σ (B → X) (λ g → fiber (precomp f X) (g ∘ f))
+  diagonal-into-fibers-of-precomp = map-section-family (λ g → g , refl)
```
VojtechStep, line 95: Unless I'm missing some implicit arguments, that lambda should be replaced with `id`

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B)
+  where
+
+  is-pushout-is-epimorphism :
+    is-epimorphism f →
+    {l : Level} → universal-property-pushout l f f (cocone-codiagonal-map f)
+  is-pushout-is-epimorphism e X =
+    is-equiv-comp-htpy
+      (cocone-map f f (cocone-codiagonal-map f))
```
VojtechStep, line 115: Missing space `( cocone-map`

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+```
+
+If the map `f : A → B` is epi, then its codiagonal is an equivalence.
+
+```agda
+  codiagonal-is-equiv-is-epimorphism :
+    is-epimorphism f → is-equiv (codiagonal-map f)
+  codiagonal-is-equiv-is-epimorphism e =
+    is-equiv-up-pushout-up-pushout
+      ( f)
```
VojtechStep, line 130: These two `f`s can be written without parantheses and joined with the first line - we do that for leading single-lettter arguments

VojtechStep (commented):
Thank you for the contribution! I left some general comments on the style. Some of them occur multiple times which I didn't mark, because I'm on my phone on my way to class 😅

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
 
 ```agda
 module _
-  {l1 l2 : Level} (l : Level) {A : UU l1} {B : UU l2}
   where
 
+  is-epimorphism-level : (A → B) → (l : Level) → UU (l1 ⊔ l2 ⊔ lsuc l)
```
EgbertRijke, line 52: Usually we declare universe levels before anything else

EgbertRijke: (if they're explicit, that is)

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,97 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+### Epimorphisms, pushouts and codiagonals
+
+```text
+If the map `f : A → B` is epi, then the commutative square
+
+      f
+    A → B
+  f ↓   ↓ id
+    B → B
+      id
```
EgbertRijke, line 70: Whooo this feels like old-school nlab style 🤩

src/foundation/epimorphisms.lagda.md
tomdjong: I thought arguments were indented. Could you explain the precise convention (later) please?

src/foundation/epimorphisms.lagda.md
tomdjong: But it's `λ g → (g , refl)`. Should I write parentheses to clarify?

src/foundation/epimorphisms.lagda.md
EgbertRijke: For your reference, usually we make our diagrams with ascii art, like here:

https://unimath.github.io/agda-unimath/foundation-core.commuting-squares-of-maps.html

But yours is fine too 👍

src/foundation/epimorphisms.lagda.md
VojtechStep: Oh, right - in this case I'd appreciate it if you could wrap the `g, refl` in paramtheses

src/foundation/epimorphisms.lagda.md
VojtechStep: Parantheses and argument indentation have a funny relationship. Arguments are indeed indented, but you'll notice that in
```agda
( function
    ( argument))
```
we would have four spaces between the first non-whitespace character on one line and the first non-whitespace character on the other line. I wasn't around when this convention was introduced, but I rationalize it to myself by considering the opening paranthesis and the following name as one token; then this convention is consistent with indenting arguments, because here
```agda
( function
  ( argument))
```
The `( function` "is a function" and the `( argument` "is an argument".
Ultimately I think this is just what I tell myself so that I can write code which doesn't run rapidly to the right and still sleep at night.

tomdjong: Thank you!

VojtechStep: I think it's not linkable because of the colon in the heading

tomdjong: Ah excellent, thanks. I had only checked `CONTRIBUTING.md` so far. Anyway, I think the code conforms to the conventions now.

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+  is-epimorphism-level l f = (X : UU l) → is-emb (precomp f X)
+
   is-epimorphism : (A → B) → UUω
-  is-epimorphism f = {l : Level} (X : UU l) → is-emb (precomp f X)
+  is-epimorphism f = {l : Level} → is-epimorphism-level l f
+```
+
+## Properties
+
+### Epimorphisms, pushouts and codiagonals
```
EgbertRijke, line 61: Usually we use section headers to say what property is being proved. Since this is quite a large equivalence, I would have make sections like `### The codiagonal of an epimorphism is an equivalence` or `### A map is an epimorphism if and only if its codiagonal is an equivalence`

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+  equiv-fibers-of-precomp-cocone :
+    Σ (B → X) (λ g → fiber (precomp f X) (g ∘ f)) ≃ cocone f f X
+  equiv-fibers-of-precomp-cocone =
+    equiv-comp
+      ( equiv-Σ _
+        ( id-equiv)
+        ( λ g → equiv-Σ _ id-equiv (λ h → equiv-funext)))
+      ( equiv-Σ _
+        ( id-equiv)
+        ( λ g → equiv-fiber (precomp f X) (g ∘ f)))
```
EgbertRijke, line 89: The map 

```
equiv-Σ _
  ( id-equiv)
  ( λ g → equiv-fiber (precomp f X) (g ∘ f))
```

has an equivalence in the base, so there is a quicker definition available for this purpose: `equiv-tot`. If I'm not mistaken, this entire argment can be replaced with

```suggestion
      ( equiv-tot (λ g → equiv-fiber (precomp f X) (g ∘ f)))
```

EgbertRijke, line 89: Oh, actually the entire equivalence can be done like that:

```suggestion
    equiv-tot
      ( λ g →
        ( equiv-tot (λ h → equiv-funext))) ∘e
        ( equiv-fiber (precomp f X) (g ∘ f)))
```

Note that this is a risky github suggestion to accept, since I did not typecheck it. But hopefully the idea is clear.

tomdjong: Yes, thanks! I'm implementing it myself now to ensure that things typecheck.

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+    is-epimorphism f →
+    {l : Level} → universal-property-pushout l f f (cocone-codiagonal-map f)
+  is-pushout-is-epimorphism e X =
+    is-equiv-comp-htpy
+      ( cocone-map f f (cocone-codiagonal-map f))
+      ( map-equiv (equiv-fibers-of-precomp-cocone f X))
+      ( diagonal-into-fibers-of-precomp f X)
+      ( refl-htpy)
+      ( is-equiv-diagonal-into-fibers-of-precomp-is-epimorphism f X e)
+      ( is-equiv-map-equiv (equiv-fibers-of-precomp-cocone f X))
```
EgbertRijke, line 119: Since the commuting square commutes by `refl-htpy` you can just use `is-equiv-comp` with a lot fewer arguments

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+      ( λ g →
+        is-proof-irrelevant-is-prop
+          ( is-prop-map-is-emb (e X) (g ∘ f))
+          ( g , refl))
+
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B)
+  where
+
+  is-pushout-is-epimorphism :
```
VojtechStep, line 109: `is-pushout` is actually an existing property that we have in the library, and it's slightly different from `universal-property-pushout` (it's small and specialized to the standard pushout), so I'd call this `universal-property-pushout-is-epimorphism`.

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+    is-epimorphism f →
+    {l : Level} → universal-property-pushout l f f (cocone-codiagonal-map f)
+  is-pushout-is-epimorphism e X =
+    is-equiv-comp-htpy
```
VojtechStep, line 113: This is just a minor nitpick - `is-equiv-comp-htpy` where the homotopy is trivial can be replaced with `is-equiv-comp`, but I don't think it's necessary if it's not immediately clear that the maps compose definitionally (but in that case I think it would be useful to have the triangle as a separate definition to at least capture the type, even if the definition itself will be `refl-htpy`)

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
 
 ```agda
```
EgbertRijke, line 47: Probably it will be good to split this code block up into two sections:

````agda
### Epimorphisms with respect to a specified universe

```agda
is-epimorphism-Level : ...
```

### Epimorphisms

```agda
is-epimorphism : ...
```
````

File: src/foundation/epimorphisms.lagda.md [hunk 16]
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+  equiv-fibers-of-precomp-cocone =
+    equiv-comp
+      ( equiv-Σ _
+        ( id-equiv)
+        ( λ g → equiv-Σ _ id-equiv (λ h → equiv-funext)))
+      ( equiv-Σ _
+        ( id-equiv)
+        ( λ g → equiv-fiber (precomp f X) (g ∘ f)))
+
+  diagonal-into-fibers-of-precomp :
```
VojtechStep, line 91: I'm personally not a fan of `-into-` and `-of-` in names, I think we use them primarily in module names, but I don't have better alternatives in mind here, so feel free to ignore me on this one.

src/foundation/epimorphisms.lagda.md
```diff
@@ -30,11 +46,94 @@ is an [embedding](foundation-core.embeddings.md) for every type `X`.
...
+      ( diagonal-into-fibers-of-precomp f X)
+      ( refl-htpy)
+      ( is-equiv-diagonal-into-fibers-of-precomp-is-epimorphism f X e)
+      ( is-equiv-map-equiv (equiv-fibers-of-precomp-cocone f X))
+```
+
+If the map `f : A → B` is epi, then its codiagonal is an equivalence.
+
+```agda
+  codiagonal-is-equiv-is-epimorphism :
```
EgbertRijke, line 125: This entry concludes `is-equiv codiagonal` from `is-epimporhism`, so the agda-unimath name for it is `is-equiv-codiagonal-is-epimorphism`

EgbertRijke: Actually, perhaps we should disambiguate a bit more and make the name

```
is-equiv-codiagonal-map-is-epimorphism
```

EgbertRijke:
Btw sorry that we're splitting hairs now. Your PR is great! And one reason that we're giving you so many comments is to teach you the ways of our library:)

src/foundation/epimorphisms.lagda.md
EgbertRijke: I agree. At least `fibers-of-precomp` can easily be replaced with `fiber-precomp` since that's what it looks like when you remove the arguments and the spaces. I'd have to think about what a better name for `diagonal-into` is, but we can leave it as it is for now.

src/foundation/epimorphisms.lagda.md
tomdjong: I adopted `is-equiv-comp` following Egbert's suggestion, but I agree with your caveats.

src/foundation/epimorphisms.lagda.md
tomdjong: I adopted your suggested name and added the proper `is-pushout-is-epimorphism` separately.

tomdjong:
No worries! This is exactly what I was asking for when I created the PR.

EgbertRijke:
Awesome, let's go!

PR #858, opened 2023-10-18
A map is an embedding if and only if it has contractible fibers at values

tomdjong:
The versions with `fiber'` likely come in useful too, so I included them separately.

### DIFF
--- a/src/foundation/embeddings.lagda.md
+++ b/src/foundation/embeddings.lagda.md
@@ -357,7 +357,8 @@ module _
 ### A map is an embedding if and only if it has contractible fibers at values
 
 ```agda
-module _ {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B)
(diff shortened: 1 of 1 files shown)


src/foundation/embeddings.lagda.md
```diff
@@ -351,3 +353,38 @@ module _
       ( is-equiv-map-inv-is-equiv L)
       ( M)
 ```
+
+### A map is an embedding if and only if it has contractible fibers at values
+
+```agda
+module _ {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B)
```
EgbertRijke, line 360: We usually put the variables of an anonymous module on a new line, and `where` on a new line under that
```suggestion
module _
  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B)
```

PR #860, opened 2023-10-18
A map is an epimorphism if its codiagonal is an equivalence

tomdjong:


fredrik-bakke:
Hi Tom, welcome (back) to agda-unimath! Thanks for the contributions today :)

EgbertRijke:
I'll review later tonight, because I have to go now.

src/foundation/epimorphisms.lagda.md
```diff
@@ -83,12 +84,18 @@ module _
   {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} (f : A → B) (X : UU l3)
   where
 
+  equiv-fibers-precomp-cocone-fiberwise :
```
EgbertRijke, line 87: The computation of the fibers of the precompostion map belongs in `foundation.functoriality-function-types` in a section `### The fibers of precomp`, directly after the section `### The fibers of `postcomp`

src/foundation/epimorphisms.lagda.md
```diff
@@ -140,6 +147,55 @@ If the map `f : A → B` is epi, then its codiagonal is an equivalence.
...
+        ( htpy-cocone-map-universal-property-pushout f f
+          ( cocone-pushout f f)
+          ( up-pushout f f)
+          ( cocone-codiagonal-map f))
+        ( e)
+        ( up-pushout f f))
+
+  is-epimorphism-is-pushout :
+    is-pushout f f (cocone-codiagonal-map f) → is-epimorphism f
+  is-epimorphism-is-pushout = is-epimorphism-is-equiv-codiagonal-map
```
EgbertRijke, line 196: The condition `is-pushout f f (cocone-codiagonal-map f)` is definitionally equal to the condition `is-equiv (codiagonal-map f)`, so is this only recorded so that the user doesn't have to think about the fact that these two conditions are literally the same?

EgbertRijke (approved):
Great pull request! I only had some minor comments, but it seems that you're already up to speed with how we do things over here. Well done!

src/foundation/epimorphisms.lagda.md
tomdjong: Yes and even though they are the same, conceptually they feel different to me.

src/foundation/epimorphisms.lagda.md
tomdjong: Good idea! I ran into a problem though.

The section you mentioned is actually in `foundation-core.functoriality-function-types`, but I'm using `coherence-square-maps` in the statement of `equiv-fibers-precomp-cocone-fiberwise` (to be renamed to `compute-fiber-precomp`) which comes from `foundation-core.commutative-squares-of-maps` which itself imports `foundation-core.functoriality-function-types`. How do you avoid this cyclic dependency?

tomdjong: Should I put the result about the fibers of precomposition in `foundation.functoriality-function-types` for this reason maybe?

src/foundation/epimorphisms.lagda.md
EgbertRijke: Here’s to people saying that univalent foundations can’t distinguish between equivalent concepts 🍻

src/foundation/epimorphisms.lagda.md
EgbertRijke: Yeah, that's fine.

These are indications to me that `postcomp` and `precomp` should be factored out into their own files, but you don't have to do that here.

tomdjong:
Reorganised as requested.

EgbertRijke:
Will be merged when the checks are done

PR #862, opened 2023-10-19
Level-universe compatibility patch

fredrik-bakke:
Makes the repository compatible with the new `--level-universe` flag introduced with Agda 2.6.4.


EgbertRijke:
Excellent, thank you for doing this!

PR #863, opened 2023-10-19
Characteristic subgroups

EgbertRijke:
This PR adds a definition of characteristic subgroups to the library. I proof of the expected properties is expected to follow in a later PR

### DIFF
--- a/src/group-theory/characteristic-subgroups.lagda.md
+++ b/src/group-theory/characteristic-subgroups.lagda.md
@@ -1,0 +1,60 @@
+# Characteristic subgroups
+
+```agda
+{-# OPTIONS --allow-unsolved-metas #-}
+
+module group-theory.characteristic-subgroups where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.propositions
+open import foundation.universe-levels
+
+open import group-theory.groups
+open import group-theory.images-of-group-homomorphisms
+open import group-theory.isomorphisms-groups
+open import group-theory.subgroups
+```
+
+</details>
+
+## Idea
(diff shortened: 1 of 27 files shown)


src/group-theory/characteristic-subgroups.lagda.md
```diff
@@ -0,0 +1,75 @@
...
+
+</details>
+
+## Idea
+
+A **characteristic subgroup** of a [group](group-theory.groups.md) `G` is a
+[subgroup](group-theory.subgroups.md) `H` of `G` such that `e H ⊆ H` for every
+[isomorphism](group-theory.isomorphisms-groups.md) `e : G ≅ G`. The seemingly
+stronger condition, which asserts that `e H ＝ H` for every isomorphism
+`e : G ≅ G` is equivalent.
```
fredrik-bakke, line 27: The character `e` is used frequently for the unit of the group, which defines a particular automorphism, so it seems best to use a different symbol here.

```suggestion
[subgroup](group-theory.subgroups.md) `H` of `G` such that `ϕ H ⊆ H` for every
[isomorphism](group-theory.isomorphisms-groups.md) `ϕ : G ≅ G`. The seemingly
stronger condition, which asserts that `ϕ H ＝ H` for every isomorphism
`ϕ : G ≅ G` is equivalent.
```

EgbertRijke: Ok, I'll use a nongreek letter though

fredrik-bakke: Haha, that's fine

PR #861, opened 2023-10-18
Subcategories

fredrik-bakke:

PR #864, opened 2023-10-19
Nonunital precategories

fredrik-bakke:
Defines nonunital precategories and refactors out composition operations on binary families of sets. Builds on #861.


fredrik-bakke:
This one will have to be rebased after #861 is merged before it should be reviewed.

EgbertRijke:
Awesome! I'll work towards getting #861 merged.

fredrik-bakke:
How much would you be against the idea of reassociating the definition of precategories so that it's a sigma type of composition operations that are associative and unital, instead of an associative composition operation that is unital? @EgbertRijke

fredrik-bakke:
I've already factored out composition operations on binary families of sets into a new file to define nonunital precategories, so there will be some friction when merging other PRs that are doing category theory in parallel to this one regardless.

EgbertRijke:
> How much would you be against the idea of reassociating the definition of precategories so that it's a sigma type of composition operations that are associative and unital, instead of an associative composition operation that is unital? @EgbertRijke

In the rest of the library, in the algebra part in particular, we also first have associative multiplication structure and on top of that unitality structure. To me an associative multiplicative/composition structure seems worthwhile to factor out in that way, because there might be related associatve multiplicative structures which aren't unital. Especially in the light of your recent thoughts on nonunital categories I don't really see why you would want to undo this nice setup. 

But that doesn't mean that you don't have your reasons. Can you tell me why you were thinking of doing this?

EgbertRijke:
If your concern is that it is now a subtype of a subtype of composition structures, then my response would be that instead of reassociating this sigma we should write better infrastructure for handling subtypes of subtypes. We shouldn't feel like we have to avoid subtypes of subtypes, they're a completely natural occurrence in mathematics. Instead, we should add stuff to our library so that we can handle them smoothly.

fredrik-bakke:
Alright, let me try to lay out my reasons.

1. With the current definition, it seems like associativity in some sense takes precedence over unitality. Indeed, as you point out, I am considering nonunital precategories, and in algebra associativity is maybe more commonly considered than unitality. I still find it strange that we give associativity precedence over unitality. After all, unitality for set-level structures is independent to associativity, and when we go to higher structures, we could still consider objects which are first unital and then associative in a way that coheres with unitality.
2. By reassociating, we clearly separate the structure of precategories from the properties.
3. I've always found it a little awkward that we have to define an associative composition operation before we can consider unitality of the composition structure. They should be independent, so we should be able to first define the unitality. (This can be fixed irrespective of the reassociation, but is still bundled in the problem)

fredrik-bakke:
I already resolve item 3 in this PR btw

EgbertRijke:
Excellent, thank you for explaining why you were considering this.

Your point in 1 is well taken. Indeed, with many structures unitality comes before associativity and the other associahedra.

Regarding 2, it is actually more common to put the identity morphisms among the structure. Is that what you propose? Because being unital is currently formulated as a property, so the identity morphisms are part of that property. I don't think I feel much for the old approach where you put identity morphisms among the structure, I would actually be against this.

Regarding 3, you are always free what you define first. In fact, it is anyway good practice to define the components of the sigma type separately first before you bundle them up into the structure of a category. So I don't really see much merit to this point.

I think I would still be against this refactoring.

fredrik-bakke:
I was not proposing putting identity morphisms among structure although it's an interesting discussion point, since the preservation of identities is not automatic for maps that preserve composition. And about point 3, that was just a very minor point. The previous definition of unitality required that the composition operation was associative.

EgbertRijke:
> I was not proposing putting identity morphisms among structure although it's an interesting discussion point, since the preservation of identities is not automatic for maps that preserve composition. And about point 3, that was just a very minor point. The previous definition of unitality required that the composition operation was associative.

That's in fact a good argument to put it under structure, hehe.

EgbertRijke:
I was thinking that if 1 is a convincing argument, then the type could be

```
Sigma ((comp,unital) : Sigma composition-structure is-unital) (is-associative comp)
```

fredrik-bakke:
Another purely aesthetic reason is that I'm not such a fan of weirdly associated sigma types.

fredrik-bakke:
I'd prefer `Sigma (comp) ((unitality) x (associativity))` or  `Sigma (comp) ((associativity) x (unitality))`

fredrik-bakke:
With `Sigma (comp) ((unitality) x (associativity))` we can be agnostic about whether we consider unitality as structure or property :)

EgbertRijke:
I'd still prefer not changing it

fredrik-bakke:
Alright, then let's not. Thanks for still hearing me out!

EgbertRijke:
Yeah, thanks for bringing it up! I'll think about it some more later on.

fredrik-bakke:
Here's another point, although I'm sure you won't care much for it (I still do). By reassociating to one of my proposed definitions, it will be more natural to omit defining `unital-composition-operation-...` or `associative-composition-operation-...` for any particular precategory or category. These definitions are mostly useless aside from as intermediate definitions in defining (pre)categories, and bloat the files and make the formalization process longer than it strictly has to.

fredrik-bakke:
Yet another point is I think one of the proposed definitions would be easier to memorize.

EgbertRijke:
> Yet another point is I think one of the proposed definitions would be easier to memorize.

The infrastructure should be so good that you don't need to memorize the definition.

fredrik-bakke:
Never mind. It seems like we're talking past each other at this point.

Anyways, this PR is ready for review. Hopefully, it won't be as hard to review as the last one. Sorry to be putting so much on your plate at the moment.

EgbertRijke:
Is it ready for review?:)

fredrik-bakke:
Yep :)

src/category-theory/categories.lagda.md
```diff
@@ -33,7 +36,9 @@ them, by the J-rule. A precategory is a category if this function, called
 being a category is a [proposition](foundation-core.propositions.md) since
 `is-equiv` is a proposition.
 
-## Definition
+## Definitions
+
+### The predicate of being a category on precategories
```
EgbertRijke, line 41: I still find this sentence construction very hard to parse. If you need to specify what the predicate is on, would it be better to say "The predicate on ... of being a ..."?

```suggestion
### The predicate on precategories of being a category
```

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
```diff
@@ -0,0 +1,195 @@
+# Composition operations on binary families of sets
```
EgbertRijke, line 1: I think this belongs to `foundation`

fredrik-bakke: Really? That seems strange to me. How is this different from the case of putting `cocones-under-spans` in synthetic homotopy theory, or, god forbid, sigma types in `structured-types`?

fredrik-bakke: To me, category theory starts with the study of composition operations on binary families

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+### Being unital is a property of composition operations in binary families of sets
+
+Suppose `e e' : (x : A) → hom-set x x` are both right and left units with regard
+to composition. It is enough to show that `e e'` since the right and left unit
+laws are propositions (because all hom-types are sets). By function
+extensionality, it is enough to show that `e x ＝ e' x` for all `x : A`. But by
+the unit laws we have the following chain of equalities:
+`e x ＝ (e' x) ∘ (e x) ＝ e' x.`
```
EgbertRijke, line 139:
```suggestion
**Proof:** Suppose `e e' : (x : A) → hom-set x x` are both right and left units with regard
to composition. It is enough to show that `e e'` since the right and left unit
laws are propositions (because all hom-types are sets). By function
extensionality, it is enough to show that `e x ＝ e' x` for all `x : A`. But by
the unit laws we have the following chain of equalities:
`e x ＝ (e' x) ∘ (e x) ＝ e' x.`
```

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
```diff
@@ -0,0 +1,195 @@
...
+      ( is-associative-composition-operation-binary-family-Set hom-set comp-hom)
+  is-prop-is-associative-composition-operation-binary-family-Set =
+    is-prop-type-Prop
+      is-associative-prop-composition-operation-binary-family-Set
+```
+
+### Being unital is a property of composition operations in binary families of sets
+
+Suppose `e e' : (x : A) → hom-set x x` are both right and left units with regard
+to composition. It is enough to show that `e e'` since the right and left unit
```
EgbertRijke, line 135:
```suggestion
to composition. It is enough to show that `e e' ??` since the right and left unit
```

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
EgbertRijke: Isn't a simpler proof just that `e = e e' = e'`, and this suffices because the unit laws are properties?

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
fredrik-bakke: `e = e'` 😅

src/category-theory/nonunital-precategories.lagda.md
```diff
@@ -0,0 +1,185 @@
...
+  {l1 l2 : Level} (C : Nonunital-Precategory l1 l2)
+  {x y : obj-Nonunital-Precategory C}
+  (f : hom-Nonunital-Precategory C x y)
+  (z : obj-Nonunital-Precategory C) →
+  hom-Nonunital-Precategory C z x → hom-Nonunital-Precategory C z y
+postcomp-hom-Nonunital-Precategory C f z =
+  comp-hom-Nonunital-Precategory C f
+```
+
+### The predicate of being unital on nonunital precategories
```
EgbertRijke, line 134:
```suggestion
### The predicate on nonunital precategories of being unital
```

src/category-theory/precategories.lagda.md
```diff
@@ -189,59 +218,3 @@ module _
...
-## Properties
-
-### The property of having identity morphisms is a proposition
-
-Suppose `e e' : (x : A) → hom x x` are both right and left units with regard to
-composition. It is enough to show that `e = e'` since the right and left unit
-laws are propositions (because all hom-types are sets). By function
-extensionality, it is enough to show that `e x = e' x` for all `x : A`. But by
-the unit laws we have the following chain of equalities:
-`e x = (e' x) ∘ (e x) = e' x.`
```
EgbertRijke, line 202: Now I see where the proof came from:)

fredrik-bakke: Haha

fredrik-bakke:
Awesome! What do you think of having a comments section with open questions regarding a definition?

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
EgbertRijke: > Really? That seems strange to me. How is this different from the case of putting `cocones-under-spans` in synthetic homotopy theory, or, god forbid, sigma types in `structured-types`?

Oh yeah I forgot that I have some bargaining chips next time you want to propose a refactoring:)

EgbertRijke: > To me, category theory starts with the study of composition operations on binary families

Alright, then it can stay here. But there are lots of composition operations that aren't really categorical or even associative. By having this file in category theory, the person who is formalizing such operations has to think to look this file up in category theory.

We also have unital operations in foundation, because not all unital operations are monoids.

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
fredrik-bakke: Yeah, that works, but then we have to show the individual unit laws are properties first, and I don't feel like doing that at the moment.

EgbertRijke: Sure! Let's leave it like it is, since the point here was moving an existing proof to a different file, not to formalize a new proof.

src/category-theory/composition-operations-on-binary-families-of-sets.lagda.md
fredrik-bakke: Sounds like the file on unital operations should be in `structured-types` to be honest.

fredrik-bakke: Also, if such a person ended up defining their own version of composition operations I wouldn't mind it, even if the underlying types are definitionally equal.

EgbertRijke: No, types equipped with unital operations should be in `structured-types`. Indeed they already are there.

EgbertRijke:
> Awesome! What do you think of having a comments section with open questions regarding a definition?

Yeah, I saw that. That's very nice!

EgbertRijke:
Does auto-merge automatically disable itself if you would make new commits in this PR?

fredrik-bakke:
I'd assume so. Do you want me to try?

PR #866, opened 2023-10-20
Rename Axiom L to Preunivalence

fredrik-bakke:
Also tries out some new names for the statements of the axioms of univalence, K, and, well, L.


src/foundation-core/sets.lagda.md
```diff
@@ -55,24 +55,28 @@ module _
 ### A type is a set if and only if it satisfies Streicher's axiom K
 
 ```agda
-axiom-K : {l : Level} → UU l → UU l
-axiom-K A = (x : A) (p : x ＝ x) → refl ＝ p
+statement-axiom-K : {l : Level} → UU l → UU l
+statement-axiom-K A = (x : A) (p : x ＝ x) → refl ＝ p
```
EgbertRijke, line 59: This feels double to me. Axioms are already statements.

fredrik-bakke: Yeah, `statement-axiom-K` is kind of awkward. How about `axiom-type-K`?

fredrik-bakke:
This PR is quite self-contained in its current state, so I'll just mark it as ready for review immediately :)

EgbertRijke:
Great!

src/foundation-core/sets.lagda.md
EgbertRijke: That's even more awkward...

I think I mentioned this before, that I used `statement` only in moments of lack of creativity, where I couldn't come up with a better name. I'm not really fond of it, and I would prefer to see less `statement`-names in the library, rather than more, unless you propose a clear scheme on how to use `statement`.

EgbertRijke: On the other hand, I really like that you're getting rid of the all-caps entries in this PR

EgbertRijke: In the case of axiom K, I think just `axiom-K` suffices as a name. There's no real reason to fiddle with this good name.

src/foundation/preunivalence.lagda.md
```diff
@@ -0,0 +1,82 @@
...
+K imply preunivalence.
+
+## Definition
+
+```agda
+statement-preunivalence : {l : Level} (X Y : UU l) → UU (lsuc l)
+statement-preunivalence X Y = is-emb (equiv-eq {A = X} {B = Y})
+
+axiom-preunivalence : (l : Level) → UU (lsuc l)
+axiom-preunivalence l = (X Y : UU l) → statement-preunivalence X Y
```
EgbertRijke, line 39: Maybe I have an entirely different suggestion to handle this situation. How about

```suggestion
based-preunivalence : {l : Level} (X : UU l) → UU (lsuc l)
based-preunivalence X = (Y : UU l) → is-emb (equiv-eq {A = X} {B = Y})

axiom-preunivalence-Level : (l : Level) → UU (lsuc l)
axiom-preunivalence-Level l = (X Y : UU l) → statement-preunivalence X Y

axiom-preunivalence : UUomega
axiom-preunivalence = {l : Level} → axiom-preunivalence l
```

This corresponds to how I advised Tom recently to set up the definitions for epimorphisms. The predicate of being an epimorphism with respect to a specified universe level is called `is-epimorphism-Level` and the predicate of being an epimorphism (with respect to all universe levels) is called `is-epimorphism`.

The based variant seems more sensible than the statement approach. Currently `statement-preunivalence` *isn't the statement of preunivalence* so it really can't be accepted as a name for that entry.

EgbertRijke: We can also go with `axiom-based-preunivalence` if you like that.

EgbertRijke: Also I take the blame for the setup with `UNIVALENCE`. That's how I formalized 5 years ago:)

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -34,9 +34,9 @@ open import foundation-core.propositions
 
 ## Idea
 
-The universal property of identity types characterizes families of maps out of
-the identity type. This universal property is also known as the type theoretic
-Yoneda lemma.
+The **universal property of identity types** characterizes families of maps out
+of the [identity type](foundation-core.identity-types.md). This universal
+property is also known as the type theoretic Yoneda lemma.
```
EgbertRijke, line 39:
```suggestion
property is also known as the **type theoretic Yoneda lemma**.
```

src/foundation-core/sets.lagda.md
fredrik-bakke: The issue is that `axiom-K` is taken

src/foundation/preunivalence.lagda.md
fredrik-bakke: There's nothing to blame, I'm just glad we're changing the name 🙃
I still don't see why you don't want to have the definition of preunivalence between two types. It seems worthwhile to consider whether preunivalence holds between two particular types at times.

src/foundation-core/sets.lagda.md
```diff
@@ -55,24 +55,28 @@ module _
...
+
+axiom-K : (l : Level) → UU (lsuc l)
+axiom-K l = (A : UU l) → statement-axiom-K A
```
EgbertRijke, line 62: How about:

```suggestion
based-axiom-K : {l : Level} → UU l → UU l
based-axiom-K A = (x : A) (p : x ＝ x) → refl ＝ p

axiom-K-Level : (l : Level) → UU (lsuc l)
axiom-K-Level l = (A : UU l) → based-axiom-K A

axiom-K : UUomega
axiom-K = {l : Level} → axiom-K-Level l
```

I like to avoid `statement-axiom-K` the way you propose to use it, because it isn't the statement of axiom K. The statement of axiom K is a universal quantification over all types.

Indeed, before this pull request it was wrong too:)

src/foundation/preunivalence.lagda.md
EgbertRijke: I'm trying to come up with a sensible setup. Yours is better than the old one, but now that we're thinking about it I think there could be an even better setup. 

Instead of `statement-preunivalence` perhaps `instance-preunivalence` would be more accurate.

src/foundation-core/sets.lagda.md
fredrik-bakke: Based.

EgbertRijke: Perhaps `instance-axiom-K` could also be a suggestion, like with the instance of preunivalence I suggested in the other convo.

fredrik-bakke: `instance` is good.  like that!

EgbertRijke: Yes, absolutely. But let's do this one first without also doing universal properties. But I agree, I've been thinking about this for universal properties too.

fredrik-bakke:
Done :)

EgbertRijke:
I'm really happy with how this one came out. Thank you so much for taking the initiative on this!

src/foundation/univalence.lagda.md
```diff
@@ -39,7 +39,7 @@ In this file we postulate the univalence axiom. Its statement is defined in
 ## Postulate
 
 ```agda
-postulate univalence : {l : Level} (A B : UU l) → UNIVALENCE A B
+postulate univalence : {l : Level} → axiom-univalence-Level l
```
VojtechStep, line 42: Is there a particular reason why this isn't `postulate univalence : axiom-univalence`?

EgbertRijke: Good catch! It should be judgmentally equal.

fredrik-bakke: The particular reason is that I didn't catch it, since you're wondering

PR #867, opened 2023-10-20
Strict categories

fredrik-bakke:
Define strict categories and show they are preunivalent.


fredrik-bakke:
May this be my first PR without a deletion? 🤔

EgbertRijke:
Awesome! This merge is a no-brainer 👍

PR #870, opened 2023-10-20
Formatting of postulates

EgbertRijke:
I changed the formatting of postulates to make it consistent with how we format abstract definitions, i.e., with a line break immediately after the keyword `postulate`

I also changed the way univalence is assumed, following @VojtechStep's suggestion.

### DIFF
--- a/src/foundation/replacement.lagda.md
+++ b/src/foundation/replacement.lagda.md
@@ -33,7 +33,8 @@ Replacement l =
 ## Postulate
 
 ```agda
-postulate replacement : {l : Level} → Replacement l
+postulate
+  replacement : {l : Level} → Replacement l
 ```
 
 ```agda
--- a/src/primitives/characters.lagda.md
+++ b/src/primitives/characters.lagda.md
@@ -23,7 +23,8 @@ manipulate them. Characters are written between single quotes, e.g. `'a'`.
 ## Definitions
 
 ```agda
-postulate Char : UU lzero
+postulate
+  Char : UU lzero
 {-# BUILTIN CHAR Char #-}
 
 primitive
--- a/src/primitives/characters.lagda.md
+++ b/src/primitives/characters.lagda.md
@@ -25,6 +25,7 @@ manipulate them. Characters are written between single quotes, e.g. `'a'`.
 ```agda
 postulate
   Char : UU lzero
+
 {-# BUILTIN CHAR Char #-}
 
 primitive
--- a/src/primitives/floats.lagda.md
+++ b/src/primitives/floats.lagda.md
@@ -28,7 +28,8 @@ manipulate them. Floats can be written as usual, using dots as separators, e.g.
 ## Definitions
 
 ```agda
-postulate Float : UU lzero
+postulate
+  Float : UU lzero
 {-# BUILTIN FLOAT Float #-}
 
 primitive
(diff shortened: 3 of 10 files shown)


src/foundation/replacement.lagda.md
```diff
@@ -33,7 +33,8 @@ Replacement l =
 ## Postulate
 
 ```agda
-postulate replacement : {l : Level} → Replacement l
+postulate
+  replacement : {l : Level} → Replacement l
```
fredrik-bakke, line 37: Should `Replacement` be refactored like the other axiom statements?

src/primitives/machine-integers.lagda.md
```diff
@@ -23,7 +23,8 @@ functions to manipulate them.
 ## Definitions
 
 ```agda
-postulate Word64 : UU lzero
+postulate
+  Word64 : UU lzero
 {-# BUILTIN WORD64 Word64 #-}
```
fredrik-bakke, line 28: For some reason these look better with a line break in between
```suggestion
postulate
  Word64 : UU lzero

{-# BUILTIN WORD64 Word64 #-}
```

src/primitives/floats.lagda.md
```diff
@@ -28,7 +28,8 @@ manipulate them. Floats can be written as usual, using dots as separators, e.g.
 ## Definitions
 
 ```agda
-postulate Float : UU lzero
+postulate
+  Float : UU lzero
 {-# BUILTIN FLOAT Float #-}
```
fredrik-bakke, line 33:
```suggestion
postulate
  Float : UU lzero

{-# BUILTIN FLOAT Float #-}
```

src/primitives/characters.lagda.md
```diff
@@ -23,7 +23,8 @@ manipulate them. Characters are written between single quotes, e.g. `'a'`.
 ## Definitions
 
 ```agda
-postulate Char : UU lzero
+postulate
+  Char : UU lzero
 {-# BUILTIN CHAR Char #-}
```
fredrik-bakke, line 28:
```suggestion
postulate
  Char : UU lzero

{-# BUILTIN CHAR Char #-}
```

src/primitives/strings.lagda.md
```diff
@@ -30,7 +30,8 @@ manipulate them. Strings are written between double quotes, e.g.
 ## Definitions
 
 ```agda
-postulate String : UU lzero
+postulate
+  String : UU lzero
 {-# BUILTIN STRING String #-}
```
fredrik-bakke, line 35:
```suggestion
postulate
  String : UU lzero

{-# BUILTIN STRING String #-}
```

fredrik-bakke (approved):
I don't see why it's necessary to line-wrap postulates, but I won't object

src/foundation/replacement.lagda.md
EgbertRijke: Yeah, great idea!

EgbertRijke:
> I don't see why it's necessary to line-wrap postulates, but I won't object

Here are my two reasons:
1. It looks cleaner if the name of the entry is the first thing on a line, when possible.
2. The grammar of `postulate` is the same as for `private` or `abstract`, so it would be consistent to format it the same.

fredrik-bakke:
I just found out we can put block keywords like `abstract` together with the module declaration like this:

image

This way avoids a little bit of indentation, and also forces us to group our abstract definitions more together, which is nice. The drawback is that it hides the `abstract` keyword away a little, but maybe that's good if people more easily ignore them this way.

fredrik-bakke:
There's something unreasonably satisfying to me about this keyword placement. 🤔

PR #871, opened 2023-10-21
Rename `is-contr-total` to `is-torsorial`

EgbertRijke:
Since the definition of `is-torsorial` is that the total space is contractible, this pull request proposes to replace the clumsy-looking `is-contr-total` with `is-torsorial`.

### DIFF
--- a/src/category-theory/full-large-subprecategories.lagda.md
+++ b/src/category-theory/full-large-subprecategories.lagda.md
@@ -208,8 +208,8 @@ module _
   is-large-category-large-precategory-is-large-category-Full-Large-Subprecategory
     is-large-category-C X =
     fundamental-theorem-id
-      ( is-contr-total-Eq-subtype
-        ( is-contr-total-iso-Large-Category
+      ( is-torsorial-Eq-subtype
+        ( is-torsorial-iso-Large-Category
           ( make-Large-Category C is-large-category-C)
           ( inclusion-subtype P X))
         ( is-prop-is-in-subtype P)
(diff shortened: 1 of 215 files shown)


fredrik-bakke:
I have a concern that the term "torsorial" may be less known or intuitive, making the entries harder to find and understand in context. The phrasing "contr-total" is after all very descriptive and clear, although, as you put it, also perhaps clumsy-looking to someone who knows the more advanced terminology.

EgbertRijke:
I'm glad you're saying that, especially in the context of the discussion about cotowers.

fredrik-bakke:
I've made my case for why I think `cotower` is better elsewhere, and there's nuance to the subject that is worth discussing case by case. In either case, I'm not clearly against `torsorial`. Maybe it is just my own ignorance that has led to my concern. If we establish the terminology well enough, then this can just be an experience of learning for users who didn't know the term before rather than confusion.

fredrik-bakke (approved):

EgbertRijke:
Btw, this pull request did simply a search-and-replace. So I would be ok with it if it gets rejected. But it's definitely worth discussing how much we want to use/avoid terminology that is mostly used in more advanced mathematics. Perhaps we can come up with some sort of guideline on when to use advanced terminology or when to avoid it.

(As I type this message, I see that you already merged it)

EgbertRijke:
Now I think we can use cotower too:)

To the learning experience 🍻

fredrik-bakke:
Cheers 🍻

(For real, I don't want this to be a bargaining game though)

PR #873, opened 2023-10-21
Improve computational behaviour of `iso-eq`

fredrik-bakke:
Changes the definition of `iso-eq` to compute definitionally to `hom-eq` on the first component for every identification. This way, we can use lemmas about `hom-eq` more easily, if they were to exist. After all, `is-iso` is a proposition.


fredrik-bakke:
This is a very minor change, but I think it's more elegant this way. And who knows, maybe it will save someone a little headache down the road.

Anyways, it's ready for review.

EgbertRijke:
Awesome! Nice one!

PR #874, opened 2023-10-21
Preunivalence holds in univalent foundations

fredrik-bakke:
This is another simple one. It refines the implications between the axioms, and records that preunivalence indeed holds.


src/foundation/preunivalence.lagda.md
```diff
@@ -57,33 +56,47 @@ emb-map-preunivalence L X Y =
...
-preunivalence-univalence :
-  {l : Level} → axiom-univalence-Level l → axiom-preunivalence-Level l
-preunivalence-univalence ua A B = is-emb-is-equiv (ua A B)
+module _
+  {l : Level} (A B : UU l)
+  where
+
+  preunivalence-axiom-K :
+    instance-axiom-K (UU l) → instance-axiom-K A → instance-axiom-K B →
+    instance-preunivalence A B
```
EgbertRijke, line 71: I wonder if this should be called `instance-preunivalence-instance-axiom-K`, and that `preunivalence-axiom-K` should be of the form `axiom-K -> preunivalence`.

src/foundation/preunivalence.lagda.md
```diff
@@ -57,33 +56,47 @@ emb-map-preunivalence L X Y =
...
 
-### Preunivalence generalizes axiom K
+### Preunivalence generalizes univalence
+
+```agda
+module _
+  {l : Level} (A B : UU l)
+  where
+
+  preunivalence-univalence :
```
EgbertRijke, line 85: Similar comment here

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -113,39 +113,47 @@ In this composite, we used preunivalence at the second step.
...
 ```agda
 module _
-  {l : Level} (L : axiom-preunivalence-Level l) (A : UU l)
+  {l : Level} (A : UU l)
+  (L : (a x y : A) → instance-preunivalence (Id x y) (Id a y))
   where
 
-  is-emb-Id-axiom-preunivalence-Level : is-emb (Id {A = A})
-  is-emb-Id-axiom-preunivalence-Level a =
+  emb-fiber-Id-preunivalent-Id :
```
EgbertRijke, line 120: Similarly, shouldn't this be called `eb-fiber-Id-instance-preunivalence-Id`?

fredrik-bakke: Do we have to be that rigid? I find the current name reads quite fine.

fredrik-bakke: (I think)

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -113,39 +113,47 @@ In this composite, we used preunivalence at the second step.
...
-                ( emb-equiv equiv-funext))))
-        ( emb-equiv (inv-equiv (equiv-fiber Id (Id a))))
+
+module _
+  {l : Level} (L : axiom-preunivalence-Level l) (A : UU l)
+  where
+
+  is-emb-Id-axiom-preunivalence-Level : is-emb (Id {A = A})
+  is-emb-Id-axiom-preunivalence-Level =
+    is-emb-Id-preunivalent-Id A (λ a x y → L (Id x y) (Id a y))
```
EgbertRijke, line 156: I would have a preference for the non-level-restricted versions of such statements, but feel free to ignore that.

fredrik-bakke: No, I think you're right. We're already generalizing, so it doesn't make that much sense to only go halfway. It would be weird to have a formal system that only satisfied some axiom at a certain universe level (looking at you, `UUomega`).

EgbertRijke: `UUomega` be like I'm no universe level =)

src/foundation/universal-property-identity-types.lagda.md
EgbertRijke: My instinct was that this name seems somewhat situational and not following the expected scheme, but if you insist on keeping it then let's keep it.

src/foundation/preunivalence.lagda.md
fredrik-bakke: how about `instance-preunivalence-axiom-K`?

EgbertRijke: I don't know... my suggestion obviously fits in our naming scheme, but with your suggestion you have to squint your eyes:)

fredrik-bakke: fiine

fredrik-bakke: instance is just such a long word

fredrik-bakke: We could do `fiberwise-preunivalence-axiom-K` :)

fredrik-bakke: just kidding, I'll use your suggestion

EgbertRijke: I don't think we'll be using that lemma a whole lot, so I don't think it's essential to save on characters

fredrik-bakke: I ended up calling the latter `axiom-preunivalence-axiom-K` (and similar for univalence). Looks to me like these read better, and they're consistent with the other definitions.

EgbertRijke:
Ready?:)

fredrik-bakke:
Ready 😎

PR #878, opened 2023-10-22
Refactor funext

fredrik-bakke:
Minor refactor of funext to follow the new naming scheme.
There are still some things that could be improved:
- The file on homotopy induction
- Refine the connection between weak funext and funext
- A more detailed treatment of the connection between univalence and funext. Currently, it is not clear where exactly univalence is even needed in the argument, and that `A` and `B` are assumed to live in the same universe is not commented on.

### DIFF
--- a/src/foundation-core/function-extensionality.lagda.md
+++ b/src/foundation-core/function-extensionality.lagda.md
@@ -36,11 +36,24 @@ module _
   {l1 l2 : Level} {A : UU l1} {B : A → UU l2}
   where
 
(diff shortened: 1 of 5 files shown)


EgbertRijke:
LGTM

EgbertRijke:
And I will be in a little bit of trouble because that synthetic homotopy theory refactor refactored funext too:P

fredrik-bakke:
See! Would've been nice to have!

PR #883, opened 2023-10-24
Extending pushouts by equivalences

tomdjong:
These lemmas will later be used to characterise the fibers of the cogap map (and the codiagonal in particular).
Suggestions for better names are particularly encouraged :slightly_smiling_face:


src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -830,3 +830,180 @@ module _
...
+  cocone-left-extended =
+    horizontal-map-cocone f g c ∘ j ,
+    vertical-map-cocone f g c ,
+    ( λ z →
+      equational-reasoning
+        (horizontal-map-cocone f g c ∘ j ∘ f') z
+        ＝ (horizontal-map-cocone f g c ∘ f ∘ i) z
+          by ap (horizontal-map-cocone f g c) (coh z)
+        ＝ (vertical-map-cocone f g c ∘ g ∘ i) z
+          by coherence-square-cocone f g c (i z))
```
EgbertRijke, line 870: Usually we use copattern matching for definitions by components of iterated Sigma-types, i.e. `C-c C-c` on an empty goal.

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -830,3 +830,180 @@ module _
...
+  cocone-top-extended : cocone (f ∘ i) g' X
+  cocone-top-extended = horizontal-map-cocone f g c ,
+                        vertical-map-cocone f g c ∘ j ,
+                        ( λ z →
+                          equational-reasoning
+                            (horizontal-map-cocone f g c ∘ f ∘ i) z
+                            ＝ (vertical-map-cocone f g c ∘ g ∘ i) z
+                              by coherence-square-cocone f g c (i z)
+                            ＝ (vertical-map-cocone f g c ∘ j ∘ g') z
+                              by ap (vertical-map-cocone f g c) (coh z))
```
EgbertRijke, line 922: Try copattern matching on this one, without indenting all the way to the left. I know this is common agda-practice, but we don't do it in this library.

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -830,3 +830,180 @@ module _
...
+  { l1 l2 l3 l4 l5 l6 : Level}
+  { S : UU l1} {A : UU l2} {B : UU l3} {X : UU l4} {S' : UU l5} {A' : UU l6}
+  ( f : S → A) (g : S → B) (i : S' → S) (j : A' → A) (f' : S' → A')
+  ( c : cocone f g X)
+  ( up-c : {l : Level} → universal-property-pushout l f g c)
+  ( coh : coherence-square-maps i f' f j)
+  where
+
+  cocone-left-extended : cocone f' (g ∘ i) X
+  cocone-left-extended =
```
EgbertRijke, line 861: This definition should definitely be in the file about cocones.

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -830,3 +830,180 @@ module _
...
+module _
+  { l1 l2 l3 l4 l5 l6 : Level}
+  { S : UU l1} {A : UU l2} {B : UU l3} {X : UU l4} {S' : UU l5} {B' : UU l6}
+  ( f : S → A) (g : S → B) (i : S' → S) (j : B' → B) (g' : S' → B')
+  ( c : cocone f g X)
+  ( up-c : {l : Level} → universal-property-pushout l f g c)
+  ( coh : coherence-square-maps g' i j g)
+  where
+
+  cocone-top-extended : cocone (f ∘ i) g' X
```
EgbertRijke, line 913: This definition should be in the file about cocones

EgbertRijke: In fact, we have horizontal and vertical composition of cocones. Can't you reuse the existing formalization?

EgbertRijke (commented):
Some initial comments

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
EgbertRijke: The reason we don't indent like this in agda-unimath, is that we use indentation levels to display the tree structure of a definition. This is explained here:

https://unimath.github.io/agda-unimath/CODINGSTYLE.html

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -830,3 +830,180 @@ module _
...
+
+```text
+       i       g
+   S' ---> S ----> B
+   |   ≃   |       |
+f' |       | f     |
+   v   ≃   v    ⌜  v
+   A' ---> A ----> X
+       j
+```
```
EgbertRijke, line 848: I think it would be helpful if you note in the informal text that this is a special case of the pushout pasting property

EgbertRijke: And perhaps it would be even better if the horizontal version goes directly after the horizontal pushout pasting property, and the vertical version goes directly after the vertical pushout pasting property.

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
EgbertRijke: To be clear, I don't mean that we shouldn't have this definition. What I mean is that this definition should use horizontal composition of cocones in its definition.

tomdjong:
Thanks @EgbertRijke; that was very helpful! Things should be ready for another round of reviewing now.

src/synthetic-homotopy-theory/cocones-under-spans.lagda.md
```diff
@@ -268,3 +291,73 @@ pr2 (pr2 (cocone-comp-vertical f i k c d)) =
...
+```text
+   S' ---> B'
+   |       |
+   |       |
+   v       v
+   A' ---> X
+```
+
+```agda
+cocone-comp-vertical-horizontal :
```
EgbertRijke, line 347: This looks like it should be thought of as composition of a cocone with a morphism of spans. In that light, perhaps a better name would be `comp-cocone-hom-span`. Unfortunately, we don't have morphisms of spans formalized in the library. If we had a notion of morphisms of spans, then you wouldn't have to write `coherence-square-maps` twice in the specification of this entry, and you wouldn't have to give so much data explicitly.

Perhaps an informal sentence should be in place here, that this is really composition of cocones and morphisms of spans.

And if you feel inspired to create morphisms of spans, that can go in a new file `foundation.morphisms-spans`:)

EgbertRijke: (But for this PR you don't have to do that. Just an informal sentence is enough)

src/synthetic-homotopy-theory/universal-property-pushouts.lagda.md
```diff
@@ -706,6 +750,119 @@ module _
...
+    universal-property-pushout-rectangle-universal-property-pushout-top i g' f
+      ( g , j , coh)
+      ( c)
+      ( universal-property-pushout-is-equiv i g' (g , j , coh) ie je)
+      ( up-c)
+```
+
+### Extending pushouts by equivalences of cocones
+
+Given a commutative diagram where i, j and k are equivalences,
```
EgbertRijke, line 800: Similarly, this is really an equivalence of spans.

EgbertRijke:
Awesome! Thanks for the changes and for the PR.

PR #880, opened 2023-10-22
Opposite categories, gaunt categories, replete subprecategories, large Yoneda, and miscellaneous additions

fredrik-bakke:
### Additions to category theory
- Replace yoneda lemmas by universe polymorphic ones and do some refactorings
- Additions opposite categories
- Gaunt categories
- Replete subprecategories
- Subcategories are categories
- Prove some identities for composition of functors (This is developed further in #886)
- Fix an oversight in the definition of sub(pre)categories

### Other additions
- Add missing lemmas regarding sets and 1-types
- iterated version of `is-trunc-succ-is-trunc`
- Fix DOI links

~Depends on #891.~

PR #886, opened 2023-10-26
Fun with functors

fredrik-bakke:
### Additions
- Set-magmoids
- Wide subprecategories
- Wide subcategories
- Core precategories
- Core categories
- Restrictions of functors on cores
- Essential fibers of functors
- Essentially surjective functors
- Split essentially surjective functors
- Pseudomonic functors
- Essentially injective functors
- Conservative functors

including the results
- Wide subcategories are categories
- Cores of categories are groupoids
- Fully faithful functors are pseudomoinic
- Pseudomonic functors are conservative

### A regression
I'm not so happy with how many files maps of (pre)categories have generated. It should have been enough to define them for binary families of sets and then use definitions of functors everywhere else. In particular, according to the structure identity principle, I don't think the 2-precategory of set-magmoids and maps will be univalent. Rather, the 2-precategory of binary families of sets and maps should be univalent. In response to our discussion on when functoriality is automatic, the small amount of research I've done seems to suggest it is more fruitful to ask when a map that preserves composition also preserves units, or when a map that preserves units also preserves composition than to assume neither and ask for both.

Depends on #880 and ~#891~.


VojtechStep:
I think you meant to reference #880, right?

fredrik-bakke:
Oh, yep, thanks

fredrik-bakke:
This one is ready for review when #880 is merged.

src/foundation/universal-property-truncation.lagda.md
fredrik-bakke, line 1: The code that was commented in this file had all been moved elsewhere, so we are not losing anything through this deletion.

src/category-theory/structure-equivalences-set-magmoids.lagda.md
```diff
@@ -0,0 +1,190 @@
...
+
+  structure-equiv-eq-Set-Magmoid :
+    (A B : Set-Magmoid l1 l2) →
+    A ＝ B → structure-equiv-Set-Magmoid A B
+  structure-equiv-eq-Set-Magmoid A .A refl =
+    id-structure-equiv-Set-Magmoid A
+```
+
+The rest remains to be figured out. We need the fact that binary families of
+equivalences of sets are torsorial.
```
fredrik-bakke, line 171: This should be easy to prove using that binary families of sets form a category. I'll finish this in another PR.

src/foundation/set-truncations.lagda.md
```diff
@@ -133,41 +127,34 @@ universal-property-trunc-Set :
...
-  (type-trunc-Set A → type-Set B) ≃ (A → type-Set B)
-equiv-universal-property-trunc-Set = equiv-universal-property-trunc
-
-apply-universal-property-trunc-Set :
-  {l1 l2 : Level} {A : UU l1} (t : type-trunc-Set A) (B : Set l2) →
-  (A → type-Set B) → type-Set B
-apply-universal-property-trunc-Set t B f = map-universal-property-trunc B f t
+module _
+  {l1 l2 : Level} {A : UU l1} (B : Set l2)
+  where
```
EgbertRijke, line 132: I think the mismatch is introduced here, by putting everything in the same module. Perhaps you can put `apply-universal-property-trunc-Set` in its own module.

fredrik-bakke: Thanks for pointing that out, I actually wrote an issue on this. But I will revert this change, because it does not resolve any issue.

fredrik-bakke:
In case you suddenly decide to merge this PR, be careful with the final commit message. I wonder if Mabel and Vojtech will be attributed as co-authors purely for how I updated the PR.

EgbertRijke:
Lol, how did you read my intentions?:)

EgbertRijke:
Anyway, this PR contains a lot of good work that obviously needs to get merged. I didn't have any concrete suggestions for improvements, and thought that if any improvements need to be made, we can do so while working with the code.

fredrik-bakke:
> Lol, how did you read my intentions?:)

I know you too well 😉

fredrik-bakke:
> Anyway, this PR contains a lot of good work that obviously needs to get merged. I didn't have any concrete suggestions for improvements, and thought that if any improvements need to be made, we can do so while working with the code.

100%! Although there is no rush, I understand you don't want to keep this hanging

fredrik-bakke:
Did you check the commit message?

EgbertRijke:
Yup, it should be clean of intruders

PR #891, opened 2023-10-28
Implicit function extensionality

fredrik-bakke:
Adds a series of basic lemmas for implicit functions related to function extensionality. Also adds type-theoretic principle of choice for them for better computation.

This is factored out from my other PRs, as I need it for all of them, so they all now depend on this one to be merged first. If I don't get any objections quickly, I will probably just merge this one without review.

### DIFF
--- a/src/foundation/implicit-function-types.lagda.md
+++ b/src/foundation/implicit-function-types.lagda.md
@@ -53,7 +53,7 @@ module _
   pr2 equiv-explicit-implicit-Π = is-equiv-explicit-implicit-Π
 ```
 
-### Equality if explicit functions is equality of implicit functions
+### Equality of explicit functions is equality of implicit functions
 
 ```agda
 module _
--- a/src/foundation/implicit-function-types.lagda.md
+++ b/src/foundation/implicit-function-types.lagda.md
@@ -1,0 +1,81 @@
+# Implicit function types
+
+```agda
+module foundation.implicit-function-types where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.dependent-pair-types
+open import foundation.universe-levels
+
+open import foundation-core.equivalences
+open import foundation-core.homotopies
+open import foundation-core.identity-types
+```
+
+</details>
+
+## Properties
+
+### Dependent function types taking implicit arguments are equivalent to dependent function types taking explicit arguments
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2}
+  where
+
+  implicit-explicit-Π : ((x : A) → B x) → {x : A} → B x
+  implicit-explicit-Π f {x} = f x
+
+  explicit-implicit-Π : ({x : A} → B x) → (x : A) → B x
+  explicit-implicit-Π f x = f {x}
+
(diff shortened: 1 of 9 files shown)


fredrik-bakke:
Just a note, I will review this PR myself before it is merged

src/foundation/implicit-function-types.lagda.md
```diff
@@ -0,0 +1,81 @@
...
+  equiv-implicit-explicit-Π : ((x : A) → B x) ≃ ({x : A} → B x)
+  pr1 equiv-implicit-explicit-Π = implicit-explicit-Π
+  pr2 equiv-implicit-explicit-Π = is-equiv-implicit-explicit-Π
+
+  equiv-explicit-implicit-Π : ({x : A} → B x) ≃ ((x : A) → B x)
+  pr1 equiv-explicit-implicit-Π = explicit-implicit-Π
+  pr2 equiv-explicit-implicit-Π = is-equiv-explicit-implicit-Π
+```
+
+### Equality if explicit functions is equality of implicit functions
```
VojtechStep, line 56:
```suggestion
### Equality of explicit functions is equality of implicit functions
```

src/foundation/type-theoretic-principle-of-choice.lagda.md
```diff
@@ -30,26 +30,26 @@ distributivity of Π over Σ, is also known as the type theoretic principle of
 choice. Indeed, it is the Curry-Howard interpretation of (one formulation of)
 the axiom of choice.
 
-## Definitions
+## Distributivity of Π over Σ
```
VojtechStep, line 33: I'd rather keep the file structure consistent with other files, i.e. have `Definitions > Distributivity of Pi over Sigma`, `Definitions > Distributivity or implicit Pi over Sigma`, `Properties > ...`; what motivated this change?

fredrik-bakke: Yup, I agree the structure of this file is not yet good

fredrik-bakke: The issue is I wanted to structure it as one part about explicit functions, and one about implicits. Each of them have a "statement" and "theorem" section.

fredrik-bakke: Maybe the file can start with `## Theorem` and then `### Distributivity of Pi over Sigma`,  and then `#### Statement` and go from there

fredrik-bakke: Eh, I guess having both statements at the start works fine. It's not like the definitions are very large.

fredrik-bakke:
Thanks for the comments. Let me know if the new file structure is more to your liking

VojtechStep:
Looks good, thanks 👌. I think I'd keep the `Properties` heading instead of replacing it with `Lemma`, since we usually put characterizing identity types under that, but it's not a hill I'm dying on, so feel free to merge

fredrik-bakke:
In this instance, I think "lemma" is helpful to let readers know that the good stuff is further down in the file, so I'll merge it like this :)

PR #899, opened 2023-11-02
Equality in iterated Sigma types

maybemabeline:
Computed some equalities in iterated Sigma types. Maybe this can be generalized some more, but this is what I currently need for my work with pushouts, where some cocone structures involve equality of a quadruple. Input on the names is welcome as usual.

### DIFF
--- a/src/foundation-core/equivalences.lagda.md
+++ b/src/foundation-core/equivalences.lagda.md
@@ -588,6 +588,10 @@ module _
     (e : A ≃ B) (x y : A) → (x ＝ y) ≃ (map-equiv e x ＝ map-equiv e y)
   pr1 (equiv-ap e x y) = ap (map-equiv e)
   pr2 (equiv-ap e x y) = is-emb-is-equiv (is-equiv-map-equiv e) x y
+
+  map-equiv-ap :
+    (e : A ≃ B) (x y : A) → (map-equiv e x ＝ map-equiv e y) → (x ＝ y)
+  map-equiv-ap e x y = map-equiv (inv-equiv (equiv-ap e x y))
 ```
 
 ## Equivalence reasoning
--- a/src/foundation-core/equivalences.lagda.md
+++ b/src/foundation-core/equivalences.lagda.md
@@ -589,9 +589,9 @@ module _
   pr1 (equiv-ap e x y) = ap (map-equiv e)
   pr2 (equiv-ap e x y) = is-emb-is-equiv (is-equiv-map-equiv e) x y
 
-  map-equiv-ap :
+  inv-map-equiv-ap :
     (e : A ≃ B) (x y : A) → (map-equiv e x ＝ map-equiv e y) → (x ＝ y)
-  map-equiv-ap e x y = map-equiv (inv-equiv (equiv-ap e x y))
+  inv-map-equiv-ap e x y = map-equiv (inv-equiv (equiv-ap e x y))
 ```
 
 ## Equivalence reasoning
--- a/src/foundation-core/equivalences.lagda.md
+++ b/src/foundation-core/equivalences.lagda.md
@@ -589,9 +589,9 @@ module _
   pr1 (equiv-ap e x y) = ap (map-equiv e)
   pr2 (equiv-ap e x y) = is-emb-is-equiv (is-equiv-map-equiv e) x y
 
-  inv-map-equiv-ap :
+  map-inv-equiv-ap :
     (e : A ≃ B) (x y : A) → (map-equiv e x ＝ map-equiv e y) → (x ＝ y)
-  inv-map-equiv-ap e x y = map-equiv (inv-equiv (equiv-ap e x y))
+  map-inv-equiv-ap e x y = map-inv-equiv (equiv-ap e x y)
 ```
 
 ## Equivalence reasoning
--- a/src/foundation/equality-dependent-pair-types.lagda.md
+++ b/src/foundation/equality-dependent-pair-types.lagda.md
@@ -15,6 +15,12 @@ open import foundation.dependent-identifications
 open import foundation.dependent-pair-types
 open import foundation.transport-along-identifications
 open import foundation.universe-levels
+open import foundation.equivalences
+open import foundation.functoriality-dependent-pair-types
+open import foundation.homotopies
+open import foundation.function-extensionality
+open import foundation.contractible-types
+open import foundation.type-arithmetic-dependent-pair-types
 
 open import foundation-core.function-types
 open import foundation-core.identity-types
@@ -113,7 +119,7 @@ module _
 
 ```agda
 module _
-  { l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2} {Y : UU l3} (f : Σ A B → Y)
+  {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2} {Y : UU l3} (f : Σ A B → Y)
   where
 
   compute-ap-eq-pair-Σ :
@@ -127,7 +133,7 @@ module _
 
 ```agda
 module _
-  { l1 l2 : Level} {A : UU l1} (B : A → UU l2)
+  {l1 l2 : Level} {A : UU l1} (B : A → UU l2)
   where
 
   triangle-eq-pair-Σ :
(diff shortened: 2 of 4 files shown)


src/foundation/equality-dependent-pair-types.lagda.md
```diff
@@ -137,6 +143,62 @@ module _
...
+          ( pr2 t)))
+      ( equiv-pair-eq-Σ (pr1 s) (pr1 t))
+      ( λ p →
+        ( equiv-tr
+          ( λ q → dependent-identification C q (pr2 s) (pr2 t))
+          ( inv (is-section-pair-eq-Σ (pr1 s) (pr1 t) p))))) ∘e
+    ( equiv-pair-eq-Σ s t)
+
+  triple-eq-Σ :
+    (s t : Σ (Σ A B) C) →
```
fredrik-bakke, line 175: `triple` is associated in the other order, so this name isn't quite representative. I guess the "consistent" name would be `triple-eq-Σ'`

src/foundation/type-arithmetic-dependent-pair-types.lagda.md
```diff
@@ -305,6 +305,14 @@ module _
   pr1 interchange-Σ-Σ = map-interchange-Σ-Σ
   pr2 interchange-Σ-Σ = is-equiv-map-interchange-Σ-Σ
 
+  interchange-iterated-Σ-Σ :
```
fredrik-bakke, line 308: We generally use `iterated` for constructions that are generalized over all natural numbers, so I would try to find a different name for this one

src/foundation-core/equivalences.lagda.md
```diff
@@ -588,6 +588,10 @@ module _
     (e : A ≃ B) (x y : A) → (x ＝ y) ≃ (map-equiv e x ＝ map-equiv e y)
   pr1 (equiv-ap e x y) = ap (map-equiv e)
   pr2 (equiv-ap e x y) = is-emb-is-equiv (is-equiv-map-equiv e) x y
+
+  inv-map-equiv-ap :
```
fredrik-bakke, line 592: `map-inv-equiv-ap` (although I'm not a fan of `equiv-ap` to begin with. But that's for another day  🙃)

src/foundation/equality-dependent-pair-types.lagda.md
maybemabeline: I was going off of the name `pair-eq-Σ` that works analogously as this. I guess the logic is that we obtain a pair from an equality in a Σ type, but I'm not sure if I like that either.

src/foundation/equality-dependent-pair-types.lagda.md
```diff
@@ -137,6 +143,62 @@ module _
...
+      ( λ q → dependent-identification C (eq-pair-Σ' q) (pr2 s) (pr2 t)))
+  triple-eq-Σ s t = map-equiv (equiv-triple-eq-Σ s t)
+
+  coh-triple-Σ :
+    {s t : Σ A (λ x → Σ (B x) λ y → C (x , y))} (p : s ＝ t) →
+    eq-base-eq-pair-Σ p ＝
+    eq-base-eq-pair-Σ (eq-base-eq-pair-Σ (ap (map-inv-associative-Σ A B C) p))
+  coh-triple-Σ refl = refl
+
+  dependent-eq-snd-eq-iterated-Σ :
```
fredrik-bakke, line 188: This is not a general "iterated" theorem, so a better name is something like `-double-Σ` or `-Σ²`

fredrik-bakke: Also, where does the `snd` come from?

src/foundation-core/equivalences.lagda.md
```diff
@@ -588,6 +588,10 @@ module _
...
+    (e : A ≃ B) (x y : A) → (map-equiv e x ＝ map-equiv e y) → (x ＝ y)
+  inv-map-equiv-ap e x y = map-equiv (inv-equiv (equiv-ap e x y))
```
VojtechStep, line 594: `map-equiv . inv-equiv` should be `map-inv-equiv`

src/foundation/equality-dependent-pair-types.lagda.md
maybemabeline: Good point on the use of the word iterated. The word `snd` was supposed to signify that the dependent equality is happening inside the second component of the Sigma but yeah, it was a placeholder.

src/foundation/equality-dependent-pair-types.lagda.md
```diff
@@ -137,6 +143,62 @@ module _
...
+          ( λ q → dependent-identification C q (pr2 s) (pr2 t))
+          ( inv (is-section-pair-eq-Σ (pr1 s) (pr1 t) p))))) ∘e
+    ( equiv-pair-eq-Σ s t)
+
+  triple-eq-Σ :
+    (s t : Σ (Σ A B) C) →
+    (s ＝ t) →
+    ( Σ
+      ( Eq-Σ (pr1 s) (pr1 t))
+      ( λ q → dependent-identification C (eq-pair-Σ' q) (pr2 s) (pr2 t)))
```
fredrik-bakke, line 179: This deserves its own definition `Eq-Σ²'`. Not a fan of that name, but at least it is consistent.

src/foundation/type-arithmetic-dependent-pair-types.lagda.md
VojtechStep: I'd go with `interchange-Sigma-Sigma-Sigma`, and leave the `iterated` name for when we figure out how to represent the general scheme with iterated sigmas

src/foundation/equality-dependent-pair-types.lagda.md
maybemabeline: Ok! Why the `'` at the end of the name though?

src/foundation/equality-dependent-pair-types.lagda.md
fredrik-bakke: The thing is, you could also consider
```agda
  triple-eq-Σ :
    (s t : Σ A (Σ B C)) →
```
and indeed, in this case, `s` and `t` would be terms of the form `s = triple a b c`. While in your case, the elements are of the form `triple' a b c`

src/foundation/equality-dependent-pair-types.lagda.md
```diff
@@ -137,6 +143,62 @@ module _
...
+
+```agda
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : A → UU l2} {C : Σ A B → UU l3}
+  where
+
+  equiv-triple-eq-Σ :
+    (s t : Σ (Σ A B) C) →
+    (s ＝ t) ≃
+    ( Σ
```
VojtechStep, line 156: Please format the Sigma into two lines instead of three - I'm not providing a GH suggestion because I'm on my phone.

src/foundation/equality-dependent-pair-types.lagda.md
```diff
@@ -137,6 +143,62 @@ module _
...
+      ( λ p →
+        ( equiv-tr
+          ( λ q → dependent-identification C q (pr2 s) (pr2 t))
+          ( inv (is-section-pair-eq-Σ (pr1 s) (pr1 t) p))))) ∘e
+    ( equiv-pair-eq-Σ s t)
+
+  triple-eq-Σ :
+    (s t : Σ (Σ A B) C) →
+    (s ＝ t) →
+    ( Σ
```
VojtechStep, line 177: Same formatting change here

VojtechStep (commented):
Thanks for the PR! I feel like eventually we'll want to formalize these computations in full generality with iterated sigmas, but for now it's nice to have them 👌

src/foundation/equality-dependent-pair-types.lagda.md
maybemabeline: Oh, sorry, I didn't even know we had these constructions already. I will see if I can use them somewhere here.

src/foundation/equality-dependent-pair-types.lagda.md
maybemabeline: I understand now :sweat_smile:.

src/foundation/type-arithmetic-dependent-pair-types.lagda.md
fredrik-bakke: The "general" scheme for Sigmas would be to index over binary trees instead of `ℕ`, and I'm not sure how useful/doable that is. We could also go with iterated versions for left-associated and right-associated sigmas only.

fredrik-bakke: Since it is most natural to consider iterated theorems for right-associated Sigmas, it still seems useful to record once-iterated versions for left-associated Sigmas.

src/foundation/equality-dependent-pair-types.lagda.md
fredrik-bakke: We don't characterize equality in `Σ A (Σ B C)` anywhere I believe

src/foundation/type-arithmetic-dependent-pair-types.lagda.md
VojtechStep: I was thinking only about left- and right- associative Sigmas

maybemabeline:
Regarding the coherence proof, I wasn't sure if it could be done without path induction or not.
The problem is this: when you have an equality in a Sigma type of the form `Σ A (Σ B C)`, you can obtain an equality in the first component in two ways; either take the first component of the equality immediately, or first associate to the left and then take the first component twice.

maybemabeline:
This is everything I wanted to do with this PR, so it can be merged if the renamings are ok

fredrik-bakke:
I'm not a huge fan of some of the names that are introduced in this PR, but I don't have any better suggestions. So to not hold you up, I will merge the PR in its current state. we can always change the names later anyways :)

PR #900, opened 2023-11-03
Retracts of maps

fredrik-bakke:
- Defines retracts of maps
- Shows that sections, retractions, and equivalences are closed under retracts of maps.
- Shows that fibers of the retract of a map are retracts of that map's fibers
- Shows that the fiber projections are map-retracts as well

Also fixes a misordering in the definition of commuting squares of homotopies.

### DIFF
--- a/src/foundation/commuting-squares-of-homotopies.lagda.md
+++ b/src/foundation/commuting-squares-of-homotopies.lagda.md
@@ -24,11 +24,11 @@ A square of [homotopies](foundation-core.homotopies.md)
       |         |
  left |         | right
       v         v
-      z ------> i
+      h ------> i
         bottom
 ```
 
-is said to **commute** if there is a homotopy `left ∙h bottom ~ top ∙h right`.
+is said to **commute** if there is a homotopy `top ∙h right ~ left ∙h bottom`.
 Such a homotopy is called a **coherence** of the square.
 
 ## Definition
@@ -39,7 +39,7 @@ module _
   where
 
   coherence-square-homotopies :
-    (left : f ~ h) (bottom : h ~ i) (top : f ~ g) (right : g ~ i) → UU (l1 ⊔ l2)
-  coherence-square-homotopies left bottom top right =
+    (top : f ~ g) (left : f ~ h) (right : g ~ i) (bottom : h ~ i) → UU (l1 ⊔ l2)
+  coherence-square-homotopies top left right bottom =
     left ∙h bottom ~ top ∙h right
 ```
--- a/src/foundation/commuting-squares-of-homotopies.lagda.md
+++ b/src/foundation/commuting-squares-of-homotopies.lagda.md
@@ -28,7 +28,7 @@ A square of [homotopies](foundation-core.homotopies.md)
         bottom
 ```
 
-is said to **commute** if there is a homotopy `top ∙h right ~ left ∙h bottom`.
+is said to **commute** if there is a homotopy `left ∙h bottom ~ top ∙h right `.
 Such a homotopy is called a **coherence** of the square.
 
 ## Definition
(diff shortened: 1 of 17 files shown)


fredrik-bakke:
Feel free to merge this if you want. Otherwise, I will probably continue working on it in the morning adding more theorems about retracts of maps.

src/orthogonal-factorization-systems/retracts-of-maps.lagda.md
```diff
@@ -0,0 +1,503 @@
...
+
+1. Section 4.7 of Univalent Foundations Project, _Homotopy Type Theory –
+   Univalent Foundations of Mathematics_ (2013)
+   ([website](https://homotopytypetheory.org/book/),
+   [arXiv:1308.0729](https://arxiv.org/abs/1308.0729))
+
+## External links
+
+- [retract#In arrow categories](https://ncatlab.org/nlab/show/retract#in_arrow_categories)
+  at nlab
```
fredrik-bakke, line 501: How do you caption external links to subsections of a web page?

fredrik-bakke:
I'm done adding things to this one now.

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -39,7 +39,7 @@ module _
   where
 
   coherence-square-homotopies :
-    (left : f ~ h) (bottom : h ~ i) (top : f ~ g) (right : g ~ i) → UU (l1 ⊔ l2)
-  coherence-square-homotopies left bottom top right =
+    (top : f ~ g) (left : f ~ h) (right : g ~ i) (bottom : h ~ i) → UU (l1 ⊔ l2)
```
EgbertRijke, line 42: Ah, thank you for spotting this!

src/orthogonal-factorization-systems/retracts-of-maps.lagda.md
EgbertRijke: I don't know, but why does the text that appear on our website have to be of this form `retract#In arrow categories`? Wouldn't it be better if it is a pointer written in natural language, rather than some partial html string?

```suggestion
- [Retracts in arrow categories](https://ncatlab.org/nlab/show/retract#in_arrow_categories)
  at nlab
```

src/orthogonal-factorization-systems/retracts-of-maps.lagda.md
```diff
@@ -0,0 +1,524 @@
+# Retracts of maps
```
EgbertRijke, line 1: Very nice! My only comment would be that this file wouldn't be out of place in `foundation`, because it is just something about maps. I'm actually refactoring `equivalences`, `sections` and `retractions` right now, and I've been making a file `retracts-of-types` in `foundation`. So it would be quite natural if `retracts-of-maps` sits right next to it.

EgbertRijke (approved):
Excellent pull request!

src/orthogonal-factorization-systems/retracts-of-maps.lagda.md
EgbertRijke: Actually I will take care of this myself in my upcoming 6-for-2 pull request.

fredrik-bakke: Okay, I'm fine with either :)

fredrik-bakke:
Apologies, I forgot to change back `has-section` and `has-retraction` to `retraction` and `section` in the `retracts-of-maps` file. Would you mind fixing that since you're already moving the file?

src/orthogonal-factorization-systems/retracts-of-maps.lagda.md
fredrik-bakke: I don't know, that was why I was asking. It feels strange to give our own caption to third-party sources though.

PR #910, opened 2023-11-08
Additions during work on material set theory in HoTT

elisabethstenholm:
This pull request contains code that was added to the library as needed during work on [material set theory in HoTT](https://git.app.uib.no/hott/hott-set-theory)


EgbertRijke:
Awesome!

src/category-theory/precategories-with-attributes.lagda.md
```diff
@@ -0,0 +1,154 @@
...
+record CwA {i j} (C : Precategory i j) (k : Level) : UU (i ⊔ j ⊔ lsuc k) where
+  field
+    Ty-F : functor-Precategory (opposite-Precategory C) (Set-Precategory k)
+    ext : functor-Precategory (element-Precategory C Ty-F) C
+    p : natural-transformation-Precategory (element-Precategory C Ty-F) C ext (proj₁-functor-element-Precategory C Ty-F)
+    is-pullback-p : (x y : obj-Precategory (element-Precategory C Ty-F)) (f : hom-Precategory (element-Precategory C Ty-F) x y) →
+      is-pullback-Precategory C _ _ _ _ _ _ _ _
+        (naturality-natural-transformation-Precategory (element-Precategory C Ty-F) C ext (proj₁-functor-element-Precategory C Ty-F) p f)
+
+  -- Notation
```
fredrik-bakke, line 53: This should probably be a header instead. I am also unsure how to approach the special syntax you introduce here. At the very least, I think the notation below should be declared private to the file.

fredrik-bakke:
Welcome back, Elisabeth, and thanks for the PR! It seems to contain some unfinished work, is this PR still a draft? If so, please mark it as such until you want a full review.

Some preliminary remarks:
- I noticed that you use the old style for file titles and module imports in some of your new files.
- We generally try to avoid abbreviated names and acronyms. I would probably rename `CwF` to `Precategory-With-Attributes` and `Acc` to `Accessible-Relation` (if that is the correct thing)
- We place binary operators at the end of the previous line, as opposed to the start of the new line. E.g. we write
   ```agda
     A →
     B
   ```
   as opposed to
    ```agda
     A
     → B
   ```

fredrik-bakke (changes requested):

elisabethstenholm:
Thanks @fredrik-bakke for the comments! I created this pull request quickly yesterday, so yes, there is probably some things to fix. I just checked that it compiles. I will look at the pr and fix things as soon as I have a few spare moments :)

fredrik-bakke:
> Thanks @fredrik-bakke for the comments! I created this pull request quickly yesterday, so yes, there is probably some things to fix. I just checked that it compiles. I will look at the pr and fix things as soon as I have a few spare moments :)

Super!

src/category-theory/precategories-with-attributes.lagda.md
fredrik-bakke: I'm not entirely sure what you mean by that. Do you want to define it in a module that spans the entire file? (That would be against our guidelines)

elisabethstenholm:
@fredrik-bakke I am unsure how to read the output from the CI checks. Could you help me figure out what I need to fix for the checks to pass? I have run `make check` locally and it compiles.

VojtechStep:
The files on precategories with families and precategory of elements of a presheaf don't have a markdown title, i.e. a `# Heading`

fredrik-bakke:
Oh, but beware, you seem to have placed some code outside of agda code blocks in your PR and the markdown formatting hook will interpret that code as prose and mangle it.

elisabethstenholm:
> Oh, but beware, you seem to have placed some code outside of agda code blocks in your PR and the markdown formatting hook will interpret that code as prose and mangle it.

I cannot find the code you are talking about. Do you remember where you saw it?

fredrik-bakke:
I saw some import statements being mangled and some code guards that were misplaced. I can look again a little later

src/order-theory/well-founded-orders.lagda.md
```diff
@@ -0,0 +1,112 @@
...
+open import foundation.action-on-identifications-functions
+open import foundation.binary-relations
+open import foundation.dependent-pair-types
+open import foundation.function-extensionality
+open import foundation.identity-types
+open import foundation.negation
+open import foundation.propositions
+open import foundation.universe-levels
+```
+</details>
```
VojtechStep, line 16: I think the parser is getting confused here, try this change:

````suggestion
```agda
module order-theory.well-founded-orders where
```

<details><summary>Imports</summary>

```agda
open import foundation.action-on-identifications-functions
open import foundation.binary-relations
open import foundation.function-extensionality
open import foundation.negation
open import foundation.propositions
```
</details>
````

Comment (fredrik-bakke) [reply src/order-theory/well-founded-orders.lagda.md:16]: Ah, yes, we encountered a similar issue in the past. It boils down to that XML tags are only supported if they are succeeded and preceded by empty lines and it's actually a design decision about the language.

**Comment by elisabethstenholm** [comment]
> I saw some import statements being mangled and some code guards that were misplaced. I can look again a little later

I found some leftover block comment with code, outside an agda code-block. I have removed it :)

File: src/category-theory/precategories-with-attributes.lagda.md (hunk 1 above)
Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:55]: Oh, I see, this does seem like something that contradicts the library's conventions, but I still don't know how to approach it. It is probably best if @EgbertRijke responds to this.

**Comment by fredrik-bakke** [comment]
> > I saw some import statements being mangled and some code guards that were misplaced. I can look again a little later
> 
> I found some leftover block comment with code, outside an agda code-block. I have removed it :)

Great! It was the code that Vojtech pointed out that originally caught my attention 😅

File: src/category-theory/precategories-with-attributes.lagda.md (hunk 1 above)
Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: As @elisabethstenholm says this indeed does not pollute the global namespace. Maybe this library should adopt this type of notations more?

Maybe I'm missing something and have been using the library wrongly, but one experience I've had when proving things about categories in this library is that the goal often turns into a soup of pr1/pr2 and I have to figure out on paper what I'm actually supposed to prove. I've not had the same experience when proving things about categories in Coq UniMath and Cubical Agda which both use a bit more notations (see e.g. https://github.com/agda/cubical/blob/master/Cubical/Categories/Category/Base.agda#L15 for how we implemented them there).

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: Thanks for the input, Anders! I do agree with some of your sentiments, and here's my two cents on the topic.

As you and Elisabeth demonstrate, (overloaded) special syntax can at least be implemented in a reasonable way. Thanks for that! I think, however, that there is a discussion to be had about the extent to which special syntax should be used, and what that syntax should be/look like. Personally, I am not currently against the proposition.

With regards to soup, it seems to me like this sort of translation between iterated projections and named projections is a thing that Agda should handle for us. I don't know, can some language feature get us there? As far as I understand, the `syntax` keyword won't do as it doesn't accept a type specification.

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: This is completely out of my depth, but is there perhaps some pretty printing feature of Agda we can use?

Comment (VojtechStep) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: We might consider making the projections `opaque` with Agda 2.6.4 and unfold as necessary, but I'm not sure if that would improve the ergonomics

Comment (EgbertRijke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: The problem with modules is that they can't be used for a uniform naming scheme. With the current proposal, we can call a few things `Precategory-With-Families.Something` but we can't add new things to this module. All the stuff that follows would be called `Precategory-With-Families-more-things` or `more-thigs-Precategor-With-Families`. This leads to very ununiform naming.

The module/record approach to naming is _not_ a naming scheme. Therefore I find it unsuitable from a library-wide perspective. I prefer to have a manually defined but _uniform_ naming scheme that looks the same for all entries.

Comment (EgbertRijke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: The naming scheme for precategory with families that would fit the current agda-unimath approach to naming would be

```
entry-Precategory-With-Families
```

for everything. Eg.

```
context-Precategory-With-Families
type-Precategory-With-Families
substitution-Precategory-With-Families
substitution-set-Precategory-With-Families
id-substitution-Precategory-With-Families
comp-substitution-Precategory-With-Families
```

and so on.

Comment (EgbertRijke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: For instance `precategory-Precategory-With-Families` would be the underlying precategory of a precategory with families. The name `type-presheaf-Precategory-With-Families` breaks down as follows: Given a precategory with families this returns a presheaf, hence `presheaf-Precategory-With-Families` is telling what this operation does. Then the name of that presheaf is that it is the presheaf of types. This name goes in front because it takes the highest significance: so the name becomes `type-presheaf-Precategory-With-Families`.

The entry `type-Precategory-With-Families` would return elements of the presheaf of types. Eg. If I assume a type in context gamma, I would assume `A : type-Precategory-With-Families Gamma`.

Comment (EgbertRijke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: I agree that we can improve the way Agda displays goals in our library. That's a very fair point!

Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: > The problem with modules is that they can't be used for a uniform naming scheme. With the current proposal, we can call a few things `Precategory-With-Families.Something` but we can't add new things to this module. All the stuff that follows would be called `Precategory-With-Families-more-things` or `more-thigs-Precategor-With-Families`. This leads to very ununiform naming.
> 
> The module/record approach to naming is _not_ a naming scheme. Therefore I find it unsuitable from a library-wide perspective. I prefer to have a manually defined but _uniform_ naming scheme that looks the same for all entries.

I don't understand this objection. Modules/records are supposed to be opened which leads to lots of flexibility in how to handle naming without interfering with the global name space (https://agda.readthedocs.io/en/latest/language/module-system.html#name-modifiers). Another nice thing you can do is that if you have multiple CwA's around you can write `module C1 = CwA C k` and `module C2 = CwA D l` and then you can write for example `C1.Ty` and `C2.Ty` without Agda getting confused. We use this a lot for Algebra in the cubical library: https://github.com/agda/cubical/blob/master/Cubical/Algebra/Ring/Base.agda#L127

Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: > > The problem with modules is that they can't be used for a uniform naming scheme. With the current proposal, we can call a few things `Precategory-With-Families.Something` but we can't add new things to this module. All the stuff that follows would be called `Precategory-With-Families-more-things` or `more-thigs-Precategor-With-Families`. This leads to very ununiform naming.
> > The module/record approach to naming is _not_ a naming scheme. Therefore I find it unsuitable from a library-wide perspective. I prefer to have a manually defined but _uniform_ naming scheme that looks the same for all entries.
> 
> I don't understand this objection. Modules/records are supposed to be opened which leads to lots of flexibility in how to handle naming without interfering with the global name space (https://agda.readthedocs.io/en/latest/language/module-system.html#name-modifiers). Another nice thing you can do is that if you have multiple CwA's around you can write `module C1 = CwA C k` and `module C2 = CwA D l` and then you can write for example `C1.Ty` and `C2.Ty` without Agda getting confused. We use this a lot for Algebra in the cubical library: https://github.com/agda/cubical/blob/master/Cubical/Algebra/Ring/Base.agda#L127

Compare this with: https://unimath.github.io/agda-unimath/ring-theory.homomorphisms-rings.html#3665

Isn't it nicer to use a bit of modules/records  and infix notations to make things look like normal mathematics?

Comment (EgbertRijke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: I had no idea of this possibility 🙂

Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: (Just to avoid confusion about our slightly funny definition of ring hom, there is a smart constructor that makes it look like you expect: https://github.com/agda/cubical/blob/master/Cubical/Algebra/Ring/Base.agda#L314 

Having redundant data in definitions sometimes make formalization nicer. A standard formalization trick is to duplicate the associativity field for categories to get C^op^op = C definitionally, see https://unimath.github.io/doc/UniMath/4dd5c17/UniMath.CategoryTheory.Categories.html#is_precategory. If you're not already doing this you probably want to if you want to formalize limits/colimits nicely with less duplication)

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: Hi Anders, thanks for taking the time again! Would you mind explaining what the line `module R = RingStr R` does? It's a little hard to parse coming from agda-unimath. 😅 Does it overload the symbol `R` in some way?

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: > Having redundant data in definitions sometimes make formalization nicer. A standard formalization trick is to duplicate the associativity field for categories to get C^op^op = C definitionally, see https://unimath.github.io/doc/UniMath/4dd5c17/UniMath.CategoryTheory.Categories.html#is_precategory. If you're not already doing this you probably want to if you want to formalize limits/colimits nicely with less duplication)

We did consider this briefly, but no one has gotten around to doing anything with it I believe. Do you do something similar for equivalences?

Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: > Hi Anders, thanks for taking the time again! Would you mind explaining what the line `module R = RingStr R` does? It's a little hard to parse coming from agda-unimath. 😅 Does it overload the symbol `R` in some way?

Oh yeah, I'm so used to doing this that I forgot it looks a bit funny. It's explained here in the manual: https://agda.readthedocs.io/en/latest/language/module-system.html#parameterised-modules

So we define a new module which happens to have the same name as the original ring R and its definition is the RingStr of the original R (a RingStr is the **data** of a ring (0,1,+,*,-) without any properties). This gives us access to all the data of R by just writing `R.1`, `x R.+ y`, etc. 

If you want to have a unique name for the ring data you could write for example `module RD = RingStr R` and then the operations would be `x RD.+ y`, etc. This quickly gets old and I like to just use the same name for `R` (similarly to how one often writes `foo {x = x}` when naming implicit arguments when `foo {x = y}` would be less confusing)

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: Right. I mean equivalences of types, and whichever you use as your standard one. For instance, do you prefer using half-adjoint equivalences(coherently invertible maps in agda-unimath) because of its symmetry?

Comment (mortberg) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: > Right. I mean equivalences of types, and whichever you use as your standard one. For instance, do you prefer using half-adjoint equivalences(coherently invertible maps in agda-unimath) because of its symmetry?

Not really... I guess my favorite is contractible fibers because it's the easiest to remember the definition of :-)

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: I see. I always found biinvertibilty the easiest to remember, but I guess that is because it's the first one I learned about and because of the motivation for its definition 😄

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: I have another question with regards to your comment on Discord. Am I correct to interpret it in the form that when we define the underlying type of a proposition as just that, the underlying type of a proposition, then that definition in some way has to carry with it the proof that the type is a proposition, and that is what causes the overhead?

Comment (fredrik-bakke) [reply src/category-theory/precategories-with-attributes.lagda.md:73]: E.g. in this definition:

```agda
  is-category-prop-Precategory : Prop (l1 ⊔ l2)
  is-category-prop-Precategory =
    Π-Prop
      ( obj-Precategory C)
      ( λ x →
        Π-Prop
          ( obj-Precategory C)
          ( λ y → is-equiv-Prop (iso-eq-Precategory C x y)))

  is-category-Precategory : UU (l1 ⊔ l2)
  is-category-Precategory = type-Prop is-category-prop-Precategory
```
would you suggest that the code would be more performant if we were to define `is-category-Precategory` as a separate type with no reference to any proposition?

mortberg: > E.g. in this definition:
> 
> ```agda
>   is-category-prop-Precategory : Prop (l1 ⊔ l2)
>   is-category-prop-Precategory =
>     Π-Prop
>       ( obj-Precategory C)
>       ( λ x →
>         Π-Prop
>           ( obj-Precategory C)
>           ( λ y → is-equiv-Prop (iso-eq-Precategory C x y)))
> 
>   is-category-Precategory : UU (l1 ⊔ l2)
>   is-category-Precategory = type-Prop is-category-prop-Precategory
> ```
> 
> would you suggest that the code would be more performant if we were to define `is-category-Precategory` as a separate type with no reference to any proposition?

I don't think it matters for performance in this particular example... I was more thinking of for example the definition of precategories. Currently properties and data are very much mixed. I would split it into precategory_data (objects, morphisms, id, comp) and then have is_precategory which is parametrized by precategory_data. The nice thing with this is that a lot of things only depend on the data (like an equation `f o g o (h o id) = (f o g) o h` only needs the data). Another nice thing with this is that you can make the is_precategory proofs opaque/abstract everywhere making you sure that things that shouldn't get unfolded doesn't. Another thing to keep in mind is that you could have precategory_data be a record while is_precategory is an iterated Sigma and precategory is a Sigma of these two types. This way you get the data to be a nice record where you can have nice names for the projections making goals display nicely and Agda's module system at hand when opening things to handle name space issues, etc, AND you can use general lemmas about Sigma types to get that is_precategory is a proposition and to characterize the equality of precategories... 

Regarding bundling things into Prop's I like to be on the safe side and try to keep data and properties separate also for Prop's, so I would not bundle them up like you do here. I don't know how much this matters for efficiency, but keep in mind that if you build them bundled like this it's of course harder to make the is-prop part abstract/opaque which might indeed impact performance. (Of course things are a bit subtle in univalent math where propositions are proof relevant and might carry computational content. Obviously we should not make all propositions abstract, e.g. we often need the center of contraction of is-contr even though it's a prop, but if one is careful about it one can make a lot of props abstract without hiding too much)

elisabethstenholm: Have you considered a naming convention similar to Lean? There they put almost everything within a namespace in order to allow for name overloading. So one can then write something like 'Ring.unit' and 'Semiring.unit' to get the corresponding unit elements. (It is possible to open a namespace, but this should only be done very selectively, so far as I recall from my limited Lean knowledge.) 

Lean's naming convention is actually very similar to the naming convention of the library at the moment, where we would write 'context-Precategory-With-Families'. Instead, with a style similar to Lean, we would put things into named modules and one would write the module name followed by '.' and then the term. For example, 'Precategory-With-Families.context'. The good thing with this approach is that a user has more flexibility with the names. For example, when a user is importing the module 'Precategory-With-Families' they could rename it to 'CwF' ('PwF'?) or something they prefer. And then the names would be 'CwF.context' etc. One can also apply parametrised modules, as discussed above. For example, one could write something like 'module myCwF = Precategories-With-Families myPrecategory' and then use 'myCwF.context'.

As it is now, it is possible to rename things, if one wants shorter/different names, but that requires to manually rename every entry when importing the module. Which is 😢

With this said, I don't know who would have the time/energy/motivation to refactor this in the library. I recognise it would be a lot of work!

fredrik-bakke: I still haven't seen a way to add additional definitions to a namespace in a later file. Is this actually possible?

elisabethstenholm: I was thinking about this. I don't think it is possible to extend a module in another file, in Agda. That is a downside with the Lean style approach in our case. Maybe too much of a downside... 🤔

fredrik-bakke: It is certainly one of the reasons we are not already doing this. 😄 The discussion has moved to the Discord server by the way, so it's probably best to continue any discussion there :)

fredrik-bakke: > With this said, I don't know who would have the time/energy/motivation to refactor this in the library. I recognise it would be a lot of work!

It's a good thing I'm a sucker for refactoring with all these change proposals :)

EgbertRijke:
This PR contains many excellent changes so I would like to merge it soon. Since we had a lot of changes recently, causing merge conflicts, I thought I'd help resolve them, and also help formatting the code to satisfy the reviewers. I have worked through the entire PR and made a pull request over at her fork of the library:

https://github.com/elisabethstenholm/agda-unimath/pull/1

I think when she merges that one, it should be ready to merge over here too, because my PR over at her fork is basically an extended review together with a resolution of all my comments.

PR #914, opened 2023-11-11
Cubical compatibility

fredrik-bakke:
Enables the `--cubical-compatible` flag library-wide, and disables the `UnsupportedIndexedMatch` warning.

The flag could not be set literally for the entire library, as it for some reason ate up my entire memory in the `type-theory` module. Since there is no `--no-cubical-compatible` flag, this means that it must be selectively set everywhere else instead of being selectively disabled in the problem files.

The `UnsupportedIndexedMatch` warning states the following:

```text
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor ***,
        which is not yet supported
```

Where `***` can, for instance, be `pair`, `inl`, or `inr`.

This branch type checks with Agda 2.6.4.


EgbertRijke:
Isn't the library super slow with `--cubical-compatible`?

When I made the decision between `--without-K` and `--cubical-compatible` I noticed that the library typechecks a lot faster with `--without-K`, and that's why we've not converted to `--cubical-compatible`. Has this changed?

fredrik-bakke:
I didn't notice a big difference, but the library is already quite slow to type check so that might be why. The reason why I'm posting this PR is I'm doing some cubical shenanigans and wanted to still be able to use the library. If I recall correctly, last time we spoke of cubical compatibility it didn't work, so I thought since it does now, we could maybe reconsider it.

fredrik-bakke:
We can compare times once the CI finishes, I guess

EgbertRijke:
> I didn't notice a big difference, but the library is already quite slow to type check so that might be why. The reason why I'm posting this PR is I'm doing some cubical shenanigans and wanted to still be able to use the library. If I recall correctly, last time we spoke of cubical compatibility it didn't work, so I thought since it does now, we could maybe reconsider it.

Well it gave some warnings back then, not errors. They were warnings of the kind you mentioned in the PR message above.

If the library will be significantly slower I will not be in favor of merging this PR, I'm sorry. Like you say, it is already slow enough as it is.

fredrik-bakke:
10 minutes, so that's 3 minutes slower.

fredrik-bakke:
> > I didn't notice a big difference, but the library is already quite slow to type check so that might be why. The reason why I'm posting this PR is I'm doing some cubical shenanigans and wanted to still be able to use the library. If I recall correctly, last time we spoke of cubical compatibility it didn't work, so I thought since it does now, we could maybe reconsider it.
> 
> Well it gave some warnings back then, not errors. They were warnings of the kind you mentioned in the PR message above.
> 
> If the library will be significantly slower I will not be in favor of merging this PR, I'm sorry. Like you say, it is already slow enough as it is.

That's alright, I figured I could open a PR for discussion since I wanted to implement this version to use for my experiment regardless.

fredrik-bakke:
If there was a `--no-cubical-compatible` flag, we could make it super-easy to enable and disable `--cubical-compatible`

EgbertRijke:
It looks like it is about as fast as it was previously... so perhaps you are right that there is no significant difference. In that case I have no objections.

fredrik-bakke:
Huh, is that right?

EgbertRijke:
> If there was a `--no-cubical-compatible` flag, we could make it super-easy to enable and disable `--cubical-compatible`

This question could be raised with the developers

EgbertRijke:
Yeah, actually I see some advantages too. For instance that we could unpostulate univalence, function extensionality, and pushouts:)

fredrik-bakke:
Oh, because the main branch also uses caching when type checked?

EgbertRijke:
I honestly don't know... but the compile times on here are dependent on chaches anyway, so it is difficult to compare. But if it's just 3 minutes, then maybe it is ok. 

I'll be annoyed with the warnings though, because I'll see them every time I run `make check` in my terminal, which I do very often.

fredrik-bakke:
> Yeah, actually I see some advantages too. For instance that we could unpostulate univalence, function extensionality, and pushouts:)

Don't worry, my experiments do not involve the Kan-involution `~_` ;)

fredrik-bakke:
> I honestly don't know... but the compile times on here are dependent on chaches anyway, so it is difficult to compare. But if it's just 3 minutes, then maybe it is ok.
> 
> I'll be annoyed with the warnings though, because I'll see them every time I run `make check` in my terminal, which I do very often.

As far as I can tell from testing they will not show up anywhere

EgbertRijke:
It's ugly though that you have to enable `--cubical-compatible` in every file. Are you sure there is no other solution?

fredrik-bakke:
I'm not completely sure, no. I can have a second look.

EgbertRijke:
Ok! I'll be offline again. That will give you some time to think about it until at least tomorrow

fredrik-bakke:
There is absolutely no rush, I was not expecting this PR to be merged at all to be honest. It seems to me that if you actually care about cubical compatibility, you will want to use a different library anyway.

fredrik-bakke:
I type-checked this branch vs. master a second time on my laptop. Here are the times
```text
master:              542.87 real       532.49 user         8.37 sys
cubical-compatible:  583.49 real       572.36 user         9.72 sys
```
So using `--cubical-compatible` seems to be approximately 7,5% slower.

EgbertRijke:
I'm still thinking about whether this is a good idea. I would really like to say that the results in this library hold in the internal language of an arbitrary higher topos. This pull request seems to fly in the face of that philosophy. Alternatively, we could have a cubical compatible version on a separate branch, i.e., not on master. 

Then I don't care so much. Since you mentioned that it is only for your experiments, that seems to be appropriate. Or alternatively you can do this just on your own fork too. I think this PR changes what our library is about, so it needs some careful consideration.

fredrik-bakke:
Although this change does not strictly alter the semantics of the library (it just turns on some bookkeeping features of Agda to enable the use of cubical later), I agree with you that this feels like a departure from the principles of agda-unimath given the invasiveness of the change. Hence, I think it is best to keep it in a separate branch in its current form. However, should Agda in the future add functionality equivalent to a `--no-cubical-compatible` flag, I would not be against revisiting this discussion. In that context, the change would be more akin to a slight alteration of our Agda configuration.

PR #915, opened 2023-11-12
Miscellaneous Agda configuration maintenance

fredrik-bakke:
### Summary
- ~Adds built-in binding to `add-ℕ` and `mul-ℕ`~ This leads to the library [not type-checking](https://github.com/UniMath/agda-unimath/actions/runs/6841375267/job/18601660143) for some reason.
- Adds built-in binding to `ℤ`
- Defines `refl` as an instance
- Removes a duplicate paragraph in `foundation.telescopes`

### DIFF
--- a/src/elementary-number-theory/integers.lagda.md
+++ b/src/elementary-number-theory/integers.lagda.md
@@ -44,6 +44,8 @@ negative whole numbers.
 ```agda
 ℤ : UU lzero
 ℤ = ℕ + (unit + ℕ)
+
+{-# BUILTIN INTEGER ℤ #-}
 ```
 
 ### Inclusion of the negative integers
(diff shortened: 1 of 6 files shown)


src/elementary-number-theory/integers.lagda.md
```diff
@@ -44,6 +44,8 @@ negative whole numbers.
 ```agda
 ℤ : UU lzero
 ℤ = ℕ + (unit + ℕ)
+
+{-# BUILTIN INTEGER ℤ #-}
```
EgbertRijke, line 48: Oh! This works now? It used to not work

fredrik-bakke: 🤷I don't know if it has much effect

PR #922, opened 2023-11-16
Displayed precategories

fredrik-bakke:
Defines displayed precategories.


fredrik-bakke:
I'm marking this PR as ready for review without finishing the construction of the fiber category.

src/category-theory/precategories.lagda.md
```diff
@@ -316,6 +317,69 @@ module _
       ( nonunital-precategory-Precategory C)
 ```
 
+### Higher coherences in a precategory
```
EgbertRijke, line 320: I would prefer section titles that express precisely what higher coherences are being defined, because given our conventions the section title might be suggesting a definition of a general higher coherence.

EgbertRijke (approved):
This is a nice pull request. I had only two minor comments. Thanks for adding displayed categories to the library!

fredrik-bakke:
Thanks for the review!

PR #936, opened 2023-11-23
Several facts about `k`-equivalences

tomdjong:
The highlights are the following two results:

1. If `g ∘ f` is `(k+1)`-connected, then `f` is `k`-connected if and only if `g` is `(k+1)`-connected. This is an instance of Proposition 2.31 in [CORS'20] for the truncation modality.
2. Every `k`-equivalence with a section is `k`-connected. This follows from the above as observed by @UlrikBuchholtz.

### DIFF
--- a/src/foundation/truncation-equivalences.lagda.md
+++ b/src/foundation/truncation-equivalences.lagda.md
@@ -17,7 +17,11 @@ open import foundation.fibers-of-maps
 open import foundation.functoriality-dependent-pair-types
 open import foundation.functoriality-truncation
 open import foundation.identity-types
+open import foundation.propositional-truncations
+open import foundation.sections
+open import foundation.surjective-maps
 open import foundation.truncations
+open import foundation.type-arithmetic-dependent-pair-types
 open import foundation.universal-property-dependent-pair-types
 open import foundation.universal-property-truncation
 open import foundation.universe-levels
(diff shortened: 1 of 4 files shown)


EgbertRijke:
Awesome, thanks! I'm just gonna merge this one because I will clean up our open PR backlog tomorrow

src/foundation/truncation-equivalences.lagda.md
```diff
@@ -106,14 +120,393 @@ is-truncation-equivalence-is-equiv-precomp k {A} {B} f H =
...
+      ( map-equiv g)
+      ( f)
+      ( is-equiv-map-equiv g)
+      ( ef)
+```
+
+### The map on dependent pair types induced by the unit of the `(k+1)`-truncation is a `k`-equivalence
+
+This is an instance of Lemma 2.27 in Christensen, Opie, Rijke & Scoccola
+[CORS'20] listed below.
```
fredrik-bakke, line 265: Hey @tomdjong, I don't know how this wasn't picked up by the link-check before your PR was merged because it's [currently blocking mine](https://github.com/UniMath/agda-unimath/actions/runs/6982639824/job/19002239719) (@VojtechStep): Writing text in square brackets `[CORS'20]` is recognized as a link, and makes our link-checker unhappy. For future reference, to write actual square brackets they should be escaped, i.e., `\[CORS'20\]`. Of course, it would also be nice if `[CORS'20]` was a clickable link that sent the user to the references section. This can be accomplished using markdown anchors, but we don't currently have a convention on this.

tomdjong: Thanks @fredrik-bakke! In the meantime, it appears to have gotten fixed upstream by someone else :slightly_smiling_face:

PR #940, opened 2023-11-24
A map is `k`-epic iff its codiagonal is `k`-connected iff its codiagonal is a `k`-equivalence

tomdjong:
Next week I'll make new files on `k`-acyclic types and maps and use this PR to show that being `k`-epic is equivalent to being `k`-acyclic.

The name `is-equiv-horizontal-map-cocone-is-truncation-equivalence-codiagonal-map` is a bit too long for comfort, but I don't know what else to call it.


EgbertRijke:
Thanks! I'll merge this one, and after that I will merge my update on the composition files. So the location of in particular precomposition and postcomposition will then be in their own files. This update should solve the problem that it was nonobvious where lemmas about composition should go.

EgbertRijke:
Auto-merge is on:)

PR #896, opened 2023-10-31
Indiscrete precategories

fredrik-bakke:
This is a small(er) one. It
- Defines indiscrete precategories and proves some basic facts
- Defines the terminal category as an indiscrete precategory and proves some basic facts
- Defines the initial category as an indiscrete precategory and proves some basic facts
- Defines the corollaries of the Yoneda lemma when both functors are representable

Depends on #886 and #880.

PR #941, opened 2023-11-24
Subterminal precategories and constant functors

fredrik-bakke:
Based on the review of #896.

### DIFF
--- a/src/category-theory/constant-functors.lagda.md
+++ b/src/category-theory/constant-functors.lagda.md
@@ -26,9 +26,9 @@ open import foundation.universe-levels
 
 ## Idea
 
-A **constant functor** is a
-[functor](category-theory.functors-categories.lagda.md) `F : C → D` that is
-constant at an object `d ∈ D` and the identity morphism at that object.
+A **constant functor** is a [functor](category-theory.functors-categories.md)
+`F : C → D` that is constant at an object `d ∈ D` and the identity morphism at
+that object.
 
 ## Definition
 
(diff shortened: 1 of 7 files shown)


src/category-theory/constant-functors.lagda.md
```diff
@@ -0,0 +1,107 @@
+# Constant functors
```
fredrik-bakke, line 1: I went with a slightly unconventional structure and put all the different notions of constant functors in one file.

fredrik-bakke: Note that this is actually similar to how some other files are organized, like `terminal-category`.

src/category-theory/indiscrete-precategories.lagda.md
```diff
@@ -44,7 +45,8 @@ hom-[sets](foundation-core.sets.md) are
 
 This construction demonstrates one essential aspect about precategories: While
 it displays `obj-Precategory` as a [retraction](foundation-core.retractions.md),
-up to weak categorical equivalence, every indiscrete precategory is subterminal.
+every indiscrete precategory is
+[subterminal](category-theory.subterminal-precategories.lagda.md).
```
fredrik-bakke, line 49:
```suggestion
[subterminal](category-theory.subterminal-precategories.md).
```

src/category-theory/constant-functors.lagda.md
EgbertRijke: Very nice, I like it. I think it makes good sense with such simple definitions. Although I can see the file becoming more complicated when we add predicates for `is-constant-functor-Categrory` etc (not asking you to do this).

EgbertRijke:
Beautiful!

PR #943, opened 2023-11-24
Reverse the direction of the equivalences on hom-sets in adjunctions

EgbertRijke:


EgbertRijke:
Pretty straightforward PR that hopefully finds a swift merge

src/category-theory/adjunctions-large-categories.lagda.md
```diff
@@ -28,23 +28,28 @@ Two [functors](category-theory.functors-large-categories.md) `F : C → D` and
...
 ```
 
 In this case we say that `F` is **left adjoint** to `G` and `G` is **right
-adjoint** to `F` and write this as `F ⊣ G`.
+adjoint** to `F`, and write this as `F ⊣ G`.
+
+**Note:** The direction of the equivalence `ϕ X Y` is chosen in such a way that
+it often coincides with the direction of the natural map. For example, in the
+[abelianization adjunction](group-theory.abelianization.md), the natural
+candidate for an equivalence is given by precomposition
```
VojtechStep, line 52: Will you elaborate on the precomposition here as you do below?

EgbertRijke: Ay, that looks like a copy-paste error. Thanks for spotting it!

VojtechStep:
Sorry that was a misclick

EgbertRijke:
Thanks for the merge! 😄

PR #947, opened 2023-11-26
Define (sequentially) compact types

fredrik-bakke:


src/synthetic-homotopy-theory/compact-types.lagda.md
```diff
@@ -0,0 +1,55 @@
...
+## Definitions
+
+### The predicate of being a compact type
+
+```agda
+module _
+  {l1 : Level} (X : UU l1)
+  where
+
+  is-compact : UUω
```
EgbertRijke, line 41: I think `is-sequentially-compact` would be a better global name, because there are very many notions of compactness.

PR #954, opened 2023-11-29
Characterization of acyclic types in terms of `const` to identity types being an equivalence

tomdjong:


src/synthetic-homotopy-theory/acyclic-maps.lagda.md
```diff
@@ -151,6 +154,52 @@ module _
             ( e X)))
 ```
 
+### A type is acyclic if and only if the constant map from any identity type is an equivalence
+
+More precisely, `A` is acyclic if and only if for all types `X and elements `x,y
```
VojtechStep, line 159: The backticks ran off

src/synthetic-homotopy-theory/acyclic-maps.lagda.md
```diff
@@ -151,6 +154,52 @@ module _
...
+```
+
+is an equivalence.
+
+```agda
+module _
+  {l : Level} (A : UU l)
+  where
+
+  is-equivalence-const-Id-is-acyclic :
```
VojtechStep, line 173: `is-equivalence` -> `is-equiv`, unless there is some potential for confusion similar to `is-truncated-acyclic` that I'm not seeing? Same thing for other occurrences of `is-equivalence` in this PR

src/synthetic-homotopy-theory/truncated-acyclic-maps.lagda.md
```diff
@@ -159,6 +162,58 @@ module _
             ( e X)))
 ```
 
+### A type is `k`-acyclic if and only if the constant map from any identity type of any `k`-type is an equivalence
+
+More precisely, `A` is `k`-acyclic if and only if for all `k`-types
+`X and elements `x,y : X``, the map
```
VojtechStep, line 168: Backticks

VojtechStep (commented):
I'm reviewing this on my phone, so no GH suggestions. Thanks for the PR!

src/synthetic-homotopy-theory/acyclic-maps.lagda.md
tomdjong: It was just an oversight, thanks!

src/foundation/constant-maps.lagda.md
```diff
@@ -160,3 +162,18 @@ const-injection :
 pr1 (const-injection A B a) = const A B
 pr2 (const-injection A B a) = is-injective-const A B a
 ```
+
+### The action on identifications of the constant map is another constant map
```
EgbertRijke, line 166: This description makes it look like the entry to be defined will be `ap-const`, which is defined in `foundation.action-on-identifications-functions`. But that's not what is going on here. I think it would be clearer to describe this in terms of diagonals:

```text
  B -> (A -> B)
```

which are defined as `const` but somehow that is more confusing in this setting.

tomdjong: Ah, you're right. I had missed `ap-const`. What about giving it the following header?
```
### The action on identifications of a diagonal map is another diagonal map
```

EgbertRijke: Yes, that would be much better. Could you then also adjust the name of the entry to something reasonable?

tomdjong: Would `htpy-diagonal-Id-ap-diagonal-htpy-eq` (changed `const` to `diagonal`) be reasonable?

tomdjong: I implemented this as I'm about to log off (but can change it if necessary).

PR #955, opened 2023-11-29
Adding more tags

EgbertRijke:


EgbertRijke:
In this PR a new phenomenon props up: multiple agda definitions for a single concept!:)

@vojtech do you think this should be possible?

src/graph-theory/directed-graphs.lagda.md
```diff
@@ -18,34 +19,46 @@ open import foundation.universe-levels
 
 ## Idea
 
-A **directed graph** consists of a type of vertices equipped with a binary, type
-valued relation of edges. Alternatively, one can define a directed graph to
-consist of a type `V` of **vertices**, a type `E` of **edges**, and a map
-`E → V × V` determining the **source** and **target** of each edge.
+A
+{{#concept "directed graph" Agda={Directed-Graph,Directed-Graph'} WD="Directed graph" WDID=Q1137726}}
```
VojtechStep, line 23: Multiple `Agda` definitions like this won't be recognized right now and will give a warning (because the preprocessor will be looking for the literal identifier `{Directed-Graph,Directed-Graph'}`)

src/graph-theory/edge-coloured-undirected-graphs.lagda.md
```diff
@@ -20,14 +20,17 @@ open import graph-theory.undirected-graphs
 
 ## Idea
 
-An **edge-coloured undirected graph** is an
-[undirected graph](graph-theory.undirected-graphs.md) equipped with a family of
-maps `E p → X` from the edges at
+An
+{{#concept "edge-coloured undirected graph" Agda=Edge-Coloured-Undirected-Graph WD="Graph coloring" WDID=Q504843}}
```
VojtechStep, line 24: "we use US English" - would it be reasonable to change "colour" to "color" while you're touching these files?

EgbertRijke: Yeah! I noticed too that there was a problem here but I didn’t address it. Good call. I will address it.

VojtechStep:
(wrong Vojtěch - sorry for the ping)
I'd need to think about what this means for the database and for Mathswitch. For the time being I'd stick to one definition per concept

src/graph-theory/directed-graphs.lagda.md
EgbertRijke: If possible I would like to avoid having to go through all of these files multiple times for entering the metadata. Would it be reasonable to keep the multiple Agda definitions in there so that I don’t have to go through all of the files again later to see if there were multiple Agda definitions in some of the files?

EgbertRijke: Or would it be better to use the Agda= field multiple times?

EgbertRijke: Or would it be better to use the Agda= field multiple times?

VojtechStep: If you want the syntax to work now, then repeating the `Agda` field should work.
However - if we're gonna allow splitting `{{#definition}}` and `{{#concept-data}}`, and if we'll be adding more features, then I assume we'll revisit these tags anyway, right?

fredrik-bakke:
Is the mathswitch label correct to add for these PRs?
Also, a newbie question, how does the macro discern between for instance Wikidata "wiki" entries and "entity" entries. Are the wikidata id's unique across these categories?
Finally, I think I saw someone recently state that the macro should error if the agda field is omitted. I don't think that should be the case, as there occasionally appear files with only prose. Alternatively, we should have a separate macro for those scenarios.

EgbertRijke:
I think that *if* the Agda field is provided, then an error should be generated if an erroneous name is entered in the Agda field. It would be a pain to maintain if this isn’t checked automatically. I do agree that the Agda field should be optional

fredrik-bakke:
Ah, yeah, that makes sense.

EgbertRijke:
This PR is getting stale and can't be merged until we enhance our infrastructure for the concept macro, so I will just close it for now.

PR #959, opened 2023-11-30
Play with `hom-arrow` as a record

fredrik-bakke:
So after seeing Egbert's proof of the fact that every retract of maps also defines a retract on fiber inclusions, I wanted to put records to the test to see what the fuzz was all about. This PR is purely for demonstration and I am not trying to outright advocate for using records, but we can already see a pretty good mileage out of changing just one definition to a record. I.e., perhaps it is good to consider using records for just very particular definitions? For instance, the mentioned proof becomes 48 lines shorter after the change.

### DIFF
--- a/src/foundation/cones-over-cospans.lagda.md
+++ b/src/foundation/cones-over-cospans.lagda.md
@@ -73,12 +73,12 @@ module _
   coherence-square-cone = pr2 (pr2 c)
 
   hom-arrow-cone : hom-arrow vertical-map-cone g
-  pr1 hom-arrow-cone = horizontal-map-cone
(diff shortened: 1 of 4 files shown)


EgbertRijke:
Note that this pull request breaks the torsoriality of `htpy-hom-arrow`, so it should not be an open pull request.

fredrik-bakke:
Right. As stated, it was posted to open for discussion, not to function as a serious PR.

EgbertRijke:
In my opinion, the real way to make this proof stronger is by developing a library of wild categories. The lemma that is made shorter here should be a direct application of the theory of wild categories, and it would be even shorter than what is proposed here.

PR #960, opened 2023-12-01
Closure properties of acyclic maps and types

tomdjong:
I also formalized that we have symmetric pushout squares (for pullbacks this was already in the library), i.e. the pushout of two maps `f` and `g` is also the pushout of `g` and `f`.


src/synthetic-homotopy-theory/acyclic-maps.lagda.md
```diff
@@ -289,6 +302,227 @@ module _
...
+    apply-universal-property-trunc-Prop h
+      ( is-acyclic-Prop A)
+      ( λ a →
+        is-acyclic-is-acyclic-map-terminal-map A
+          ( is-acyclic-map-left-factor
+            ( terminal-map)
+            ( point a)
+            ( is-acyclic-map-terminal-map-is-acyclic unit is-acyclic-unit)
+            ( λ b → is-acyclic-equiv (fiber-const a b) (l-ac a b))))
+```
```
fredrik-bakke, line 524: It seems appropriate to me to write a new file on locally acyclic types, where in particular `is-locally-acyclic` is defined

fredrik-bakke (approved):
Very nice pull request, you're picking up steam with these!

It's a bit unfortunate that you have results about acyclic types in the file about acyclic maps, but I see there's a dependency issue.

tomdjong:
> It's a bit unfortunate that you have results about acyclic types in the file about acyclic maps, but I see there's a dependency issue.

Yes, I agree. I have some ideas on how to resolve this, but I will check them when I meet with Egbert next week. For the moment, I would be happy to get all closure properties merged and then I can reorganize later.

src/synthetic-homotopy-theory/acyclic-maps.lagda.md
tomdjong: If I had a few more results about locally acyclic types, or if it turned out to be a very interesting notion on its own, then I would probably agree with you, but since this is the only result I would honestly prefer not to create a separate file.

PR #885, opened 2023-10-26
Rebase infrastructure for pushouts to span diagrams

EgbertRijke:
In this PR I refactor the definitions of spans, morphisms of spans, and equivalences of spans (previously called homotopies of spans), and I will work on a better integration of this infrastructure in the theory of pushouts.

Since this touches upon earlier work of @VojtechStep and @tomdjong, perhaps you two can have a look at this PR when it's ready. (That's not to exclude @fredrik-bakke from also having a look)

PR #961, opened 2023-12-01
Refactor universal property of suspensions

fredrik-bakke:
Refactors universal properties of suspensions to use the `Level`-naming scheme.


src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
```diff
@@ -87,22 +88,58 @@ suspension-structure-suspension :
...
+[equivalence](foundation-core.equivalences.md). Note that this proposition is
+[small](foundation-core.small-types.md), whereas
+[the universal property](foundation-core.universal-property-pullbacks.md) is a
+large proposition.
+
+```agda
+is-suspension :
+  {l1 l2 : Level} {X : UU l1} {Y : UU l2} →
+  suspension-structure X Y → UU (l1 ⊔ l2)
+is-suspension s = is-equiv (cogap-suspension-structure-suspension s)
```
fredrik-bakke, line 129: @morphismz While refactoring the universal properties for suspensions, I noticed you had given a different definition of `is-suspension`. The convention is to define it as I have done here. The point is that from this small proposition, we can still derive the large universal property. :)

EgbertRijke:
Wouldn't it be better convert to stating universal properties in `UUomega` only? I don't really know what the level-wise universal properties are good for

fredrik-bakke:
> Wouldn't it be better convert to stating universal properties in `UUomega` only? I don't really know what the level-wise universal properties are good for

Backward compatibility, I suppose.

fredrik-bakke:
Also, sometimes it is used in Sigma types. E.g.

```agda
  universal-property-pushout-extension-by-equivalences :
    {l : Level} → is-equiv i → is-equiv j → is-equiv k →
    Σ (cocone f' g' X) (λ d → universal-property-pushout-Level f' g' d l)
```

src/foundation/unit-type.lagda.md
```diff
@@ -49,7 +49,7 @@ ind-unit p star = p
 
 ```agda
 terminal-map : {l : Level} {A : UU l} → A → unit
-terminal-map = const _ unit star
+terminal-map {A = A} = const A unit star
```
EgbertRijke, line 52: I'm actually not entirely comfortable with it if you go through the library and change the underscores to `{X = X}` notation.

EgbertRijke, line 52: I'm actually not entirely comfortable with it if you go through the library and change the underscores to `{X = X}` notation.

EgbertRijke: A better way to do it, is to open an anonymous module with parameters `{l : Level} {A : UU l}`. That puts `A` in the context.

fredrik-bakke: I see. For a minute there I thought you were contesting the other aspect of the change.

src/synthetic-homotopy-theory/dependent-universal-property-suspensions.lagda.md
```diff
@@ -43,16 +43,21 @@ pr1 (dependent-ev-suspension s B f) =
...
 
 module _
-  (l : Level) {l1 l2 : Level} {X : UU l1} {Y : UU l2}
+  {l1 l2 : Level} {X : UU l1} {Y : UU l2}
   (s : suspension-structure X Y)
   where
 
-  dependent-universal-property-suspension : UU (l1 ⊔ l2 ⊔ lsuc l)
-  dependent-universal-property-suspension =
+  dependent-universal-property-suspension-Level :
```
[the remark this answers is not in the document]
fredrik-bakke: Your point is that if one wants to define Sigma types of suspensions, then one should use the small proposition `is-suspension` rather than this type?

src/foundation/universal-property-cartesian-product-types.lagda.md
```diff
@@ -36,24 +36,30 @@ product
 ### The universal property of cartesian products as pullbacks
 
 ```agda
-universal-property-product :
+map-up-product :
```
EgbertRijke, line 39: I'm not sure if I like `up` very much either. I'm sorry that I don't like very much, but `map-universal-property-product` reads much better to me.

src/foundation/universal-property-cartesian-product-types.lagda.md
```diff
@@ -36,24 +36,30 @@ product
...
+  ((x : X) → A x × B x) → (((x : X) → A x) × ((x : X) → B x))
+pr1 (map-up-product f) x = pr1 (f x)
+pr2 (map-up-product f) x = pr2 (f x)
+
+up-product :
   {l1 l2 l3 : Level} {X : UU l1} {A : X → UU l2} {B : X → UU l3} →
   ((x : X) → A x × B x) ≃ (((x : X) → A x) × ((x : X) → B x))
-pr1 universal-property-product f = (λ x → pr1 (f x)) , (λ x → pr2 (f x))
-pr2 universal-property-product =
+pr1 up-product = map-up-product
```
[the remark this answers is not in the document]
fredrik-bakke: Your particular tastes is what keeps this library afloat

src/synthetic-homotopy-theory/circle.lagda.md
```diff
@@ -433,7 +433,7 @@ pr2 (pr2 dependent-suspension-structure-sphere-1-circle-sphere-1) =
 sphere-1-circle-sphere-1 : section sphere-1-circle
 pr1 sphere-1-circle-sphere-1 = circle-sphere-1
 pr2 sphere-1-circle-sphere-1 =
-  map-inv-dependent-up-suspension
+  map-inv-dup-suspension
```
EgbertRijke, line 436: `dup` is even worse than `up`. But it's an imporovement over `dependent-up`:)

src/foundation/universal-property-cartesian-product-types.lagda.md
fredrik-bakke: Your particular tastes is what keeps this library afloat

src/synthetic-homotopy-theory/dependent-universal-property-suspensions.lagda.md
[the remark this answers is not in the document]
EgbertRijke: No, my point is that we should not have any "universal properties" that are restricted to one universe level. We should just state universal properties directly in `UUomega`

src/synthetic-homotopy-theory/dependent-universal-property-suspensions.lagda.md
```diff
@@ -43,16 +43,21 @@ pr1 (dependent-ev-suspension s B f) =
...
+    (l : Level) → UU (l1 ⊔ l2 ⊔ lsuc l)
+  dependent-universal-property-suspension-Level l =
     (B : Y → UU l) → is-equiv (dependent-ev-suspension s B)
+
+  dependent-universal-property-suspension : UUω
```
EgbertRijke, line 58: This one is good. It should be defined directly, not in terms of `dependent-universal-property-suspension-Level`.

src/synthetic-homotopy-theory/dependent-universal-property-suspensions.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: I see. Well, then I concur, and propose the solution that if one wants to define a Sigma type of suspensions (or other), then one should use the associated small proposition that entails the large universal property.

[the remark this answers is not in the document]
EgbertRijke: I think I am completely missing your point about Sigma-types of suspensions. Could you explain it to me in more detail?

EgbertRijke:
> Reference in new issue

Ah, your point is here. That code doesn't look very good to me. Looks like I have another refactoring job to do.

EgbertRijke:
Wait I think I am already doing that refactoring in #885. Mind you I might have been addressing some of these universal properties there too.

EgbertRijke:
I think #885 is already much deeper in than you are here. Since this pr is <200 lines of code, perhaps it would be better to abandon this one. I'm really sorry.

fredrik-bakke:
Well, 200 LOC is what I've pushed...

fredrik-bakke:
Also, there doesn't seem to be any overlap in what I've actually pushed. Am I missing something?

fredrik-bakke:
I decided to just ignore the `up` problem for now, as it seems to be a more delicate problem than I first anticipated.

src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
morphismz: @fredrik-bakke  Oh, sorry about that. I forget even writing the definition of `is-suspension`. I'll think about how to phrase things this way for the universal property of fibers.

src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
```diff
@@ -87,22 +88,58 @@ suspension-structure-suspension :
...
+### The cogap map of a suspension structure
+
+```agda
+module _
+  {l1 l2 : Level} {X : UU l1} {Y : UU l2} (s : suspension-structure X Y)
+  where
+
+  cogap-suspension-structure-suspension : suspension X → Y
+  cogap-suspension-structure-suspension =
+    cogap-suspension (cocone-suspension-structure X Y s)
```
morphismz, line 113: I'm not quite sure I understand the point of `cogap-suspension-structure-suspension`, since below we have `map-inv-up-suspension` which serves the same purpose. We should at least have a lemma that relates these two: i.e. say that `cogap-suspension-structure-suspension c` is the same map as `map-inv-up-supension (suspension-structure-suspension-cocone c)`.

Could you explain a bit about why we need this?

fredrik-bakke: oh, that's because I didn't notice it was the same map. What I have defined follows our naming conventions more closely, however, and as it is not defined as the inverse of a composite of equivalences, it may happen that it enjoys better computational properties. I don't know for sure though.

I defined it because we conventionally define `is-universal-thing` predicates as the predicate that a particular (co)gap map is an equivalence.

morphismz: I'm not sure what computational properties `cogap-suspension-structure-suspension` will have, since `cogap` itself is defined as the inverse of `equiv-up-pushout` (which is postulated to be an equivalence, so the inverse is essentially postulated). And `equiv-up-suspension` is not a composite of equivalences; its underlying map is just `ev-suspension`, which has very nice computational properties. We just show that `ev-suspension` is an equivalence by showing it is homotopic to a composite of equivalences.

I may protest a bit against getting rid of `map-inv-up-suspension`, unless we really have a reason, for the above reasons. Additionally:

(i) a lot of the characterization of homotopies between maps out of a suspension relies on having `ev-suspension` as the inverse of `map-inv-up-suspension`, and also relies on the dependent version of this

(ii) we already have proofs computing how  `map-inv-up-suspension` behaves applied to `north-suspension`, ect.

(iii) there is code in the suspensions file and the suspensions structures file that essentially establishes an equivalence between the universal property of suspensions stated in terms of suspension structures and stated in terms of cocones. This is how `up-suspension` was constructed. This gives an easy way to convert between the two universal properties. Why by pass all that? 

Couldn't we just rename `map-inv-up-suspension` to `cogap`?

fredrik-bakke: Excellent, I'll have a closer look momentarily

fredrik-bakke: I'm purely going by staring at the code currently, but one computational property that `cogap-suspension-structure-suspension` has that I'm not sure `map-inv-up-suspension` does is that it is definitionally equal to `cogap-suspension`, hence there will never be a need to have homotopy lemmas between the two. The definition of `map-inv-up-suspension` in comparison seems rather ad-hoc.

Since `ev-suspension` is shown to be an equivalence by being homotopic to a composite of equivalences, the definition of its inverse will compute (with the current definitions) to be composite of the sections of those two equivalences I believe. However, you should not rely on the computational properties of `map-inv-equiv` as it is subject to change and will be very unreliable. Doing that systematically is likely to lead to a maintenance nightmare and make the code unrefactorable. Indeed, I believe Egbert has been playing with the idea of making `map-inv-equiv` abstract.

As you point out, the same problem does not occur for the other definitions of cogap maps, as their underlying(/overarching?) equivalences are only postulated, hence wouldn't reduce regardless.

fredrik-bakke: Nothing stops us from showing that `cogap-suspension-structure-suspension` is the inverse to `ev-suspension`, and I can't see anything that suggests that `map-inv-up-suspension` is a better choice of inverse.

fredrik-bakke: I take issue with the "if it ain't broke, don't fix it" mentality as it doesn't take into consideration that others may want to read or use the code in the future. Moreover, it goes against the library's goal of being maintainable and scalable.

morphismz: Maybe I'm missing your point, but it seems like the main reason you want to get rid of `map-inv-up-suspension` is possible reduction behavior. But I'm just not seeing how such a replacement will help.

> Yep, I just verified that the two functions are not definitionally equal, as the following does not typecheck
> 
> ```agda
>   eq-cogap-suspension-structure-suspension :
>     {l : Level} (Z : UU l) →
>     cogap-suspension-structure-suspension {l2 = l} {X} {Z} ＝ map-inv-up-suspension {l} Z
>   eq-cogap-suspension-structure-suspension Z = refl
> ```

This just seems to say that the map we currently have is not definitionally the same as the map you wish to replace it with. Unless I am missing something, I do not see how this suggests either has better computational properties than the other. It just says they don't have the same such properties. 

> one computational property that `cogap-suspension-structure-suspension` has that I'm not sure `map-inv-up-suspension`
> does is that it is definitionally equal to `cogap-suspension`

Similarly, this just seems to say they don't have the same computational properties. Could you say more about why being definitionally equal to `cogap-suspension` is a virtue? After all, we have not needed `cogap-suspension` in the library before. And I don't yet understand why its being added now (it seems like its being added as part of a refactoring, rather than being added to serve some purpose/prove some theorem). 

> Since `ev-suspension` is shown to be an equivalence by being homotopic to a composite of equivalences, the 
> definition of its inverse will compute (with the current definitions) to be composite of the sections of those two 
> equivalences I believe. However, you should not rely on the computational properties of `map-inv-equiv` as it is subject
> to change and will be very unreliable. Doing that systematically is likely to lead to a maintenance nightmare and make 
> the code unrefactorable. Indeed, I believe Egbert has been playing with the idea of making `map-inv-equiv` abstract.

I suppose what I should have said earlier is this: we can never truly benefit from computational properties of inverse of an equivalence defined by a postulated universal property. Simply because the inverse is postulated; it won't really have any reduction behavior (unless I'm misunderstanding things). So, it seems that using `cogap` versus `map-inv-up-suspension` will make limited difference, because both are postulated inverses, and have no reduction behavior.

So, to your point that the inverse of `ev-suspension` will be a composite of two sections: I don't think this is having much effect on the way we prove things. The inverse doesn't have any interesting reduction behavior, since one of these sections is postulated. If we made `map-inv-equiv` abstract, I don't think it would have too much effect. Taking a look at where we prove things about `map-inv-equiv`, it doesn't look like we use any reduction behavior of the map in question.

I guess a summary or tl;dr of my qualms is this: I am not really worried about the underlying map of the inverse, since this will ultimately be a postulated thing, which will not ave any interesting reduction behavior. But, what I really want to keep is that `ev-suspension` is (definitionally) the underlying map of the equivalence. Does this happen the way you suggest reworking things? I don't think it does.

Having `ev-suspension` as the underlying map has been very helpful in my experience with using suspensions in the library. Things such as the characterization of homotopies between maps out of a suspension use thing to their advantage. So does the suspension loop space adjunction. A fair amount of the code I have written relies on using `ev-suspension` and `dependent-ev-suspension` as the map of the equivalence.

I don't yet understand why we should move away from this? I don't think we should move away from this to try to get a better inverse, since the inverse will ultimately be (or have a part of it that is) postulated.


Sorry for causing such a fuss over something so small. Maybe, at the end of the day, this change doesn't affect that much.

fredrik-bakke: By definition of the universal property of suspensions, the underlying map of `up-suspension` _must_ be `ev-suspension`

fredrik-bakke: my point is not that `cogap` should reduce to something that "computes" to even more basic primitives, but that other definitions may compute at least in terms of `cogap`.

fredrik-bakke: This may become useful, as we often use `cogap` for other constructions, so then we can more easily relate our constructions to one another. Perhaps I can draw the comparison to the importance of `ev-suspension` to begin with.

fredrik-bakke: You are correct that I don't have a particular problem that my suggestion solves in mind at present. However, this is part of taking a principled approach, and serves to reduce the probability of headaches down the line. Indeed, whole libraries have been rewritten in the past because the authors made less principled decisions about their definitions, even though they were all equivalent.

morphismz: I see. I think you have a better sense of how these things will pan out, so I will go with your decisions here.

fredrik-bakke: Aside from renaming `map-inv-up-suspension`, you shouldn't have to do anything

morphismz: I spent a little more time looking at this and realized a few things: 

(i) the definition of `up-suspension` could have just been:

```agda
  up-suspension :
    {l : Level} →
    universal-property-suspension l X
      ( suspension X)
      ( suspension-structure-suspension X)
  up-suspension Z =
      ( is-equiv-map-equiv
        ( ( equiv-suspension-structure-suspension-cocone X Z) ∘e
          ( equiv-up-pushout terminal-map terminal-map Z)))
```

(ii) I think I pin pointed what prevents `map-inv-up-suspension` from being definitionally the same as `cogap-suspension-structure-suspension`. By definition, `cogap-suspenion-structure-suspension` is the composite of `cogap` and `cocone-suspension-structure`, while `map-inv-up-suspension` is the composite of `cogap` and `map-inv-equiv suspension-structure-cocone`. And `cocone-suspension-structure` is not definitionally `map-inv-equiv suspension-structure-cocone`, but is homotopic to it.

That is all to say, you were totally right :)

fredrik-bakke: I'm glad this could be a learning experience for you. 😄 Indeed, it made me reflect on things again that I mostly take for granted nowadays, and when I take things for granted I tend to end up being wrong in the end.

I want to emphasize the point that we are very meticulous about when we invoke reduction behaviors in agda-unimath. Indeed, this is why we mark so many things `abstract` and why one of Egbert's mantras is that we like to display our proofs by logical arguments rather than by letting the machine sort it out for us. I think this is one of the great things about agda-unimath, although others lament over this approach. One of the features of this method is that it forces us to be very explicit and methodical about how we write our constructions, precisely as demonstrated by how I've refactored your code today. Indeed, if we did not mark so many things abstract, there's a chance that the two definitions would have coincided by luck, rather than by intention.

fredrik-bakke: Anyway, I think I'm done with this PR now. I would be happy if you reviewed it before it is merged.

morphismz: @fredrik-bakke It looks pretty good to me! I don't have any other comments. I can give the code a more thorough look through later tonight or tomorrow moring if need be.

src/species/composition-cauchy-series-species-of-types.lagda.md
```diff
@@ -88,8 +88,8 @@ module _
             ( λ V →
               ( equiv-prod
                 ( id-equiv)
-                ( ( inv-equiv universal-property-product) ∘e
-                  ( equiv-prod id-equiv equiv-ev-pair))) ∘e
+                ( inv-equiv equiv-up-product ∘e
+                  equiv-prod id-equiv equiv-ev-pair)) ∘e
```
VojtechStep, line 92: Don't we usually keep the operands of a multiline infix application paranthesized?
```suggestion
                ( ( inv-equiv equiv-up-product) ∘e
                  ( equiv-prod id-equiv equiv-ev-pair))) ∘e
```

src/synthetic-homotopy-theory/circle.lagda.md
```diff
@@ -470,8 +463,8 @@ apply-up-suspension-meridian-one-suspension-circle-sphere-1-circle =
   ( identification-left-whisk
     ( inv (ap circle-sphere-1 (meridian-suspension (one-Fin 1))))
     ( inv
-      ( up-suspension-meridian-suspension
-        (sphere 0) 𝕊¹ suspension-structure-sphere-0-𝕊¹ (one-Fin 1)))) ∙
+      ( compute-meridian-cogap-suspension-structure
+          suspension-structure-sphere-0-𝕊¹ (one-Fin 1)))) ∙
```
VojtechStep, line 467:
```suggestion
          ( suspension-structure-sphere-0-𝕊¹)
          ( one-Fin 1)))) ∙
```

src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
```diff
@@ -224,83 +259,84 @@ dependent-up-suspension l {X = X} B =
...
+pr1 (equiv-dup-suspension {X = X} B) =
+  dependent-ev-suspension (suspension-structure-suspension X) B
+pr2 (equiv-dup-suspension B) = dup-suspension B
 
 module _
   {l1 l2 : Level} {X : UU l1} (B : suspension X → UU l2)
   where
 
-  map-inv-dependent-up-suspension :
+  map-inv-dup-suspension :
```
VojtechStep, line 274: TILI: this could be called `dependent-cogap-suspension`

src/species/composition-cauchy-series-species-of-types.lagda.md
fredrik-bakke: Yes, thanks for spotting that. Egbert recently changed his mind about this regarding top-level expressions of identity type formation, but I don't yet understand where to draw the line precisely.

src/synthetic-homotopy-theory/suspensions-of-types.lagda.md
fredrik-bakke: Sorry, what does TILI mean? Thanks for the suggestion

VojtechStep: Take it or leave it - a remark more than anything, so that you can mark the conversation as resolved without any action.

fredrik-bakke: Ah, gotcha. I'll rename it, because it matches the other refactor I did :)

fredrik-bakke: Done!

PR #846, opened 2023-10-16
Compatibility patch for Agda 2.6.4

fredrik-bakke:
Makes the library compatible with Agda 2.6.4.

PR #962, opened 2023-12-02
Update reflection for 2.6.4

FernandoChu:
Hello, sorry for the delay an non-communication, I had to do some other stuff. This PR just uncomments some of the new BUILTINs and replaces the old `AGDATCMBLOCKONMETA` that was deleted in https://github.com/UniMath/agda-unimath/pull/846.


fredrik-bakke (approved):
Nice! Thanks for the help

PR #969, opened 2023-12-04
The `is-sequential-colimit` predicate

### LINKED ISSUE #966: Add `is-sequential-colimit`
In line with the other `is-<colimit>` properties - it should be a property of cocones, defined as the map `cogap-sequential-colimit c : sequential-colimit A -> X` being an equivalence. Then `is-sequential-colimit c` should be logically equivalent to `universal-property-sequential-colimit c`.

fredrik-bakke:
Resolves #966. Builds on #919.


EgbertRijke:
Excellent PR!

PR #973, opened 2023-12-05
Action on homotopies of functions

fredrik-bakke:
### Summary
- Defines a file for the action on homotopies of functions
- Refactors the universal property of contractible types to a new file (this was initially to avoid cyclic dependencies)
- Split function extensionality into core again, so we can use the lemmas in `foundation.commuting-squares-of-maps` for efficient proofs of different functoriality properties of `eq-htpy`. Unless we decide to create a file called `functoriality-function-extensionality`, I think it is unavoidable to have a core file as well.


fredrik-bakke:
Let me know if I'm using the concept tags wrong btw!

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -0,0 +1,283 @@
...
+
+</details>
+
+## Idea
+
+The
+{{#concept "function extensionality axiom" Agda=function-extensionality Agda=funext}}
+asserts that [identifications](foundation-core.identity-types.md) of (dependent)
+functions are [equivalently](foundation-core.equivalences.md) described as
+pointwise equalities between them. In other words, a function is completely
```
EgbertRijke, line 33: This sentence should probably mention homotopies instead of pointwise equalities, because then we can link to `foundation-core.homotopies.md`

src/foundation/action-on-homotopies-functions.lagda.md
```diff
@@ -0,0 +1,127 @@
...
+
+```text
+  F : ((x : A) → B x) → C
+```
+
+we thus get an action of type
+
+```text
+  f ~ g → F f ＝ F g.
+```
```
EgbertRijke, line 43: So the action on homotopies of the identity function is `eq-htpy` :)

src/foundation/action-on-homotopies-functions.lagda.md
```diff
@@ -0,0 +1,127 @@
...
+
+  compute-action-htpy-function-refl-htpy :
+    (f : (x : A) → B x) → action-htpy-function refl-htpy ＝ refl
+  compute-action-htpy-function-refl-htpy f =
+    ap (ap F) (eq-htpy-refl-htpy f)
+```
+
+## Properties
+
+### The action on homotopies of a constant map is constant
```
EgbertRijke, line 79: I would put this in the file about constant maps. My reasoning is as follows:

If you introduce a general concept, then the `## Properties` section should more or less be about properties at the same level of generality. The level of generality of the concept of "action on homotopies" is clearly greater than the level of generality of the concept of "constant map". On the other hand, the lemma about constant maps shown here is exactly at the same level of generality of the concept of "constant map", so it seems more natural to me to put it there.

All of the above are of course vague heuristics.

src/foundation/function-extensionality.lagda.md
```diff
@@ -26,243 +29,103 @@ open import foundation-core.precomposition-functions
...
-  where
-
-  htpy-eq-concat :
-    (p : f ＝ g) (q : g ＝ h) → htpy-eq (p ∙ q) ＝ (htpy-eq p ∙h htpy-eq q)
-  htpy-eq-concat refl refl = refl
+    ( coherence-square-htpy-eq-ap-precomp f g h)
 ```
 
-### `eq-htpy` preserves concatenation of homotopies
+### `eq-htpy` preserves inverses
```
EgbertRijke, line 110: Perhaps an informal statement of what is being shown here would be nice

src/foundation/universal-property-contractible-types.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+open import foundation-core.identity-types
+open import foundation-core.singleton-induction
+```
+
+</details>
+
+## Idea
+
+The
+{{#concept "dependent universal property of [contractible types](foundation-core.contractible-types.md)" Agda=dependent-universal-property-contr}}
```
EgbertRijke, line 28: I don't think we should have links inside a concept name. It is probably fine if we link to contractible type in the sentence where we say that a type satisfies the universal property of contractible types if and only if it is contractible.

EgbertRijke: Perhaps @VojtechStep has more knowledge on this. Are links in concept identifiers ok?

src/foundation/universal-property-contractible-types.lagda.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+is an [equivalence](foundation-core.equivalences.md) for every type family
+`P : A → 𝒰`.
+
+That `ev-point` is a [section](foundation-core.sections.md) corresponds to that
+contractible types satisfy
+[singleton induction](foundation-core.singleton-induction.md), while the
+[retraction](foundation-core.retractions.md) corresponds to uniqueness of this
+induction principle.
```
EgbertRijke, line 42: This paragraph is written in a confusing manner, because it seems to highlight a contrast:

"That `f` is a section ..., while the retraction"

The first part of the sentence is also not entirely correct, because actually singleton induction corresponds to `ev-point` _having_ a section, not _being_ a section. Note also that the sentence "f is a section" leaves implicit what `f` is a section of. Sometimes this is clear from the context, but in this paragraph it is definitely not clear what `ev-point` is conditioned to be section of. It would be clearer to be explicit and say that "f is a section of g".

````suggestion
The condition that `ev-point` has a [section](foundation-core.sections.md)

```text
  P a → ((x : A) → P x)
```

is another way of phrasing that the type satisfies
[singleton induction](foundation-core.singleton-induction.md). Furthermore, the condition that `ev-point` has a [retraction](foundation-core.retractions.md) asserts that all dependent functions `(x : A) → P x` are fully determined by their value at `a`.
````

File: src/foundation/homotopy-induction.lagda.md [hunk 7]
```diff
@@ -28,13 +35,30 @@ The principle of **homotopy induction** asserts that homotopies form an
...
+    (C : (Σ ((x : A) → B x) (f ~_)) → UU l3) →
+    coherence-triangle-maps
+      ( ev-point (f , refl-htpy))
+      ( ev-refl-htpy (λ g H → C (g , H)))
+      ( ev-pair)
+  triangle-ev-refl-htpy C F = refl
+```
 
+### The homotopy induction principle
+
```
EgbertRijke, line 60: It would be nice to write the informal statement of homotopy induction here, or in the idea section.

EgbertRijke:
Great pull request. Thank you for making all those changes!

src/foundation/action-on-homotopies-functions.lagda.md
```diff
@@ -0,0 +1,127 @@
...
+</details>
+
+## Idea
+
+Applying the
+[action on identifications](foundation.action-on-identifications-functions.md)
+to [identifications](foundation-core.identity-types.md) arising from the
+[function extensionality axiom](foundation.function-extensionality.md) gives us
+the
+{{#concept "action on homotopies" Disambiguation="of functions" Agda=action-htpy-function}}.
```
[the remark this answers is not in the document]
EgbertRijke: Its purpose is to be able to add disambiguations for concepts that could make sense in multiple settings.

src/foundation/universal-property-contractible-types.lagda.md
EgbertRijke: Ok, thank you Vojtech! That's awesome.

Then we can discuss separately whether it is a good idea to put links in the defined terms. Personally I find this practice a somewhat unaesthetic and I don't use it.

src/foundation/action-on-homotopies-functions.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: To the machine, to the reader, or to the implementer? If it's also to the reader, then we should think about how that information is displayed to them.

src/foundation/universal-property-contractible-types.lagda.md
fredrik-bakke: Thanks, Egbert!

fredrik-bakke:
I've addressed your comments as far as I was able to now. Thank you for the nice review!

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -0,0 +1,283 @@
...
+
+Consider a map `f : A → B` and two dependent functions `g h : (b : B) → C b`.
+Then the square
+
+```text
+                     ap (precomp-Π f C)
+       (g ＝ h) ---------------------------> (g ∘ f ＝ h ∘ f)
+          |                                         |
+  htpy-eq |                                         | htpy-eq
+          V                                         V
```
fredrik-bakke, line 176: We should prefer lowercase `v` in diagrams for arrow-heads btw, since they are more similar to `<`, `>` and ´^´. I'll fix it repo-wide as part of a larger refactor later on though.

src/foundation/function-extensionality.lagda.md
```diff
@@ -26,243 +29,115 @@ open import foundation-core.precomposition-functions
...
-### Naturality of `htpy-eq` for dependent functions
+### Naturality of `eq-htpy` for dependent functions
 
 Consider a map `f : A → B` and two dependent functions `g h : (b : B) → C b`.
 Then the square
 
 ```text
                      ap (precomp-Π f C)
        (g ＝ h) ---------------------------> (g ∘ f ＝ h ∘ f)
+          ^                                         ^
```
VojtechStep, line 48: Why should the squares in this file go upwards?

VojtechStep (commented):
I have some nitpicks, but otherwise LGTM 👌

src/foundation-core/function-extensionality.lagda.md
VojtechStep: Could you not do it as part of a larger refactor, but rather as its own thing that we can merge fast?

fredrik-bakke: Oh, yeah, sure. I meant it will be part of a larger project to refactor tons of things

src/foundation/function-extensionality.lagda.md
fredrik-bakke: I thought it was reasonable as this gives it the same shape as the converse. I don't feel strongly about this though

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -0,0 +1,283 @@
...
+The
+{{#concept "function extensionality axiom" Agda=function-extensionality Agda=funext}}
+asserts that [identifications](foundation-core.identity-types.md) of (dependent)
+functions are [equivalently](foundation-core.equivalences.md) described as
+[homotopies](foundation-core.homotopies.md) between them. In other words, a
+function is completely determined by its values.
+
+## Definitions
+
+### Equalities induce homotopies
```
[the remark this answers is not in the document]
fredrik-bakke: We actually use "equalities" more often in headers than "identifications". I believe the discussion you are remembering is about that we prefer "identifications" over "paths" or perhaps even "identities" (note that we do write "characterizing identity of bla bla" from time to time). I think the current formulation is just fine. We can consider forming a convention on this later if you feel strongly about this though.

src/foundation/function-extensionality.lagda.md
fredrik-bakke: I should say contrasting* shape to the converse

PR #977, opened 2023-12-08
small addendum to copartial elements, and proposal of erased -> denied

EgbertRijke:
This PR adds a small lemma of copartial elements, and proposes to rename erased to denied.

Calling `f` to be erased at `a` suggests that a value of `f` at `a` existed, but this is not accurate. Instead we propose to say that "evaluation of `f` at `a` is denied". The main example is to deny evaluation of the fraction `1/0` is denied, because division by 0 is not possible.


VojtechStep:
LGTM

PR #993, opened 2024-01-09
Make type argument explicit for `terminal-map`

fredrik-bakke:
Agda can't always infer the domain of `terminal-map`, and since we usually use `terminal-map` without evaluating it, I suggest we make the type argument explicit. This is in line with how we use `const`, `precomp`, and `postcomp`.

### DIFF
--- a/src/synthetic-homotopy-theory/tangent-spheres.lagda.md
+++ b/src/synthetic-homotopy-theory/tangent-spheres.lagda.md
@@ -71,10 +71,10 @@ module _
                 Σ ( C → X)
                   ( λ i →
                     Σ ( coherence-square-maps
-                          ( j)
-                          ( terminal-map (type-mere-sphere n T))
-                          ( i)
-                          ( point x))
+                        ( j)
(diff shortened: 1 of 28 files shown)


src/synthetic-homotopy-theory/tangent-spheres.lagda.md
```diff
@@ -70,9 +70,16 @@ module _
               ( λ j →
                 Σ ( C → X)
                   ( λ i →
-                    Σ ( coherence-square-maps j terminal-map i (point x))
+                    Σ ( coherence-square-maps
+                          ( j)
+                          ( terminal-map (type-mere-sphere n T))
+                          ( i)
+                          ( point x))
```
EgbertRijke, line 77:
```suggestion
                    Σ ( coherence-square-maps
                        ( j)
                        ( terminal-map (type-mere-sphere n T))
                        ( i)
                        ( point x))
```

EgbertRijke:
Thank you for doing this. I agree it's better to have the type explicit.

PR #979, opened 2023-12-10
Basic properties of orthogonal maps

fredrik-bakke:
- Coproducts of pullbacks are pullbacks
- Gives some equivalent characterizations of orthogonal maps
- Orthogonal maps are closed under homotopy
- Equivalences are left and right orthogonal to any map
- Right orthogonal maps are closed under
  - Composition
  - Left cancellation
  - Dependent products
  - Postcomp exponentiation
  - Products
  - Base change
- Left orthogonal maps are closed under
  - Composition
  - Right cancellation
  - Dependent sums
  - Coproducts
- Local (dependent) types are closed under
  - equivalent types
  - function homotopy
- A type is `f`-local if and only if its terminal map is
- A type is `f`-local if and only if its terminal map is `f`-orthogonal
- A dependent type is `f`-local if its fibers are `f`-null
- Code formatting and refactoring for pullbacks and miscellaneous.

PR #994, opened 2024-01-10
Use `point x` instead of `const unit X x`

fredrik-bakke:
Also adds the fact `is-injective-point`, and fixes misnamings in `foundation.constant-maps`.

### DIFF
--- a/src/foundation/booleans.lagda.md
+++ b/src/foundation/booleans.lagda.md
@@ -124,9 +124,11 @@ eq-Eq-bool :
 eq-Eq-bool {true} {true} star = refl
 eq-Eq-bool {false} {false} star = refl
 
-neq-false-true-bool :
-  false ≠ true
+neq-false-true-bool : false ≠ true
 neq-false-true-bool ()
+
+neq-true-false-bool : true ≠ false
+neq-true-false-bool ()
 ```
 
 ## Structure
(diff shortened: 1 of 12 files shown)


fredrik-bakke:
On inspection this PR is going to be annoying to merge with #979

fredrik-bakke:
Eh, it's probably fine.

src/foundation/booleans.lagda.md
```diff
@@ -255,26 +257,12 @@ pr1 equiv-neg-bool = neg-bool
 pr2 equiv-neg-bool = is-equiv-neg-bool
 ```
 
-### The constant function `const bool bool b` is not an equivalence
+### The constant functions `const bool bool b` are not equivalences
```
EgbertRijke, line 260: This change seems to make the sentence grammatically incorrect, because `const bool bool b` is one function.

src/foundation/constant-maps.lagda.md
```diff
@@ -55,115 +55,115 @@ module _
   compute-action-htpy-function-const c H = ap-const c (eq-htpy H)
 ```
 
-### A type is `k+1`-truncated if and only if all constant maps into it are `k`-truncated
+### A type is `k+1`-truncated if and only if all point inclusions are `k`-truncated
```
EgbertRijke, line 58:
```suggestion
### A type is `k+1`-truncated if and only if all point inclusions are `k`-truncated maps
```

The word "map" dropped in the revision, but I think it is clearer if it is mentioned

EgbertRijke (approved):
Good pull request! Thank you for making these improvements. I had only two minor remarks.

fredrik-bakke:
Thanks for the comments. I've resolved them now :)

EgbertRijke:
Excellent. Enabling auto-merge

PR #995, opened 2024-01-11
Make type family implicit for `is-torsorial-Eq-structure` and `is-torsorial-Eq-Π`

fredrik-bakke:


fredrik-bakke:
This one is bound to cause a bit of a merging mess, so I'm fine with leaving it open until #885 is merged.

EgbertRijke:
I’m surprised that this is possible. Hats off for finding such a big improvement!

PR #998, opened 2024-01-11
Make type arguments implicit for `eq-equiv`

fredrik-bakke:


EgbertRijke:
Looks good to me! I'll turn on auto-merge

EgbertRijke:
I'll wait with merging until tomorrow, because it seems you're still making new changes to this pull request.

fredrik-bakke:
Sorry, I'm done now.

src/foundation/univalence.lagda.md
```diff
@@ -157,13 +157,13 @@ eq-equiv-fam {B = B} {C} = map-inv-is-equiv (is-equiv-equiv-eq-fam B C)
...
+  equiv-eq q ∘e equiv-eq p ＝ equiv-eq (p ∙ q)
 compute-equiv-eq-concat refl refl = eq-equiv-eq-map-equiv refl
 
 compute-eq-equiv-comp-equiv :
-  {l : Level} (A B C : UU l) (f : A ≃ B) (g : B ≃ C) →
-  ((eq-equiv f) ∙ (eq-equiv g)) ＝ eq-equiv (g ∘e f)
-compute-eq-equiv-comp-equiv A B C f g =
+  {l : Level} {A B C : UU l} (f : A ≃ B) (g : B ≃ C) →
+  eq-equiv f ∙ eq-equiv g ＝ eq-equiv (g ∘e f)
+compute-eq-equiv-comp-equiv f g =
```
EgbertRijke, line 166: With identifications like this I don't think we can be so sure that we don't ever need to know the way they are associated, because I can imagine some complicated synthetic homotopy theory where it would become relevant. I thought we had the convention to not omit parentheses in that case, didn't we?

fredrik-bakke: Ah, yep, you're right. I was a bit eager there.

fredrik-bakke: Fixed now

EgbertRijke:
Thanks for the quick fix. Auto-merge is back on

PR #999, opened 2024-01-16
Extensions of types

EgbertRijke:
This pull request proposes a definition for extension types, following the discussion in #885. I think this is a useful definition to have, and it could help easing off the load of definitions in `surjective-maps` by being able to separate it into their own concepts, such as `surjective-extensions-types` and friends.

This PR is intentionally kept small and is ready for review.

### DIFF
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -17,7 +17,7 @@ open import foundation.universe-levels
 
 Consider a type `X`. An
 {{#concept "extension" Disambiguation="type" Agda=extension-type}} of `X` is an
-object of the coslice category of `X`, i.e., it consists of a type `Y` and a map
+object in the [coslice](foundation.coslice.md) under `X`, i.e., it consists of a type `Y` and a map
 `f : X → Y`.
 
 In the above definition of extensions of types our aim is to capture the most
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -15,7 +15,7 @@ open import foundation.universe-levels
 
 ## Idea
 
-Consider a type `X`. An {{#concept "extension" Disambiguation="type"}} of `X` is
+Consider a type `X`. An {{#concept "extension" Disambiguation="type" Agda=extension-type}} of `X` is
 an object of the coslice category of `X`, i.e., it consists of a type `Y` and a
 map `f : X → Y`.
 
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -1,7 +1,7 @@
 # Extensions of types
 
 ```agda
-module foundation.extensions-of-types where
+module foundation.extensions-types where
 ```
 
 <details><summary>Imports</summary>
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -17,8 +17,8 @@ open import foundation.universe-levels
 
 Consider a type `X`. An
 {{#concept "extension" Disambiguation="type" Agda=extension-type}} of `X` is an
-object in the [coslice](foundation.coslice.md) under `X`, i.e., it consists of a type `Y` and a map
-`f : X → Y`.
+object in the [coslice](foundation.coslice.md) under `X`, i.e., it consists of a
+type `Y` and a map `f : X → Y`.
 
 In the above definition of extensions of types our aim is to capture the most
 general concept of what it means to be an extension of a type. Similarly, in any
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -20,6 +20,14 @@ Consider a type `X`. An
 object of the coslice category of `X`, i.e., it consists of a type `Y` and a map
 `f : X → Y`.
 
+In the above definition of extensions of types our aim is to capture the most
+general concept of what it means to be an extension of a type. Similarly, in any
+[category](category-theory.categories.md) we would say that an extension of an
+object `X` consists of an object `Y` equipped with a morphism `f : X → Y`.
+
+Our notion of extensions of types are not to be confused with extension types of
+cubical or simplicial homotopy type theory.
+
 ## Definitions
 
 ### Extensions of types
--- a/src/foundation/extensions-types.lagda.md
+++ b/src/foundation/extensions-types.lagda.md
@@ -26,7 +26,8 @@ general concept of what it means to be an extension of a type. Similarly, in any
 object `X` consists of an object `Y` equipped with a morphism `f : X → Y`.
 
 Our notion of extensions of types are not to be confused with extension types of
-cubical or simplicial homotopy type theory.
+cubical type theory or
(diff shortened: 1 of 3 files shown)


VojtechStep:
Could we find a slightly different name for `extension-type`? I like the title "Extensions of types", but I'm a little worried that `extension-type` might get confused for the concept of extensions types from cubical HoTT

EgbertRijke:
That is a very good point. Thank you for bringing it up.

EgbertRijke:
Another thing to think about is whether the dual notion would also make sense, i.e., would it make sense to call the objects of the slice category over an object "lifts of objects". I think I could see it making sense. 

A lift of types would be a type `Y` equipped with a map into the type that is being lifted. Then a lift of maps would go along a lift of types and that seems sensible conceptually.

src/foundation/extensions-types.lagda.md
```diff
@@ -0,0 +1,39 @@
...
+```agda
+open import foundation.dependent-pair-types
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Consider a type `X`. An {{#concept "extension" Disambiguation="type"}} of `X` is
```
fredrik-bakke, line 18:
```suggestion
Consider a type `X`. An {{#concept "extension" Disambiguation="type" Agda=extension-type}} of `X` is
```

src/foundation/extensions-types.lagda.md
```diff
@@ -0,0 +1,39 @@
...
+an object of the coslice category of `X`, i.e., it consists of a type `Y` and a
+map `f : X → Y`.
+
+## Definitions
+
+### Extensions of types
+
+```agda
+extension-type : {l1 : Level} (l2 : Level) (X : UU l1) → UU (l1 ⊔ lsuc l2)
+extension-type l2 X = Σ (UU l2) (λ Y → X → Y)
```
fredrik-bakke, line 28: This seems like the most sensible definition you can give of an extension of a type to me. However, we don't directly recover extensions of maps when we interpret this definition in the arrow category.

fredrik-bakke: It seems to me like we agree about what an "extension" should be in that they keep reference to the object they extend. This is different to what you call an extension of a span diagram.

You write

````md
## Idea

Consider a [span diagram](foundation.span-diagrams.md) `s` given by

```text
       f       g
  A <----- S -----> B
```

and maps `i : A → A'` and `j : B → B'`. The
{{#concept "extension" Disambiguation="span diagram"}} of `s` by `i` and `j` is
the span diagram

```text
       i ∘ f     j ∘ g
  A' <------- S -------> B'.
```
````

In this instance, I would instead say that _`(i , j)` is an extension of `(i ∘ f , j ∘ g)` along `(f , g)` by `(refl-htpy , refl-htpy)`_.

**Comment by VojtechStep** [comment]
> I thought extension types were original to simplicial HoTT

That's very much possible, I'm not all that well versed in this stuff, and I've only encountered them in cubical settings.

File: src/foundation/extensions-types.lagda.md (hunk 2 above)
Comment (fredrik-bakke) [reply src/foundation/extensions-types.lagda.md:28]: Perhaps they don't need to be. We can still call them an extension.

Comment (EgbertRijke) [reply src/foundation/extensions-types.lagda.md:28]: Indeed we are in the process of establishing what it means to be an extension of an object. What I wrote in that file (and the file name) will be updated once we have established what an extension of objects is. I apologize for the temporary inconsistency.

Comment (fredrik-bakke) [reply src/foundation/extensions-types.lagda.md:28]: Ah, okay, I thought we were still debating whether extensions of span diagrams should be what you defined them as. 😅

Comment (EgbertRijke) [reply src/foundation/extensions-types.lagda.md:28]: > This seems like the most sensible definition you can give of an extension of a type to me. However, I want to remark we don't directly recover extensions of maps when we interpret this definition in the arrow category.

When you interpret the proposed definition in the slice category, then you get the bundled version of extensions of maps. Since extensions of maps are unbundled, they will still be conceptually distinct from extensions of objects in the slice category, but the difference is merely bundling them up. 

Are you happy with this explanation, or will you really insist that the arrow category has to be used?

Comment (fredrik-bakke) [reply src/foundation/extensions-types.lagda.md:28]: Then, similarly, an extension of a span diagram as I phrased it will be a morphism in the double coslice category `Coslice × Coslice`! This makes great sense! 😄

**Review by fredrik-bakke — APPROVED** [review]

File: src/foundation/extensions-types.lagda.md (hunk 2 above)
Comment (EgbertRijke) [reply src/foundation/extensions-types.lagda.md:28]: > Ah, your motivation is that an extension of a map `f` along a map `g` is a morphism `g -> f` in the coslice.
> 
> > Are you happy with this explanation, or will you really insist that the arrow category has to be used?
> 
> I don't appreciate how you are talking to me. I was merely trying to understand your motivation.

I'm sorry that my question came across negatively? I've actually been filled with anxiety in our recent conversations, and especially that time you suggested that I was taking your PR hostage was very triggering to me. So I'm all the time worried "Oh what will Fredrik think of this?" and generally I don't have a sense that you appreciate my comments on your pull requests. So I was really trying to ask you if that explanation was making you happy because I was afraid that it didn't and that you really wanted to have the arrow category in my explanation. It wasn't meant passive-aggressively, even though I realise now that it could easily come across that way.

**Comment by EgbertRijke** [comment]
I think I'll make a comment in the `## Idea` section explaining that we aim to define the _general concept_ of an extension. The extension types in cubical/simplicial type theory are merely called that way because they're _examples_ of extensions and in their context there was less of a need to be more disambiguative.

**Comment by EgbertRijke** [comment]
I have added some clarifying remarks to the `## Idea` section. Does anyone know what to link to for extension types in cubical Agda or simplicial hott?

**Comment by fredrik-bakke** [comment]
> I have added some clarifying remarks to the `## Idea` section. Does anyone know what to link to for extension types in cubical Agda or simplicial hott?

The reference for extension types in simplicial type theory is https://arxiv.org/abs/1705.07442
and the relevant formalization can be found here https://rzk-lang.github.io/sHoTT/simplicial-hott/03-extension-types.rzk/

**Comment by EgbertRijke** [comment]
I added only the link to the arxiv article because our page about extension types not really a page about the concept of extension types of simplicial type theory and therefore I don't feel compelled to give all possible links. My purpose is just to make people aware that there are different things that people call extension types.

**Comment by fredrik-bakke** [comment]
Sounds good. I provided both as I wasn't sure what you were asking for.

**Comment by EgbertRijke** [comment]
Yep, thank you for providing them!

**Comment by VojtechStep** [comment]
LGTM, enabling automerge

File: src/foundation/extensions-types.lagda.md [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Consider a type `X`. An
+{{#concept "extension" Disambiguation="type" Agda=extension-type}} of `X` is an
+object of the coslice category of `X`, i.e., it consists of a type `Y` and a map
```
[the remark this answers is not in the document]
fredrik-bakke: Sorry for blocking the PR, but don't you agree a here is better?

PR #1001, opened 2024-01-16
Lifts of types

EgbertRijke:
In this pull request, dual to #999 I propose a definition of lifts of types.

### DIFF
--- a/src/foundation/lifts-types.lagda.md
+++ b/src/foundation/lifts-types.lagda.md
@@ -1,0 +1,42 @@
+# Lifts of types
+
+```agda
+module foundation.lifts-types where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.dependent-pair-types
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Consider a type `X`. A {{#concept "lift" Disambiguation="type" Agda=lift-type}}
+of `X` is an object of the slice category of `X`, i.e., it consists of a type
+`Y` and a map `f : Y → X`.
+
+In the above definition of lifts of types our aim is to capture the most general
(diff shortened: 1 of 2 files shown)


EgbertRijke:
Note that there is already some work in `slice` about this definition, and we would face the question whether it is desirable to rephrase type duality in terms of lifts of types.

src/foundation/lifts-types.lagda.md
```diff
@@ -0,0 +1,42 @@
...
+open import foundation.dependent-pair-types
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Consider a type `X`. A {{#concept "lift" Disambiguation="type" Agda=lift-type}}
+of `X` is an object of the slice category of `X`, i.e., it consists of a type
```
fredrik-bakke, line 19:
```suggestion
of `X` is an object in the [slice](foundation.slice.md) over `X`, i.e., it consists of a type
```

fredrik-bakke (approved):
Sorry for the delayed review. Looks good to me!

src/foundation/lifts-types.lagda.md
fredrik-bakke: (or some other formulation in which a link to the slice file makes sense)

fredrik-bakke:
I'm still a bit confused about how these relate to lifts and extensions of maps, but I don't think that should block these PRs.

fredrik-bakke:
Before I merge, what do you think about coauthorship on these prs?

EgbertRijke:
I’m happy to share. This PR came out of a discussion we had

fredrik-bakke:
Okay, great 😄

Issue #641, opened 2023-06-04
Port from using ASCII diagrams to using `<svg>`s


fredrik-bakke:
This is a follow-up to a monologue I had on Discord.

Following Emily Riehl, Jonathan Weinberger, and Nikolai Kudasov's example over at <https://github.com/emilyriehl/yoneda>, I think we should port to using `<svg>`s rather than ASCII diagrams, as these look wildly better and are far more versatile.

The original example I brought up can be seen here: <https://emilyriehl.github.io/yoneda/simplicial-hott/05-segal-types.rzk/> (source: <https://raw.githubusercontent.com/emilyriehl/yoneda/master/src/simplicial-hott/05-segal-types.rzk.md>)
Screenshot:
![image](https://github.com/UniMath/agda-unimath/assets/5176294/ac70be28-ea14-410e-82b1-4df853391e73)

And works flawlessly on our website as well when pasted over.
Example screenshot:
![image](https://github.com/UniMath/agda-unimath/assets/5176294/fc8af30b-4da8-4c14-acd3-bd01b8568156)

VojtechStep:
> (Observe that the arrows are missing their heads. It seems they may have added some extra CSS code to get the arrowheads.)

The arrow heads are defined at the bottom of the file - a `marker-end` attribute references a marker, which is defined in the svg `defs` tag as e.g. `<marker id="arrow" ...><path ... /></marker>`

VojtechStep:
Another problem with the current ASCII/Unicode diagrams is that if we ever try to alter the markup (so that we can e.g. use subscript capital letters, which don't exist in Unicode), then the source code and output wouldn't match in terms of the number of characters, so the source diagrams wouldn't be aligned anymore (and we would always need to look at the output to see if it renders correctly, which removes the strongest argument for ASCII/Unicode diagrams)

fredrik-bakke:
Here's a repository that may be useful for when we want to implement latex support:
- https://github.com/gboyd068/obsidian-swiftlatex-render

PR #1002, opened 2024-01-16
Render ASCII diagrams as SVGs using svgbob

fredrik-bakke:
This is a test PR to report my findings using `mdbook-svgbob2`. It is a simple wrapper for ASCII diagrams to render them as SVGs on the website. I have had some success so far, but I wouldn't say it is satisfactory yet. I will try it out a bit more and report some findings here.

This PR works towards resolving #641.

### Success
The code
````text
```svgbob
          hA
    A ---------> A'
    |\           |\
    | \ h   ⇗    | \ h'
    |  \      f' |  \
    |   V        |   V
  f | ⇐ B ------ | -> B'
    |   /   hB   | ⇐ /
    |  / g       |  / g'
    | /     ⇗    | /
    VV           VV
    C ---------> C' ,
          hC
```
````

is rendered as
image

### Failure
The code
````text
```svgbob
          B
      h  ^| \  g
       /  |   \
     /  f | ⇑   V
    A ---------> C
    |     | hB   |
    | ⇗   V   ⇗  |
 hA |     B'     | hC
    | h' ^  \ g' |
    |  /  ⇑   \  |
    V/          VV
    A' --------> C' .
          f'
```
````
renders as
image

## Why use Svgbob
This option has a couple of benefits over possible alternatives:
1. It fails elegantly, meaning that ill-formatted ASCII diagrams will just display as ill-formatted SVGs, rather than causing errors.
2. The code is readable and intuitive.
3. It does not stray too far from our current approach, hopefully making the switch easy.

## Some references
- [Svgbob editor](https://ivanceras.github.io/svgbob-editor/)
- [Svgbob Specification](https://ivanceras.github.io/content/Svgbob/Specification.html)
- [`mdbook-svgbob2` on GitHub](https://github.com/matthiasbeyer/mdbook-svgbob2)

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -84,6 +84,12 @@ jobs:
           crate: mdbook-catppuccin
           version: '1.2.0'
 
+      - name: Install mdbook-svgbob2
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: mdbook-svgbob2
+          version: '0.3.0'
+
       - uses: actions/setup-python@v4
         with:
           python-version: '3.8'
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -11,6 +11,7 @@
     "ms-python.vscode-pylance",
     "oderwat.indent-rainbow",
     "brunnerh.insert-unicode",
-    "editorconfig.editorconfig"
+    "editorconfig.editorconfig",
+    "tamasfe.even-better-toml"
   ]
 }
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,8 @@ install-website-dev:
 	@cargo install mdbook-katex@0.5.7
 	@cargo install mdbook-pagetoc@0.1.7
 	@cargo install mdbook-catppuccin@1.2.0
+	@cargo install mdbook-svgbob2@0.3.0
+
 
 .PHONY: unused-imports
 unused-imports:
--- a/book.toml
+++ b/book.toml
@@ -24,6 +24,23 @@ macros = "website/latex-macros.txt"
 [preprocessor.catppuccin]
 assets_version = "1.2.0" # DO NOT EDIT: Managed by `mdbook-catppuccin install`
 
+[preprocessor.svgbob2] # NOTE: Currently the default settings
+font_size = "14"
+font_family = "monospace"
+fill_color = "black"
+background = "transparent"
+stroke_color = "var(--fg)" # variable from the mdbook css files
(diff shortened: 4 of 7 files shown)


EgbertRijke:
I think the ASCII diagram wasn't ill formatted. It was quite glorious actually. But the svg does look ill formatted and less glorious:P

EgbertRijke:
On a more serious note, would there be a way to typeset those diagrams so that they would renderer properly with svgbob?

fredrik-bakke:
Yes, this may be a deal-breaker for svgbob. The spacing will be important for diagonal lines, and there's only one possible diagonal slope, unless you want dotted lines instead for some reason.

fredrik-bakke:
> On a more serious note, would there be a way to typeset those diagrams so that they would renderer properly with svgbob?

That is one of the next questions I want to answer. In the meantime, here are all the possible alternatives for sloped lines:

image

EgbertRijke:
What a strange design choice to only allow those angles and not even give them all proper arrow heads:P

EgbertRijke:
It could be that the best way would be just to integrate some form of tikz-cd into the website. I have no idea how to do this though. But if svg diagrams are so limited that makes me wonder whether we should convert to them at all.

fredrik-bakke:
It is not SVGs that are limited, but svgbob. I suspect that the limitations come from the fact that it is taking an ASCII diagram as an input.

VojtechStep:
I think in the long run we'd like to be able to typeset our diagrams using quiver, but it doesn't support SVG export at the moment https://github.com/varkor/quiver/issues/8

VojtechStep:
I know 1lab's diagrams are created in quiver, but IIRC the resource pipeline isn't very straight-forward

EgbertRijke:
I agree that if the result doesn't look good and we can't quite get the diagrams we would like to show, then it's better to stick to ascii diagrams until we find a real solution. Thank you for investigating though!

BTW why is it important that it has to be svg? What other outputs does quiver give?

VojtechStep:
I think currently it just gives you the LaTeX code. They have another issue for other image format exports, but it looks like they decided to first implement SVGs, and build other exports on top of that, which is reasonable

fredrik-bakke:
Hmm, maybe I could make an mdbook preprocessor following 1labs recipe...

fredrik-bakke:
But I don't want to do that if the proposal will be blocked on the grounds that we have to build the website to verify the diagrams are correct, as is our stance on using LaTeX in general.

PR #1004, opened 2024-01-17
Reformatting commented blocks of code

EgbertRijke:
This is a trivial pull request that merely reformats commented code to use `text` blocks. The purpose is that the commented code then displays better on the website, i.e., as a block of unhighlighted code that is uninterrupted with `--` on every line.


fredrik-bakke (approved):
Looks good, thanks for cleaning this up!

PR #1010, opened 2024-01-25
Fix "The predicate of" section headers

EgbertRijke:
This PR gives section headers that start as "The predicate ..." a uniform appearance:

- Sentences of the form "The predicate of ... on ..." are changed to "The predicate on ... of ...".
- Predicates are not on single elements of types, so in cases where this was suggested I reformulated it to plural.


fredrik-bakke (approved):
Sorry for my bad habit! 😣

EgbertRijke:
That's ok! I have other bad habits:)

PR #1011, opened 2024-01-28
Inline `pair`

fredrik-bakke:
Sets the pragma
```agda
{-# INLINE pair #-}
```
https://agda.readthedocs.io/en/latest/language/pragmas.html#inlining-constructor-right-hand-sides
Presumably, this gives us the performance benefit of copattern matched definitions, while also giving us the flexibility to use `_,_` and `pair` to make definitions.

### DIFF
--- a/src/foundation/dependent-pair-types.lagda.md
+++ b/src/foundation/dependent-pair-types.lagda.md
@@ -31,6 +31,7 @@ record Σ {l1 l2 : Level} (A : UU l1) (B : A → UU l2) : UU (l1 ⊔ l2) where
 open Σ public
 
 {-# BUILTIN SIGMA Σ #-}
+{-# INLINE pair #-}
 
 infixr 3 _,_
 pattern _,_ a b = pair a b


EgbertRijke:
Note that we already had the flexibility of using `_,_` and `pair` to make definitions. We just had conventions that we usually didn't do that:)

Anyway, that's just an unimportant detail. Thank you for making this PR, and I'll turn on auto-merge:)

fredrik-bakke:
We should also discuss if we want to remove that convention.

fredrik-bakke:
I, for one, think that copattern-matched definitions tend to be overly verbose, making the content harder to identify.

fredrik-bakke:
I should say, I wouldn't suggest instating the opposite convention at this moment

EgbertRijke:
I believe that our copattern matching conventions are quite solid and are serving us well, so I'm not very inclined to change anything about them. Note that we also state in our conventions that if our convention impedes readability, then the priority should be with writing readable code. This is the case with the 5-tuples defined in `binary-type-duality`, where the code gets a lot more readable if we directly define them as 5-tuples.

Another mechanism, which I have also suggested before but it met with some resistance, by which we can make our copattern matching definitions a lot more readable is by defining their components separately, before the entry that we wish to define by copattern matching. When we do this, we would already be using our naming scheme to name the components of the copattern matching definition. When we use these names in the components of a copattern matching definition, they would be descriptive of the component. In other words, it would make it super easy to read what that component is about. An example where I have used this way of writing a readable copattern matching definition is when I defined composition of morphisms of arrows:

https://unimath.github.io/agda-unimath/foundation.morphisms-arrows.html#composition-of-morphisms-of-arrows

It's only a little extra effort, but it makes the copattern matching definition a lot more readable.

So I don't suggest that we should change our conventions. When copattern matching definitions are two or three components long, then it should definitely be the standard to define them by copattern matching. If they are longer, then there is the option to refactor it or to give it as an explicit pair.

PR #1013, opened 2024-01-28
Transport-split and preunivalent type families

fredrik-bakke:
This is the follow-up PR to #1009.

### Summary
- Transport-split type families are type families `B` such that `equiv-tr B` admits fiberwise sections
  - This notion is equivalent to univalent type families by a corollary of the fundamental theorem of identity types
- Some infrastructure and basic lemmas for univalent type families
- Preunivalent type families are type families `B` such that `equiv-tr B` is a fiberwise embedding
  - 3-for-2 for preunivalent type families using  the preunivalence axiom
  - Some other infrastructure
- `Fin` is preunivalent

### Future work
- Closure properties for univalent and preunivalent type families

Co-authored-by: Egbert Rijke <e.m.rijke@gmail.com>


src/foundation/transport-split-type-families.lagda.md
```diff
@@ -0,0 +1,97 @@
...
+  {P : {x y : A} → B x ≃ B y → UU l3}
+  where
+
+  ind-equiv-is-transport-split' :
+    (f : {x y : A} (p : x ＝ y) → P (equiv-tr B p))
+    {x y : A} (e : B x ≃ B y) → P e
+  ind-equiv-is-transport-split' f {x} {y} e =
+    tr P
+      ( is-section-map-section (equiv-tr B) (s x y) e)
+      ( f (map-section (equiv-tr B) (s x y) e))
```
fredrik-bakke, line 83: I've been struggling a bit to identify precisely what extra data is needed for this map to satisfy the expected computation rule. Surely I'm missing something obvious. For instance, I would expect univalent type families to satisfy it. Anyway, I'll have to look at it again later.

EgbertRijke: What exactly do you mean by "equivalence induction" here? Do you mean that

```text
λ y → B x ≃ B y
```

is an identity system for every `x`?

EgbertRijke: Ah, it seems that this should give you directly that transport-split type families are univalent:

https://unimath.github.io/agda-unimath/foundation.fundamental-theorem-of-identity-types.html#the-fundamental-theorem-of-identity-types-using-sections

fredrik-bakke:
> The terminology "equivalence injective type family" is still a bit subobtimal, because it is so close to suggesting the definition of injective type family that I'm not really fond of. So I've been thinking about whether there could be more interesting concepts lurking under here.
>   
>   The family of maps in equivalence injective type families are converse to the transport maps `tr B : (x = y) -> (B x <~> B y)`. The main example of equivalence injective type families is the family `Fin`, but for `Fin` something more holds: this map has a section. So how about abolishing any suggestion of "injective type family" altogether, and call these type families `transport-split`? 
>   
>   In other words, a type family `B` would be called `transport-split` if the map `tr B : (x = y) -> (B x <~> B y)` has a section for each `x y : A`. In particular, if a type family is transport split, we have a family of maps `(B x <~> B y) -> (x = y)` indexed by `x y : A`, and moreover, this family of maps sends `id-equiv (B x)` to `refl`.

_Originally posted by @EgbertRijke in https://github.com/UniMath/agda-unimath/pull/1009#discussion_r1468441070_

What you're saying here seems to suggest that `Fin` is a univalent type family, i.e. that `(n ＝ m) ≃ (Fin n ≃ Fin m)`, since every transport-split type family is univalent. However, this cannot be true as `(n ＝ m)` is a proposition while `(Fin n ≃ Fin m)` is not for `n = m > 1`. Instead, `is-equivalence-injective-Fin` is only a retraction of `equiv-tr`, and not a section.

EgbertRijke:
> > The terminology "equivalence injective type family" is still a bit subobtimal, because it is so close to suggesting the definition of injective type family that I'm not really fond of. So I've been thinking about whether there could be more interesting concepts lurking under here.
> > The family of maps in equivalence injective type families are converse to the transport maps `tr B : (x = y) -> (B x <~> B y)`. The main example of equivalence injective type families is the family `Fin`, but for `Fin` something more holds: this map has a section. So how about abolishing any suggestion of "injective type family" altogether, and call these type families `transport-split`?
> > In other words, a type family `B` would be called `transport-split` if the map `tr B : (x = y) -> (B x <~> B y)` has a section for each `x y : A`. In particular, if a type family is transport split, we have a family of maps `(B x <~> B y) -> (x = y)` indexed by `x y : A`, and moreover, this family of maps sends `id-equiv (B x)` to `refl`.
> 
> _Originally posted by @EgbertRijke in [#1009 (comment)](https://github.com/UniMath/agda-unimath/pull/1009#discussion_r1468441070)_
> 
> What you're saying here seems to suggest that `Fin` is a univalent type family, i.e. that `(n ＝ m) ≃ (Fin n ≃ Fin m)`, since every transport-split type family is univalent. However, this cannot be true as `(n ＝ m)` is a proposition while `(Fin n ≃ Fin m)` is not for `n = m > 1`. Instead, `is-equivalence-injective-Fin` is only a retraction of `equiv-tr`, and not a section.

Oh! `fundamental-theorem-id-section` worked? I deleted that comment because I made it late at night and thought I was making a mistake. That's nice!

I certainly did not intend to suggest that `Fin` is a univalent type family.

fredrik-bakke:
> Oh! `fundamental-theorem-id-section` worked? I deleted that comment because I made it late at night and thought I was making a mistake. That's nice!

Yes :) Luckily you couldn't hide the comment from me

fredrik-bakke:
Sorry for the accusatory tone of my message. I just wanted to clarify as I was confused for a bit. This begs the question if we should care about type families such that `equiv-tr B` admits retractions however, which is equivalent to your observation that `is-equivalence-injective-Fin refl = id-equiv` I believe

EgbertRijke:
> Sorry for the accusatory tone of my message. I just wanted to clarify as I was confused for a bit. This begs the question if we should care about type families such that `equiv-tr B` admits retractions however, which is equivalent to your observation that `is-equivalence-injective-Fin refl = id-equiv` I believe

I see, yes I must have misspoken when I said section/retraction and perhaps I meant the other. I'm glad you were able to sort it out, and the proof that transport-split families are univalent is certainly worthy of being recorded in the library.

fredrik-bakke:
Here's something I think may be possible to prove using preunivalence: `type-Set` is a preunivalent type family.

fredrik-bakke:
> Here's something I think may be possible to prove ~using preunivalence~ without using univalence: `type-Set` is a preunivalent type family.

Okay, that was too easy. I bet it can be done without univalence or preunivalence. Anyway, I'll leave that for another time.

fredrik-bakke:
Alright, I'm opening this PR up for review now. Btw, I believe you should have co-authorship over this PR, Egbert, given the ideas you contributed to it. Thank you for that!

src/foundation-core/sets.lagda.md
```diff
@@ -147,4 +147,21 @@ abstract
...
+module _
+  {l1 l2 : Level} (A : Set l1) (B : Set l2)
+  where
+
+  type-equiv-Set : UU (l1 ⊔ l2)
+  type-equiv-Set = type-Set A ≃ type-Set B
+
+  equiv-Set : Set (l1 ⊔ l2)
+  pr1 equiv-Set = type-equiv-Set
+  pr2 equiv-Set = is-set-equiv-is-set (is-set-type-Set A) (is-set-type-Set B)
```
EgbertRijke, line 166: Given that we moved to `hom-Set A B : UU _` and `hom-set-Set A B : Set _`, perhaps it would make sense to also rename

- `type-equiv-Set` -> `equiv-Set`
-  `equiv-Set` -> `equiv-set-Set`. The reading of this is that it takes as input sets (hence `Set`), it returns a set (hence `set`, and it is the set of equivalences.

src/foundation/preunivalent-type-families.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+
+```agda
+is-preunivalent :
+  {l1 l2 : Level} {A : UU l1} → (A → UU l2) → UU (l1 ⊔ l2)
+is-preunivalent {A = A} B = (x y : A) → is-emb (λ (p : x ＝ y) → equiv-tr B p)
+```
+
+## Properties
+
+### Type families are preunivalent if and only if they are faithful as maps
```
EgbertRijke, line 54: Perhaps it is better to say: Assuming the preunivalence axiom, type families are...

src/foundation/preunivalent-type-families.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+    is-preunivalent-is-0-map H =
+      is-preunivalent-is-faithful (is-faithful-is-0-map H)
+```
+
+### Families of sets are preunivalent if `equiv-tr` is fiberwise injective
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (B : A → UU l2)
+  (is-set-B : (x : A) → is-set (B x))
```
EgbertRijke, line 117: Here I would certainly assume `B` is just a family of sets over `A`, rather than split that hypothesis in two.

src/foundation/preunivalent-type-families.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+  is-preunivalent-is-0-map (is-0-map-pr1 (is-set-type-Set ∘ S))
+
+is-preunivalent-inclusion-subuniverse :
+  {l1 l2 : Level} (S : subuniverse l1 l2) →
+  is-preunivalent (inclusion-subuniverse S)
+is-preunivalent-inclusion-subuniverse S =
+  is-preunivalent-projection-Type-With-Set-Element (set-Prop ∘ S)
+```
+
+## See also
```
EgbertRijke, line 154: Beautiful work!

src/univalent-combinatorics/standard-finite-types.lagda.md
```diff
@@ -427,22 +432,53 @@ leq-nat-succ-Fin (succ-ℕ k) (inr star) =
...
+  {n : ℕ} → is-equivalence-injective-Fin {n} {n} id-equiv ＝ refl
+compute-is-equivalence-injective-Fin-id-equiv {zero-ℕ} = refl
+compute-is-equivalence-injective-Fin-id-equiv {succ-ℕ n} =
+  ap² succ-ℕ
+    ( ( ap is-equivalence-injective-Fin compute-equiv-equiv-Maybe-id-equiv) ∙
+      ( compute-is-equivalence-injective-Fin-id-equiv {n}))
+```
+
+### `Fin` is a preunivalent type family
+
```
EgbertRijke, line 463: Could you perhaps remark on whether or not you used the preunivalence axiom in this proof? It looks to me like you didn't use it, and that is certainly noteworthy.

EgbertRijke (approved):
This is a PR of the highest quality! Thank you Fredrik!

src/foundation/preunivalent-type-families.lagda.md
fredrik-bakke: Great suggestion!

fredrik-bakke:
Thank you for the generous words! I believe I've resolved all of your comments now.

src/foundation/propositional-extensionality.lagda.md
```diff
@@ -118,6 +118,8 @@ pr2 (Prop-Set l) = is-set-type-Prop
 
 ### The canonical type family over `Prop` is univalent
 
+The proof does not rely on [the univalence axiom](foundation.univalence.md).
```
EgbertRijke, line 121: The way we formalized it, it actually does rely on the univalence axiom.

fredrik-bakke: Oh? That's surprising

fredrik-bakke: I'll look at it tomorrow

EgbertRijke: It's not so surprising. You'll need some sort of mechanism that allows you to conclude that equivalent propositions are equal. If you go down `is-torsorial-equiv`, which is used in this proof, you'll quickly run into univalence.

Propositional extensionality is generally understood to be a restricted form of univalence, and it definitely doesn't just follow from MLTT. It's typically either assumed as propositional extensionality (i.e., directly as a postulate) or proven to be a consequence of univalence.

fredrik-bakke: Ah, I thought it was a consequence of funext. Sorry for the mix-up!

EgbertRijke: I'll just remove this sentence via github suggestions and turn on the automerge

```suggestion
```

EgbertRijke:
Thank you for including me as a coauthor btw!

src/foundation/propositional-extensionality.lagda.md
```diff
@@ -118,6 +118,7 @@ pr2 (Prop-Set l) = is-set-type-Prop
 
 ### The canonical type family over `Prop` is univalent
```
EgbertRijke, line 120:
```suggestion
```

src/foundation/propositional-extensionality.lagda.md
fredrik-bakke: Great, thanks

PR #1014, opened 2024-01-29
Refactor files about identity types and homotopies

EgbertRijke:
The following files were significantly cleaned up:
- `foundation.path-algebra`

The following files were created:
- `foundation.action-on-higher-identifications-functions`
- `foundation.whiskering-higher-homotopies`
- `foundation.whiskering-identifications`

The following other actions were performed:
- Adding detailed informal explanations to many entries affected by this pull request.
- Replace `whisk` with `whisker` throughout the library.
- Making the order of arguments uniform for various operations on commuting squares and triangles of identifications.
- Fixing names so that operator names go before the entity that they act on:
  * `htpy-left-whisker` to `htpy-left-whisker`
  * `coherence-square-identifications-horizontal-inv` to `horizontal-inv-coherence-square-identifications`
  * `coherence-square-identifications-left-paste` to `left-concat-identification-coherence-square-identification`, and variants
- Replace `ap (ap f)` with `ap² f` throughout the library. I also renamed the various `nat-sq`-entries about `ap²`.
- Moving the entries about squares of identifications from `path-algebra` to `commuting-squares-of-identifications`.
- There were duplicate entries that did the same pasting lemmas of commuting squares of identifications, spread out over several files including `commuting-squares-of-identifications` and `path-algebra`. These separate formalization attempts have been unified.
- Removing "iterated inverse laws" from `path-algebra` since they were duplicate entries and they were not actually iterated.
- Rename `ap-concat-eq` to `map-coherence-triangle-identifications`, change the order of its arguments to match with `coherence-triangle-identifications`, and move it to `commuting-triangles-of-identifications`.
- Rename `coherence-square-identifications-ap` to `map-coherence-square-identifications`.
- Remove entries `inv-htpy-left-whisk-inv-htpy` and `inv-htpy-right-whisk-inv-htpy`.
- Rename `ap-left-whisk-htpy` to `left-whisker-htpy²`, and `ap-right-whisk-htpy` to `right-whisker-htpy²` and move them to `whiskering-higher-homotopies`.
- Correcting a statement that `refl-htpy` is a unit element on the homotopy side for whiskering. Instead, it is an absorbing element.


EgbertRijke:
I thought this would be a quicky but I'm encountering a lot of substandard programming while editing the entries about whiskering, so I better put my shoulders under this and do the refactoring well. I'll mark this PR as a draft for now.

EgbertRijke:
I have explained some of the work that I've been doing in this pull request in the PR description. Hopefully that clears some of the confusion. Apologies that I didn't do this earlier.

fredrik-bakke:
Thanks for doing this! I've been contemplating doing something similar recently, but I'm glad someone else did it. Would you mind also renaming `htpy-precomp` to `precomp-htpy` and `htpy-postcomp` to `postcomp-htpy`?

fredrik-bakke:
> - Replace `ap (ap f)` with `ap²` f throughout the library. I also renamed the various `nat-sq`-entries about `ap²`.

Keep in mind that, at times, `ap (ap f)` may be more naturally understood as a law about left whiskering of homotopies, so there may be a naming conflicts arising.

EgbertRijke:
I am opening up this pull request for review. There are likely a few more small things that I will come across before a review will come in, but I don't expect any more big changes.

EgbertRijke:
I added the documentation tag to this PR because I have written a lot of informal explanations for all the entries that I edited.

fredrik-bakke:
I'll review it a little later today 😊

EgbertRijke:
Note that I keep catching things to improve during self-review, but I will stop looking for improvements once the an actual review is in. Otherwise this can go on forever with this PR.

EgbertRijke:
So a thing got broken, because there was a problem:

The same naming scheme was used for whiskering homotopies with fucntions, and whiskering 2-homotopies with homotopies. Furthermore, there was not a single comment on the fact that there were two different concepts at play. So I needed to figure out how to establish a good naming scheme that will meaningfully distinguish between all the variations of whiskering that exist, including whiskering higher identifications and homotopies. Clearly, this was a nontrivial problem, so I can't blame anyone that this wasn't done before.

The thing I came up with is a general notion of whiskering: Here I defined whiskering _with respect to a binary operation_, and this establishes a useful naming scheme:
- `left-whisker-concat` and `right-whisker-concat` for whiskering with respect to `concat` (i.e. on identifications).
- `left-whisker-concat-htpy` and `right-whisker-concat-htpy` for wihiskering with respect to `concat-htpy`
- `left-whisker-comp` and `right-whisker-comp` for whiskering with respect to composition (i.e. on functions)
- `left-whisker-comp²` and `right-whisker-comp²` for whiskering 2-homotopies with respect to composition.

Clearly, our contributors including me have had difficulty in the past with coming up with names for many operations involved in whiskering and higher path algebra. Hopefully this will alleviate some of the difficulty.

Also, I had to deal with this for #885 because I was running into those things.

src/finite-group-theory/delooping-sign-homomorphism.lagda.md
```diff
@@ -1146,34 +1145,35 @@ module _
...
                           ( eq-pair-Σ p
-                            ( eq-is-prop is-prop-type-trunc-Prop)))) ∙
-                      ( ap
+                            ( eq-is-prop is-prop-type-trunc-Prop))))
+                      ( inv (inv (eq-counting-equivalence-class-R n))) ∙
+                      ( left-whisker-concat
                         ( ap
                           ( equivalence-class ∘ R (n +ℕ 2))
                           ( eq-pair-Σ p
-                            ( eq-is-prop is-prop-type-trunc-Prop)) ∙_)
```
EgbertRijke, line 1176: This notation should not have happened by the way. When you read the construction as a tree, there is no hint that the identification constructed with `ap` is going to be be an argument to a partially applied concatenation operation. I don't consider this notation kind to the reader.

fredrik-bakke:
Should I wait with reviewing this PR?

EgbertRijke:
I'm happy to have your feedback on this PR so far, but the problem is that it breaks something in `lifting-squares` that is not easy to fix because that reflexivity homotopy is so hard-coded. Given that it breaks things in the library I would also totally understand if you want to wait with looking at this PR. I'm working hard to fix it.

src/foundation-core/identity-types.lagda.md
```diff
@@ -14,11 +14,27 @@ open import foundation.universe-levels
...
-theory, we introduce the identity type as an inductive family of types, where
-the induction principle can be understood as expressing that the identity type
-is the least reflexive relation.
+The equality relation on a type is introduced as binary family of types
+
+```text
+  Id : A → A → 𝒰
+```
+
+equipped with a proof of reflexivity
```
fredrik-bakke, line 23:
```suggestion
equipped with a _proof of reflexivity_
```

fredrik-bakke: or concept?

src/foundation-core/identity-types.lagda.md
```diff
@@ -87,9 +121,17 @@ ind-Id :
 ind-Id x B b y refl = b
 ```
 
-## Structure
+## Operations on the identity type
+
+The identity types form a weak groupoidal structure on types. Thus they come
+equipped with **concatenation** `(x ＝ y) → (y ＝ z) → (x ＝ z)` and an
+**inverse operation** `(x ＝ y) → (y ＝ x)`.
```
fredrik-bakke, line 128: `#concept` ?

src/foundation-core/identity-types.lagda.md
```diff
@@ -234,6 +418,22 @@ step-equational-reasoning p z q = p ∙ q
 syntax step-equational-reasoning p z q = p ＝ z by q
 ```
 
+**Note.** Equational reasoning is a convenient way to construct identifications.
+However, in some situations it may not be the fastest or cleanest mechanism to
+construct an identification. Some constructions of identifications naturally
+involve computations that are more deeply nested in the terms. Furthermore,
+proofs by equational reasoning tend to require a lot of reassociation.
+
+Some tools that allow us to perform faster comutations are the transpositions
```
fredrik-bakke, line 427:
```suggestion
Some tools that allow us to perform faster computations are the transpositions
```

src/foundation-core/identity-types.lagda.md
```diff
@@ -234,6 +418,22 @@ step-equational-reasoning p z q = p ∙ q
...
+defined above, the transpositions and splicing operations defined in
+[commuting squares of identifications](foundation.commuting-squares-of-identifications.md)
+and
+[commuting triangles of identifications](foundation.commuting-triangles-of-identifications.md),
+and the higher concatenation operations defined in
+[path algebra](foundation.path-algebra.md). Each of these operations has good
+computational behavior, so there is infrastructure for reasoning about
+identifications that are constructed using them.
```
fredrik-bakke, line 435: Maybe it is appropriate to also mention that there is additional infrastructure for reasoning with homotopies that should be preferred when working with them? Generally, I tend to find that proofs about homotopies that reduce to equational reasoning are less conceptual and longer than the alternative.

src/foundation-core/whiskering-homotopies-composition.lagda.md
```diff
@@ -30,8 +30,8 @@ assumes a diagram of the form
       g
 ```
 
-and is defined to be a function `h ·l_ : (f ~ g) → (h ∘ f ~ h ∘ g)`. The **right
-whiskering** operation assumes a diagram of the form
+and is defined to be a function `H ↦ h ·l H : (f ~ g) → (h ∘ f ~ h ∘ g)`. The
+**right whiskering** operation assumes a diagram of the form
```
fredrik-bakke, line 34:
```suggestion
{{#concept "right whiskering" Disambiguation="homotopies with respect to composition" Agda=_·r_}} operation assumes a diagram of the form
```

src/foundation-core/identity-types.lagda.md
```diff
@@ -131,30 +205,144 @@ module _
...
+  is-retraction-inv-concat' refl refl = refl
+
+  is-section-inv-concat' :
+    {x y z : A} (q : y ＝ z) (r : x ＝ z) → (r ∙ inv q) ∙ q ＝ r
+  is-section-inv-concat' refl refl = refl
+```
+
 ### Transposing inverses
 
+Consider a triangle of identifications
```
fredrik-bakke, line 325:
```suggestion
Consider a [triangle of identifications](foundation-core.commuting-triangles-of-identifications.md)
```

src/foundation/action-on-higher-identifications-functions.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+
+Any map `f : A → B` has an
+{{#concept "action on higher identifications" Disambiguation="functions" Agda=ap²}},
+which is a map
+
+```text
+  ap² f : (p ＝ q) → (ap f p ＝ ap f q)
+```
+
+Here `p q : x ＝ y` are identifications in the type `A`. The action of `f` on
```
fredrik-bakke, line 33:
```suggestion
Here `p q : x ＝ y` are [identifications](foundation-core.identity-types.md) in the type `A`. The action of `f` on
```

src/foundation/action-on-higher-identifications-functions.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+                      ap² f (horizontal-inv-Id² α)
+        ap f (inv p) ------------------------------> ap f (inv q)
+             |                                            |
+  ap-inv f p |                                            | ap-inv f q
+             ∨                                            ∨
+        inv (ap f p) ------------------------------> inv (ap f q)
+                      horizontal-inv-Id² (ap² f α)
+```
+
+commutes.
```
fredrik-bakke, line 72:
```suggestion
[commutes](foundation-core.commuting-squares-of-identifications.md).
```

src/foundation/action-on-higher-identifications-functions.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+
+```text
+                 ap² (const b) α
+  ap (const b) p ---------------> ap (const b) q
+                 \              /
+      ap-const b p \          / ap-const b q
+                     ∨      ∨
+                       refl
+```
+
```
fredrik-bakke, line 170:
```suggestion

[commutes](foundation-core.commuting-triangles-of-identifications.md).

```

src/foundation-core/whiskering-homotopies-composition.lagda.md
```diff
@@ -270,18 +257,17 @@ definitionally.
...
 module _
-  {l1 l2 l3 l4 : Level}
-  {A : UU l1} {B : A → UU l2}
+  {l1 l2 l3 l4 : Level} {A : UU l1} {B : A → UU l2}
   {C : {x : A} → B x → UU l3} {D : {x : A} → B x → UU l4}
   {f g : {x : A} (y : B x) → C y}
   where
 
-  coherence-left-right-whisk-htpy :
+  coherence-double-whisker-comp :
```
fredrik-bakke, line 265: I think this name should contain both `left` and `right` somehow, to make it clear the coherence is not about a double one-sided whiskering.

src/foundation/commuting-triangles-of-homotopies.lagda.md
```diff
@@ -8,32 +8,36 @@ module foundation.commuting-triangles-of-homotopies where
...
+    f ----> g
+     \     /
+ left \   / right
+       ∨ ∨
+        h
 ```
 
-is said to commute if there is a homotopy between the homotopy on the left
-`f ~ h` and the composite homotopy `f ~ g ~ h`.
+is said to commute if there is a homotopy `left ~ top ∙h right`.
```
fredrik-bakke, line 36:
```suggestion
is said to {{#concept "commute" Disambiguation="triangle of homotopies" Agda=coherence-triangle-homotopies}} if there is a homotopy `left ~ top ∙h right`.
```

src/foundation/function-extensionality.lagda.md
```diff
@@ -69,7 +69,7 @@ coherence-square-eq-htpy-ap-precomp-Π f {C = C} g h =
     ( equiv-funext)
     ( equiv-funext)
     ( precomp-Π f (eq-value g h))
-    ( coherence-square-htpy-eq-ap-precomp-Π f g h)
+    ( coherence-square-homotopies-eq-ap-precomp-Π f g h)
```
fredrik-bakke, line 72: I think this thing should be called `coherence-square-homotopies-htpy-eq-ap-precomp-Π` as the `htpy`-part served two purposes. Sorry for the bad naming!

src/foundation/morphisms-arrows.lagda.md
```diff
@@ -361,10 +361,14 @@ module _
...
+    ( horizontal-pasting-coherence-square-identifications
+      ( htpy-codomain-htpy-hom-arrow f g α β H (f a))
+      ( htpy-codomain-htpy-hom-arrow f g β γ K (f a))
       ( coh-hom-arrow f g α a)
       ( coh-hom-arrow f g β a)
       ( coh-hom-arrow f g γ a)
+      ( (g ·l htpy-domain-htpy-hom-arrow f g α β H) a)
+      ( (g ·l htpy-domain-htpy-hom-arrow f g β γ K) a)
       ( coh-htpy-hom-arrow f g α β H a)
       ( coh-htpy-hom-arrow f g β γ K a))
```
fredrik-bakke, line 373: Couldn't you keep the arguments implicit?

src/foundation/morphisms-arrows.lagda.md
```diff
@@ -454,36 +461,40 @@ module _
...
+    ( right-whisker-coherence-square-identifications
       ( ap
         ( map-codomain-hom-arrow g h γ)
         ( htpy-codomain-htpy-hom-arrow f g α β H (f a)))
       ( ap (map-codomain-hom-arrow g h γ) (coh-hom-arrow f g α a))
+      ( ap (map-codomain-hom-arrow g h γ) (coh-hom-arrow f g β a))
+      ( ap
+        ( map-codomain-hom-arrow g h γ ∘ g)
+        ( htpy-domain-htpy-hom-arrow f g α β H a))
       ( coh-hom-arrow g h γ (map-domain-hom-arrow f g β a))
```
fredrik-bakke, line 473: Couldn't you keep the arguments implicit?

src/foundation/whiskering-higher-homotopies-composition.lagda.md
```diff
@@ -0,0 +1,65 @@
...
+
+Consider two dependent functions `f g : (x : A) → B x` equipped with two
+homotopies `H H' : f ~ g`, and consider a family of maps
+`h : (x : A) → B x → C x`. Then we obtain a map
+
+```text
+  α ↦ ap h ·l α : H ~ H' → h ·l H ~ h ·l H'
+```
+
+This operation is called the {{#concept "left whiskering of 2-homotopies"}}.
```
fredrik-bakke, line 29:
```suggestion
This operation is called the {{#concept "left whiskering of 2-homotopies" Agda=left-whisker-comp²}}.
```

src/foundation/whiskering-higher-homotopies-composition.lagda.md
```diff
@@ -0,0 +1,65 @@
...
+Alternatively the left whiskering operation of 2-homotopies can be defined using
+the
+[action on higher identifications of functions](foundation.action-on-higher-identifications-functions.md)
+by
+
+```text
+  α x ↦ ap² h (α x).
+```
+
+## Definitions
```
fredrik-bakke, line 39: Maybe also briefly mention right whiskering?

fredrik-bakke (commented):
I got through about half the files, and the changes look amazing so far! Can't wait to start using it. I'll review the rest later.

src/foundation-core/coherently-invertible-maps.lagda.md
```diff
@@ -139,24 +138,19 @@ module _
...
+              ( right-whisker-htpy (is-retraction-is-invertible H) f)
+              ( is-section-is-invertible H x)) ∙
+            ( ap
+              ( concat' _ (is-retraction-is-invertible H (f x)))
+              ( ( ap-comp f
+                  ( map-inv-is-invertible H ∘ f)
+                  ( is-section-is-invertible H x)) ∙
+                ( inv
+                  ( ap
+                    ( ap f)
```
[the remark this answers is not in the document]
EgbertRijke: I didn't do that here because it will induce cyclic module dependencies.

src/foundation-core/identity-types.lagda.md
EgbertRijke: Yes, and I could say "reflexivity element" ^^

src/foundation-core/whiskering-homotopies-composition.lagda.md
EgbertRijke: It will be easy to disambiguate with double whiskering and _iterated_ whiskering. I'll make a general naming scheme by adding it to the general whiskering operations.

src/foundation/function-extensionality.lagda.md
EgbertRijke: Thank you for that clarification!

src/foundation-core/coherently-invertible-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: ah, okay 👍

src/foundation/morphisms-arrows.lagda.md
EgbertRijke: It was annoying that it was so situational which arguments were implicit. I couldn't remember which arguments were expected, and find it easier if its just uniform: stuff about pasting always has the same arguments ordered `top-left`, `top-right`, `left`, `middle`, `right`, `bottom-left`, `bottom-right`.

Filling in a few extra arguments isn't really something that slows someone down in the formalization, but having to functions that all do something similar but have their arguments handled differently is a pain to the maintainer.

fredrik-bakke:
🥲

EgbertRijke:
We're compiling again! So I declare this PR done.

Our spaghetti code is now served with sauce.

fredrik-bakke:
Very nice! I'll see if I can review it during my layover in Belgrade tomorrow 😄🍴

src/foundation-core/coherently-invertible-maps.lagda.md
```diff
@@ -139,24 +138,19 @@ module _
...
+                    ( coh-is-coherently-invertible-id
+                      ( is-section-is-invertible H) x)))))))
```
fredrik-bakke, line 153: Since this entire proof is wrapped in an `inv`, I think it is pertinent to record `inv-coherence-map-inv-is-invertible` first. This way, if the inverse is needed, we can avoid `inv-inv`, and the bulk of the proof will be less indented.

src/foundation-core/commuting-squares-of-homotopies.lagda.md
```diff
@@ -0,0 +1,46 @@
...
+          top
+      f ------> g
+      |         |
+ left |         | right
+      v         v
+      h ------> i
+        bottom
+```
+
+is said to **commute** if there is a homotopy `left ∙h bottom ~ top ∙h right `.
```
fredrik-bakke, line 32:
```suggestion
is said to **commute** if there is a homotopy `left ∙h bottom ~ top ∙h right`.
```

src/foundation-core/commuting-squares-of-homotopies.lagda.md
```diff
@@ -0,0 +1,46 @@
...
+Such a homotopy is called a **coherence** of the square.
```
fredrik-bakke, line 33:
```suggestion
Such a homotopy is called a {{#concept "coherence" Disambiguation="square of homotopies"}} of the square.
```

src/foundation-core/commuting-squares-of-maps.lagda.md
```diff
@@ -24,15 +24,24 @@ open import foundation-core.whiskering-homotopies
...
+       A --------> X
+       |           |
+  left |           | right
+       ∨           ∨
+       B --------> Y
+          bottom
 ```
 
-is said to commute if there is a homotopy between both composites.
+is said to **commute** if there is a homotopy
```
fredrik-bakke, line 36:
```suggestion
is said to {{#concept "commute" Disambiguation="square of maps"}} if there is a [homotopy](foundation-core.homotopies.md)
```

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -214,15 +214,16 @@ Consider a map `f : A → B` and two functions `g h : B → C`. Then the square
 commutes.
 
 ```agda
-coherence-square-htpy-eq-ap-precomp :
+coherence-square-homotopies-eq-ap-precomp :
```
fredrik-bakke, line 217: `coherence-square-maps-htpy-eq-ap-precomp` or `coherence-htpy-eq-ap-precomp`

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -186,15 +186,15 @@ Then the square
 [commutes](foundation-core.commuting-squares-of-maps.md).
 
 ```agda
-coherence-square-htpy-eq-ap-precomp-Π :
+coherence-square-homotopies-htpy-eq-ap-precomp-Π :
```
fredrik-bakke, line 189: Ah, actually, this is a coherence square of maps. so it should be `coherence-square-maps-htpy-eq-ap-precomp-Π`. But we could also consider just calling it `coherence-htpy-eq-ap-precomp-Π` as I don't think the shape or type of the edges add much information to the name. That should probably be part of a different refactoring though.

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -243,15 +244,15 @@ Consider a map `f : {x : A} → B x → C x` and two functions
 commutes.
 
 ```agda
-coherence-square-htpy-eq-ap-postcomp-Π :
+coherence-square-homotopies-eq-ap-postcomp-Π :
```
fredrik-bakke, line 247: `coherence-square-maps-htpy-eq-ap-postcomp-Π` or `coherence-htpy-eq-ap-postcomp-Π`

src/foundation-core/function-extensionality.lagda.md
```diff
@@ -271,16 +272,16 @@ Consider a map `f : B → C` and two functions `g h : A → B`. Then the square
 commutes.
 
 ```agda
-coherence-square-htpy-eq-ap-postcomp :
+coherence-square-homotopies-eq-ap-postcomp :
```
fredrik-bakke, line 275: `coherence-square-maps-htpy-eq-ap-postcomp-Π` or `coherence-htpy-eq-ap-postcomp-Π`

src/foundation-core/identity-types.lagda.md
```diff
@@ -131,30 +205,144 @@ module _
...
+```text
+  (p ∙ q) ∙ inv q ＝ p
+  (p ∙ inv q) ∙ q ＝ p.
+```
+
+In
+[`foundation.identity-types`](foundation.identity-types.md)`we will use these families of identifications to conclude that`concat
+p z`and`concat' x
+q`are [equivalences](foundation-core.equivalences.md) with inverses`concat (inv
+p) z`and`concat' x (inv q)`, respectively.
```
fredrik-bakke, line 299:
```suggestion
In [`foundation.identity-types`](foundation.identity-types.md) we will use these families of identifications to conclude that `concat p z` and `concat' x q` are [equivalences](foundation-core.equivalences.md) with inverses `concat (inv p) z` and `concat' x (inv q)`, respectively.
```

src/foundation-core/identity-types.lagda.md
```diff
@@ -131,30 +205,144 @@ module _
...
+  {l : Level} {A : UU l}
+  where
+
+  is-retraction-inv-concat :
+    {x y z : A} (p : x ＝ y) (q : y ＝ z) → (inv p ∙ (p ∙ q)) ＝ q
+  is-retraction-inv-concat refl q = refl
+
+  is-section-inv-concat :
+    {x y z : A} (p : x ＝ y) (r : x ＝ z) → (p ∙ (inv p ∙ r)) ＝ r
+  is-section-inv-concat refl refl = refl
```
fredrik-bakke, line 312: I don't see why you would need to pattern-match on `r` in this proof.

```suggestion
  is-section-inv-concat :
    {x y z : A} (p : x ＝ y) (r : x ＝ z) → (p ∙ (inv p ∙ r)) ＝ r
  is-section-inv-concat refl r = refl
```

src/foundation-core/identity-types.lagda.md
```diff
@@ -182,9 +370,56 @@ module _
   double-transpose-eq-concat' r refl refl q α = right-unit ∙ (α ∙ right-unit)
 ```
 
-The fact that `left-transpose-eq-concat` and `right-transpose-eq-concat` are
-equivalences is recorded in
-[`foundation.identity-types`](foundation.identity-types.md).
+### Splicing and unsplicing concatenations of identifications
```
fredrik-bakke, line 373: Nice addition!

src/foundation/commuting-squares-of-maps.lagda.md
```diff
@@ -218,30 +213,26 @@ module _
...
-    v                    v
-    B -----------------> Y
-            bottom
+            top
+       A --------> X
+       |           |
+  left |           | right
+       ∨           ∨
+       B --------> Y
+          bottom
```
fredrik-bakke, line 222: I don't see how this is an improvement

src/foundation/commuting-squares-of-maps.lagda.md
```diff
@@ -218,30 +213,26 @@ module _
...
-        v                         v
-     (S → B) ----------------> (S → Y)
-             postcomp S bottom
+                        postcomp S top
+              (S → A) ------------------> (S → X)
+                 |                           |
+ postcomp S left |                           | postcomp S right
+                 ∨                           ∨
+              (S → B) ------------------> (S → Y).
+                       postcomp S bottom
```
fredrik-bakke, line 235: I don't see how this is an improvement

src/foundation/commuting-triangles-of-homotopies.lagda.md
```diff
@@ -75,59 +79,68 @@ module _
...
-  {left : f ~ h} (right : g ~ h) {top : f ~ g}
+  {f g h i : (x : A) → B x} (H : i ~ f)
+  (left : f ~ h) (right : g ~ h) (top : f ~ g)
   where
 
-  right-whisk-htpy-coherence-triangle-homotopies :
-    {i : (x : A) → B x}
-    (T : coherence-triangle-homotopies left right top) (H : i ~ f) →
+  left-whisker-concat-coherence-triangle-homotopies :
+    (T : coherence-triangle-homotopies left right top) →
```
fredrik-bakke, line 123: You've somehow switched what's left and right whiskering

src/foundation/commuting-triangles-of-homotopies.lagda.md
```diff
@@ -75,59 +79,68 @@ module _
...
 
-  right-whisk-coherence-triangle-homotopies :
+  left-whisker-comp-coherence-triangle-homotopies :
     {l3 : Level} {X : UU l3}
     (T : coherence-triangle-homotopies left right top) (i : X → A) →
     coherence-triangle-homotopies
       {f = f ∘ i} {g ∘ i} {h ∘ i}
       (left ·r i) (right ·r i) (top ·r i)
-  right-whisk-coherence-triangle-homotopies T i = T ∘ i
+  left-whisker-comp-coherence-triangle-homotopies T i = T ∘ i
```
fredrik-bakke, line 145: this is right whiskering

src/foundation/commuting-triangles-of-homotopies.lagda.md
```diff
@@ -75,59 +79,68 @@ module _
...
+  right-whisker-comp-coherence-triangle-homotopies :
     {l3 : Level} {X : UU l3} (i : B → X)
     (T : coherence-triangle-homotopies left right top) →
     coherence-triangle-homotopies
       {f = i ∘ f} {i ∘ g} {i ∘ h}
       (i ·l left) (i ·l right) (i ·l top)
-  left-whisk-coherence-triangle-homotopies i =
-    distributivity-left-whisk i left right top
+  right-whisker-comp-coherence-triangle-homotopies i =
+    distributivity-left-whisker i left right top
```
fredrik-bakke, line 110: This is left whiskering

src/foundation/commuting-triangles-of-identifications.lagda.md
```diff
@@ -57,145 +61,355 @@ module _
...
+     (left ＝ top ∙ right) → left ＝ (top ∙ p) ∙ (p⁻¹ ∙ right).
+   ```
+
+   Similarly, we have a map
+
+   ```text
+     (left ＝ top ∙ right) → left ＝ (top ∙ p⁻¹) ∙ (p ∙ right).
+   ```
+
+Because concatenation of identifications is an equivalence, it follows that all
```
fredrik-bakke, line 134:
```suggestion
Because concatenation of identifications is an [equivalence](foundation-core.equivalences.md), it follows that all
```

src/foundation/commuting-triangles-of-identifications.lagda.md
```diff
@@ -57,145 +61,355 @@ module _
...
+
+Furthermore, in the case where the identification `right` is `refl` we obtain an
+identification
+
+```text
+  inv right-unit ＝
+  map-coherence-triangle-identifications p refl p (inv right-unit)
+```
+
+and in the case where the identification `top` is refl` we obtain
```
fredrik-bakke, line 387:
```suggestion
and in the case where the identification `top` is `refl` we obtain
```

src/foundation/homotopies-morphisms-arrows.lagda.md
```diff
@@ -0,0 +1,538 @@
...
+      ( comp-hom-arrow f g h γ β)
+  pr1 left-whisker-comp-hom-arrow =
+    htpy-domain-left-whisker-comp-hom-arrow
+  pr1 (pr2 left-whisker-comp-hom-arrow) =
+    htpy-codomain-left-whisker-comp-hom-arrow
+  pr2 (pr2 left-whisker-comp-hom-arrow) =
+    coh-left-whisker-comp-hom-arrow
+```
+
+#### Right whiskering
```
fredrik-bakke, line 319:
```suggestion
#### Right whiskering of homotopies of homotopies of morphisms of arrows with respect to composition
```

src/foundation/postcomposition-functions.lagda.md
```diff
@@ -206,3 +208,26 @@ module _
...
+    eq-htpy-refl-htpy (h ∘ f)
+
+  compute-ap-postcomp-eq-htpy :
+    (H : f ~ g) →
+    eq-htpy (h ·l H) ＝ ap (postcomp A h) (eq-htpy H)
+  compute-ap-postcomp-eq-htpy H =
+    ( ap
+      ( λ K → eq-htpy (h ·l K))
+      ( inv (is-section-eq-htpy H))) ∙
+    ( compute-ap-postcomp-htpy-eq (eq-htpy H))
```
fredrik-bakke, line 232: These are already defined in `foundation.function-extensionality` I think.

src/foundation/precomposition-functions.lagda.md
```diff
@@ -151,3 +151,27 @@ module _
...
+
+  compute-ap-precomp-eq-htpy :
+    ( H : f ~ g) →
+    eq-htpy (H ·r h) ＝ ap (precomp h C) (eq-htpy H)
+  compute-ap-precomp-eq-htpy H =
+    ( ap
+      ( λ K → eq-htpy (K ·r h))
+      ( inv (is-section-eq-htpy H))) ∙
+    ( compute-ap-precomp-htpy-eq (eq-htpy H))
+```
```
fredrik-bakke, line 177: Aren't these already defined in `foundation(-core).function-extensionality`?

fredrik-bakke:
I'm about half-way done again. So much to review 😵‍💫

EgbertRijke:
> I'm about half-way done again. So much to review 😵‍💫

Thank you for doing it so quickly. I’m sorry that this pull request got so big. There were a lot of problems with this part of the library (duplications, bad names, confusing specifications, unrefactored refactorable code, and generally little coherence in how we went about higher path algebra, no explanations). Once I addressed those problems it broke later parts of the library, so I had to go fix those too.

src/foundation/commuting-squares-of-maps.lagda.md
EgbertRijke: The original diagram was massive. This doesn’t happen in latex either. So I reduced the size without making the diagram appear cramped.

I switched ordinarily labelling of arrows, as they are more standard.

This diagram appears many times in our informal explanations, and now its appearance matches with its other occurrences, making our diagram formatting more uniform.

That’s three reasons:)

fredrik-bakke: While I'm not entirely sold on using `⇗` when the squares are not lax, I don't see why the diagrams should be more uniform, and I think the previous version was more aesthetically pleasing. Moreover, I don't see how the previous versions could be characterized as "massive". If anything, I think they are still kind of small:

image

(Note that the font size of `text` blocks is smaller than that of `agda`-blocks. Something we should fix, because it looks a little weird when non-type-checked code is placed next to type-checked code.)

Either way, I don't think it is worthwhile fighting over this, as I still very much consider our current solution to be temporary, so I will mark this conversation as resolved. Maybe I will finally have some time to implement a more permanent solution soon.

src/foundation-core/coherently-invertible-maps.lagda.md
EgbertRijke: This is a small request. It is no problem for me to implement it.

src/synthetic-homotopy-theory/dependent-coforks.lagda.md
```diff
@@ -358,7 +351,14 @@ module _
...
                               ( d)
-                              ( inl a))))) ∙
+                              ( inl a))))
+                        ( coherence-square-dependent-cocone
+                            ( vertical-map-span-cocone-cofork f g)
+                            ( horizontal-map-span-cocone-cofork f g)
+                            ( cocone-codiagonal-cofork f g e)
+                            ( P)
+                            ( d)
+                            ( inl a))) ∙
```
fredrik-bakke, line 361:
```suggestion
                        ( coherence-square-dependent-cocone
                          ( vertical-map-span-cocone-cofork f g)
                          ( horizontal-map-span-cocone-cofork f g)
                          ( cocone-codiagonal-cofork f g e)
                          ( P)
                          ( d)
                          ( inl a))) ∙
```

src/synthetic-homotopy-theory/dependent-cocones-under-spans.lagda.md
```diff
@@ -173,11 +173,11 @@ module _
...
       pr1 (pr2 (htpy-eq-dependent-cocone d' p))
 
-    coherence-square-htpy-eq-dependent-cocone :
+    coherence-square-homotopies-eq-dependent-cocone :
       coherence-htpy-dependent-cocone d'
         ( horizontal-htpy-eq-dependent-cocone)
         ( vertical-htpy-eq-dependent-cocone)
-    coherence-square-htpy-eq-dependent-cocone =
+    coherence-square-homotopies-eq-dependent-cocone =
       pr2 (pr2 (htpy-eq-dependent-cocone d' p))
```
fredrik-bakke, line 181: I'm not entirely sure this renaming is correct

src/synthetic-homotopy-theory/dependent-coforks.lagda.md
EgbertRijke: Formatting errors in files not directly related to the subject of the PR definitely starts to feel out of scope, but I will implement this change.

EgbertRijke: Ah, perhaps it was a formatting error that I introduced. Then I definitely need to address it. Thanks for catching it!

src/foundation/postcomposition-functions.lagda.md
EgbertRijke: You're right! I moved these over from another file, so I didn't catch that they were already duplicate.

src/synthetic-homotopy-theory/dependent-cocones-under-spans.lagda.md
EgbertRijke: Good catch

src/foundation-core/function-extensionality.lagda.md
fredrik-bakke, line 1: Lol, we've been fighting over whether this file should exist or not. But seems you may have found a better solution? I'll have to see during the review again. Anyway. Since there are no formalizations left in this file I think you should delete it.

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -34,8 +34,11 @@ A square of [homotopies](foundation-core.homotopies.md)
         bottom
 ```
 
-is said to **commute** if there is a homotopy `left ∙h bottom ~ top ∙h right `.
-Such a homotopy is called a **coherence** of the square.
+is said to be a {{#concept "commuting sqsuare" Disambiguation="homotopies"}} of
```
fredrik-bakke, line 37:
```suggestion
is said to be a {{#concept "commuting square" Disambiguation="homotopies"}} of
```

src/foundation/commuting-triangles-of-maps.lagda.md
```diff
@@ -196,3 +198,66 @@ module _
...
+    coherence-triangle-maps diagonal right top →
+    coherence-square-maps top left right bottom
+  horizontal-pasting-coherence-triangle-maps diagonal H K = H ∙h K
+
+  compute-refl-htpy-horizontal-pasting-coherence-triangle-maps :
+    (diagonal : A → Y) →
+    (H : coherence-triangle-maps' diagonal bottom left) →
+    (K : coherence-triangle-maps diagonal right top) →
+    (horizontal-pasting-htpy-coherence-triangle-maps refl-htpy H K) ~
+    (horizontal-pasting-coherence-triangle-maps diagonal H K)
```
fredrik-bakke, line 260:
```suggestion
    ( horizontal-pasting-htpy-coherence-triangle-maps refl-htpy H K) ~
    ( horizontal-pasting-coherence-triangle-maps diagonal H K)
```

src/foundation/composition-algebra.lagda.md
```diff
@@ -78,7 +78,7 @@ module _
     (h : C → A) (H : f ~ g) (S : UU l4) →
     precomp h S ·l htpy-precomp H S ~ htpy-precomp (H ·r h) S
   inv-distributive-htpy-precomp-right-whisker h H S i =
-    coherence-square-eq-htpy-ap-precomp h (i ∘ f) (i ∘ g) (i ·l H)
+    compute-eq-htpy-ap-precomp h (i ·l H)
```
fredrik-bakke, line 81: Yeah, that's better 🙂

src/foundation/evaluation-functions.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (B : A → UU l2) (a : A)
+  where
+
+  ev-dependent-function : ((x : A) → B x) → B a
+  ev-dependent-function = ev a
+```
+
+### The evaluation function for nondependent functions
```
fredrik-bakke, line 54: TIL
image

(I've been using a hyphen, but apparantly that's wrong)

src/foundation/evaluation-functions.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2} (a : A)
+  where
+
+  ev-implicit-function : ({x : A} → B x) → B a
+  ev-implicit-function f = f {a}
+```
+
+### The evaluation function of implicit function, specified with an explicit type family
```
fredrik-bakke, line 76:
```suggestion
### The evaluation function of implicit functions, specified with an explicit type family
```

src/foundation/evaluation-functions.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+For any two functions `f g : A → B`, the action on identifications of
+`ev-function B a`
+
+```text
+  ap (ev-function B a) : (f ＝ g) → (f a ＝ g a)
+```
+
+is homotopic to the map `p ↦ htpy-eq p a`, where `htpy-eq` is the operation that
+constructs a homotopy from an identification of functions, which we constructed
+in [foundation.function-extensionality. In other words, the triangle of maps
```
fredrik-bakke, line 100:
```suggestion
in [`foundation.function-extensionality`](foundation.function-extensionality.md). In other words, the triangle of maps
```

src/foundation/evaluation-functions.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+
+[commutes](foundation-core.commuting-triangles-of-maps.md).
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (a : A)
+  where
+
+  ap-ev : {f g : A → B} (p : f ＝ g) → (ap (ev-function B a) p) ＝ htpy-eq p a
+  ap-ev refl = refl
```
fredrik-bakke, line 119: Maybe I'm losing it, but what if we used `ap (ev-function B a) p` as the definition for `htpy-eq`? I could foresee us getting lots of nice definitional equalities w.r.t. whiskering and so on if we did.

fredrik-bakke: Not that this is relevant to your PR of course

src/foundation-core/function-extensionality.lagda.md
EgbertRijke: I know... but last time you created the file we didn't have files for pre- and postcomposition yet. It seemed more natural to put those lemmas there, and to keep `function-extensionality` really light on dependencies. In some other files I also reduced the dependencies by not using simply replacable definitions from later files, and this reduced a lot of the tension that generates cyclic dependencies every now and then.

src/foundation/evaluation-functions.lagda.md
EgbertRijke: I can do a quick search

src/foundation/transposition-identifications-along-equivalences.lagda.md
```diff
@@ -0,0 +1,230 @@
...
+
+of [identity types](foundation-core.identity-types.md). There are two ways of
+constructing this equivalence. One way uses the fact that `e⁻¹` is a
+[section](foundation-core.sections.md) of `e`, from which it follows that
+
+```text
+ (e x ＝ y) ≃ (e x ＝ e e⁻¹ y) ≃ (x ＝ e⁻¹ y).
+```
+
+The other way usex the fact that `e⁻¹` is a
```
fredrik-bakke, line 40:
```suggestion
The other way uses the fact that `e⁻¹` is a
```

src/foundation/transposition-identifications-along-equivalences.lagda.md
fredrik-bakke, line 1: Very nice file. I just hope people will find it

src/graph-theory/equivalences-directed-graphs.lagda.md
```diff
@@ -24,11 +23,14 @@ open import foundation.homotopies
 open import foundation.identity-types
 open import foundation.structure-identity-principle
 open import foundation.torsorial-type-families
+open import foundation.transposition-identifications-along-equivalences
 open import foundation.type-arithmetic-dependent-pair-types
 open import foundation.type-theoretic-principle-of-choice
 open import foundation.univalence
 open import foundation.universe-levels
 
+open import foundation-core.equivalences
```
fredrik-bakke, line 32: I don't see why you would need to import core here
```suggestion
open import foundation.equivalences
```

src/structured-types/initial-pointed-type-equipped-with-automorphism.lagda.md
```diff
@@ -14,13 +14,15 @@ open import foundation.action-on-identifications-functions
...
 open import foundation.dependent-pair-types
-open import foundation.equivalences
 open import foundation.homotopies
 open import foundation.identity-types
 open import foundation.iterating-automorphisms
+open import foundation.transposition-identifications-along-equivalences
 open import foundation.unit-type
 open import foundation.universe-levels
 
+open import foundation-core.equivalences
```
fredrik-bakke, line 24:
```suggestion
open import foundation.equivalences
```
Is this a greedy search and replace?

src/foundation/evaluation-functions.lagda.md
EgbertRijke: That's a very nice idea! We should use the dependent version, of course, and check that it satisfies the expected judgmental equality at `refl`, but I'm definitely open to it. Good find!

src/structured-types/initial-pointed-type-equipped-with-automorphism.lagda.md
EgbertRijke: No that was actually me not realizing that I was outside of foundation:)

src/foundation-core/function-extensionality.lagda.md
EgbertRijke: To be sure, I'm not opposed to having a file in foundation-core if it is necessary for some reason. But if it is not necessary then it is my preference to not split content over two files.

src/graph-theory/equivalences-directed-graphs.lagda.md
EgbertRijke: My mistake. Thanks for catching it!

EgbertRijke:
I finished working through this round of review comments.

src/foundation/evaluation-functions.lagda.md
EgbertRijke: So, I actually put it in the library, and it made one step somewhere shorter:)

src/foundation/action-on-higher-identifications-functions.lagda.md
```diff
@@ -0,0 +1,189 @@
+# The action of functions on higher identifications
```
fredrik-bakke, line 1: Can you add a `## See also` link to this file in `action-on-identifications-functions`?

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -28,18 +34,258 @@ A square of [homotopies](foundation-core.homotopies.md)
...
+      f ------> g                    f -------------> g
+      |         |                    |                |
+ left |         | right    ≃    left |                | right ∙h H
+      ∨         ∨                    ∨                ∨
+      h ------> i                    h -------------> j
+        bottom                          bottom ∙h H
+```
+
+This is the
+{{#concept "right whiskering" Disambiguation="commuting squares of homotopies with respect to concatenation"}}
```
fredrik-bakke, line 72:
```suggestion
{{#concept "right whiskering" Disambiguation="commuting squares of homotopies with respect to concatenation" Agda=right-whisker-concat-coherence-square-homotopies}}
```

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -28,18 +34,258 @@ A square of [homotopies](foundation-core.homotopies.md)
...
+      f ------> g                         e ----------> g
+      |         |                         |             |
+ left |         | right    ≃    H ∙h left |             | right
+      ∨         ∨                         ∨             ∨
+      h ------> i                         h ----------> i
+        bottom                                bottom
+```
+
+This is the
+{{#concept "left whiskering" Disambiguation="commuting squares of homotopies with respect to concatenation"}}
```
fredrik-bakke, line 137:
```suggestion
{{#concept "left whiskering" Disambiguation="commuting squares of homotopies with respect to concatenation" Agda=left-whisker-concat-coherence-square-homotopies}}
```

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -28,18 +34,258 @@ A square of [homotopies](foundation-core.homotopies.md)
...
```
fredrik-bakke, line 202:
```suggestion
{{#concept "left whiskering" Disambiguation="commuting squares of homotopies with respect to concatenation" Agda=double-whisker-coherence-square-homotopies}}
```

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -28,18 +34,258 @@ A square of [homotopies](foundation-core.homotopies.md)
...
+  { l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {C : UU l3}
+  ( f : B → C) {g h h' k : A → B}
+  ( H : g ~ h) (H' : g ~ h') {K : h ~ k} {K' : h' ~ k}
+  where
+
+  ap-left-whisker-concat-coherence-square-homotopies :
+    coherence-square-homotopies H H' K K' →
+    coherence-square-homotopies (f ·l H) (f ·l H') (f ·l K) (f ·l K')
+  ap-left-whisker-concat-coherence-square-homotopies α a =
+    map-coherence-square-identifications f (H a) (H' a) (K a) (K' a) (α a)
```
fredrik-bakke, line 279: Should this be `map-...` instead?

src/foundation/commuting-squares-of-homotopies.lagda.md
```diff
@@ -28,18 +34,258 @@ A square of [homotopies](foundation-core.homotopies.md)
...
   where
 
-  coherence-square-homotopies :
-    (top : f ~ g) (left : f ~ h) (right : g ~ i) (bottom : h ~ i) → UU (l1 ⊔ l2)
-  coherence-square-homotopies top left right bottom =
-    left ∙h bottom ~ top ∙h right
+  ap-right-whisker-concat-coherence-square-homotopies :
+    coherence-square-homotopies H H' K K' →
+    coherence-square-homotopies (H ·r f) (H' ·r f) (K ·r f) (K' ·r f)
+  ap-right-whisker-concat-coherence-square-homotopies α = α ·r f
```
fredrik-bakke, line 290: This also? `map-...`

src/foundation/commuting-squares-of-identifications.lagda.md
```diff
@@ -22,197 +26,1654 @@ open import foundation-core.identity-types
...
+           top
+       x -------> y
+       |          |
+  left |          | right
+       ∨          ∨
+       z -------> w
+          bottom
 ```
 
 is said to **commute** if there is an identification
```
fredrik-bakke, line 38:
```suggestion
is said to {{#concept "commute" Disambiguation="commuting square of identifications"}} if there is an identification
```

src/foundation/commuting-squares-of-identifications.lagda.md
```diff
@@ -22,197 +26,1654 @@ open import foundation-core.identity-types
...
+  where
+
+  equiv-left-whisker-concat-coherence-square-identifications :
+    (p : u ＝ x)
+    (top : x ＝ y) (left : x ＝ z) (right : y ＝ w) (bottom : z ＝ w) →
+    coherence-square-identifications top left right bottom ≃
     coherence-square-identifications (p ∙ top) (p ∙ left) right bottom
-  coherence-square-identifications-top-left-whisk' refl sq = sq
+  equiv-left-whisker-concat-coherence-square-identifications refl
+    top left right bottom = id-equiv
```
fredrik-bakke, line 702:
```suggestion
  equiv-left-whisker-concat-coherence-square-identifications
    refl top left right bottom =
    id-equiv
```

src/foundation/function-extensionality.lagda.md
```diff
@@ -2,200 +2,276 @@
...
+### `htpy-eq` preserves concatenation of identifications
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2} {f g h : (x : A) → B x}
+  where
+
+  htpy-eq-concat :
+    (p : f ＝ g) (q : g ＝ h) → htpy-eq (p ∙ q) ＝ htpy-eq p ∙h htpy-eq q
+  htpy-eq-concat refl refl = refl
```
fredrik-bakke, line 203: This should probably be swapped with an instantiation of the same result for `ap`

src/foundation/function-extensionality.lagda.md
```diff
@@ -2,200 +2,276 @@
...
                 inv-htpy
 ```
 
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2} {f g : (x : A) → B x}
+  where
+
+  compute-htpy-eq-inv : inv-htpy {f = f} {g} ∘ htpy-eq ~ htpy-eq ∘ inv
+  compute-htpy-eq-inv refl = refl
```
fredrik-bakke, line 244: This should probably be swapped with an instantiation of the same result for `ap`

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -75,173 +114,332 @@ pushout-product:
...
-  pr1 cone-pullback-hom' = postcomp B g
-  pr1 (pr2 cone-pullback-hom') = precomp f X
-  pr2 (pr2 cone-pullback-hom') = refl-htpy
+  cone-pullback-hom : cone (precomp f Y) (postcomp A g) (B → X)
+  pr1 cone-pullback-hom = postcomp B g
+  pr1 (pr2 cone-pullback-hom) = precomp f X
+  pr2 (pr2 cone-pullback-hom) = refl-htpy
+
+  gap-cone-pullback-hom : (B → X) → type-standard-pullback-hom f g
+  gap-cone-pullback-hom = gap (precomp f Y) (postcomp A g) cone-pullback-hom
```
fredrik-bakke, line 242: Why not `gap-pullback-hom`?

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -75,173 +114,332 @@ pushout-product:
...
+  (f : A → B) (g : X → Y)
+  where
+
+  triangle-pullback-hom :
+    coherence-triangle-maps'
+      ( gap-cone-pullback-hom f g)
+      ( map-compute-pullback-hom f g)
+      ( pullback-hom f g)
+  triangle-pullback-hom j =
+    eq-pair-Σ refl (eq-pair-Σ refl (is-retraction-eq-htpy refl))
```
fredrik-bakke, line 340: I've been using `eq-pair-eq-pr2` for these kinds of proofs recently since the type of the second component reduces in this case. Maybe you want to weigh in on wheter we want to rename it, but otherwise I think it should be used.

fredrik-bakke:
```suggestion
    eq-pair-eq-pr2 (eq-pair-eq-pr2 (is-retraction-eq-htpy refl))
```

src/orthogonal-factorization-systems/orthogonal-maps.lagda.md
```diff
@@ -224,20 +225,24 @@ module _
...
-      ( map-fibered-map-type-standard-pullback-hom f g)
-      ( gap (precomp f Y) (postcomp A g) (cone-pullback-hom' f g))
-      ( H)
-      ( is-equiv-map-equiv (equiv-fibered-map-type-standard-pullback-hom f g))
+  is-orthogonal-is-orthogonal-pullback-condition =
+    is-equiv-top-map-triangle
+      ( gap-cone-pullback-hom f g)
+      ( map-compute-pullback-hom f g)
+      ( pullback-hom f g)
+      ( inv-htpy (triangle-pullback-hom f g))
```
fredrik-bakke, line 244: regarding our discussion earlier, these two proofs would reduce if `eq-htpy refl-htpy ≐ refl`

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -75,173 +114,332 @@ pushout-product:
...
+  pr2 (pr2 (map-inv-compute-pullback-hom h)) =
+    coh-standard-pullback-hom f g h
+
+  is-section-map-inv-compute-pullback-hom :
+    is-section map-compute-pullback-hom map-inv-compute-pullback-hom
+  is-section-map-inv-compute-pullback-hom h =
+    eq-pair-Σ refl
+      ( eq-pair-Σ refl
+        ( is-retraction-eq-htpy (eq-coh-standard-pullback-hom f g h)))
+
```
fredrik-bakke, line 278:
```suggestion
    eq-pair-eq-pr2
      ( eq-pair-eq-pr2
        ( is-retraction-eq-htpy (eq-coh-standard-pullback-hom f g h)))

```

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -75,173 +114,332 @@ pushout-product:
...
+  is-retraction-map-inv-compute-pullback-hom :
+    is-retraction map-compute-pullback-hom map-inv-compute-pullback-hom
+  is-retraction-map-inv-compute-pullback-hom h =
+    eq-pair-Σ refl
```
fredrik-bakke, line 282:
```suggestion
    eq-pair-eq-pr2 (eq-pair-eq-pr2 (is-section-eq-htpy (coh-hom-arrow f g h)))
```

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -75,173 +114,332 @@ pushout-product:
...
+      ( eq-pair-Σ refl (is-section-eq-htpy (coh-hom-arrow f g h)))
```
fredrik-bakke, line 283:
```suggestion
    eq-pair-eq-pr2 (eq-pair-eq-pr2 (is-section-eq-htpy (coh-hom-arrow f g h)))
```

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -270,15 +485,31 @@ module _
...
+      ( map-domain-right-whisker-hom-arrow α)
+      ( map-codomain-right-whisker-hom-arrow α)
+  coh-right-whisker-hom-arrow α =
+    coh-hom-arrow f g α ·r h
+
+  right-whisker-hom-arrow :
+    hom-arrow f g → hom-arrow (f ∘ h) g
+  pr1 (right-whisker-hom-arrow α) = map-domain-right-whisker-hom-arrow α
+  pr1 (pr2 (right-whisker-hom-arrow α)) = map-codomain-right-whisker-hom-arrow α
+  pr2 (pr2 (right-whisker-hom-arrow α)) = coh-right-whisker-hom-arrow α
```
fredrik-bakke, line 506: This hardly looks like the right place for these definitions

src/orthogonal-factorization-systems/lifting-squares.lagda.md
```diff
@@ -8,280 +8,323 @@ module orthogonal-factorization-systems.lifting-squares where
...
 open import orthogonal-factorization-systems.lifts-of-maps
+open import orthogonal-factorization-systems.pullback-hom
 ```
 
 </details>
 
 ## Idea
 
-A **lifting square** is a commuting square
+A **lifting square** of a commuting square
```
fredrik-bakke, line 45:
```suggestion
A {{#concept "lifting square" Disambiguation="maps of types" Agda=lifting-square}} of a commuting square
```

src/orthogonal-factorization-systems/lifting-squares.lagda.md
```diff
@@ -8,280 +8,323 @@ module orthogonal-factorization-systems.lifting-squares where
...
+    coherence-htpy-lifting-square f g α k k
+      ( htpy-diagonal-map-refl-htpy-lifting-square)
+  coh-refl-htpy-lifting-square =
+    right-unit-law-concat-htpy-hom-arrow f g α
+      ( pullback-hom f g (diagonal-map-lifting-square f g α k))
+      ( is-lifting-diagonal-map-lifting-square f g α k)
+
+  refl-htpy-lifting-square : htpy-lifting-square f g α k k
+  pr1 refl-htpy-lifting-square = htpy-diagonal-map-refl-htpy-lifting-square
+  pr2 refl-htpy-lifting-square = coh-refl-htpy-lifting-square
```
fredrik-bakke, line 229: 😍

src/orthogonal-factorization-systems/lifting-squares.lagda.md
```diff
@@ -8,280 +8,323 @@ module orthogonal-factorization-systems.lifting-squares where
...
+
+  is-torsorial-htpy-lifting-square :
+    is-torsorial (htpy-lifting-square f g α l)
+  is-torsorial-htpy-lifting-square =
+    is-torsorial-Eq-structure
+      ( is-torsorial-htpy _)
+      ( diagonal-map-lifting-square f g α l , refl-htpy)
+      ( is-torsorial-htpy-htpy-hom-arrow f g α
+        ( pullback-hom f g (diagonal-map-lifting-square f g α l))
+        ( _))
```
fredrik-bakke, line 313: 😶‍🌫️

src/orthogonal-factorization-systems/lifting-squares.lagda.md
fredrik-bakke, line 1: Absolutely marvelous. Thank you so much for doing this!

src/orthogonal-factorization-systems/orthogonal-maps.lagda.md
fredrik-bakke: Actually, this is a great reason not to make this a rewrite rule. We will not have as good control anymore of where we are applying function extensionality.

src/orthogonal-factorization-systems/pullback-hom.lagda.md
```diff
@@ -253,15 +451,32 @@ module _
...
+      ( f)
+      ( h ∘ g)
+      ( map-codomain-left-whisker-hom-arrow α)
+  coh-left-whisker-hom-arrow α = h ·l (coh-hom-arrow f g α)
+
+  left-whisker-hom-arrow :
+    hom-arrow f g → hom-arrow f (h ∘ g)
+  pr1 (left-whisker-hom-arrow α) = map-domain-left-whisker-hom-arrow α
+  pr1 (pr2 (left-whisker-hom-arrow α)) = map-codomain-left-whisker-hom-arrow α
+  pr2 (pr2 (left-whisker-hom-arrow α)) = coh-left-whisker-hom-arrow α
```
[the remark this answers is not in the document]
EgbertRijke: I just edited definitions that were already there:)

src/orthogonal-factorization-systems/lifting-squares.lagda.md
EgbertRijke: The power of refactoring!

src/orthogonal-factorization-systems/lifting-squares.lagda.md
EgbertRijke: Would you mind if I change the terminology to "lifting of a square"? It sounds more accurate to me, because the commuting square is in the context.

src/orthogonal-factorization-systems/pullback-hom.lagda.md
EgbertRijke: I agree but I will let the original author of these definitions deal with this:)

src/orthogonal-factorization-systems/lifting-squares.lagda.md
fredrik-bakke: How about lifting _in_ a square? lifting of a square sounds like a different operation to me intuitively speaking

src/orthogonal-factorization-systems/pullback-hom.lagda.md
EgbertRijke: What is `eq-pair-eq-pr2`? That name sounds so unintuitive to me. I think that if you want me to use it, then I'm gonna rename that function 😅

src/foundation/whiskering-higher-homotopies-composition.lagda.md
```diff
@@ -0,0 +1,83 @@
...
+the
+[action on higher identifications of functions](foundation.action-on-higher-identifications-functions.md)
+by
+
+```text
+  α x ↦ ap² h (α x).
+```
+
+Similarly, the {{#concept "right whiskering" Disambiguation="2-homotopies with
+respect to composition" Agda=right-whisker-comp²]} is defined to be the
```
fredrik-bakke, line 41:
```suggestion
Similarly, the {{#concept "right whiskering" Disambiguation="2-homotopies with respect to composition" Agda=right-whisker-comp²}} is defined to be the
```

src/orthogonal-factorization-systems/pullback-hom.lagda.md
fredrik-bakke: It's `eq-pair-Σ refl` :) I think the name looks awful, but I wouldn't call it unintuitive.

src/orthogonal-factorization-systems/orthogonal-maps.lagda.md
EgbertRijke: Sure, but that's such a minor benefit and we'd have to break the semantics for it. Learning how to deal with identifications is worth much more than giving in to what-if-this-equality-was-judgmental desires.

src/foundation/whiskering-homotopies-composition.lagda.md
```diff
@@ -40,15 +45,20 @@ whiskering** operation assumes a diagram of the form
                h
 ```
 
-and is defined to be a function `_·r f : (g ~ h) → (g ∘ f ~ h ∘ f)`.
+and is defined to be a function `H ↦ H ·r f : (g ~ h) → (g ∘ f ~ h ∘ f)`.
```
fredrik-bakke, line 48:
```suggestion
and is defined to be the function `H ↦ H ·r f : (g ~ h) → (g ∘ f ~ h ∘ f)`.
```

src/foundation/whiskering-homotopies-composition.lagda.md
```diff
@@ -30,8 +33,10 @@ assumes a diagram of the form
       g
 ```
 
-and is defined to be a function `h ·l_ : (f ~ g) → (h ∘ f ~ h ∘ g)`. The **right
-whiskering** operation assumes a diagram of the form
+and is defined to be a function `H ↦ h ·l H : (f ~ g) → (h ∘ f ~ h ∘ g)`. The
```
fredrik-bakke, line 36:
```suggestion
and is defined to be the function `H ↦ h ·l H : (f ~ g) → (h ∘ f ~ h ∘ g)`. The
```

src/foundation/whiskering-homotopies-concatenation.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+</details>
+
+## Idea
+
+Consider a homotopy `H : f ~ g` and a homotopy `K : I ~ J` between two
+homotopies `I J : g ~ f`. The
+{{#concept "left whiskering" Disambiguation="homotopies with respect to concatenation" Agda=left-whisker-concat-htpy}}
+of `H` and `K` is a homotopy `H ∙ I ~ H ∙ J`. In other words, left whiskering of
```
fredrik-bakke, line 26:
```suggestion
of `H` and `K` is a homotopy `H ∙h I ~ H ∙h J`. In other words, left whiskering of
```

src/orthogonal-factorization-systems/orthogonal-maps.lagda.md
fredrik-bakke: I'm not entirely convinced there isn't an easy argument that shows this doesn't break semantics, although this certainly isn't my expertise, and we should certainly make sure this is the case before this is ever implemented. Anyway, I was mainly answering your question from earlier about whether it would ever be useful.

src/foundation/whiskering-homotopies-concatenation.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+We implement the left whiskering operation of homotopies with respect to
+concatenation as an instance of a general left whiskering operation.
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2}
+  where
+
+  left-whisker-concat-htpy :
+    left-whiskering-operation ((x : A) → B x) _~_ _∙h_ _~_
```
fredrik-bakke, line 61:
```suggestion
    left-whiskering-operation ((x : A) → B x) (_~_) (_∙h_) (_~_)
```

src/foundation/whiskering-homotopies-concatenation.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+We implement the right whiskering operation of homotopies with respect to
+concatenation as an instance of a general right whiskering operation.
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2}
+  where
+
+  right-whisker-concat-htpy :
+    right-whiskering-operation ((x : A) → B x) _~_ _∙h_ _~_
```
fredrik-bakke, line 82:
```suggestion
    right-whiskering-operation ((x : A) → B x) (_~_) (_∙h_) (_~_)
```

src/foundation/whiskering-homotopies-concatenation.lagda.md
```diff
@@ -0,0 +1,157 @@
...
+of `H` and `K` is a homotopy `H ∙ I ~ H ∙ J`. In other words, left whiskering of
+homotopies with respect to concatenation is a
+[whiskering operation](foundation.whiskering-operations.md)
+
+```text
+  (H : f ~ g) {I J : g ~ h} → I ~ J → H ∙h I ~ H ∙h K.
+```
+
+Similarly, we introduce
+{{#concept 'right whiskering" Disambiguation="homotopies with respect to concatenation' Agda=right-whisker-concat-htpy}}
```
fredrik-bakke, line 35:
```suggestion
{{#concept r"ight whiskering" Disambiguation="homotopies with respect to concatenation" Agda=right-whisker-concat-htpy}}
```

src/foundation/whiskering-identifications-concatenation.lagda.md
```diff
@@ -0,0 +1,366 @@
...
+1. The square commutes by naturality of the homotopy
+   `α ↦ left-whisker-concat α β`.
+2. The transposed square commutes by the naturality of the homotopy
+   `β ↦ right-whisker-concat α β`.
+
+These two ways in which the square commutes are inverse to each other.
+
+**Note.** The following statements could have been formalized using
+[commuting squares of identifications](foundation.commuting-squares-of-identifications.md).
+However, in order to avoid cyclic module dependencies in the library we avoid
```
fredrik-bakke, line 219: 😓

src/orthogonal-factorization-systems/pullback-hom.lagda.md
fredrik-bakke: nice

src/foundation/function-extensionality.lagda.md
EgbertRijke: I tried and didn't get it to work. Perhaps I don't quite understand what you have in mind.

src/foundation/whiskering-identifications-concatenation.lagda.md
EgbertRijke: :)

src/foundation-core/function-extensionality.lagda.md
fredrik-bakke: You didn't delete this file yet?

src/foundation/function-extensionality.lagda.md
```diff
@@ -2,200 +2,276 @@
...
 
-Consider a map `f : A → B` and two dependent functions `g h : (x : B) → C x`.
-Then the square
+```text
+              htpy-eq
+    (f ＝ g) ----------> (f ~ g)
+           \            /
+  ap (ev a) \          / ev a
+             ∨        ∨
+            (f a ＝ g a)
```
fredrik-bakke, line 51: This diagram doesn't type check

fredrik-bakke: oh, wait, it does...

EgbertRijke:
I have addressed all the comments now. I defer all the future comments to future pull requests

EgbertRijke:
Also I am happy to share this PR with @fredrik-bakke, for his heroic efforts of reviewing this monster. Thank you!

fredrik-bakke:
Hooray! 🎉

PR #1015, opened 2024-01-31
Judgmentally involutive identity types

fredrik-bakke:


fredrik-bakke:
Turns out I could also make the concatenation operation on the judgmentally involutive identity types judgmentally left unital by being a little meticulous with definitions. 😄

EgbertRijke:
Note that we are both editing `foundation-core.identity-types` right now

fredrik-bakke:
Oh, apologies. Thanks for reminding me!

EgbertRijke:
It's ok. I can work your changes into mine if you finish before me. If I finish before you then you'll have to do it :)

EgbertRijke:
This pull request is going to be an amazing addition to the library, and you should also think about writing up a preprint with this stuff. This is real progress.

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,562 @@
...
+  compositional-eq-eq p z q = q ∙ᵣ p
+
+  eq-compositional-eq : x ＝ᶜ y → x ＝ y
+  eq-compositional-eq f = f x refl
+
+  is-section-eq-compositional-eq :
+    is-section compositional-eq-eq eq-compositional-eq
+  is-section-eq-compositional-eq f =
+    eq-multivariable-htpy 2
+      ( λ z p → inv (commutative-concatr-refl-Id-compositional-Id f p))
```
fredrik-bakke, line 158: If I could somehow get this proof to reduce to `refl` when `f = refl-compositional-Id`, then we would also get a definitional computation rule for the induction principle of the computational identity type and the yoneda identity type.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
...
+```
+
+The computational identity types are
+[equivalent](foundation-core.equivalences.md) to the standard identity types,
+but satisfy the following algebraic laws judgmentally:
+
+- `inv refl = refl`
+- `inv (inv p) = p`
+- `(p ∙ q) ∙ r = p ∙ (q ∙ r)`
+- `refl ∙ p = p` or `p ∙ refl = p`.
```
EgbertRijke, line 57: Would you mind using `\doteq` for judgmental equality, the same way I do it in my book?

The equality symbol `=` should mean propositional equality, and the fact that Agda reserves it for `:=` is not enough of a reason for me to start using it to mean judgmental equality.

There is already precedent for using `≐` in the library for judgmental equality in a couple of other files.

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -73,6 +78,11 @@ equiv-ev-refl' :
 equiv-ev-refl' a {B} =
   ( equiv-ev-refl a) ∘e
   ( equiv-Π-equiv-family (λ x → equiv-precomp-Π (equiv-inv a x) (B x)))
+
+is-equv-ev-refl' :
```
EgbertRijke, line 82: `is-equv` should be `is-equiv`

src/foundation-core/identity-types.lagda.md
```diff
@@ -93,6 +93,33 @@ The identity types form a weak groupoidal structure on types.
...
+3. We can define `_∙_` by induction on both `x ＝ y` and `y ＝ z`. This only
+   gives us the computation rule `refl ∙ refl = refl`.
+
+While the third option may be preferred by some for its symmetry, for practical
+reasons, we use the first alternative by convention.
+
+See also
+
+- [judgmentally right unital concatenation operation on indentifications](foundation.judgmentally-right-unital-concatenation-identifications.md)
+- [judgmentally compositional identity types](foundation.judgmentally-compositional-identity-types.md)
```
EgbertRijke, line 121: These other operations `concatr` and `concatl` are defined just using the identity type. It would make more sense to me to introduce them here, since you talk about them here anyway.

EgbertRijke: Nevermind, I just looked more closely at those files and they are actually really nice.

.vscode/agda.code-snippets
```diff
@@ -8,11 +8,27 @@
   "Full width equals sign (＝)": {
     "body": ["＝"],
     "description": "Full width equals sign",
-    "prefix": ["Id", "equals"]
```
EgbertRijke, line 11: If you're removing it here, can you also remove it from the `HOWTO-INSTALL` documentation?

.vscode/agda.code-snippets
```diff
@@ -8,11 +8,27 @@
...
+    "prefix": ["Id"]
   },
   "Yoneda embedding (ょ)": {
     "body": ["ょ"],
     "description": "Yoneda embedding",
     "prefix": ["yoneda"]
+  },
+
+  "Equational reasoning": {
```
EgbertRijke, line 19: What do these things do and are they worth documenting in the files that cover equality reasoning, homotopy reasoning, and equivalence reasoning? If you're adding new functionality then it should be documented somewhere, so that people can know about it.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
...
+open import foundation-core.torsorial-type-families
+```
+
+</details>
+
+## Idea
+
+The [standard definition of identity types](foundation-core.identity-types.md)
+suffer the limitation that many of the basic operations only satisfy algebraic
+laws _weakly_. In this file, we consider the
```
EgbertRijke, line 36:
```suggestion
The standard definition of [identity types](foundation-core.identity-types.md)
has the limitation that many of the basic operations only satisfy algebraic
laws _weakly_. In this file, we consider the
```

.vscode/agda.code-snippets
fredrik-bakke: I'd say this functionality is self-documenting. Whenever a vscode user starts writing "equational-reasoning", the editor will suggest inserting this pattern

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
...
+- `refl ∙ p = p` or `p ∙ refl = p`.
+
+**Note.** The computational identity types do not satisfy the judgmental laws
+
+- `refl ∙ p = p` and `p ∙ refl ＝ p` simultaneously,
+- `inv p ∙ p = refl`, or
+- `p ∙ inv p = refl`,
+
+and they do not have a judgmental computation property for their induction
+principle.
```
EgbertRijke, line 66: This comment makes me wonder whether "computational" is the right adjective for these identity types. These identity types satisfies some laws judgmentally, but the Martin-Löf identity types also satisfy some (other) laws judgmentally. In this sense they are both computational, so it is not clear why one should be called computational and the other not.

fredrik-bakke: I agree. I consider the current name temporary. How about "judgmentally algebraic"?

EgbertRijke: I should also say that this comment does not necessarily needs to be resolved before merging the PR. It's something to think about, but I'm happy to merge this PR without addressing it.

EgbertRijke: That doesn't have a right sound to it, because all the type constructors in type theory are (essentially) algebraic.

.vscode/agda.code-snippets
EgbertRijke: Ah ok. Then it is fine:)

src/foundation/computational-identity-types.lagda.md
EgbertRijke: Let's just keep it as it is for now, or we could leave a short note saying that the name is considered subject of discussion and why.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
+# The computational identity types
```
EgbertRijke, line 1:
```suggestion
# Computational identity types
```

For example, the file about identity types is also called "Identity types" and not "The identity types"

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
...
+      ( is-torsorial-Id x)
+
+  dependent-universal-property-identity-system-computational-Id :
+    dependent-universal-property-identity-system
+      ( computational-Id x)
+      ( refl-computational-Id)
+  dependent-universal-property-identity-system-computational-Id =
+    dependent-universal-property-identity-system-is-torsorial
+      ( refl-computational-Id)
+      ( is-torsorial-computational-Id)
```
EgbertRijke, line 184: I would give each of these a separate header. Headers are good:)

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: Algebraically reducing identity types?

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,518 @@
...
+```agda
+module _
+  {l : Level} {A : UU l}
+  where
+
+  infixl 15 _∙ₗʲ_
+  _∙ₗʲ_ : {x y z : A} → x ＝ʲ y → y ＝ʲ z → x ＝ʲ z
+  (z , p , q) ∙ₗʲ (z' , p' , q') = (z' , p' , q' ∙ʸ invr-yoneda-Id p ∙ʸ q)
+
+  concatl-computational-Id : {x y : A} → x ＝ʲ y → (z : A) → y ＝ʲ z → x ＝ʲ z
```
EgbertRijke, line 296: the name part "concatl" is slightly unusual in the sense that we usually don't attach letters to words in this way, but we would use new words and hyphenations.

It is acceptible for me if you use it here, but if we would start using this naming style more widely throughout the library I could see the names becoming unreadable very quickly.

src/foundation/computational-identity-types.lagda.md
EgbertRijke: I think I'm less fond of using the adjective algebraic.

Another alternative would be the Bakke identity type:)

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
+# The judgmentally compositional identity types
```
EgbertRijke, line 1:
```suggestion
# Judgmentally compositional identity types
```

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
...
+open import foundation-core.sections
+open import foundation-core.torsorial-type-families
+```
+
+</details>
+
+## Idea
+
+The [standard definition of identity types](foundation-core.identity-types.md)
+suffer the limitation that many of the basic operations only satisfy algebraic
```
EgbertRijke, line 32:
```suggestion
The standard definition of [identity types](foundation-core.identity-types.md)
has the limitation that many of the basic operations only satisfy algebraic
```

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
...
+laws _weakly_. In this file, we consider the
+{{#concept "judgmentally compositional identity types" Agda=yoneda-Id}}
```
EgbertRijke, line 34: An explanation of why this identity type is called `yoneda-Id` would be very fitting for the `## Idea` section.

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
...
+  distributive-inv-Id-yoneda-Id f g =
+    inv (inv-distributive-inv-Id-yoneda-Id f g)
+```
+
+### The judgmentally compositional identity types are equivalent to the standard identity types
+
+We define the equivalence `yoneda-eq-eq : x ＝ y → x ＝ʸ y` using the
+judgmentally right unital concatenation operation on identifications. While this
+makes the proof it is a section non-judgmental, it instead gives us the
+judgmental computation rules
```
EgbertRijke, line 132: This sentence is unclear. For example, I do not know what it means for a proof to be non-judgmental.

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
...
+and
+
+```text
+  eq-yoneda-eq refl-yoneda-Id ＝ refl.
+```
+
+The proof that it is a retraction requires the
+[function extensionality axiom](foundation.function-extensionality.md). However,
+function extensionality will become indispensable regardless when we move to
+proving miscellaneous algebraic laws of the type family.
```
EgbertRijke, line 147:
```suggestion
proving miscellaneous algebraic laws of the Yoneda identity type.
```

?

src/foundation/judgmentally-compositional-identity-types.lagda.md
```diff
@@ -0,0 +1,572 @@
...
+  equiv-yoneda-eq-eq : (x ＝ y) ≃ (x ＝ʸ y)
+  pr1 equiv-yoneda-eq-eq = yoneda-eq-eq
+  pr2 equiv-yoneda-eq-eq = is-equiv-yoneda-eq-eq
+
+  equiv-eq-yoneda-eq : (x ＝ʸ y) ≃ (x ＝ y)
+  pr1 equiv-eq-yoneda-eq = eq-yoneda-eq
+  pr2 equiv-eq-yoneda-eq = is-equiv-eq-yoneda-eq
+```
+
+The reflexivity witnesses are mapped to reflexivity witnesses judgmentally.
```
EgbertRijke, line 191: Would "reflexivity element" be an acceptable name for you?

fredrik-bakke: Yep

src/foundation/judgmentally-compositional-identity-types.lagda.md
fredrik-bakke: I haven't yet explained "yoneda". I'll do that

src/foundation/judgmentally-compositional-identity-types.lagda.md
fredrik-bakke: Good catch

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: Haha, I don't like that 😅

src/foundation/judgmentally-compositional-identity-types.lagda.md
EgbertRijke: It was mainly about what "the type family" refers to. I would also be happy with

```suggestion
proving miscellaneous algebraic laws of the judgmentally compositional identity type.
```

(unless "the type family" referred to something else of course)

EgbertRijke (approved):
This is a very nice pull request. Well done!

One side remark, perhaps for future work is that since you are combining several ideas in the computational identity types, it would make sense to develop these separate ideas at the level of identity systems. For example, you apply to idea of making an identity system strictly involutive to the Yoneda identity type, which is an identity system but it is not refactored in that way.

On the other hand, your constructions clearly work, and it is not guaranteed to work if you generalize everything to identity systems first. My remark could therefore be utter nonsense.

Your PR is very nice, and this is a very excellent addition to the library!

src/foundation/computational-identity-types.lagda.md
EgbertRijke: One naming scheme that is more `agda-unimath`-like would be using `left-strict-concat` for what you call `concatl` and `right-strict-concat` for what you call `concatr`. Then we would have `left-strict-inv` and `right-strict-inv` as well.

fredrik-bakke:
So, your remark is just partial nonsense. There are parts of the constructions that apply to every identity system, and there are parts that rely on the reductional behaviours of the standard identity type. Therefore, it would make sense to develop parallel files that apply to identity systems more broadly. In particular, you could get judgmentally involutional identity systems that are also judgmentally associative. But I don't believe they will satisfy `inv refl = refl` in general.

fredrik-bakke:
I just realized that while the induction principles don't compute generally, we can make them do it when the type family doesn't depend on the identity element!

EgbertRijke:
Can you also add your new files to the table about identity types?

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: I'm more inclined to just remove `invl` to be honest. It shouldn't be used because it's _just worse_ ™️. But your naming suggestion is otherwise good!

fredrik-bakke:
I believe this PR is ready for review now. Please don't merge if the prose is not perfect.

fredrik-bakke:
On second thought, I'm not such a big fan of using "strict" and "strictly" everywhere, because a common reading of "strictly XYZ" is "property XYZ holds and no other".

EgbertRijke:
> On second thought, I'm not such a big fan of using "strict" and "strictly" everywhere, because a common reading of "strictly XYZ" is "property XYZ holds and no other".

That is not a reading of strict that I find common. In fact, I think it is almost always false that the type of properties that holds for an object is contractible in the sense you mention above.

Besides, it is simply not true that we are using the adjective "strict" everywhere, or that this is suggested.

src/foundation-core/identity-types.lagda.md
```diff
@@ -139,6 +139,33 @@ table given above.
...
 
+The
+{{#concept "concatenation operation on identifications" Agda=_∙_ Agda=_∙'_ Agda=concat}}
+is a map
+
+```text
+  _∙_ : x ＝ y → y ＝ z → x ＝ z
+```
+
+for all `x y z : A`. However, there are essentially three different ways we can
```
EgbertRijke, line 150:
```suggestion
is a family of binary operations

```text
  _∙_ : x ＝ y → y ＝ z → x ＝ z
```

indexed by `x y z : A`. However, there are essentially three different ways we can
```

File: src/foundation-core/identity-types.lagda.md [hunk 19]
```diff
@@ -139,6 +139,33 @@ table given above.
...
+3. We can define `_∙_` by induction on both `x ＝ y` and `y ＝ z`. This only
+   gives us the computation rule `refl ∙ refl = refl`.
+
+While the third option may be preferred by some for its symmetry, for practical
+reasons, we use the first alternative by convention.
+
+See also
+
+- [The definitionally right unital concatenation operation on indentifications](foundation.definitionally-right-unital-concatenation-identifications.md)
+- [The yoneda identity types](foundation.yoneda-identity-types.md)
```
EgbertRijke, line 167: It is a little bit strange to have a separate "See also" section here. Wouldn't it be better to have it below in the actual `## See also` section?

At the very least, it would be better to make it a proper header, i.e., `#### See also`.

src/foundation-core/identity-types.lagda.md
```diff
@@ -139,6 +139,33 @@ table given above.
...
+for all `x y z : A`. However, there are essentially three different ways we can
+define concatenation of identifications, all with different computational
+behaviours.
+
+1. We can define concatenation by induction on the equality `x ＝ y`. This gives
+   us the computation rule `refl ∙ q = q`.
+2. We can define concatenation by induction on the equality `y ＝ z`. This gives
+   us the computation rule `p ∙ refl = p`.
+3. We can define `_∙_` by induction on both `x ＝ y` and `y ＝ z`. This only
+   gives us the computation rule `refl ∙ refl = refl`.
```
EgbertRijke, line 159: Use `\doteq` for any judgmental equalities

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+open import foundation-core.torsorial-type-families
+```
+
+</details>
+
+## Idea
+
+The standard definition of [identity types](foundation-core.identity-types.md)
+has the limitation that many of the basic operations only satisfy algebraic laws
+_weakly_. In this file, we consider the
```
EgbertRijke, line 41: Perhaps "in this file" should be "on this page", since the website is the main product of agda-unimath

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+{{#concept "computational identity types" Agda=computational-Id}} `x ＝ʲ y`.
+These are defined using the construction of the
+[strictly involutive identity types](foundation.strictly-involutive-identity-types.md):
```
EgbertRijke, line 44: I must say, I find the word "strict" very well-chosen here. I don't understand your objection.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+```
+
+but using the [yoneda identity types](foundation.yoneda-identity-types.md)
+(`_＝ʸ_`) as the underlying identity types:
+
+```text
+  (x ＝ʲ y) := Σ (z : A) ((z ＝ʸ y) × (z ＝ʸ x))
+```
+
+The yoneda identity types are defined as
```
EgbertRijke, line 57: Should Yoneda be capitalized (everywhere)?

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+
+```text
+  (x ＝ʸ y) := (z : A) → (z ＝ x) → (z ＝ y).
+```
+
+The yoneda identity types are equivalent to the standard identity types, but
```
EgbertRijke, line 63: hyperlink equivalent

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+- `inv reflʲ ≐ reflʲ`.
+
+While the last three equalities hold by the same computations as for the
+strictly involutive identity types using the fact that `inv reflʸ ≐ reflʸ`,
+strict associativity relies on the strict associativity of the underlying yoneda
+identity types. See the file about strictly involutive identity types for
+further details on these computations.
+
+In addition to these strict algebraic laws, we can also define a recursion
+principle for the computational identity types that computes judgmentally.
```
EgbertRijke, line 81: Instead of "we can also define" write "we also define" to be more assertive about what is being done in this file.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+
+  computational-Id : (x y : A) → UU l
+  computational-Id x y = Σ A (λ z → (z ＝ʸ y) × (z ＝ʸ x))
+
+  infix 6 _＝ʲ_
+  _＝ʲ_ : A → A → UU l
+  (a ＝ʲ b) = computational-Id a b
+
+  refl-computational-Id : {x : A} → x ＝ʲ x
+  refl-computational-Id {x} = (x , refl-yoneda-Id , refl-yoneda-Id)
```
EgbertRijke, line 109: Should we also introduce `reflʲ`, since you have the notation `_＝ʲ_` for this identity type?

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+```
+
+## Properties
+
+### The computational identity types are equivalent to the yoneda identity types
+
+Similarly to the strictly involutive identity types, this equivalence is a
+strict retraction and preserves the reflexivities strictly.
```
EgbertRijke, line 117: hyperlink retraction

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+
+  equiv-computational-eq-yoneda-eq : {x y : A} → (x ＝ʸ y) ≃ (x ＝ʲ y)
+  pr1 equiv-computational-eq-yoneda-eq = computational-eq-yoneda-eq
+  pr2 equiv-computational-eq-yoneda-eq = is-equiv-computational-eq-yoneda-eq
+
+  equiv-yoneda-eq-computational-eq : {x y : A} → (x ＝ʲ y) ≃ (x ＝ʸ y)
+  pr1 equiv-yoneda-eq-computational-eq = yoneda-eq-computational-eq
+  pr2 equiv-yoneda-eq-computational-eq = is-equiv-yoneda-eq-computational-eq
+```
+
```
EgbertRijke, line 174: A sentence explaining what will be shown in the following additional code block would be helpful.

EgbertRijke: Eg., "In the following code block we show that the equivalence from the Yoneda identity type to the computational identity type strictly preserves `refl` in both directions."

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+    {x : A} →
+    computational-eq-yoneda-eq (refl-yoneda-Id {x = x}) ＝ refl-computational-Id
+  preserves-refl-computational-eq-yoneda-eq = refl
+```
+
+### The computational identity types are equivalent to the standard identity types
+
+We define the equivalence as the composite `(x ＝ y) ≃ (x ＝ʸ y) ≃ (x ＝ʲ y)`.
+Since each of these equivalences preserves the groupoid structure so does the
+composite. For the same reason, it preserves the reflexivities strictly.
```
EgbertRijke, line 195: In one case you say "preserves" and in the other case you say "preserves strictly". If you mean that in both cases the preservation is strict, then it is better to say "preserves strictly" in both cases (to not create an apparent contrast).

EgbertRijke: Also, if in one case you really mean that it is nonstrict, then it is worthwhile to be even more explicit about it.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+  is-equiv-computational-eq-eq :
+    {x y : A} → is-equiv (computational-eq-eq {x} {y})
+  is-equiv-computational-eq-eq = is-equiv-map-equiv equiv-computational-eq-eq
+
+  is-equiv-eq-computational-eq :
+    {x y : A} → is-equiv (eq-computational-eq {x} {y})
+  is-equiv-eq-computational-eq = is-equiv-map-equiv equiv-eq-computational-eq
+```
+
+The reflexivity elements are preserved judgmentally.
```
EgbertRijke, line 239: We should probably stick to "preserves strictly" so that readers who are new to this matter are not put off by the different terminologies.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+- `(p ∙ʲ q) ∙ʲ r ≐ p ∙ʲ (q ∙ʲ r)`
+- `reflʲ ∙ʲ p ≐ p` or `p ∙ʲ reflʲ ≐ p`
+- `inv (inv p) ≐ p`
+- `inv reflʲ ≐ reflʲ`.
+
+### Inverting computational identifications
+
+The construction and computations are the same as for the strictly involutive
+identity types. The inversion operation is defined by swapping the positions of
+the two yoneda identifications
```
EgbertRijke, line 368: I like the terminology "Yoneda identification". Similarly, we say in the `## Idea` section above that elements of the computational identity type are called `{{#concept "computational identifications" Agda=computational-Id}}`

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+  preserves-inv-eq-computational-eq (z , f , g) =
+    ( ap (g y) (left-unit-concatr)) ∙
+    ( distributive-inv-Id-yoneda-Id g f) ∙
+    ( ap (λ r → inv (f x r)) (inv left-unit-concatr))
+```
+
+### The concatenation operations on computational identifications
+
+There is both a judgmentally left unital and a judgmentally right unital
+concatenation operation, while both are strictly associative.
```
EgbertRijke, line 419: You are mixing "judgmental" and "strict" in one sentence here. It is better to only use "strict"

EgbertRijke: Same for the rest of this text below.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+```agda
+module _
+  {l : Level} {A : UU l}
+  where
+
+  infixl 15 _∙ᵣʲ_
+  _∙ᵣʲ_ : {x y z : A} → x ＝ʲ y → y ＝ʲ z → x ＝ʲ z
+  (w , p , q) ∙ᵣʲ (w' , p' , q') = (w , p ∙ʸ inv-yoneda-Id q' ∙ʸ p' , q)
+
+  concatr-computational-Id : {x y : A} → x ＝ʲ y → (z : A) → y ＝ʲ z → x ＝ʲ z
```
EgbertRijke, line 512: I still like `right-strict-concat` better than `concatr`, and I'm worried that it unnecessarily sets a bad precedent to name things like this.

src/foundation/computational-identity-types.lagda.md
```diff
@@ -0,0 +1,800 @@
...
+  equiv-htpy-computational-eq = equiv-funext ∘e equiv-eq-computational-eq
+
+  equiv-computational-eq-htpy : (f ~ g) ≃ (f ＝ʲ g)
+  equiv-computational-eq-htpy = equiv-computational-eq-eq ∘e equiv-eq-htpy
+
+  funext-computational-Id : is-equiv htpy-computational-eq
+  funext-computational-Id = is-equiv-map-equiv equiv-htpy-computational-eq
+```
+
+### Univalence with respect to computational identifications
```
EgbertRijke, line 766: Question: Can it be shown, whithout using function extensionality that this form of univalence implies the usual form of univalence?

Note: the equivalence of `computational-Id` and `Id` uses function extensionality.

EgbertRijke: Similar questions can be asked about function extensionality, and whether this form of univalence implies function extensionality, or function extensionality wrt the computation identity type.

src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md
```diff
@@ -0,0 +1,178 @@
+# The definitionally right unital concatenation operation on identifications
```
EgbertRijke, line 1: I think we're mixing too many adjectives: "strict", "definitional", and "judgmental". This is definitely my biggest concern for this PR.

src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md
```diff
@@ -0,0 +1,178 @@
...
+
+The
+{{#concept "concatenation operation on identifications" Agda=_∙_ Agda=_∙'_ Agda=concat}}
+is a map
+
+```text
+  _∙_ : x ＝ y → y ＝ z → x ＝ z
+```
+
+for all `x y z : A`. However, there are essentially three different ways we can
```
EgbertRijke, line 28:
````suggestion
is family of binary operations

```text
  _∙_ : x ＝ y → y ＝ z → x ＝ z
```

indexed by `x y z : A`. However, there are essentially three different ways we can
````

File: src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md [hunk 37]
```diff
@@ -0,0 +1,178 @@
...
+define concatenation of identifications, all with different computational
+behaviours:
+
+1. We can define concatenation by induction on the equality `x ＝ y`. This gives
+   us the computation rule `refl ∙ q = q`.
+2. We can define concatenation by induction on the equality `y ＝ z`. This gives
+   us the computation rule `p ∙ refl = p`.
+3. We can define `_∙_` by induction on both `x ＝ y` and `y ＝ z`. This only
+   gives us the computation rule `refl ∙ refl = refl`.
```
EgbertRijke, line 37: Use `\doteq` for all judgmental equalities

src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md
```diff
@@ -0,0 +1,178 @@
...
+  p ∙ᵣ refl = p
+
+  concatr : {x y : A} → x ＝ y → (z : A) → y ＝ z → x ＝ z
+  concatr p z q = p ∙ᵣ q
+
+  concatr' : (x : A) {y z : A} → y ＝ z → x ＝ y → x ＝ z
+  concatr' x q p = p ∙ᵣ q
+```
+
+### Translating between the left and right unital versions of concatenation
```
EgbertRijke, line 66: Dropping the adjective strict/judgmental/definitional from this header makes it seem sloppy, because all concatenation operations are both left and right unital in a nonstrict sense.

src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md
```diff
@@ -0,0 +1,178 @@
...
+  is-injective-concatr refl s = (inv left-unit-concatr ∙ s) ∙ left-unit-concatr
+
+  is-injective-concatr' :
+    {x y z : A} (r : y ＝ z) {p q : x ＝ y} → p ∙ᵣ r ＝ q ∙ᵣ r → p ＝ q
+  is-injective-concatr' refl s = s
+```
+
+## See also
+
+- [Yoneda identity types](foundation.yoneda-identity-types.md)
```
EgbertRijke, line 178: And also the computational identity type?

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+where we use a superscript `i` to distinguish the strictly involutive identity
+type from the standard identity type.
+
+In addition, we maintain the following judgmental laws
+
+- `inv reflⁱ ≐ reflⁱ`
+- `ind-Id f reflⁱ ≐ f reflⁱ`
+- `reflⁱ ∙ p ≐ p` or `p ∙ reflⁱ ≐ p`
+
+among other more specific ones considered in this file.
```
EgbertRijke, line 57:
```suggestion
among other more specific laws considered in this file.
```

EgbertRijke: Perhaps also "on this page" instead of "in this file"

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+
+The standard definition of [identity types](foundation-core.identity-types.md)
+has the limitation that many of the basic operations only satisfy algebraic laws
+_weakly_. In this file, we consider the
+{{#concept "strictly involutive identity types" Agda=involutive-Id}}
+
+```text
+  (x ＝ⁱ y) := Σ (z : A) ((z ＝ y) × (z ＝ x))
+```
+
```
EgbertRijke, line 41: Perhaps we could also define elements of the strictly involutive identity types to be `{{#concept "strictly involutive identifications" ...}}`

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+    eq-involutive-eq (inv-involutive-Id p) ＝ inv (eq-involutive-eq p)
+  preserves-inv-eq-involutive-eq (z , p , q) =
+    ap (inv p ∙_) (inv (inv-inv q)) ∙ inv (distributive-inv-concat (inv q) p)
+```
+
+### Concatenation of strictly involutive identifications
+
+We have, practically speaking, two definitions of the concatenation operation on
+strictly involutive identity types. One satisfies a strict left unit law and the
+other satisfies a strict right unit law. In both cases, we must use the
```
EgbertRijke, line 290: Could you explain why we must use definitional right unital concatenation here? It is not immediately apparent to the reader (i.e., me).

EgbertRijke: Also note that if you change "definitional right unital concatenation" to something more preferrable, then you would have to change it here too.

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+[definitionally right unital concatenation operation on standard identifications](foundation.definitionally-right-unital-concatenation-identifications.md)
+`_∙ᵣ_`, to obtain this strict one-sided unit law.
+
+The judgmentally left unital concatenation operation is defined by
+
+```text
+  (w , p , q) ∙ⁱ (w' , p' , q') := (w' , p' , (q' ∙ᵣ inv p) ∙ᵣ q),
+```
+
+and the judgmentally right unital concatenation operation is defined by
```
EgbertRijke, line 300: Here you're calling it "judgmentally right unital"

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+    ( ap
+      ( inv q ∙_)
+      ( ( eq-double-concat-concatr-left-associated p (inv q') p') ∙
+        ( assoc p (inv q') p'))) ∙
+    ( inv (assoc (inv q) p (inv q' ∙ p')))
+```
+
+### The groupoidal laws for the strictly involutive identity types
+
+The general proof-strategy is to induct on the necessary identifications to make
```
EgbertRijke, line 401: proof strategy

(the hyphen has no place here)

EgbertRijke: What are "the necessary identifications"?

src/foundation/strictly-involutive-identity-types.lagda.md
```diff
@@ -0,0 +1,509 @@
...
+the left endpoints judgmentally equal, and then proceed by reasoning with the
+groupoid-laws of the underlying identity types.
```
EgbertRijke, line 403: groupoid laws

(no hyphen)

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+notation of the yoneda identity types.
+
+The yoneda identity types are [equivalent](foundation-core.equivalences.md) to
+the standard identity types, but satisfy judgmental laws
+
+- `(p ∙ʸ q) ∙ʸ r ≐ p ∙ʸ (q ∙ʸ r)`,
+- `reflʸ ∙ʸ p ≐ p`, and
+- `p ∙ʸ reflʸ ≐ p`.
+
+This is achieved by proxiyng to function composition and utilizing its
```
EgbertRijke, line 56: typo: proxying

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+
+  infix 6 _＝ʸ_
+  _＝ʸ_ : A → A → UU l
+  (a ＝ʸ b) = yoneda-Id a b
+```
+
+We define the reflexivity to be the identity function:
+
+```agda
+  refl-yoneda-Id : {x : A} → x ＝ʸ x
```
EgbertRijke, line 85: `reflʸ`?

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+
+and from right to left by evaluation at the reflexivity
+
+```text
+  eq-yoneda-eq := f ↦ f refl   : x ＝ʸ y → x ＝ y.
+```
+
+It should be noted that we define the map `x ＝ y → x ＝ʸ y` using the
+[definitionally right unital concatenation operation](foundation.definitionally-right-unital-concatenation-identifications.md).
+While this obstructs us from showing that the homotopy
```
EgbertRijke, line 163: hyperlink homotopy

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+    {x : A} → yoneda-eq-eq (refl {x = x}) ＝ refl-yoneda-Id
+  preserves-refl-yoneda-eq-eq = refl
+
+  preserves-refl-eq-yoneda-eq :
+    {x : A} → eq-yoneda-eq (refl-yoneda-Id {x = x}) ＝ refl
+  preserves-refl-eq-yoneda-eq = refl
+```
+
+Below, we consider the alternative definition of `yoneda-eq-eq` using the
+definitionally left unital concatenation operation on standard identity types.
```
EgbertRijke, line 262: This is, I believe, the first time that you use "definitionally left unital". All the other instances were "judgmental"

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+As we can see, the retracting homotopy holds judgmentally, but now
```
EgbertRijke, line 263: hyperlink retracting

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+`yoneda-eq-eq refl` does not compute definitionally to `reflʸ`.
```
EgbertRijke, line 264: "judgmentally" and "definitionally" are used mixed here, which I believe could be a source of confusion for some readers.

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+  ind-yoneda-Id :
+    (b : B x refl-yoneda-Id) {y : A} (f : x ＝ʸ y) → B y f
+  ind-yoneda-Id b {y} f =
+    tr
+      ( B y)
+      ( is-section-eq-yoneda-eq f)
+      ( ind-Id x (λ y p → B y (yoneda-eq-eq p)) b y (eq-yoneda-eq f))
+```
+
+While the induction principle does not have the wanted reduction behaviour, the
```
EgbertRijke, line 386: "desired" instead of "wanted"

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+non-dependent eliminator does. This is simply because we no longer need to
```
EgbertRijke, line 387: nondependent

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+transport along `is-section-eq-yoneda-eq`.
```
EgbertRijke, line 388: hyperlink transport

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+We consider two ways of defining the inversion operation on yoneda
+identifications: by the definitionally right unital, and definitionally left
+unital concatenation operation on the underlying identity type respectively. The
+former enjoys the computational property
+
+```text
+  inv reflʸ ≐ reflʸ,
+```
+
+hence will be preferred going.
```
EgbertRijke, line 431: the phrase "will be preferred going" reads like a broken sentence to me.

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+    {x y : A} (f : x ＝ʸ y) →
+    inv-yoneda-Id (inv-yoneda-Id f) ＝ f
+  inv-inv-yoneda-Id {x} f =
+    eq-multivariable-htpy 2
+      ( λ _ p →
+        ( ap (p ∙ᵣ_) (ap inv left-unit-concatr ∙ inv-inv (f x refl))) ∙
+        ( inv (commutative-preconcatr-refl-Id-yoneda-Id f p)))
+```
+
+```agda
```
EgbertRijke, line 458: A sentence explaining what the next code block does would be helpful

EgbertRijke (approved):
This is an impressive pull request. It is very well written and well-documented. I went into more detail than usual with my feedback, because you specifically mentioned that you were shooting for perfection. Otherwise I would not have gone into such annoying detail.

My biggest complaint is that you use the words "definitional", "judgmental", and "strict" interchangeably. Note that the words "definitional" and "judgmental" are not considered synonymous. "Definitional equality" is equality that holds by definition, and judgmental equality is a notion of equality that encompasses judgmental equality, but also includes other strict equalities such as rewrite rules.

I would drop your extensive use of "definitionally unital" and or "judgmentally unital", which is used inconsistently, in favor of "strictly unital", which is also in line with your use of "strictly associative". You had some concerns about the word strict, but it is widely understood what is meant by "strict" in this setting, while your use of "definitional" in some places is widely understood to be inaccurate.

Other than that, I applaud you for finding these excellent definitions, and I am very grateful that you are adding them to the library. This is very awesome work!

src/foundation/definitionally-right-unital-concatenation-identifications.lagda.md
fredrik-bakke: Hmm, I guess!

src/foundation/strictly-involutive-identity-types.lagda.md
fredrik-bakke: I used "definitionally" specifically for the operation whose definition is that it is right unital. But don't worry, I will change it. I originally used "judgmental" everywhere, but I started experimenting after you agreed "strict" was better

src/foundation/strictly-involutive-identity-types.lagda.md
fredrik-bakke: The "necessary identifications" refer to the minimal number of identifications we must pattern-match on to make the first component of two strictly involutive identifications judgmentally equal.

fredrik-bakke: This is what it literally says, isn't it?

src/foundation/yoneda-identity-types.lagda.md
fredrik-bakke: okay :)

src/foundation/yoneda-identity-types.lagda.md
```diff
@@ -0,0 +1,775 @@
...
+  yoneda-eq-eq p z q = q ∙ᵣ p
+
+  eq-yoneda-eq : x ＝ʸ y → x ＝ y
+  eq-yoneda-eq f = f x refl
+```
+
+The construction of the homotopy `yoneda-eq-eq ∘ eq-yoneda-eq ~ id` requires the
+[function extensionality axiom](foundation.function-extensionality.md). And
+while we could show an analogous induction principle of the yoneda identity
+types without the use of this axiom, function extensionality will become
```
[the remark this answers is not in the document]
fredrik-bakke: It's already hyperlinked there

EgbertRijke:
Note that if my feedback is "too much" then I'm happy to be more lenient, because I am actually very happy with this pull request. It's just that you asked for the text to be "perfect" so I gave it a thorough read-through.

src/foundation/strictly-involutive-identity-types.lagda.md
EgbertRijke: The reader is asked to do that in their heads, without seeing those identifications displayed. Instead of "the necessary implications" it would be better to display the identifications you're doing induction on.

fredrik-bakke:
I don't find your review too meticulous at all! Thanks for the added attention. It's much appreciated :)

src/foundation-core/identity-types.lagda.md
fredrik-bakke: The links are only relevant to the present discussion, which is why I placed it there. I'll go with the `#### See also` :)

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: Yes, I like this suggestion

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: well, it preserves the units strictly, and sometimes it preserves some of the other structure strictly as well. Thus I don't find it fitting to state it is not strict. Do you propose something else?

EgbertRijke: My proposal is that you write very explicitly about what you're thinking of when you're omitting the word strictly, if you are omitting it intentionally. Right now it is not clear what you have in mind by omitting the word strictly.

src/foundation/strictly-involutive-identity-types.lagda.md
fredrik-bakke: While I agree with this suggestion, it creates a discrepancy with how we refer to other pages. We usually write ``in [`foundation.univalence`](...)...``, but it doesn't make sense to say "in a page".

EgbertRijke: Ok, I won't hold this PR up over this

src/foundation/strictly-involutive-identity-types.lagda.md
fredrik-bakke: Which identifications I'm doing induction on depends on the lemma

fredrik-bakke: I don't understand exactly what you want me to write here. Perhaps you can look if the new version is satisfactory?

src/foundation/computational-identity-types.lagda.md
fredrik-bakke: Okay, so I wound up writing "We define the equivalence as the composite `(x ＝ y) ≃ (x ＝ʸ y) ≃ (x ＝ʲ y)`.
Since each of these equivalences preserve the groupoid structure weakly so does
the composite. For the same reason, it preserves the reflexivities strictly." Do you agree this is clear enough?

src/foundation/strictly-involutive-identity-types.lagda.md
EgbertRijke: Ok, I will look. I will also not hold up the PR over this issue, so don't sweat it too much. Indeed I was not aware that it depends on the lemma.

fredrik-bakke:
I've addressed all of your comments to the best of my ability now.

PR #1017, opened 2024-02-04
Rename `(co)prod` to `(co)product`

fredrik-bakke:


EgbertRijke:
I added a new label "improve naming" because this PR is a quick global search-and-replace that improves our naming. I thought such a label would be more descriptive than "refactoring", because refactoring suggests that you actually factor out steps or clean up to make the code simpler. I'd be happy to put the label back if you think it is more fitting.

BTW I am a bit afraid of quick search-and-replace changes since they are (1) so quick to make but (2) easily jinx other people's work. But I think this one is not actually causing so much trouble.

fredrik-bakke:
There was one naming conflict that came up when making this refactoring. Namely, `prod-Precategory` referred to the product of two precategories, while `product-Precategory` referred to the product of two objects in a precategory. I consider this bad practice, and propose that we refer to the latter as `product-obj-Precategory` instead, suggesting a general scheme such as, for instance, `pullback-hom-Precategory` for a pullback of a morphism in a precategory. I thus renamed
- `pullback-Precategory` to `pullback-obj-Precategory`, 
- `coproduct-Precategory` to `coproduct-obj-Precategory`,
- `exponential-Precategory` to `exponential-obj-Precategory`.

I consider this PR ready for review again. I can take a quick look over it first during my layover if you prefer, as this is a quick search-and-replace job as you suggested.

fredrik-bakke:
(This also suggests that we may want to change our naming scheme for category theory modules as well. Something I suggest we have a look at when we start doing wild category theory soon.)

EgbertRijke:
> There was one naming conflict that came up when making this refactoring. Namely, `prod-Precategory` referred to the product of two precategories, while `product-Precategory` referred to the product of two objects in a precategory. I consider this bad practice, and propose that we refer to the latter as `product-obj-Precategory` instead, suggesting a general scheme such as, for instance, `pullback-hom-Precategory` for a pullback of a morphism in a precategory. I thus renamed
> 
> * `pullback-Precategory` to `pullback-obj-Precategory`,
> * `coproduct-Precategory` to `coproduct-obj-Precategory`,
> * `exponential-Precategory` to `exponential-obj-Precategory`.
> 
> I consider this PR ready for review again. I can take a quick look over it first during my layover if you prefer, as this is a quick search-and-replace job as you suggested.

Are you sure that it won't be confusing if `pullback-hom-Precategory` should mean a pullback, and not a pullback-hom?

fredrik-bakke:
We will have to disambiguate that special case somehow. Otherwise, do you disagree it is a solid scheme?

EgbertRijke:
If it's ok with you I will wait with merging this PR until #1014 is merged, because I don't look forward to hours of resolving merge conflicts whereas a much quicker solution would be that you just do your `search-and-replace` again if there are any problems.

fredrik-bakke:
> If it's ok with you I will wait with merging this PR until #1014 is merged, because I don't look forward to hours of resolving merge conflicts whereas a much quicker solution would be that you just do your `search-and-replace` again if there are any problems.

Sure thing!

EgbertRijke:
> We will have to disambiguate that special case somehow. Otherwise, do you disagree it is a solid scheme?

The scheme with `pullback-obj-Precategory` has no obvious problems, so I agree.

fredrik-bakke (commented):
I had a look at about 50 of the files. Looks good to merge to me

PR #1018, opened 2024-02-04
Definition of monads on precategories and categories (+ of whiskering and horizontal composition)

GregorPercic:
Note: I wanted to utilize implicit parameters more, but when I used them, Agda colored some text as yellow and weird and very long terms began appearing in the bottom "Agda" section of the screen, though they did not register as errors, and type checking also passed. So I decided to explicate all parameters, even the implicit ones.

Update: after adding explanation of the defined concepts in Markdown, there might be a problem with a link. I included this link to the section Whiskering:

```
[whiskering](category-theory.natural-transformations-functors-precategories.md#whiskering)
```

I don't know if this link format is supported.

### DIFF
--- a/src/category-theory/monads-on-categories.lagda.md
+++ b/src/category-theory/monads-on-categories.lagda.md
@@ -8,8 +8,8 @@ module category-theory.monads-on-categories where
 
 ```agda
 open import category-theory.categories
-open import category-theory.precategories
 open import category-theory.monads-on-precategories
+open import category-theory.precategories
 
 open import foundation.universe-levels
 ```
--- a/src/category-theory/monads-on-categories.lagda.md
+++ b/src/category-theory/monads-on-categories.lagda.md
@@ -1,0 +1,30 @@
+# Monads on categories
+
+```agda
+module category-theory.monads-on-categories where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import category-theory.categories
+open import category-theory.precategories
+open import category-theory.monads-on-precategories
+
+open import foundation.universe-levels
+```
+
+</details>
+
+## Definitions
+
+### The type of monads on categories
+
+```agda
+module _
+  {l : Level} (C : Category l l)
+  where
+
+  monad-Category : UU l
+  monad-Category = monad-Precategory l (precategory-Category C)
+```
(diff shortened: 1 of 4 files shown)


src/category-theory/monads-on-categories.lagda.md
```diff
@@ -0,0 +1,30 @@
...
+open import category-theory.categories
+open import category-theory.precategories
+open import category-theory.monads-on-precategories
+
+open import foundation.universe-levels
+```
+
+</details>
+
+## Definitions
```
EgbertRijke, line 19: Would you mind writing an `## Idea` section with the "textbook" definition of a monad?

src/category-theory/natural-transformations-functors-precategories.lagda.md
```diff
@@ -256,3 +259,95 @@ module _
       ( map-functor-Precategory C D H)
       ( map-functor-Precategory C D I)
 ```
+
+## Whiskering and horizontal composition
```
EgbertRijke, line 263: This can be split up in two sections, one about whiskering and one about horizontal compositions. In each of these sections, it would be useful to have a "textbook" description of the definitions of whiskering and horizontal composition.

src/category-theory/monads-on-categories.lagda.md
GregorPercic: OK, will do.

src/category-theory/natural-transformations-functors-precategories.lagda.md
GregorPercic: OK, will do.

src/category-theory/monads-on-categories.lagda.md
GregorPercic: Do we want to have the textbook definition written in LaTeX or just plain text (Unicode)? I can do either.

EgbertRijke: We usually use plain text, unicode, and ascii diagrams. Have a look in some of the other files to see how it is done

EgbertRijke: Thank you so much for doing this btw! We like it if our library is informative, so this is a great help!

EgbertRijke:
Awesome! Thank you so much for adding the informal explanations.

Great first pull request!

PR #1022, opened 2024-02-06
Deduplicate definitions

### LINKED ISSUE #992: Resolve duplicate definitions
We aim to have unique identifiers for all public definitions throughout the codebase, as stated in the library coding style guide. This was supposed to be enforced by recursively opening all modules of the library, so that Agda would complain about duplicate definitions when checking the Everything file.

Unfortunately we aren't properly opening the top-level modules, so while we avoided naming collisions inside top-level modules, we managed to introduce duplicates across top-level module boundaries. These should be resolved by either removing one of the definitions, and reusing the other, or by renaming one or both definitions to something less ambiguous.

So far I found the following duplicates:
- [x] Theory of finite groups is duplicated between `finite-algebra` and `finite-group-theory`. The `finite-algebra` module should build on top of the formalization from `finite-group-theory`
- [x] `foundation.separated-types.is-separated` vs `orthogonal-factorization-systems.separated-types.is-separated`
- [x] `univalent-combinatorics.standard-finite-types.Fin` vs `reflection.group-solver.Fin`
- [x] `universal-algebra.terms-over-signatures.Term` vs `reflections.terms.Term`
- [x] `foundation.path-algebra.cube` vs `univalent-combinatorics.cubes.cube`

The Everything file is generated by a script in the Makefile, so once the duplicates are resolved, we should change the script to generate `open import <module> public` lines.

fredrik-bakke:
Resolves #992.


fredrik-bakke:
gotta fix the rest of the names in `reflection` tomorrow.

fredrik-bakke:
Turns out constructor names can be overloaded. I.e. two different inductive types can have constructors with the same name. Did you guys know this?

EgbertRijke:
I did not know this, but I don't want us to exploit this in agda-unimath

EgbertRijke:
Perhaps for records I did know this

fredrik-bakke:
agreed, I was just a bit shocked to find out

fredrik-bakke:
For context, constructor overloading was used in `reflection`. I'm working to disambiguate them now.

EgbertRijke:
Awesome! Thank you for taking care of that part of the library

fredrik-bakke:
Okay, this one is ready for review now. Note that it does not do a complete refactor of `reflection`, although it improves on many of its bad practices. I did not want to touch `group-solver` in the end because it is such a hot mess.

src/finite-algebra/semisimple-commutative-finite-rings.lagda.md
```diff
@@ -85,12 +85,17 @@ module _
...
-    compute-structure-commutative-ring-𝔽 X p
-  pr2 (compute-structure-semisimple-commutative-ring-𝔽 (p , s)) = s
+  pr1
+    ( semisimple-commutative-ring-structure-semisimple-commutative-ring-𝔽
+      ( p , s)) =
+    finite-commutative-ring-structure-commutative-ring-𝔽 X p
+  pr2
+    ( semisimple-commutative-ring-structure-semisimple-commutative-ring-𝔽
+      ( p , s)) =
+    s
```
EgbertRijke, line 100: I feel like in such cases defining them as a pair is better for readability

src/finite-group-theory/finite-groups.lagda.md
```diff
@@ -282,6 +305,22 @@ module _
...
+finite-group-is-finite-Group :
+  {l : Level} → (G : Group l) → is-finite (type-Group G) → Group-𝔽 l
+pr1 (finite-group-is-finite-Group G f) =
+  finite-semigroup-is-finite-Semigroup (semigroup-Group G) f
+pr2 (finite-group-is-finite-Group G f) = is-group-Group G
+
+module _
+  {l : Level} (G : Group-𝔽 l)
+  where
+  commute-Group-𝔽 : type-Group-𝔽 G → type-Group-𝔽 G → UU l
```
fredrik-bakke, line 318:
```suggestion
  where

  commute-Group-𝔽 : type-Group-𝔽 G → type-Group-𝔽 G → UU l
```

src/finite-algebra/semisimple-commutative-finite-rings.lagda.md
EgbertRijke: Another way that could make such copattern matching definitions more readable is to 
1.  _not_ pattern match on the pair `(p , s)`
2.  assume a variable `x` of the same type as `(p , s)` in the module
3. Define the components of this entry separately, and
4. use named projections

EgbertRijke: And in this case a third approach that would reduce to the same definition would be to use `map-Sigma-map-base`

EgbertRijke (approved):
Thank you for making this PR. I think I had only one small comment, so we can merge it soon.

PR #1025, opened 2024-02-08
Strict symmetrizations of binary relations

fredrik-bakke:
Defines the strict symmetrization of a binary relation, and proves some basic properties of these. Among these,
- the construction has a unit `R -> S(R)` if `R` is reflexive
- the construction has a counit `S(R) -> R` if `R` has extensions.

This construction is useful in defining certain coherences for wild categories.


fredrik-bakke:
Just for clarity, I do want to write some good explainers for the files I'm developing here before this PR is merged, but my first priority is to find the right definitions.

fredrik-bakke:
I'm adding a definition of "wide displayed large reflexive graphs", but I do not like the terminology usage "wide" here. I think "wide" should mean "equipped with a section of the family of vertices", but currently I take it to mean "contractible family of vertices". Would love naming suggestions. Basically, what I use it as is a displayed reflexive graph that only adds structure on the edges and keeps the same vertices.

fredrik-bakke:
If we want to formalize [this nice result](https://www.cs.bham.ac.uk/~mhe/TypeTopology/WildCategories.Cones.html) about wild categories due to Jon Sterling and Mike Shulman for 1-coherent wild categories, we better require the wild identity relation to have an action on wild identifications.

fredrik-bakke:
I'm marking this one as ready for review.

src/foundation/reflexive-relations.lagda.md
```diff
@@ -1,38 +0,0 @@
-# Reflexive relations
```
EgbertRijke, line 1: Where did the contents of this file go?

src/foundation/strict-symmetrization-binary-relations.lagda.md
```diff
@@ -0,0 +1,185 @@
...
+[outer horn filler condition](foundation.outer-2-horn-filler-conditions-binary-relations.md),
+then it has a counit map `Rˢ → R`.
+
+An essential fact about the strict symmetrization of a relation is that the
+strict symmetrization of an identity relation is equivalent to the identity
+relation.
+
+**Warning.** The strict symmetrization is not the symmetric closure in general.
+For instance, if the underlying relation has an initial element then the strict
+symmetrization will be reflexive, while the symmetric closure need not be.
```
EgbertRijke, line 46: It has not been explained what an initial element is in this context, so I suggest adding a short explanation here:

```suggestion
**Warning.** The strict symmetrization is not the symmetric closure in general.
For instance, if the underlying relation has an initial element, i.e., there is an element `a` such that `R a x` is [contractible](foundation-core.contractible-types.md) for every `x`, then the strict
symmetrization will be reflexive, while the symmetric closure need not be.
```

Is contractibility really needed here?

src/foundation/strict-symmetrization-binary-relations.lagda.md
```diff
@@ -0,0 +1,185 @@
...
+`r : Rˢ x y`, we have a symmetry operation `sym r : Rˢ y x` such that
+
+```text
+  sym (sym r) ≐ r.
+```
+
+If the underlying binary relation is reflexive, then this construction has a
+unit map `R → Rˢ`. If the binary relation satisfies an
+[outer horn filler condition](foundation.outer-2-horn-filler-conditions-binary-relations.md),
+then it has a counit map `Rˢ → R`.
```
EgbertRijke, line 38: You are using the terminology of adjunctions here, but whether we actually have an adjunction is not explained. Could you say something about this?

I think if there is no adjunction then it is definitely worth remarking that while you're borrowing terminology from adjunctions there is no adjunction, so that readers are not lead astray. Obviously, if there is an adjunction that is worth remarking too.

src/foundation/strict-symmetrization-binary-relations.lagda.md
```diff
@@ -0,0 +1,185 @@
...
+
+An essential fact about the strict symmetrization of a relation is that the
+strict symmetrization of an identity relation is equivalent to the identity
+relation.
```
EgbertRijke, line 42: It seems that you are talking about the strictly involutive identity types here, without saying so. I think it would be beneficial to the reader if you name what you're talking about here, and provide a link.

EgbertRijke (commented):
Nice pull request! I just had a few minor comments.

src/foundation/strict-symmetrization-binary-relations.lagda.md
fredrik-bakke: Nope, but the conceptual thing requires contractibility. It is supposed to be a demonstrative example

src/foundation/strict-symmetrization-binary-relations.lagda.md
fredrik-bakke: I will have to think a little about whether there is some sort of adjunction.

src/category-theory/large-precategories.lagda.md
```diff
@@ -45,10 +46,10 @@ record
...
 
   hom-Large-Precategory :
-    {l1 l2 : Level}
-    (X : obj-Large-Precategory l1)
-    (Y : obj-Large-Precategory l2) →
-    UU (β l1 l2)
+      {l1 l2 : Level} →
+      obj-Large-Precategory l1 →
+      obj-Large-Precategory l2 →
+      UU (β l1 l2)
```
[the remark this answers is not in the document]
fredrik-bakke: It was not. Thanks for spotting it!

src/foundation/reflexive-relations.lagda.md
fredrik-bakke: Aha, I initially deleted it, but i was about to put it back in

fredrik-bakke:
By the way. We have three conflicting notions of "discreteness" in the library:

1. A _discrete type_ is a type with decidable equality.
2. A _discrete relation_ is a relation that is equivalent to the identity relation.
3. A _discrete precategory_ is the precategory defined by the identity types of a set.


If we change a discrete precategory to be the precategory defined by the identity types of a 1-type, then item 2 and 3 would agree, but that still leaves item 1 as a distinct notion of discreteness.

src/foundation/strict-symmetrization-binary-relations.lagda.md
fredrik-bakke: I added a comment stating that we're not implying there's an adjunction. Hopefully, that is clear enough for now.

fredrik-bakke:
I've resolved your comments now

src/foundation/outer-2-horn-filler-conditions-binary-relations.lagda.md
```diff
@@ -0,0 +1,175 @@
...
+  {l1 l2 : Level} {A : UU l1} (R : Relation l2 A)
+  where
+
+  has-extensions-Relation : UU (l1 ⊔ l2)
+  has-extensions-Relation = {x y z : A} → R x y → R x z → R y z
+```
+
+## Properties
+
+### If there is an element that relates to `y` then `y` relates to `y`
```
fredrik-bakke, line 115:
```suggestion
### If there is an element that relates to `y` and the relation has extensions, then `y` relates to `y`
```

fredrik-bakke:
I improved a bit on how the concepts of the PR are organized by splitting `outer-2-horn-filler-conditions-binary-relations` into `binary-relations-with-extensions` and `binary-relations-with-lifts`, and I also added a remark about the dual strict symmetrization, although I did not create a separate file for it.

fredrik-bakke:
I'm merging this PR, any lingering questions can be reconsidered again later.

PR #1046, opened 2024-02-28
Self-host GoatCounter script

fredrik-bakke:
Makes us self-host the GoatCounter script. Perhaps this will avoid some more tracker blockers. ~The hosted script is a minified version of the original one.~

Follow-up to #1035.


fredrik-bakke:
I'm having trouble building the website locally due to a new version of Python 🤦‍♂️

VojtechStep:
If we're pretty-printing the mangled script then maybe just include it in the original form. Also the ISC license notice from the original file has to be included at the top of the file.

fredrik-bakke:
Oh, shoot, I thought I had pushed the change

VojtechStep (approved):
LGTM

PR #1054, opened 2024-03-02
Associativity of pullbacks

fredrik-bakke:
Proves $(A ×_X B) ×_Y C ≃ A ×_X (B ×_Y C)$ for standard and non-standard pullbacks.

### DIFF
--- a/src/foundation-core/pullbacks.lagda.md
+++ b/src/foundation-core/pullbacks.lagda.md
@@ -384,12 +384,12 @@ module _
   where
 
   abstract
-    is-pullback-top-is-pullback-rectangle :
+    is-pullback-top-square-is-pullback-rectangle :
       (c : cone f g B) (d : cone (horizontal-map-cone f g c) h A) →
       is-pullback f g c →
       is-pullback f (g ∘ h) (pasting-vertical-cone f g h c d) →
       is-pullback (horizontal-map-cone f g c) h d
-    is-pullback-top-is-pullback-rectangle c d is-pb-c is-pb-dc =
+    is-pullback-top-square-is-pullback-rectangle c d is-pb-c is-pb-dc =
       is-pullback-is-fiberwise-equiv-map-fiber-vertical-map-cone
         ( horizontal-map-cone f g c)
         ( h)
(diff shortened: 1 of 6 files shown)


src/foundation-core/pullbacks.lagda.md
```diff
@@ -502,6 +502,101 @@ module _
...
+```
+
+and with pullback cones `α` and `β` respectively. Moreover, consider a cone `γ`
+over the pullbacks as in the following diagram
+
+```text
+  ∙ ------------> ∙ ------------> C
+  |               | ⌟             |
+  |       γ       |       β       | i
+  |               ∨               ∨
```
EgbertRijke, line 521: Should this be

```suggestion
  ∨               ∨               ∨
```

?

It looks like your central dots are not terminal objects here, and that this map is not an identifications.

src/foundation/standard-pullbacks.lagda.md
```diff
@@ -271,6 +271,188 @@ triangle-map-commutative-standard-pullback :
...
+
+```text
+      f       g       g       h
+  A ----> X <---- B ----> X <---- C.
+```
+
+- See also the following relevant stack exchange question:
+  [Assocaitivity of pullbacks](https://math.stackexchange.com/questions/2046276/associativity-of-pullbacks).
+
+#### The standard ternary pullback
```
EgbertRijke, line 340: We could place this under `## Definitions`. I don't think people looking for the standard ternary pullback would necessarily be looking for it here.

EgbertRijke (approved):
Great pull request!

src/foundation-core/pullbacks.lagda.md
fredrik-bakke: Thanks! The central dots are meant to be unnamed vertices. Do you find the style confusing, or was the comment mainly to justify your question?

src/foundation/standard-pullbacks.lagda.md
fredrik-bakke: Hmm. In my mind it fits better here. I see it as similar to how we define types of homotopies where we also characterize the identity types, as opposed to as an unmotivated definition in the `## Definitions` section

fredrik-bakke: Never mind, I changed my mind 😅

fredrik-bakke: One could probably argue that the standard ternary pullback should be in a new file, but then associativity will be kinda out of place

src/foundation-core/pullbacks.lagda.md
EgbertRijke: No, the style is not confusing. Perhaps you could draw an extra pullback-corner in the top-left square just to be more explicit

fredrik-bakke: But `γ` is not assumed to be a pullback

EgbertRijke: ah, ok.

fredrik-bakke:
Seeing as the conversations have been resolved, I'm merging this pullback. Thanks for the review!

PR #1063, opened 2024-03-10
Add link to benchmarks in GitHub README

fredrik-bakke:


VojtechStep:
I think I'd rather advertise a public link once the website is hosted somewhere else than on my personal Netlify account, but migrating it isn't very high on my list of priorities.

fredrik-bakke:
It is too bad if end up never sharing our benchmarking data simply because it is hosted in the wrong place. Therefore I suggest we share the link to the current hosting location until it is changed.

VojtechStep:
I don't have a strong opinion here, but a part of why migrating isn't a high priority for me is that I don't feel the need to show this data publicly. But if you do, then I guess feel free to merge this.

PR #1071, opened 2024-03-13
Let being an isomorphism be a property

EgbertRijke:
In this pull request I am testing out whether it is sensible to let isomorphisms be a property in a uniformly defined manner across the library.

Note: these changes are so far experimental, and I am unsure at the moment whether I actually want to propose to merge this line of thought.

### DIFF
--- a/src/category-theory/retractions-in-categories.lagda.md
+++ b/src/category-theory/retractions-in-categories.lagda.md
@@ -1,0 +1,67 @@
+# Retractions in categories
+
+```agda
+module category-theory.retractions-in-categories where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import category-theory.categories
+open import category-theory.retractions-in-precategories
+
+open import foundation.dependent-pair-types
+open import foundation.identity-types
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Consider a morphism `f : A → B` in a [category](category-theory.categories.md) `𝒞`. A {{#concept "retraction" Disambiguation="morphism in a category" Agda=retraction-hom-Category}} of `f` consists of a morphism `g : B → A` equipped with an [identification](foundation-core.identifications.md)
+
+```text
+  g ∘ f ＝ id.
+```
+
+## Definitions
+
+### The predicate on morphisms in a category of being a retraction
+
+```agda
+module _
+  {l1 l2 : Level} (𝒞 : Category l1 l2)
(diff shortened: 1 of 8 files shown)


fredrik-bakke:
Hey! Although I don't know what your precise motivation is, I'm curious wouldn't a better setting be to consider this in a wild category without a class of equivalences? Then the framework will apply to structures that are not set-level, and we will additionally be able to avoid function extensionality (because of the setoid relation). To me, that sounds like a better definition.

fredrik-bakke:
Of course, `is-prop-iso` will still rely on extensionality principles of the underlying type of morphisms, but many other constructions won't.

src/category-theory/isomorphisms-in-precategories.lagda.md
```diff
@@ -22,56 +24,119 @@ open import foundation.retractions
...
-      ( comp-hom-Precategory C f g ＝ id-hom-Precategory C) ×
-      ( comp-hom-Precategory C g f ＝ id-hom-Precategory C))
+module _
+  {l1 l2 : Level} (C : Precategory l1 l2)
+  {x y : obj-Precategory C} (f : hom-Precategory C x y)
+  where
+  
+  is-iso-Precategory : UU l2
+  is-iso-Precategory =
+    section-hom-Precategory C f × retraction-hom-Precategory C f
```
fredrik-bakke, line 63: Considering that morphism types in a category are sets, I'd propose we rather use the following definition

``agda
Σ ( hom-Precategory C y x)
    ( λ g →
      ( comp-hom-Precategory C f g ＝ⁱ id-hom-Precategory C) ×
      ( comp-hom-Precategory C g f ＝ⁱ id-hom-Precategory C))
```

for isomorphisms, as this type has a strictly involutive inverse operator. In fact, it would be even better to use the pairing trick that Anders showed us (I'm sorry, this means backtracking on my initiative to use the strictly involutive identity type in set-level category theory), as that definition is predicative, something that might be important for large categories. The only caveat being that it only works for set-level things.

fredrik-bakke:
@EgbertRijke This PR seems to have been abandoned. Maybe you'd want to open an issue on this question instead? I'd like to have the changes that are not about invertible morphisms in this PR merged if we can, without tackling the problem on how to define isomorphisms. That's for a different refactoring PR.

fredrik-bakke:
There's also a question on whether we prefer the terminology section/retraction vs split epi/split mono in category theory.

PR #1005, opened 2024-01-18
Basic properties of the flat modality

fredrik-bakke:
Proves a series of basic properties of the flat modality.

### General properties

- [X] The universal property of flat discrete crisp types
- [ ] The dependent universal property of flat discrete crisp types
- [X] Functoriality of flat
- [X] Flat is idempotent
- [x] A crisp type is crisply flat discrete if its counit has a crisp section

### Left exactness of flat

- [X] Flat distributes over identity types
- [X] The crisp identity types of flat discrete crisp types are flat discrete
- [X] Flat distributes over dependent pair types
- [X] Flat distributes over product types
- [x] Flat distributes over pullbacks
- [ ] Flat distributes over sequential limits
- [x] The unit type is flat discrete

### Right exactness of flat

- [X] The empty type is flat discrete
- [X] The natural numbers are flat discrete
- [X] Flat distributes over coproduct types
- [ ] Flat distributes over pushouts
- [ ] Flat distributes over coequalizers
- [ ] Flat distributes over sequential colimits

PR #1078, opened 2024-03-14
Basic properties of the flat modality

fredrik-bakke:
This is the replacement of #1005.

Proves a series of basic properties of the flat modality.

### General properties

- [X] The universal property of flat discrete crisp types
- [x] ~The dependent universal property of flat discrete crisp types~
- [X] Functoriality of flat
- [X] Flat is idempotent
- [x] A crisp type is crisply flat discrete if its counit has a crisp section

### Left exactness of flat

- [X] Flat distributes over identity types
- [X] The crisp identity types of flat discrete crisp types are flat discrete
- [X] Flat distributes over dependent pair types
- [X] Flat distributes over product types
- [x] Flat distributes over pullbacks
- [x] ~Flat distributes over sequential limits~
- [x] The unit type is flat discrete

### Right exactness of flat

- [X] The empty type is flat discrete
- [X] The natural numbers are flat discrete
- [X] Flat distributes over coproduct types
- [x] ~Flat distributes over pushouts~
- [x] ~Flat distributes over coequalizers~
- [x] ~Flat distributes over sequential colimits~

### Notes
- The constructor for the flat modality is renamed from `cons-flat` to `intro-flat`. This makes it easier to distinguish from `counit-flat`.
- In the future, we will probably want to prove `crisp-based-ind-Id` from the existence of the sharp modality rather than postulating it. The same is true for the modal induction principle of the sharp modality.
- This PR does some ground work with the sharp modality too.


fredrik-bakke:
I am marking this PR as ready for review, as it is very close to finished. What I have in mind left to complete are a couple of sections of prose that I have specifically tagged with `TODO`s. While this PR is very low priority, would you at some point be willing to review this @EgbertRijke or perhaps even @VojtechStep?

src/foundation-core/equivalences.lagda.md
```diff
@@ -212,21 +212,22 @@ module _
...
+  is-equiv-is-coherently-invertible :
+    is-coherently-invertible f → is-equiv f
+  is-equiv-is-coherently-invertible H =
+    is-equiv-is-invertible' (is-invertible-is-coherently-invertible H)
+
+  is-equiv-is-transpose-coherently-invertible :
+    is-transpose-coherently-invertible f → is-equiv f
+  is-equiv-is-transpose-coherently-invertible H =
+    is-equiv-is-invertible'
+      ( is-invertible-is-transpose-coherently-invertible H)
```
EgbertRijke, line 230: Is it not possible to leave these `abstract`?

src/foundation-core/injective-maps.lagda.md
```diff
@@ -106,26 +107,28 @@ module _
...
-    is-injective-equiv (pair f H) = is-injective-is-equiv H
+  is-injective-equiv : (e : A ≃ B) → is-injective (map-equiv e)
+  is-injective-equiv e = is-injective-is-equiv (is-equiv-map-equiv e)
 
+abstract
+  is-injective-map-inv-equiv :
+    {l1 l2 : Level} {A : UU l1} {B : UU l2} (e : A ≃ B) →
+    is-injective (map-inv-equiv e)
+  is-injective-map-inv-equiv e =
+    is-injective-is-equiv (is-equiv-map-inv-equiv e)
```
EgbertRijke, line 122: I wonder what your idea is here for making some things abstract and removing the abstract label elsewhere?

Note that I am not criticizing the changes, but merely ask you how they have come about.

src/foundation-core/postcomposition-dependent-functions.lagda.md
```diff
@@ -0,0 +1,44 @@
...
+
+open import foundation-core.function-types
+```
+
+</details>
+
+## Idea
+
+Given a type `A` and a dependent map `f : {a : A} → X a → Y a`, the
+{{#concept "dependent postcomposition function" Agda=postcomp-Π}}
```
EgbertRijke, line 20: I think it is better to use the word dependent function/map when the emphasis is on the type of the value depending on the input. Here, given that `{a : A}` is an implicit argument, it seems that the emphasis should be that it is a _family of maps_ from `X` to `Y`.

src/foundation-core/postcomposition-dependent-functions.lagda.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+is defined by `λ h x → f (h x)`.
+
+Note that, as opposed to
+[precomposition of dependent functions](foundation-core.precomposition-dependent-functions.md),
+the use-case for postcomposition of dependent functions is very limited, since
+the definition of `f` depends on the particular choice of `A`. Once we allow `A`
+to vary while keeping `f` fixed, we reduce to the case of
+[postcomposition of (nondependent) functions](foundation-core.postcomposition-functions.md).
```
EgbertRijke, line 33: We had this text before, I believe, but it always struck me as a somewhat odd thing to say that the use-case for postcomposition is very limited, suggesting that it is barely used. In fact, it is quite widely used isn't it? So I think something about this phrasing is suggesting that it is not often used or not very useful that seems a bit misleading to me.

src/foundation-core/postcomposition-dependent-functions.lagda.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Definitions
+
```
EgbertRijke, line 36:
```suggestion
## Definitions

### Postcomposing families of maps

```

src/foundation-core/postcomposition-dependent-functions.lagda.md
EgbertRijke: Would it perhaps be better to call this file `postcomposition-families-of-maps` to aim for a more useful connotation of this concept?

src/foundation-core/retractions.lagda.md
```diff
@@ -75,24 +75,39 @@ identifications.
...
-    (i : A → B) (r : B → A) (H : r ∘ i ~ id)
-    (x y : A) → ((ap-retraction i r H x y) ∘ (ap i {x} {y})) ~ id
-  is-retraction-ap-retraction i r H x .x refl = left-inv (H x)
-
-  retraction-ap :
-    (i : A → B) → retraction i → (x y : A) → retraction (ap i {x} {y})
-  pr1 (retraction-ap i (pair r H) x y) = ap-retraction i r H x y
-  pr2 (retraction-ap i (pair r H) x y) = is-retraction-ap-retraction i r H x y
+  is-injective-retraction :
+    {x y : A} → i x ＝ i y → x ＝ y
```
EgbertRijke, line 95: Is it possible to use `is-injective` here or does that cause a cyclic module dependency?

src/foundation/equivalences.lagda.md
```diff
@@ -238,14 +238,14 @@ module _
...
-    B ---> Y
-       j
+         i
+    A ------> X
+    |       ∧ |
+  f |     /   | g
+    |   h     |
+    ∨ /       ∨
+    B ------> Y
+         j
```
EgbertRijke, line 248:
```suggestion
         i
    A ------> X
    |       ∧ |
  f |     /   | g
    |   h     |
    ∨ /       ∨
    B ------> Y.
         j
```

EgbertRijke (commented):
I have reviewed the changes to `foundation`, `foundation-core`, and `elementary-number-theory` so far, which seem to have little to do with basic properties of the flat modality. I will review the other changes later.

src/foundation-core/equivalences.lagda.md
fredrik-bakke: It definitely is, they're not used anywhere currently. I unabstracted them as they satisfy very nice computational properties. It seems to me that if you don't want to compute with these, then the underlying `is-coherently-invertible` proof should be abstracted rather than these directly.

EgbertRijke: Ok, that sounds good.

If you have observed some very nice computational properties, would you mind writing up briefly what they are, so that future formalizers can see why these definitions are unabstracted?

src/foundation-core/injective-maps.lagda.md
fredrik-bakke: The work I was doing here, although most of it is removed now, was working towards being able to say that `is-injective-equiv e` was an equivalence. That required the map to unfold such that `is-injective-equiv e` would be identified with the inverse of `ap (map-equiv e)` otherwise there seemed to be no other way than to reprove everything. However, before I finished this line of formalization I came up with a more elegant proof circumventing this whole ordeal.

I believe it is fine to leave `is-injective-map-inv-equiv` abstract because then you probably want to use `ap (map-equiv e)` instead.

src/foundation-core/postcomposition-dependent-functions.lagda.md
fredrik-bakke: That's probably better. I would, however. prefer if the standard definition keeps `a : A` as an implicit argument, to keep `postcomp-Π` consistent with `postcomp`

src/foundation-core/postcomposition-dependent-functions.lagda.md
fredrik-bakke: I see, I'll try to rephrase it.

fredrik-bakke: How is this?

```md
Note that, since the definition of the family of maps `f` depends on the base
`A`, postcomposition of families of maps does not generalize
[postcomposition of functions](foundation-core.postcomposition-functions.md) in
the same way that
[precomposition of dependent functions](foundation-core.precomposition-dependent-functions.md)
generalizes
[precomposition of functions](foundation-core.precomposition-functions.md). If
`A` can vary while keeping `f` fixed, we have necessarily reduced to the case of
[postcomposition of functions](foundation-core.postcomposition-functions.md).
```

src/foundation-core/retractions.lagda.md
fredrik-bakke: It causes cyclic dependencies.

src/foundation-core/postcomposition-dependent-functions.lagda.md
fredrik-bakke: Oh, so actually I want to go back on this. The point is that we are defining a postcomposition map _on_ dependent functions, similar to how we define a precomposition map on dependent functions. There, it is not the map we are precomposing with that is dependent.

src/foundation-core/postcomposition-dependent-functions.lagda.md
fredrik-bakke: I went with `### Postcomposition of dependent functions`

src/foundation-core/postcomposition-dependent-functions.lagda.md
EgbertRijke: Ah, you are completely right. Sorry that I missed this. Perhaps this point of view could be made a little bit more apparent in the prose, it's quite useful. Thanks for finding this conceptual thought!

EgbertRijke:
Would it be reasonable a separate PR for the changes not in `modal-type-theory`? Those changes seem quite useful to me and they can be merged today independently of the rest of the PR.

One way to do this is:
- create a new branch from `master` in your clone
- go into this new branch
- use the following command to pull all the changes from a specific folder into your branch:

  ```
  git checkout flat-modality -- path/to/your/folder
  ```

  where `path/to/your/folder` ranges over the paths to `elementary-number-theory`, `foundation`, `foundation-core`, and `orthogonal-factorization-systems.
  
Note that this is only a suggestion. If you prefer to not create a separate pull request for these changes, that's ok with me too.

fredrik-bakke:
> Would it be reasonable a separate PR for the changes not in `modal-type-theory`? Those changes seem quite useful to me and they can be merged today independently of the rest of the PR.
> 
> One way to do this is:
> 
> * create a new branch from `master` in your clone
> * go into this new branch
> * use the following command to pull all the changes from a specific folder into your branch:
>   ```
>   git checkout flat-modality -- path/to/your/folder
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   where `path/to/your/folder` ranges over the paths to `elementary-number-theory`, `foundation`, `foundation-core`, and `orthogonal-factorization-systems.
> 
> Note that this is only a suggestion. If you prefer to not create a separate pull request for these changes, that's ok with me too.

Yes! I was thinking the same actually. 😅

EgbertRijke:
Btw if there is a better way to do this, I would love to know. This was just the one way I know how to create a branch with changes from specific folders, but obviously my `git` skills are nowhere near as good as the `git` skills of you and Vojtech.

fredrik-bakke:
> Btw if there is a better way to do this, I would love to know. This was just the one way I know how to create a branch with changes from specific folders, but obviously my `git` skills are nowhere near as good as the `git` skills of you and Vojtech.

That's not true. I didn't for instance know how to do this before you instructed me just now.

fredrik-bakke:
> I am marking this PR as ready for review, as it is very close to finished. What I have in mind left to complete are a couple of sections of prose that I have specifically tagged with `TODO`s. While this PR is very low priority, would you at some point be willing to review this @EgbertRijke or perhaps even @VojtechStep?

I've added the missing explanations mentioned in this comment now.

references.bib
```diff
@@ -338,6 +346,16 @@ @book{MRR88
   isbn        = {978-0-387-96640-3 978-1-4419-8640-5}
 }
 
+@misc{Mye21,
```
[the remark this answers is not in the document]
fredrik-bakke: So far I've used the format `DavidJaz/Cohesion` for GitHub repositories while articles get the style `Mye21`. For blog posts, I've additionally added a couple of terms from the title of the post, E.g. `Shu14SplittingIdempotents`. It's not a perfect system and if you have any objections feel free to air them.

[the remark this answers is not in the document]
fredrik-bakke: Sorry, I didn't mean to sound accusatory in my reply. Thanks for asking about it! It has been a conscious decision to use different styles, but the design is very much in the spirit of "something that works" rather than anything cohesive and neat.

fredrik-bakke:
Since it seems no one is willing to give a review of this PR, I'm happy to have it merged without one. If that is okay with you @EgbertRijke @VojtechStep

EgbertRijke:
I quickly glanced over this PR and it looks good to me. Sorry that this flew under the radar for so long.

PR #1083, opened 2024-03-19
chore: fix some typos in the library 

fredrik-bakke:
### Note
"Non-" is commonly spelled without a hyphen in American English.
- https://www.oxfordlearnersdictionaries.com/definition/english/non
- https://en.wiktionary.org/wiki/non-#Usage_notes

### DIFF
--- a/src/linear-algebra/functoriality-vectors.lagda.md
+++ b/src/linear-algebra/functoriality-vectors.lagda.md
@@ -1,4 +1,4 @@
-# Functoriality of vectors
+# Functoriality of the type of vectors
 
 ```agda
 module linear-algebra.functoriality-vectors where
--- a/src/linear-algebra/functoriality-vectors.lagda.md
+++ b/src/linear-algebra/functoriality-vectors.lagda.md
@@ -1,4 +1,4 @@
-# Functoriality of the type of vectors
+# Functoriality of vectors
 
 ```agda
 module linear-algebra.functoriality-vectors where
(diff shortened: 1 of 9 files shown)


src/linear-algebra/functoriality-vectors.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Functoriality of the type of vectors
+# Functoriality of vectors
```
EgbertRijke, line 1: Hm... I'm not so sure about this change. The phrase "functoriality of vectors" sounds like it is about functorial properties of vectors themselves, but this file is about functoriality of the type family `vec`.

fredrik-bakke: Indeed!

EgbertRijke (approved):
Other than my one objection, this PR looks good.

src/linear-algebra/functoriality-vectors.lagda.md
EgbertRijke: Will you change that to "Functoriality of the type of matrices" then?

EgbertRijke: > I was hoping you would comment on it, because the file for matrices was titled "Functoriality of matrices" 😅

Btw, I don't want to be annoying, but I do want to say that I don't like it very much if you intentionally make a poor change in order to provoke a reaction to that change.

fredrik-bakke: I'm sorry. I will try to avoid letting such things go through without remark from now on

PR #1086, opened 2024-03-20
chore: Janitorial work in foundation

fredrik-bakke:
This is a follow-up to #1085 with changes from #1032.

Issue #930, opened 2023-11-21
Project: Sheaves and excision with respect to regular cd-structures


EgbertRijke:
## Project description

A **complete cd-structure** consists of a class of **distinguished squares** of maps, i.e., distinguished morphisms in the arrow category, that is stable under base change. A complete cd-structure is said to be **regular** if the following three conditions hold:

1. The domain of any distinguished square is a mono.
2. Any distinguished square is cartesian
3. The diagonal of any distinguished square is distinguished.

A type `F` is said to be a **sheaf** with respect to a cd-structure if `F` is right orthogonal to `inclusion-im (cogap d)` for any distinguished square `d`.

A type `F` is said to **satisfy excision** with respect to a cd-structure if `F` is right orthogonal to `cogap d` for any distinguished square `d`.

**Goal of the project:** Show that `F` is a sheaf with respect to a regular cd-structure if and only if it satisfies excision.

**Current participants:** Reid Barton, Egbert Rijke.

## Tasks

A list of tasks will appear here.

fredrik-bakke:
I've already defined the closed modality. It's in `orthogonal-factorization-systems.closed-modality` if I remember correctly.

fredrik-bakke:
> - [X] Prepare the file `pullback-squares for real-world applications, and possibly rename it to `cartesian squares`. Alternatively, introduce a file for cartesian morphisms of arrows.

Cartesian morphisms of arrows were added in #979.

fredrik-bakke:
This issue seems to be missing any mention of orthogonal maps or fiberwise orthogonal maps. Are they not needed?

fredrik-bakke:
When you write
> - [ ]  Define base change of morphisms of arrows.

Are you looking for a file about the following construct?

```agda
base-change :
  {l1 l2 : Level} (l3 l4 : Level) {A : UU l1} {B : UU l2} → (A → B) →
  UU (l1 ⊔ l2 ⊔ lsuc l3 ⊔ lsuc l4)
base-change l3 l4 f =
  Σ (UU l3) (λ C → Σ (UU l4) (λ D → Σ (C → D) (λ g → cartesian-hom-arrow g f)))
```

Otherwise, just using cartesian morphisms of arrows seems to be working fine with me for now. Although I would probably say the same about cartesian morphisms of arrows vs. pullback cones if we didn't already have cartesian morphisms of arrows.

PR #1088, opened 2024-03-20
Null maps, null types and null type families

fredrik-bakke:
Defines `Y`-null maps, `Y`-null types and `Y`-null type families, and establishes a few basic properties of these. Also formalizes the fiber condition for orthogonal maps. Part of #930. Depends on #1086.

- If `A` is a retract of `B` and `S` is a retract of `T` then `diagonal-exponential A S` is a retract of `diagonal-exponential B T`.
- Null types
  - Null types are closed under equivalences in the base and exponent
  - Null types are closed under retracts in the base and exponent
  - A type is `Y`-null if and only if the terminal projection of `Y` is orthogonal to the terminal projection of `A`
- Null families
  - Null families are closed under equivalences in the family and exponent
  - Null families are closed under retracts in the family and exponent
- Null maps, equivalence of
  - Fiber condition
  - Pullback condition
  - Right orthogonal map to terminal projection condition
  - Local at terminal projection condition
- Fiber condition for orthogonal maps


fredrik-bakke:
I think this PR is mostly ready for review now, so I've marked it as such. 😁
I might end up wanting to use some of this in #1118 soon.

src/foundation/constant-maps.lagda.md
```diff
@@ -55,16 +64,21 @@ module _
   compute-action-htpy-function-const c H = ap-const c (eq-htpy H)
 ```
 
+### Computing the fibers of the point
```
EgbertRijke, line 67:
```suggestion
### Computing the fibers of point inclusions
```

src/foundation/diagonal-maps-of-types.lagda.md
```diff
@@ -80,3 +85,36 @@ module _
   pr1 diagonal-exponential-injection = diagonal-exponential A B
   pr2 diagonal-exponential-injection = is-injective-diagonal-exponential
 ```
+
+### The action on identifications of a constant map is a constant map
```
EgbertRijke, line 89:
```suggestion
### The action on identifications of an (exponential) diagonal is a diagonal
```

src/orthogonal-factorization-systems/null-families-of-types.lagda.md
```diff
@@ -0,0 +1,123 @@
...
+open import orthogonal-factorization-systems.orthogonal-maps
+```
+
+</details>
+
+## Idea
+
+A family of types `B : A → UU l` is said to be
+{{#concept "null" Disambiguation="family of types" Agda=is-null-family}} at `Y`,
+or **`Y`-null**, if every [fiber](foundation-core.fibers-of-maps.md) is. I.e.,
```
EgbertRijke, line 27: Linking to fibers of maps is probably not appropriate here, because `B x` is not definitionally the fiber of a map.

PR #1089, opened 2024-03-20
Colocal types

fredrik-bakke:
Defines what it means for a type to be colocal at a map, and shows a few basic properties of these.


src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,213 @@
...
+### Types colocal at dependent maps
+
+```agda
+module _
+  {l1 l2 l3 : Level} (A : UU l1) {X : A → UU l2} {Y : A → UU l3}
+  (f : {a : A} → X a → Y a)
+  where
+
+  is-dependent-map-colocal : UU (l1 ⊔ l2 ⊔ l3)
+  is-dependent-map-colocal = is-equiv (postcomp-Π A (λ {a} → f {a}))
```
VojtechStep, line 63: Does Agda have problems with
```suggestion
  is-dependent-map-colocal = is-equiv (postcomp-Π A f)
```
?

src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,213 @@
...
+
+```agda
+module _
+  {l1 l2 l3 : Level} {X : UU l1} {Y : UU l2} {A : UU l3} {f f' : X → Y}
+  where
+
+  is-colocal-htpy : (H : f ~ f') → is-colocal f' A → is-colocal f A
+  is-colocal-htpy H = is-equiv-htpy (postcomp A f') (htpy-postcomp A H)
+
+  is-colocal-htpy' : (H : f ~ f') → is-colocal f A → is-colocal f' A
```
VojtechStep, line 123: Maybe keeping with the `inv-` mixfix
```suggestion
  is-colocal-inv-htpy : (H : f' ~ f) → is-colocal f' A → is-colocal f A
```

src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,213 @@
...
+  is-colocal-htpy' H = is-equiv-htpy' (postcomp A f) (htpy-postcomp A H)
+```
+
+### If `S` is `f`-colocal then `S` is colocal at every retract of `f`
```
VojtechStep, line 127: Nit: the title calls about assuming `is-colocal f S` to show `is-colocal g s` for `f retract-of-map f`, but the formalization switches the names

src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,213 @@
...
+  where
+
+  is-colocal-is-equiv :
+    is-equiv f → {l : Level} (A : UU l) → is-colocal f A
+  is-colocal-is-equiv is-equiv-f = is-equiv-postcomp-is-equiv f is-equiv-f
+```
+
+### A contractible type is `f`-colocal if and only if `f` is an equivalence
+
+This remains to be formalized.
```
VojtechStep, line 209: Is this included in the batch of PRs to follow, or should it have an open issue?

VojtechStep (commented):
Looks good 👍

src/orthogonal-factorization-systems/colocal-types.lagda.md
fredrik-bakke: It's not following in a subsequent PR, but I am also hesitant to open an issue. It's one of those things that's trivial and not very useful as a formalized piece of code, but may assist a reader in understanding the concept better.

fredrik-bakke: Note that we have a lot of open claims in the library that do not have a corresponding issue. The series of issues I wrote in September last year targeted one particular wording of "it remains to be formalized" and hardly caught all open claims in the library. Something I only discovered a bit later.

fredrik-bakke: Well, I shouldn't say it's not useful. But it's not something I'm looking to use currently

src/orthogonal-factorization-systems/colocal-types.lagda.md
fredrik-bakke: Yes
image

src/orthogonal-factorization-systems/colocal-types.lagda.md
fredrik-bakke: Thanks! I remember justifying the discrepancy to myself a couple of times, but now I can't remember the justification, so I'll just change it 🙃👍

src/orthogonal-factorization-systems/colocal-types.lagda.md
VojtechStep: If it's easy to work out, would it make sense to open an issue with the tag "good first issue", for people looking into contributing to the library?

src/orthogonal-factorization-systems/colocal-types.lagda.md
VojtechStep: Looks like I still have garbage intuition around what Agda can and cannot infer 😅. It's interesting that it's able to infer it in `is-property-is-dependent-map-colocal`.

fredrik-bakke: You and me both 🥲
`postcomp-Π` is a bit of a weird one though

src/orthogonal-factorization-systems/colocal-types.lagda.md
fredrik-bakke: Now they're formalized so we don't have to worry about them 😇

VojtechStep: Smh

EgbertRijke:
Hi! I think we had this discussion with Tom before, when we should speak of the diagonal and when we should speak of the constant map. Eg: it is misleading to say that "the constant map is an equivalence" if it is merely meant that `A -> (B -> A)` is an equivalence. It is much clearer to say that the diagonal map is an equivalence in this case. We were fairly careful about when to use the word "diagonal" and when to use the word "constant map". 

In this pull request I see some changes from "diagonal" back to "constant map". Is the above consideration taken into account here?

src/foundation/contractible-types.lagda.md
```diff
@@ -182,24 +182,24 @@ module _
...
         ( λ f → eq-htpy (λ x → ap f (contraction H x)))
         ( λ x → refl)
 
-  equiv-diagonal-is-contr :
+  equiv-const-is-contr :
     {l : Level} (X : UU l) → is-contr A → X ≃ (A → X)
-  pr1 (equiv-diagonal-is-contr X H) = const A X
-  pr2 (equiv-diagonal-is-contr X H) = is-equiv-diagonal-is-contr H X
+  pr1 (equiv-const-is-contr X H) = const A X
+  pr2 (equiv-const-is-contr X H) = is-equiv-const-is-contr H X
```
EgbertRijke, line 204: This entry was more accurately named before: the underlying map of this equivalence is the map `X -> (A -> X)` which is, despite it being implemented using `const` is not a constant map. It is rather the diagonal map, mapping `X` into the exponent `(A -> X)`.

src/foundation/contractible-types.lagda.md
```diff
@@ -182,24 +182,24 @@ module _
...
-    is-equiv-diagonal-is-contr :
+    is-equiv-const-is-contr :
       is-contr A →
       {l : Level} (X : UU l) → is-equiv (const A X)
-    is-equiv-diagonal-is-contr H X =
+    is-equiv-const-is-contr H X =
       is-equiv-is-invertible
         ( ev-point' (center H))
         ( λ f → eq-htpy (λ x → ap f (contraction H x)))
         ( λ x → refl)
```
EgbertRijke, line 199: Similarly, this entry was more accurately named as `is-equiv-diagonal-is-contr`.

src/foundation/contractible-types.lagda.md
fredrik-bakke: Humm, I see. I would be in favor of defining `diagonal := const` then

fredrik-bakke: What is already called `diagonal` can for instance be renamed to `diagonal-product`

fredrik-bakke: It's good you bring it up, because I make the same mistake in #1088

EgbertRijke: Yes, I think that is a very good suggestion. Would you be willing to do that?

fredrik-bakke: On it!

EgbertRijke: Awesome, thank you so much!

fredrik-bakke: I've reverted the changes I made for now. I would prefer if the fix could be made as a separate pull request, as I don't currently know how large the changes will be.

src/foundation-core/empty-types.lagda.md
```diff
@@ -48,6 +48,13 @@ is-nonempty : {l : Level} → UU l → UU l
 is-nonempty A = is-empty (is-empty A)
 ```
 
+### The initial map into a type
+
+```agda
+initial-map : {l : Level} (A : UU l) → empty → A
+initial-map A = ex-falso
+```
```
EgbertRijke, line 56: Wouldn't this make more sense in `universal-property-empty-type`, which is _about_ initiality of the empty type?

src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Idea
+
+A type `A` is said to be
+{{#concept "colocal" Disambiguation="type at a map" Agda=is-colocal}} at the map
+`f : X → Y`, or **`f`-colocal**, if the
+[postcomposition map](foundation-core.postcomposition-functions.md)
+
+```text
+  f ∘_ : (A → X) → (A → Y)
```
EgbertRijke, line 42: I still think that the hyphen is more conventional in mathematical texts for this purpose than the underscore 😅

src/orthogonal-factorization-systems/colocal-types.lagda.md
```diff
@@ -0,0 +1,266 @@
...
+is an [equivalence](foundation-core.equivalences.md). This is dual to the notion
+of an [`f`-local type](orthogonal-factorization-systems.local-types.md), which
+is a type such that the
+[precomposition map](foundation-core.precomposition-functions.md)
+
+```text
+  _∘ f : (Y → A) → (X → A)
+```
+
+is an equivalence.
```
EgbertRijke, line 54: I think this exposition deserves a bit more detail. For instance, we can't expect readers to be able to infer what kind of types are colocal merely from the sentence "This is dual to the notion of local". Things to include that would improve this prose would be answerts to the following questions:
- What does it really mean for a type to be colocal at `f`?
- Does it generalize already existing concepts?
- In what context is this definition relevant?

fredrik-bakke: Hi, I'm revisiting this PR again because I'd like to do some refactoring in the `orthogonal-factorization-systems` module that requires the PRs that touch it to be merged or closed, and I'm not entirely certain what you are asking here.

> - What does it really mean for a type to be colocal at `f`?

I thought it was stated quite clearly above:

````md
A type `A` is said to be
{{#concept "colocal" Disambiguation="type at a map" Agda=is-colocal}} at the map
`f : X → Y`, or **`f`-colocal**, if the
[postcomposition map](foundation-core.postcomposition-functions.md)

```text
  f ∘_ : (A → X) → (A → Y)
```

is an [equivalence](foundation-core.equivalences.md). [...]
````

I've added two additional equivalent conditions for a type to be colocal, but it's not clear to me if this is what you meant. Is there something else you'd like me to say?

> - Does it generalize or is it related to already existing concepts?

I added that its a specialization of left orthogonal maps

> - In what context is this definition relevant?

I would honestly like to avoid this question for now as I am still a little uncertain to what extent this definition is relevant. Still, I expect there's a comparison to be made to coreflective subuniverses.

PR #1100, opened 2024-03-28
The additive group of rational numbers

malarbol:
This pull requests introduces the additive group structure on the rational numbers.

We introduce 
- the `neg-fraction-Z` and `neg-Q` functions and some basic properties with respect to addition and multiplication. 
- the difference of rational numbers and some basic properties with respect to addition and multiplication. 

We also prove associativity and commutativity of multiplication on the rational numbers.

### DIFF
--- a/src/elementary-number-theory/addition-integers.lagda.md
+++ b/src/elementary-number-theory/addition-integers.lagda.md
@@ -330,7 +330,8 @@ abstract
 ### Interchange law for addition with respect to addition
 
 ```agda
-interchange-law-add-add-ℤ : interchange-law add-ℤ add-ℤ
+interchange-law-add-add-ℤ :
+  (x y u v : ℤ) → (x +ℤ y) +ℤ (u +ℤ v) ＝ (x +ℤ u) +ℤ (y +ℤ v)
 interchange-law-add-add-ℤ =
   interchange-law-commutative-and-associative
     add-ℤ
@@ -545,7 +546,11 @@ is-zero-left-add-ℤ x y H =
       by ap (_+ℤ (neg-ℤ y)) H
     ＝ zero-ℤ
       by right-inverse-law-add-ℤ y
+```
 
+### If `x + y = x` then `y = 0`
+
+```agda
 is-zero-right-add-ℤ :
   (x y : ℤ) → x +ℤ y ＝ x → is-zero-ℤ y
 is-zero-right-add-ℤ x y H =
--- a/src/elementary-number-theory/addition-integers.lagda.md
+++ b/src/elementary-number-theory/addition-integers.lagda.md
@@ -330,37 +330,39 @@ abstract
 ### Interchange law for addition with respect to addition
 
 ```agda
-interchange-law-add-add-ℤ :
-  (x y u v : ℤ) → (x +ℤ y) +ℤ (u +ℤ v) ＝ (x +ℤ u) +ℤ (y +ℤ v)
-interchange-law-add-add-ℤ =
-  interchange-law-commutative-and-associative
-    add-ℤ
-    commutative-add-ℤ
-    associative-add-ℤ
+abstract
+  interchange-law-add-add-ℤ :
+    (x y u v : ℤ) → (x +ℤ y) +ℤ (u +ℤ v) ＝ (x +ℤ u) +ℤ (y +ℤ v)
+  interchange-law-add-add-ℤ =
+    interchange-law-commutative-and-associative
+      add-ℤ
+      commutative-add-ℤ
+      associative-add-ℤ
 ```
 
 ### Addition by `x` is a binary equivalence
 
 ```agda
-is-section-left-add-neg-ℤ :
-  (x y : ℤ) → x +ℤ (neg-ℤ x +ℤ y) ＝ y
-is-section-left-add-neg-ℤ x y =
-  equational-reasoning
-    x +ℤ (neg-ℤ x +ℤ y)
-    ＝ (x +ℤ neg-ℤ x) +ℤ y
-      by inv (associative-add-ℤ x (neg-ℤ x) y)
-    ＝ y
-      by ap (_+ℤ y) (right-inverse-law-add-ℤ x)
+abstract
+  is-section-left-add-neg-ℤ :
+    (x y : ℤ) → x +ℤ (neg-ℤ x +ℤ y) ＝ y
+  is-section-left-add-neg-ℤ x y =
+    equational-reasoning
+      x +ℤ (neg-ℤ x +ℤ y)
+      ＝ (x +ℤ neg-ℤ x) +ℤ y
+        by inv (associative-add-ℤ x (neg-ℤ x) y)
+      ＝ y
+        by ap (_+ℤ y) (right-inverse-law-add-ℤ x)
 
-is-retraction-left-add-neg-ℤ :
-  (x y : ℤ) → (neg-ℤ x) +ℤ (x +ℤ y) ＝ y
-is-retraction-left-add-neg-ℤ x y =
-  equational-reasoning
-    neg-ℤ x +ℤ (x +ℤ y)
-    ＝ (neg-ℤ x +ℤ x) +ℤ y
-      by inv (associative-add-ℤ (neg-ℤ x) x y)
-    ＝ y
-      by ap (_+ℤ y) (left-inverse-law-add-ℤ x)
+  is-retraction-left-add-neg-ℤ :
+    (x y : ℤ) → (neg-ℤ x) +ℤ (x +ℤ y) ＝ y
+  is-retraction-left-add-neg-ℤ x y =
+    equational-reasoning
+      neg-ℤ x +ℤ (x +ℤ y)
+      ＝ (neg-ℤ x +ℤ x) +ℤ y
+        by inv (associative-add-ℤ (neg-ℤ x) x y)
+      ＝ y
+        by ap (_+ℤ y) (left-inverse-law-add-ℤ x)
 
 abstract
   is-equiv-left-add-ℤ : (x : ℤ) → is-equiv (x +ℤ_)
@@ -373,29 +375,30 @@ equiv-left-add-ℤ : ℤ → (ℤ ≃ ℤ)
 pr1 (equiv-left-add-ℤ x) = add-ℤ x
 pr2 (equiv-left-add-ℤ x) = is-equiv-left-add-ℤ x
 
-is-section-right-add-neg-ℤ :
-  (x y : ℤ) → (y +ℤ neg-ℤ x) +ℤ x ＝ y
-is-section-right-add-neg-ℤ x y =
-  equational-reasoning
-    (y +ℤ neg-ℤ x) +ℤ x
-    ＝ y +ℤ (neg-ℤ x +ℤ x)
-      by associative-add-ℤ y (neg-ℤ x) x
-    ＝ y +ℤ zero-ℤ
-      by ap (y +ℤ_) (left-inverse-law-add-ℤ x)
-    ＝ y
-      by right-unit-law-add-ℤ y
+abstract
+  is-section-right-add-neg-ℤ :
+    (x y : ℤ) → (y +ℤ neg-ℤ x) +ℤ x ＝ y
+  is-section-right-add-neg-ℤ x y =
+    equational-reasoning
+      (y +ℤ neg-ℤ x) +ℤ x
+      ＝ y +ℤ (neg-ℤ x +ℤ x)
+        by associative-add-ℤ y (neg-ℤ x) x
+      ＝ y +ℤ zero-ℤ
+        by ap (y +ℤ_) (left-inverse-law-add-ℤ x)
+      ＝ y
+        by right-unit-law-add-ℤ y
 
-is-retraction-right-add-neg-ℤ :
-  (x y : ℤ) → (y +ℤ x) +ℤ neg-ℤ x ＝ y
-is-retraction-right-add-neg-ℤ x y =
-  equational-reasoning
-    (y +ℤ x) +ℤ neg-ℤ x
-    ＝ y +ℤ (x +ℤ neg-ℤ x)
-      by associative-add-ℤ y x (neg-ℤ x)
-    ＝ y +ℤ zero-ℤ
-      by ap (y +ℤ_) (right-inverse-law-add-ℤ x)
-    ＝ y
-      by right-unit-law-add-ℤ y
+  is-retraction-right-add-neg-ℤ :
+    (x y : ℤ) → (y +ℤ x) +ℤ neg-ℤ x ＝ y
+  is-retraction-right-add-neg-ℤ x y =
+    equational-reasoning
+      (y +ℤ x) +ℤ neg-ℤ x
+      ＝ y +ℤ (x +ℤ neg-ℤ x)
+        by associative-add-ℤ y x (neg-ℤ x)
+      ＝ y +ℤ zero-ℤ
+        by ap (y +ℤ_) (right-inverse-law-add-ℤ x)
+      ＝ y
+        by right-unit-law-add-ℤ y
 
 abstract
   is-equiv-right-add-ℤ : (y : ℤ) → is-equiv (_+ℤ y)
@@ -416,145 +419,152 @@ pr2 is-binary-equiv-left-add-ℤ = is-equiv-left-add-ℤ
 ### Addition by an integer is a binary embedding
 
 ```agda
-is-emb-left-add-ℤ :
-  (x : ℤ) → is-emb (x +ℤ_)
-is-emb-left-add-ℤ x =
-  is-emb-is-equiv (is-equiv-left-add-ℤ x)
+abstract
+  is-emb-left-add-ℤ :
+    (x : ℤ) → is-emb (x +ℤ_)
+  is-emb-left-add-ℤ x =
+    is-emb-is-equiv (is-equiv-left-add-ℤ x)
 
-is-emb-right-add-ℤ :
-  (y : ℤ) → is-emb (_+ℤ y)
-is-emb-right-add-ℤ y = is-emb-is-equiv (is-equiv-right-add-ℤ y)
+  is-emb-right-add-ℤ :
+    (y : ℤ) → is-emb (_+ℤ y)
+  is-emb-right-add-ℤ y = is-emb-is-equiv (is-equiv-right-add-ℤ y)
 
-is-binary-emb-add-ℤ : is-binary-emb add-ℤ
-is-binary-emb-add-ℤ =
-  is-binary-emb-is-binary-equiv is-binary-equiv-left-add-ℤ
+  is-binary-emb-add-ℤ : is-binary-emb add-ℤ
+  is-binary-emb-add-ℤ =
+    is-binary-emb-is-binary-equiv is-binary-equiv-left-add-ℤ
 ```
 
 ### Addition by `x` is injective
 
 ```agda
-is-injective-right-add-ℤ : (x : ℤ) → is-injective (_+ℤ x)
-is-injective-right-add-ℤ x = is-injective-is-emb (is-emb-right-add-ℤ x)
+abstract
+  is-injective-right-add-ℤ : (x : ℤ) → is-injective (_+ℤ x)
+  is-injective-right-add-ℤ x = is-injective-is-emb (is-emb-right-add-ℤ x)
 
-is-injective-add-ℤ : (x : ℤ) → is-injective (x +ℤ_)
-is-injective-add-ℤ x = is-injective-is-emb (is-emb-left-add-ℤ x)
+  is-injective-add-ℤ : (x : ℤ) → is-injective (x +ℤ_)
+  is-injective-add-ℤ x = is-injective-is-emb (is-emb-left-add-ℤ x)
 ```
 
 ### Negative laws for addition
 
 ```agda
-right-negative-law-add-ℤ :
-  (k l : ℤ) → k +ℤ neg-ℤ l ＝ neg-ℤ (neg-ℤ k +ℤ l)
-right-negative-law-add-ℤ (inl zero-ℕ) l =
-  equational-reasoning
-    neg-one-ℤ +ℤ neg-ℤ l
-    ＝ pred-ℤ (zero-ℤ +ℤ neg-ℤ l)
-      by left-predecessor-law-add-ℤ zero-ℤ (neg-ℤ l)
-    ＝ neg-ℤ (succ-ℤ l)
-      by pred-neg-ℤ l
-right-negative-law-add-ℤ (inl (succ-ℕ x)) l =
-  equational-reasoning
-    pred-ℤ (inl x) +ℤ neg-ℤ l
-    ＝ pred-ℤ (inl x +ℤ neg-ℤ l)
-      by left-predecessor-law-add-ℤ (inl x) (neg-ℤ l)
-    ＝ pred-ℤ (neg-ℤ (neg-ℤ (inl x) +ℤ l))
-      by ap pred-ℤ (right-negative-law-add-ℤ (inl x) l)
-    ＝ neg-ℤ (succ-ℤ (inr (inr x) +ℤ l))
-      by pred-neg-ℤ (inr (inr x) +ℤ l)
-right-negative-law-add-ℤ (inr (inl star)) l =
-  refl
-right-negative-law-add-ℤ (inr (inr zero-ℕ)) l =
-  inv (neg-pred-ℤ l)
-right-negative-law-add-ℤ (inr (inr (succ-ℕ n))) l =
-  equational-reasoning
-    succ-ℤ (in-pos-ℤ n) +ℤ neg-ℤ l
-    ＝ succ-ℤ (in-pos-ℤ n +ℤ neg-ℤ l)
-      by left-successor-law-add-ℤ (in-pos-ℤ n) (neg-ℤ l)
-    ＝ succ-ℤ (neg-ℤ (neg-ℤ (inr (inr n)) +ℤ l))
-      by ap succ-ℤ (right-negative-law-add-ℤ (inr (inr n)) l)
-    ＝ neg-ℤ (pred-ℤ ((inl n) +ℤ l))
-      by inv (neg-pred-ℤ ((inl n) +ℤ l))
+abstract
+  right-negative-law-add-ℤ :
+    (k l : ℤ) → k +ℤ neg-ℤ l ＝ neg-ℤ (neg-ℤ k +ℤ l)
+  right-negative-law-add-ℤ (inl zero-ℕ) l =
+    equational-reasoning
+      neg-one-ℤ +ℤ neg-ℤ l
+      ＝ pred-ℤ (zero-ℤ +ℤ neg-ℤ l)
+        by left-predecessor-law-add-ℤ zero-ℤ (neg-ℤ l)
+      ＝ neg-ℤ (succ-ℤ l)
+        by pred-neg-ℤ l
+  right-negative-law-add-ℤ (inl (succ-ℕ x)) l =
+    equational-reasoning
+      pred-ℤ (inl x) +ℤ neg-ℤ l
+      ＝ pred-ℤ (inl x +ℤ neg-ℤ l)
+        by left-predecessor-law-add-ℤ (inl x) (neg-ℤ l)
+      ＝ pred-ℤ (neg-ℤ (neg-ℤ (inl x) +ℤ l))
+        by ap pred-ℤ (right-negative-law-add-ℤ (inl x) l)
+      ＝ neg-ℤ (succ-ℤ (inr (inr x) +ℤ l))
+        by pred-neg-ℤ (inr (inr x) +ℤ l)
+  right-negative-law-add-ℤ (inr (inl star)) l =
+    refl
+  right-negative-law-add-ℤ (inr (inr zero-ℕ)) l =
+    inv (neg-pred-ℤ l)
+  right-negative-law-add-ℤ (inr (inr (succ-ℕ n))) l =
+    equational-reasoning
+      succ-ℤ (in-pos-ℤ n) +ℤ neg-ℤ l
+      ＝ succ-ℤ (in-pos-ℤ n +ℤ neg-ℤ l)
+        by left-successor-law-add-ℤ (in-pos-ℤ n) (neg-ℤ l)
+      ＝ succ-ℤ (neg-ℤ (neg-ℤ (inr (inr n)) +ℤ l))
+        by ap succ-ℤ (right-negative-law-add-ℤ (inr (inr n)) l)
+      ＝ neg-ℤ (pred-ℤ ((inl n) +ℤ l))
+        by inv (neg-pred-ℤ ((inl n) +ℤ l))
 ```
 
 ### Distributivity of negatives over addition
 
 ```agda
-distributive-neg-add-ℤ :
-  (k l : ℤ) → neg-ℤ (k +ℤ l) ＝ neg-ℤ k +ℤ neg-ℤ l
-distributive-neg-add-ℤ (inl zero-ℕ) l =
-  equational-reasoning
-    neg-ℤ (inl zero-ℕ +ℤ l)
-    ＝ neg-ℤ (pred-ℤ (zero-ℤ +ℤ l))
-      by ap neg-ℤ (left-predecessor-law-add-ℤ zero-ℤ l)
-    ＝ neg-ℤ (inl zero-ℕ) +ℤ neg-ℤ l
-      by neg-pred-ℤ l
-distributive-neg-add-ℤ (inl (succ-ℕ n)) l =
-  equational-reasoning
-    neg-ℤ (pred-ℤ (inl n +ℤ l))
-    ＝ succ-ℤ (neg-ℤ (inl n +ℤ l))
-      by neg-pred-ℤ (inl n +ℤ l)
-    ＝ succ-ℤ (neg-ℤ (inl n) +ℤ neg-ℤ l)
-      by ap succ-ℤ (distributive-neg-add-ℤ (inl n) l)
-    ＝ neg-ℤ (pred-ℤ (inl n)) +ℤ neg-ℤ l
-      by ap (_+ℤ (neg-ℤ l)) (inv (neg-pred-ℤ (inl n)))
-distributive-neg-add-ℤ (inr (inl star)) l =
-  refl
-distributive-neg-add-ℤ (inr (inr zero-ℕ)) l =
-  inv (pred-neg-ℤ l)
-distributive-neg-add-ℤ (inr (inr (succ-ℕ n))) l =
-  equational-reasoning
-    neg-ℤ (succ-ℤ (in-pos-ℤ n +ℤ l))
-    ＝ pred-ℤ (neg-ℤ (in-pos-ℤ n +ℤ l))
-      by inv (pred-neg-ℤ (in-pos-ℤ n +ℤ l))
-    ＝ pred-ℤ (neg-ℤ (inr (inr n)) +ℤ neg-ℤ l)
-      by ap pred-ℤ (distributive-neg-add-ℤ (inr (inr n)) l)
+abstract
+  distributive-neg-add-ℤ :
+    (k l : ℤ) → neg-ℤ (k +ℤ l) ＝ neg-ℤ k +ℤ neg-ℤ l
+  distributive-neg-add-ℤ (inl zero-ℕ) l =
+    equational-reasoning
+      neg-ℤ (inl zero-ℕ +ℤ l)
+      ＝ neg-ℤ (pred-ℤ (zero-ℤ +ℤ l))
+        by ap neg-ℤ (left-predecessor-law-add-ℤ zero-ℤ l)
+      ＝ neg-ℤ (inl zero-ℕ) +ℤ neg-ℤ l
+        by neg-pred-ℤ l
+  distributive-neg-add-ℤ (inl (succ-ℕ n)) l =
+    equational-reasoning
+      neg-ℤ (pred-ℤ (inl n +ℤ l))
+      ＝ succ-ℤ (neg-ℤ (inl n +ℤ l))
+        by neg-pred-ℤ (inl n +ℤ l)
+      ＝ succ-ℤ (neg-ℤ (inl n) +ℤ neg-ℤ l)
+        by ap succ-ℤ (distributive-neg-add-ℤ (inl n) l)
+      ＝ neg-ℤ (pred-ℤ (inl n)) +ℤ neg-ℤ l
+        by ap (_+ℤ (neg-ℤ l)) (inv (neg-pred-ℤ (inl n)))
+  distributive-neg-add-ℤ (inr (inl star)) l =
+    refl
+  distributive-neg-add-ℤ (inr (inr zero-ℕ)) l =
+    inv (pred-neg-ℤ l)
+  distributive-neg-add-ℤ (inr (inr (succ-ℕ n))) l =
+    equational-reasoning
+      neg-ℤ (succ-ℤ (in-pos-ℤ n +ℤ l))
+      ＝ pred-ℤ (neg-ℤ (in-pos-ℤ n +ℤ l))
+        by inv (pred-neg-ℤ (in-pos-ℤ n +ℤ l))
+      ＝ pred-ℤ (neg-ℤ (inr (inr n)) +ℤ neg-ℤ l)
+        by ap pred-ℤ (distributive-neg-add-ℤ (inr (inr n)) l)
 ```
 
 ### The inclusion of ℕ into ℤ preserves addition
 
 ```agda
-add-int-ℕ : (x y : ℕ) → (int-ℕ x) +ℤ (int-ℕ y) ＝ int-ℕ (x +ℕ y)
-add-int-ℕ x zero-ℕ = right-unit-law-add-ℤ (int-ℕ x)
-add-int-ℕ x (succ-ℕ y) =
-  equational-reasoning
-    int-ℕ x +ℤ int-ℕ (succ-ℕ y)
-    ＝ int-ℕ x +ℤ succ-ℤ (int-ℕ y)
-      by ap ((int-ℕ x) +ℤ_) (inv (succ-int-ℕ y))
-    ＝ succ-ℤ (int-ℕ x +ℤ int-ℕ y)
-      by right-successor-law-add-ℤ (int-ℕ x) (int-ℕ y)
-    ＝ succ-ℤ (int-ℕ (x +ℕ y))
-      by ap succ-ℤ (add-int-ℕ x y)
-    ＝ int-ℕ (succ-ℕ (x +ℕ y))
-      by succ-int-ℕ (x +ℕ y)
+abstract
+  add-int-ℕ : (x y : ℕ) → (int-ℕ x) +ℤ (int-ℕ y) ＝ int-ℕ (x +ℕ y)
+  add-int-ℕ x zero-ℕ = right-unit-law-add-ℤ (int-ℕ x)
+  add-int-ℕ x (succ-ℕ y) =
+    equational-reasoning
+      int-ℕ x +ℤ int-ℕ (succ-ℕ y)
+      ＝ int-ℕ x +ℤ succ-ℤ (int-ℕ y)
+        by ap ((int-ℕ x) +ℤ_) (inv (succ-int-ℕ y))
+      ＝ succ-ℤ (int-ℕ x +ℤ int-ℕ y)
+        by right-successor-law-add-ℤ (int-ℕ x) (int-ℕ y)
+      ＝ succ-ℤ (int-ℕ (x +ℕ y))
+        by ap succ-ℤ (add-int-ℕ x y)
+      ＝ int-ℕ (succ-ℕ (x +ℕ y))
+        by succ-int-ℕ (x +ℕ y)
 ```
 
 ### If `x + y = y` then `x = 0`
 
 ```agda
-is-zero-left-add-ℤ :
-  (x y : ℤ) → x +ℤ y ＝ y → is-zero-ℤ x
-is-zero-left-add-ℤ x y H =
-  equational-reasoning
-    x
-    ＝ x +ℤ zero-ℤ
-      by inv (right-unit-law-add-ℤ x)
-    ＝ x +ℤ (y +ℤ neg-ℤ y)
-      by inv (ap (x +ℤ_) (right-inverse-law-add-ℤ y))
-    ＝ (x +ℤ y) +ℤ neg-ℤ y
-      by inv (associative-add-ℤ x y (neg-ℤ y))
-    ＝ y +ℤ neg-ℤ y
-      by ap (_+ℤ (neg-ℤ y)) H
-    ＝ zero-ℤ
-      by right-inverse-law-add-ℤ y
+abstract
+  is-zero-left-add-ℤ :
+    (x y : ℤ) → x +ℤ y ＝ y → is-zero-ℤ x
+  is-zero-left-add-ℤ x y H =
+    equational-reasoning
+      x
+      ＝ x +ℤ zero-ℤ
+        by inv (right-unit-law-add-ℤ x)
+      ＝ x +ℤ (y +ℤ neg-ℤ y)
+        by inv (ap (x +ℤ_) (right-inverse-law-add-ℤ y))
+      ＝ (x +ℤ y) +ℤ neg-ℤ y
+        by inv (associative-add-ℤ x y (neg-ℤ y))
+      ＝ y +ℤ neg-ℤ y
+        by ap (_+ℤ (neg-ℤ y)) H
+      ＝ zero-ℤ
+        by right-inverse-law-add-ℤ y
 ```
 
 ### If `x + y = x` then `y = 0`
 
 ```agda
-is-zero-right-add-ℤ :
-  (x y : ℤ) → x +ℤ y ＝ x → is-zero-ℤ y
-is-zero-right-add-ℤ x y H =
-  is-zero-left-add-ℤ y x (commutative-add-ℤ y x ∙ H)
+abstract
+  is-zero-right-add-ℤ :
+    (x y : ℤ) → x +ℤ y ＝ x → is-zero-ℤ y
+  is-zero-right-add-ℤ x y H =
+    is-zero-left-add-ℤ y x (commutative-add-ℤ y x ∙ H)
 ```
 
 ## See also
--- a/src/elementary-number-theory/addition-integers.lagda.md
+++ b/src/elementary-number-theory/addition-integers.lagda.md
@@ -441,8 +441,8 @@ abstract
   is-injective-right-add-ℤ : (x : ℤ) → is-injective (_+ℤ x)
   is-injective-right-add-ℤ x = is-injective-is-emb (is-emb-right-add-ℤ x)
 
-  is-injective-add-ℤ : (x : ℤ) → is-injective (x +ℤ_)
-  is-injective-add-ℤ x = is-injective-is-emb (is-emb-left-add-ℤ x)
+  is-injective-left-add-ℤ : (x : ℤ) → is-injective (x +ℤ_)
+  is-injective-left-add-ℤ x = is-injective-is-emb (is-emb-left-add-ℤ x)
 ```
 
 ### Negative laws for addition
--- a/src/elementary-number-theory/addition-integers.lagda.md
+++ b/src/elementary-number-theory/addition-integers.lagda.md
@@ -330,39 +330,37 @@ abstract
 ### Interchange law for addition with respect to addition
 
 ```agda
-abstract
-  interchange-law-add-add-ℤ :
-    (x y u v : ℤ) → (x +ℤ y) +ℤ (u +ℤ v) ＝ (x +ℤ u) +ℤ (y +ℤ v)
-  interchange-law-add-add-ℤ =
-    interchange-law-commutative-and-associative
-      add-ℤ
-      commutative-add-ℤ
-      associative-add-ℤ
+interchange-law-add-add-ℤ :
+  (x y u v : ℤ) → (x +ℤ y) +ℤ (u +ℤ v) ＝ (x +ℤ u) +ℤ (y +ℤ v)
+interchange-law-add-add-ℤ =
+  interchange-law-commutative-and-associative
+    add-ℤ
+    commutative-add-ℤ
+    associative-add-ℤ
 ```
 
 ### Addition by `x` is a binary equivalence
 
 ```agda
-abstract
-  is-section-left-add-neg-ℤ :
-    (x y : ℤ) → x +ℤ (neg-ℤ x +ℤ y) ＝ y
-  is-section-left-add-neg-ℤ x y =
-    equational-reasoning
-      x +ℤ (neg-ℤ x +ℤ y)
-      ＝ (x +ℤ neg-ℤ x) +ℤ y
-        by inv (associative-add-ℤ x (neg-ℤ x) y)
-      ＝ y
-        by ap (_+ℤ y) (right-inverse-law-add-ℤ x)
+is-section-left-add-neg-ℤ :
+  (x y : ℤ) → x +ℤ (neg-ℤ x +ℤ y) ＝ y
+is-section-left-add-neg-ℤ x y =
+  equational-reasoning
+    x +ℤ (neg-ℤ x +ℤ y)
+    ＝ (x +ℤ neg-ℤ x) +ℤ y
+      by inv (associative-add-ℤ x (neg-ℤ x) y)
+    ＝ y
+      by ap (_+ℤ y) (right-inverse-law-add-ℤ x)
 
-  is-retraction-left-add-neg-ℤ :
-    (x y : ℤ) → (neg-ℤ x) +ℤ (x +ℤ y) ＝ y
-  is-retraction-left-add-neg-ℤ x y =
-    equational-reasoning
-      neg-ℤ x +ℤ (x +ℤ y)
-      ＝ (neg-ℤ x +ℤ x) +ℤ y
-        by inv (associative-add-ℤ (neg-ℤ x) x y)
-      ＝ y
-        by ap (_+ℤ y) (left-inverse-law-add-ℤ x)
+is-retraction-left-add-neg-ℤ :
+  (x y : ℤ) → (neg-ℤ x) +ℤ (x +ℤ y) ＝ y
+is-retraction-left-add-neg-ℤ x y =
+  equational-reasoning
+    neg-ℤ x +ℤ (x +ℤ y)
+    ＝ (neg-ℤ x +ℤ x) +ℤ y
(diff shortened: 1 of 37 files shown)


src/elementary-number-theory/greatest-common-divisor-integers.lagda.md
```diff
@@ -59,6 +59,22 @@ gcd-ℤ x y = int-ℕ (nat-gcd-ℤ x y)
 
 ## Properties
 
+### The greatest common divisor is invariant by neg-ℤ
```
fredrik-bakke, line 62:
```suggestion
### The greatest common divisor is invariant under negatives
```

src/elementary-number-theory/group-of-rational-numbers.lagda.md
fredrik-bakke, line 1: I'm just curious. While I agree that this is "the group" of rational numbers, where would you put the multiplicative group once that is formalized?

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -59,3 +69,113 @@ right-unit-law-mul-ℚ x =
...
+
+### Interchange law
+
+```agda
+interchange-law-mul-mul-ℚ : interchange-law mul-ℚ mul-ℚ
+interchange-law-mul-mul-ℚ =
+  interchange-law-commutative-and-associative
+    mul-ℚ
+    commutative-mul-ℚ
+    associative-mul-ℚ
```
fredrik-bakke, line 154: Very nice! Maybe it is worthwhile to spell out what the interchange law is, since it is not made explicit anywhere in the code here?

src/elementary-number-theory/positive-and-negative-integers.lagda.md
```diff
@@ -42,18 +43,12 @@ is-zero-is-nonnegative-is-nonpositive-ℤ :
...
-```
-
-### Negative integers are nonzero
-
-```agda
-is-nonzero-is-negative-ℤ : {x : ℤ} → is-negative-ℤ x → is-nonzero-ℤ x
-is-nonzero-is-negative-ℤ {inl x} H ()
+not-is-negative-is-positive-ℤ : (x : ℤ) → ¬ (is-negative-ℤ x × is-positive-ℤ x)
+not-is-negative-is-positive-ℤ (inl x) (H , K) = K
+not-is-negative-is-positive-ℤ (inr x) (H , K) = H
```
fredrik-bakke, line 51:
```suggestion
is-not-negative-and-positive-ℤ : (x : ℤ) → ¬ (is-negative-ℤ x × is-positive-ℤ x)
is-not-negative-and-positive-ℤ (inl x) (H , K) = K
is-not-negative-and-positive-ℤ (inr x) (H , K) = H
```
This reads better to me

src/elementary-number-theory/group-of-rational-numbers.lagda.md
malarbol: I did start with the name `additive-group-of-rational-numbers` but this looked better next to `group-of-integers` and I guess the multiplicative group will  be named `group-of-nonzero-rational-numbers'  (but that would require a nice `nonzero-rational-numbers` module ;p )

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -156,7 +166,66 @@ in-fraction-fraction-ℚ (pair (pair m (pair n n-pos)) p) =
...
+        ( left-zero-law-mul-ℤ (denominator-ℚ x))) ∙
+    in-fraction-fraction-ℚ zero-ℚ
+```
+
+### The rational image of the negative of an integer is the rational negative of its image
+
+```agda
+neg-ℚ-in-int-neg-ℤ : (k : ℤ) → in-int (neg-ℤ k) ＝ neg-ℚ (in-int k)
+neg-ℚ-in-int-neg-ℤ k =
+  eq-ℚ (in-int (neg-ℤ k)) (neg-ℚ (in-int k)) refl refl
```
fredrik-bakke, line 217: Oh. `in-int` is a bit strange to name the canonical inclusion of integers into rationals. Why not call it something like `rational-int` instead?

fredrik-bakke (approved):
Very nice!

src/elementary-number-theory/group-of-rational-numbers.lagda.md
fredrik-bakke: Aha. Although I'm not an authority on the matter, it seems more natural to me to call them the additive and multiplicative groups of rational numbers. It makes sense to call the additive group of integers "the group of integers" since there is no canonical alternative. On the other hand, if you were to talk of "the monoid of integers" I would not be sure which one you meant, and would expect you to clarify if it were the additive or multiplicative monoid (or some other).

fredrik-bakke: Perhaps it is better to call the file `additive-group-of-rational-numbers` then?

fredrik-bakke: We do have `monoid-of-natural-numbers-with-addition` and `monoid-of-natural-numbers-with-maximum`, but these names don't roll as well of the tongue

src/elementary-number-theory/rational-numbers.lagda.md
malarbol: It _is_ strange but it's been there for a long time. I think `rational-ℤ` would be even better to follow up with `int-ℕ`. Then I can rename my function `real-rational` to `real-ℚ` and we'll have a nice chain 
```
real-ℚ -- rational-ℤ -- int-ℕ
```

fredrik-bakke: ah, yes, that's probably better :)

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
fredrik-bakke: both please 😄

src/elementary-number-theory/group-of-rational-numbers.lagda.md
fredrik-bakke: It just seems to me like "additive group" and "multiplicative group" are more recognizable. With those names, the reader doesn't have to compare two module names against each other in order to figure out which is which. But it's not that important to me. If you prefer the current name then I'm okay with it.

malarbol: I convinced myself this was a better alternative so I may convince others, I don't know. I guess the `additive-group-of-rational-numbers` and `multiplicative-group-of-rational-numbers` names focus more of the operations while `group-of-rational-numbers` and `group-of-nonzero-rational-numbers` is about carrier-types. 
But even then, I think `multiplicative-group-of-nonzero-rational-numbers` would still be a better name: it would feel really weird to write "The rational numbers are a group under multiplication".

src/elementary-number-theory/ring-of-integers.lagda.md
```diff
@@ -52,6 +54,13 @@ pr2 ℤ-Commutative-Ring = commutative-mul-ℤ
 
 ## Properties
 
+### The ring of integers is nontrivial
+
+```agda
+is-nontrivial-ℤ-Ring : is-nontrivial-Ring ℤ-Ring
+is-nontrivial-ℤ-Ring H = is-nonzero-one-ℤ (inv H)
+```
+
```
malarbol, line 63: Do you mind if I change the `0 = 1` in the definition of trivial rings to `1 = 0`?
This would make this definition slightly cleaner and I think it would fit better with the design of the `elementary-number-theory` module which has much more `is-zero-XX` and `is-nonzero-XX` than `is-one-XX`.

src/elementary-number-theory/group-of-rational-numbers.lagda.md
fredrik-bakke: Indeed, `multiplicative-group-of-nonzero-rational-numbers` is even better

malarbol: Another problem I see is if we want to define the group structure of positive rational numbers under multiplication, it _would_ have to be called `multiplicative-group-of-positive-rational-numbers` so I'd rather go with
- `group-of-rational-numbers`
- `group-of-nonzero-rational-numbers`
- `group-of positive-rational-numbers`
 
I'm pretty sure the readers can guess which operation is used in each module. We could also include some `## See also` sections to cross-reference between each of them.

malarbol: >Indeed, multiplicative-group-of-nonzero-rational-numbers is even better

ok, we can also go with both the operation and the carrier type in the module name.

fredrik-bakke: oh, yeah, I didn't even think about the positive subgroup. That should follow from the ordering on the multiplicative group btw

fredrik-bakke: No worries, it's just good we determined the naming scheme 😄

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -59,3 +69,114 @@ right-unit-law-mul-ℚ x =
...
+  left-negative-law-mul-ℚ =
+    ( ap ( _*ℚ y) (inv (left-neg-unit-law-mul-ℚ x))) ∙
+    ( associative-mul-ℚ neg-one-ℚ x y) ∙
+    ( left-neg-unit-law-mul-ℚ (x *ℚ y))
+
+  right-negative-law-mul-ℚ : x *ℚ (neg-ℚ y) ＝ neg-ℚ (x *ℚ y)
+  right-negative-law-mul-ℚ =
+    ap ( x *ℚ_) (inv (right-neg-unit-law-mul-ℚ y)) ∙
+    inv (associative-mul-ℚ x y neg-one-ℚ) ∙
+    right-neg-unit-law-mul-ℚ (x *ℚ y)
```
fredrik-bakke, line 175:
```suggestion
    ( ap ( x *ℚ_) (inv (right-neg-unit-law-mul-ℚ y))) ∙
    ( inv (associative-mul-ℚ x y neg-one-ℚ)) ∙
    ( right-neg-unit-law-mul-ℚ (x *ℚ y))
```

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -59,3 +69,114 @@ right-unit-law-mul-ℚ x =
...
+
+negative-law-mul-ℚ : (x y : ℚ) → (neg-ℚ x) *ℚ (neg-ℚ y) ＝ x *ℚ y
+negative-law-mul-ℚ x y =
+  left-negative-law-mul-ℚ x (neg-ℚ y) ∙
+  ap neg-ℚ (right-negative-law-mul-ℚ x y) ∙
+  neg-neg-ℚ (x *ℚ y)
```
fredrik-bakke, line 181:
```suggestion
  ( left-negative-law-mul-ℚ x (neg-ℚ y)) ∙
  ( ap neg-ℚ (right-negative-law-mul-ℚ x y)) ∙
  ( neg-neg-ℚ (x *ℚ y))
```

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -156,7 +166,66 @@ in-fraction-fraction-ℚ (pair (pair m (pair n n-pos)) p) =
...
+  where
+
+  eq-ℚ : x ＝ y
+  eq-ℚ =
+    inv (in-fraction-fraction-ℚ x) ∙
+    eq-ℚ-sim-fraction-ℤ
+      ( fraction-ℚ x)
+      ( fraction-ℚ y)
+      ( ap-mul-ℤ H (inv K)) ∙
+    in-fraction-fraction-ℚ y
```
fredrik-bakke, line 185:
```suggestion
    ( inv (in-fraction-fraction-ℚ x)) ∙
    ( eq-ℚ-sim-fraction-ℤ
      ( fraction-ℚ x)
      ( fraction-ℚ y)
      ( ap-mul-ℤ H (inv K))) ∙
    ( in-fraction-fraction-ℚ y)
```

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
malarbol: oh. sorry. The precedence of the `∙` operator made these parentheses feel unnecessary so I didn't put them.
There are a lot of similar blocks everywhere. I'll try to fix them ASAP.

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -77,21 +79,21 @@ pr2 (in-fraction-ℤ x) = is-reduced-reduce-fraction-ℤ x
 ### Inclusion of the integers
 
 ```agda
-in-int : ℤ → ℚ
-in-int x = pair (pair x one-positive-ℤ) (is-one-gcd-one-ℤ' x)
+rational-ℤ : ℤ → ℚ
+rational-ℤ x = pair (pair x one-positive-ℤ) (is-one-gcd-one-ℤ' x)
```
fredrik-bakke, line 83:
```suggestion
rational-ℤ x = (x , one-positive-ℤ , is-one-gcd-one-ℤ' x)
```
For single-line expressions we prefer commas these days.

fredrik-bakke:
I noticed there's an `in-fraction-Z` function too. Should that be renamed to `rational-fraction-Z`?

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -156,7 +166,66 @@ in-fraction-fraction-ℚ (pair (pair m (pair n n-pos)) p) =
...
+    is-zero-ℤ (numerator-ℚ x) → is-zero-ℚ x
+  is-zero-is-zero-numerator-ℚ H =
+    inv (in-fraction-fraction-ℚ x) ∙
+    eq-ℚ-sim-fraction-ℤ
+      ( fraction-ℚ x)
+      ( fraction-ℚ zero-ℚ)
+      ( eq-is-zero-ℤ
+        ( ap (mul-ℤ' one-ℤ) H ∙ right-zero-law-mul-ℤ one-ℤ)
+        ( left-zero-law-mul-ℤ (denominator-ℚ x))) ∙
+    in-fraction-fraction-ℚ zero-ℚ
```
fredrik-bakke, line 209:
```suggestion
    ( inv (in-fraction-fraction-ℚ x)) ∙
    ( eq-ℚ-sim-fraction-ℤ
      ( fraction-ℚ x)
      ( fraction-ℚ zero-ℚ)
      ( eq-is-zero-ℤ
        ( ap (mul-ℤ' one-ℤ) H ∙ right-zero-law-mul-ℤ one-ℤ)
        ( left-zero-law-mul-ℤ (denominator-ℚ x)))) ∙
    ( in-fraction-fraction-ℚ zero-ℚ)
```

src/elementary-number-theory/ring-of-integers.lagda.md
fredrik-bakke: Hmm, I'm not sure about that change. While `1 = 0` may be more natural in elementary number theory, that's not the only place ring theory is used. We define nontrivial groups to be those that have an element `g` such that `1 ≠ g`. While this seems consistent with your suggestion, that is only because we use multiplicative notation for groups by convention. The underlying group of a ring is nontrivial if we have `0 ≠ r` for some `r`. Likewise, nontrivial nonunital rings would probably have the same definition. On the other hand, the underlying multiplicative monoid of a ring is only nontrivial if `1 ≠ r`, which again supports your suggestion, so I don't know.

fredrik-bakke (approved):

fredrik-bakke:
> I noticed there's an `in-fraction-Z` function too. Should that be renamed to `rational-fraction-Z`?

Oh, also, `in-neg` and `in-pos` apparently! How about `neg-int-ℕ` for the former and remove the latter? We already have `int-ℕ`

src/elementary-number-theory/ring-of-integers.lagda.md
fredrik-bakke: I'm inclined to ask that we leave it as is for now. I trust this doesn't cause trouble for you?

malarbol:
> Oh, also, `in-neg` and `in-pos` apparently! How about `neg-int-ℕ` for the former and remove the latter? We already have `int-ℕ`

And I already used the names 
```agda
positive-int-ℕ : ℕ → positive-ℤ
negative-int-ℕ : ℕ → negative-ℤ
```
so `pos-int-N` and `neg-int-N` look like a good choice.  I can also make their relationship with `positive-int-ℕ` and `negative-int-ℕ` more explicit.

src/elementary-number-theory/ring-of-integers.lagda.md
malarbol: >I'm inclined to ask that we leave it as is for now. I trust this doesn't cause trouble for you?

No trouble at all, it was more an aesthetic request. I'm still getting to know the library so this felt a bit weird but I guess inverting equality relations will always happen everywhere anyway.

src/elementary-number-theory/integer-fractions.lagda.md
```diff
@@ -67,22 +67,22 @@ is-positive-denominator-fraction-ℤ x = pr2 (positive-denominator-fraction-ℤ
 ### Inclusion of the integers
 
 ```agda
-in-fraction-ℤ-ℤ : ℤ → fraction-ℤ
-pr1 (in-fraction-ℤ-ℤ x) = x
-pr2 (in-fraction-ℤ-ℤ x) = one-positive-ℤ
+rational-fraction-ℤ-ℤ : ℤ → fraction-ℤ
+pr1 (rational-fraction-ℤ-ℤ x) = x
+pr2 (rational-fraction-ℤ-ℤ x) = one-positive-ℤ
```
fredrik-bakke, line 72: This should still be `in-fraction-ℤ-ℤ`, unless you have a better suggestion

src/real-numbers/rational-real-numbers.lagda.md
```diff
@@ -220,7 +220,7 @@ eq-real-rational-is-rational-ℝ x q H =
 
 ```agda
 rational-ℝ-rational : ℚ → Rational-ℝ lzero
-rational-ℝ-rational q = real-rational q , q , is-rational-real-rational q
+rational-ℝ-rational q = (real-ℚ q , q , is-rational-real-ℚ q)
```
fredrik-bakke, line 223: Ah, so regarding our naming scheme. So far, the capital blackboard boldface letters have been treated as "namespaces" for the different types. E.g. `ℚ` is the namespace of rational numbers, `ℝ` is the namespace of real numbers and so on. Our rules for namespaces is that they always go at the end of the entry name, and an entry has at most one namespace. E.g. `real-ℚ` is in the namespace `ℚ`. The important thing is that, namespaces are the only part of an entry name that may have unicode and capital letters. (This is of course aside from special notations.)

This means that the use of `ℝ` in this entry name is unconventional. Instead, this entry should be called `rational-real-ℚ`

fredrik-bakke: This naming scheme is not completely developed yet though. For instance, I think it is fine to have `ℚ-Group`, although one could interpret this as two namespaces.

fredrik-bakke: This is also the reason why `leq-ℚ-Prop` should be called `leq-prop-ℚ` in the future by the way!

src/elementary-number-theory/integer-fractions.lagda.md
fredrik-bakke: Or, better yet, `int-fraction-ℤ`

fredrik-bakke: By the way, which editor are you using? I suspect this error came from a naive search-and-replace of `in-fraction-ℤ`. If you're using vscode, then you can avoid matching too many things with regular expressions like, for example, the search string `in-fraction-ℤ(?!-)` will not match if `in-fraction-ℤ` is followed by a `-`. Likewise `(?<!-)in-fraction-ℤ` will not match if it is preceded by a `-`.

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
fredrik-bakke: To be honest, I'm not even sure if we maintainers agree on whether these should be parenthesized or not anymore. I always find it looks better with them, and I do believe it is stated in our code conventions that they should be there, but it is certainly not a big deal.

src/real-numbers/rational-real-numbers.lagda.md
malarbol: Thanks for all the explanation!

src/elementary-number-theory/integer-fractions.lagda.md
malarbol: I'm using [spacemacs](https://www.spacemacs.org) but this _was_ a "search and replace gone wrong" so thanks for the tip. I may ask you for more tips if you're around next time I need a regex though ;p

malarbol:
> Oh, also, in-neg and in-pos apparently! How about neg-int-ℕ for the former and remove the latter? We already have int-ℕ

Sorry, I responded a first time, then realized I had misread you and I needed a few more thoughts no this:

- `in-pos` is not a replacement for `int-N` (`in-pos` sends `zero-N` to `one-Z` and `int-N` sends it to `zero-Z`) so that doesn't really work.
- `neg-int-N` reads as it should be equal to `neg-Z ∘ int-N` which it's not, so it might be a bit confusing.

I already used `negative-int-ℕ : ℕ → negative-ℤ` and `positive-int-ℕ : ℕ → positive-ℤ` in the positive/negative-integers module. What do you think of the names:
```
in-pos --> int-positive-int-N
in-neg --> int-negative-int-N
```
so then we could have a property like this in the positive/negative integers module:
```
int-positive-int-ℕ ∼ int-positive-ℤ ∘ positive-int-ℕ
```
which is a sort of  "commuting triangles" between N, Z and positive-Z that records that `int-positive-int-N` goes into `positive-Z` and then forgets it.

fredrik-bakke:
> Sorry, I responded a first time, then realized I had misread you and I needed a few more thoughts no this [...]

Ah, indeed I misunderstood the definitions. I like your suggestion, but now I am a bit unsure if these should be in the natural numbers namespace at all. After all, they're about specific constructors for the integers, and don't have much to do with the natural numbers. But unless we come up with a better names I think we should go with your suggestions

malarbol:
> Ah, indeed I misunderstood the definitions. I like your suggestion, but now I am a bit unsure if these should be in the natural numbers namespace at all. After all, they're about specific constructors for the integers, and don't have much to do with the natural numbers. But unless we come up with a better names I think we should go with your suggestions

We can also leave them as they are until a new good name is settled. They are used in group/ring theory for properties like `integer-power-in-pos-Group`, `integer-multiple-in-pos-Ring`, `integer-multiple-in-neg-Ab` etc. which I guess should have to be renamed too, and I'm not sure `integer-power-int-positive-int-Group` would be such a good name. 

We could also go with `int-pos` and `int-neg` to keep it "simple" while fitting better with the rest of the names (where the beginning of the name kind of indicates the "returned type"). Or maybe `int-pos-N` and `int-neg-N` if you finally decide to put them in the natural numbers namespace? IMH, they still are some kind of "natural maps" from N to Z (induction principles on (one-Z, succ-Z) and (neg-one-Z, pred-Z)) so I think they could be part of the 'int-XXX-N' family.

src/elementary-number-theory/group-of-rational-numbers.lagda.md
malarbol: If you don't mind, I suggest we just keep `group-of-rational-numbers` for now. I understand your argument and I'd agree if it were `monoid-of-rational-numbers` but I'm still unconvinced that it would create any confusion on the group structure. 

And maybe we won't want to add modules for the multiplicative groups at all? Once we have a division ring structure on the rationals, the group of nonzero rational numbers is the multiplicative group of units of the ring, from which the group of positive rationals is a subgroup (contains `one-Q` and stable under `mul-Q`) .So maybe we could define things like `nonzero-Q-mul-Group`, `positive-Q-mul-Group`, `nonzero-Q-mul-Ab` and `positive-Q-mul-Ab` directly in the `discrete-field-of-rational-numbers` module?

malarbol:
I see you've opened issue #1102. I don't know if it's related to this PR but I must say I _did_ experience some performance issue on the `group-of-rational-numbers` module. The negative rationals and inverse additive law seem to slow down the type-checking quite a bit (from instantaneous / few seconds to tens of seconds).

As you know, I'm planning to follow up with two more PRs :

 - The ring structure on the rational numbers
   (add distributivity of multiplication over addition, the rest is already all here)
 - The discrete field of rational numbers
   (inverse of nonzero rational numbers; this adds quite a few more code, we'll talk about it later)

and the performance issue seem even bigger (e.g. more than a minute to load the `discrete-field-[...]` module).

I don't know if t's normal (after all, the complexity of the types involved takes a few steps up from `Z --> fraction-Z --> Q`), or if there is something wrong with my code (maybe some optimizations I'm missing?) ; I don't know either if having "equality proofs abstracted in the `elementary-number-theory` module" would solve these issues, but I thought you should know.

fredrik-bakke:
> I see you've opened issue #1102. I don't know if it's related to this PR but I must say I _did_ experience some performance issue on the `group-of-rational-numbers` module. The negative rationals and inverse additive law seem to slow down the type-checking quite a bit (from instantaneous / few seconds to tens of seconds).
> 
> As you know, I'm planning to follow up with two more PRs :
> 
> * The ring structure on the rational numbers
>   (add distributivity of multiplication over addition, the rest is already all here)
> * The discrete field of rational numbers
>   (inverse of nonzero rational numbers; this adds quite a few more code, we'll talk about it later)
> 
> and the performance issue seem even bigger (e.g. more than a minute to load the `discrete-field-[...]` module).
> 
> I don't know if t's normal (after all, the complexity of the types involved takes a few steps up from `Z --> fraction-Z --> Q`), or if there is something wrong with my code (maybe some optimizations I'm missing?) ; I don't know either if having "equality proofs abstracted in the `elementary-number-theory` module" would solve these issues, but I thought you should know.

Oh, thanks for the input! It was indeed something that came to mind while looking over your PR. Then I believe for sure that we should start to incorporate `abstract`, but that can be done in a separate PR. I am also happy to do this, but I don't want to do it while you have an open pull request, as I expect it will cause some merging conflicts. It sounds for sure like `abtract`ion will help with the issues you're experiencing.

I will be traveling next week btw, so I'm unsure how much I will be available then.

fredrik-bakke:
In general, almost all proofs of propositions can be `abstract`ed without issue, so that's what you should look for first.

fredrik-bakke:
> > Ah, indeed I misunderstood the definitions. I like your suggestion, but now I am a bit unsure if these should be in the natural numbers namespace at all. After all, they're about specific constructors for the integers, and don't have much to do with the natural numbers. But unless we come up with a better names I think we should go with your suggestions
> 
> We can also leave them as they are until a new good name is settled. They are used in group/ring theory for properties like `integer-power-in-pos-Group`, `integer-multiple-in-pos-Ring`, `integer-multiple-in-neg-Ab` etc. which I guess should have to be renamed too, and I'm not sure `integer-power-int-positive-int-Group` would be such a good name.
> 
> We could also go with `int-pos` and `int-neg` to keep it "simple" while fitting better with the rest of the names (where the beginning of the name kind of indicates the "returned type"). Or maybe `int-pos-N` and `int-neg-N` if you finally decide to put them in the natural numbers namespace? IMH, they still are some kind of "natural maps" from N to Z (induction principles on `(one-Z, succ-Z)` and `(neg-one-Z, pred-Z)`) so I think they could be part of the 'int-XXX-N' family.

I think your suggestion is at least better than the current names. `in-pos` and `in-neg` seem very out of place to me.

malarbol:
>In general, almost all proofs of propositions can be abstracted without issue, so that's what you should look for first.

ok. I'll abstract some of the equality properties used for the additive group structure (associativity, unit laws, etc.), see if it helps. Maybe we can do the full `elementary-number-theory` module in a separate PR after we merge this one.

I can also add the full commutative ring structure in this PR if you want (it's just a few hundreds LOC more for the distributivity rules), then we `abstract` all the necessary identities to solve the remaining performance issues, and I'll PR the `discrete-field-of-rational-numbers` module after that? (all this after your back from travelling of course).

fredrik-bakke:
If this PR is getting ready we can try to have it merged tonight? I'll be back in about 3 hours, and can make a final review then if it works for you

malarbol:
> If this PR is getting ready we can try to have it merged tonight? I'll be back in about 3 hours, and can make a final review then if it works for you

I'll go out for dinner too. Feel free to apply any change you deem necessary in the meantime. I'll address the rest of your review when I come back so we can merge soon.

malarbol:
> > > Ah, indeed I misunderstood the definitions. I like your suggestion, but now I am a bit unsure if these should be in the natural numbers namespace at all. After all, they're about specific constructors for the integers, and don't have much to do with the natural numbers. But unless we come up with a better names I think we should go with your suggestions
> > 
> > 
> > We can also leave them as they are until a new good name is settled. They are used in group/ring theory for properties like `integer-power-in-pos-Group`, `integer-multiple-in-pos-Ring`, `integer-multiple-in-neg-Ab` etc. which I guess should have to be renamed too, and I'm not sure `integer-power-int-positive-int-Group` would be such a good name.
> > We could also go with `int-pos` and `int-neg` to keep it "simple" while fitting better with the rest of the names (where the beginning of the name kind of indicates the "returned type"). Or maybe `int-pos-N` and `int-neg-N` if you finally decide to put them in the natural numbers namespace? IMH, they still are some kind of "natural maps" from N to Z (induction principles on `(one-Z, succ-Z)` and `(neg-one-Z, pred-Z)`) so I think they could be part of the 'int-XXX-N' family.
> 
> I think your suggestion is at least better than the current names. `in-pos` and `in-neg` seem very out of place to me.

Which suggestion 😅 ? I'd personally go with `int-pos-N` and then rename `integer-power-in-pos-Group` to `integer-power-int-pos-Group` (or maybe `int-power-int-pos-Group`?), etc. There's also this N-namespace issue that's still unclear to me. Change it however you feel like.

src/elementary-number-theory/group-of-rational-numbers.lagda.md
malarbol: I let you rename the module, if you want, and close the issue.

fredrik-bakke:
Hey again, sorry, but I'm a bit more exhausted than I anticipated and won't be able to do a good review tonight. I will get back to you tomorrow though

fredrik-bakke:
Hey! I'm looking through the files now, and, to keep things short, I'm pushing directly to your branch. That doesn't mean that I think every change I make is an improvement without question. Do let me know if you disagree with any of my changes. Otherwise, let me know when you've looked over them and then we can merge. (I will submit a review when I am done)

fredrik-bakke:
I'm having trouble type-checking `difference-rational-numbers` in that it is taking extremely long. Is this one of your problem files wrt. performance?

fredrik-bakke:
These modules are taking so long to type check now that I don't think it is a good idea to merge this PR before we have this resolved issue. I'm looking to see if abstracting definitions helps at the moment.

fredrik-bakke:
Alright, I had a go at abstracting most definitions, but it did not resolve the issue. However, enabling the `lossy-unification` flag did. This may mean that there are excessively large terms that are being unfolded before Agda is able to determine that two things are the same. If I am not mistaken, this may be part of the process of resolving implicit arguments. Anyway, that provides you with two possible hints to what is causing the performance issue. For some other time, there is always Agda's [performance profiling functionality](https://agda.readthedocs.io/en/latest/tools/performance.html) to help determine the issue. In particular, the flag `--profile=internal` will give a detailed view of how much time each step of the type-checking process is taking.

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -69,29 +71,31 @@ module _
 ### Inclusion of fractions
 
 ```agda
-in-fraction-ℤ : fraction-ℤ → ℚ
-pr1 (in-fraction-ℤ x) = reduce-fraction-ℤ x
-pr2 (in-fraction-ℤ x) = is-reduced-reduce-fraction-ℤ x
+rational-fraction-ℤ : fraction-ℤ → ℚ
+pr1 (rational-fraction-ℤ x) = reduce-fraction-ℤ x
+pr2 (rational-fraction-ℤ x) = is-reduced-reduce-fraction-ℤ x
```
fredrik-bakke, line 76: I'm not sure about this change. I wonder if it is better to call it `in-fraction-ℚ`

fredrik-bakke:
I'm liking all the headers you are adding btw!

src/elementary-number-theory/rational-numbers.lagda.md
fredrik-bakke: There seem to be two naming schemes conflicting. I have to think a bit if there is a way to unify them.

fredrik-bakke:
I won't be able to finish this review tonight, and sorry for making a bit of a mess. The way I saw it, things had to be abstracted now if the PR was to be merged. But since it was resolved by `lossy-unification`, we can consider reverting"`abstract`ageddon".

malarbol:
Sorry, I was a bit busier than expected this week-end and I couldn't answer you sooner.

>I'm having trouble type-checking difference-rational-numbers in that it is taking extremely long. Is this one of your problem files wrt. performance?

I believe the problem already appeared with negatives and "inverse additive laws" but I'm not sure.

>These modules are taking so long to type check now that I don't think it is a good idea to merge this PR before we have this issue resolved. I'm looking to see if abstracting definitions helps at the moment.

>Alright, I had a go at abstracting most definitions, but it did not resolve the issue. However, enabling the lossy-unification flag did. This may mean that there are excessively large terms that are being unfolded before Agda is able to determine that two things are the same. If I am not mistaken, this may be part of the process of resolving implicit arguments. Anyway, that provides you with two possible hints to what is causing the performance issue. For some other time, there is always Agda's [performance profiling functionality](https://agda.readthedocs.io/en/latest/tools/performance.html) to help determine the issue. In particular, the flag --profile=internal will give a detailed view of how much time each step of the type-checking process is taking.

>But since it was resolved by lossy-unification, we can consider reverting"abstractageddon".

Thanks a lot for digging deeper into this! And thanks for the tip on profiling, I'll try to look into it.
Let me know what you want to do with all the `abstract`ed definitions then. I can go removing them.


>Hey! I'm looking through the files now, and, to keep things short, I'm pushing directly to your branch. That doesn't mean that I think every change I make is an improvement without question. Do let me know if you disagree with any of my changes.

No problem. We'll keep discussing the changes until we both agree.

src/elementary-number-theory/rational-numbers.lagda.md
malarbol: I think this is one is fine and fits well in the chain
```
N --int-N--> Z --int-fraction-Z--> fraction-Z --rational-fraction-Z--> Q --real-Q--> R 
```
So I'd put back `int-fraction-ℤ <- in-fraction-ℤ` (to be parsed as `int-fraction` `ℤ`, so a function from `ℤ` to `fraction-ℤ`) and leave this one like it is (a function from `fraction-ℤ` to `Q`.

This puts `in-(neg|pos)-ℤ` in a weird place but I think this is because their "true" place in this naming scheme would be the functions `int-positive-N` and `int-negative-N` (as functions from `N` to `positive-Z` and `negative-Z`). They could fit in this scheme as `int-XXX-N` where `XXX` describes the mappings (hence my suggestions of `int-(pos|neg)-N` or maybe `int-in-(pos|neg)-N`).

This about it, or maybe ask around to the other mainteners/contributors, and we'll settle on something.

malarbol: We could also consider removing them; now that we have modules for positive and negative integers it could also benefit some other modules. For exemple, maybe the properties

```agda
  integer-power-in-pos-Group :
    (n : ℕ) (g : type-Group G) →
    integer-power-Group G (in-pos-ℤ n) g ＝
    power-Group G (succ-ℕ n) g

  integer-power-in-neg-Group :
    (n : ℕ) (g : type-Group G) →
    integer-power-Group G (in-neg-ℤ n) g ＝
    inv-Group G (power-Group G (succ-ℕ n) g)
```

could be refactored using `positive-Z` and `negative-Z` (and abelian groups, rings, etc.). This would imply a lot of changes though, so even if you end up choosing to remove them, I suggest to leave it for another Issue/PR. In the meantime, we can chose a "not-too-bad" name for them and keep on.

malarbol:
I know you're travelling this week so there's no hurry. We'll resume this conversation when you're back.

src/elementary-number-theory/multiplication-integers.lagda.md
```diff
@@ -423,6 +461,8 @@ linear-diff-right-mul-ℤ x y z =
   ( inv (right-distributive-mul-add-ℤ x (neg-ℤ y) z))
```
malarbol, line 461: I think it would be cleaner to call this 
`### Multiplication on integer fractions distributes over the difference`
and have the equalities reversed (they are mostly used with `inv` anyway). Is this ok for you?

fredrik-bakke: Sure

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -49,13 +59,124 @@ left-unit-law-mul-ℚ x =
...
 right-unit-law-mul-ℚ x =
   ( eq-ℚ-sim-fraction-ℤ
     ( mul-fraction-ℤ (fraction-ℚ x) one-fraction-ℤ)
     ( fraction-ℚ x)
     ( right-unit-law-mul-fraction-ℤ (fraction-ℚ x))) ∙
-  ( in-fraction-fraction-ℚ x)
+  ( is-retraction-rational-fraction-ℚ x)
+```
+
+### Negative unit laws for multiplication on rational numbers
```
fredrik-bakke, line 73: These aren't exactly unit laws.


```suggestion
### Computing multiplication by `-1`
```

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -135,32 +147,115 @@ eq-ℚ-sim-fraction-ℤ x y H =
...
+          ( left-zero-law-mul-ℤ (denominator-ℚ x)))) ∙
+      ( is-retraction-rational-fraction-ℚ zero-ℚ)
+```
+
+### The rational image of the negative of an integer is the rational negative of its image
+
+```agda
+abstract
+  neg-rational-ℤ-neg-ℤ : (k : ℤ) → rational-ℤ (neg-ℤ k) ＝ neg-ℚ (rational-ℤ k)
+  neg-rational-ℤ-neg-ℤ k =
```
fredrik-bakke, line 224: The name of this one doesn't quite make sense to me. The theorem is showing that `rational-ℤ` preserves negatives. Perhaps `preserves-neg-rational-ℤ` is a good name?

src/elementary-number-theory/rational-numbers.lagda.md
fredrik-bakke: It's not at all obvious to me that we would want to remove these maps ever. I think the current name is fine for now.

src/elementary-number-theory/strict-inequality-integers.lagda.md
```diff
@@ -179,3 +179,74 @@ reflects-le-right-add-ℤ :
 reflects-le-right-add-ℤ z x y =
   is-positive-eq-ℤ (left-translation-diff-ℤ y x z)
 ```
+
+### An integer `x` is positive if and only if `le-ℤ zero-ℤ x`
```
fredrik-bakke, line 183: I don't see why you wouldn't prefer:

```suggestion
### An integer `x` is positive if and only if `0 < x`
```

src/elementary-number-theory/strict-inequality-integers.lagda.md
```diff
@@ -179,3 +179,74 @@ reflects-le-right-add-ℤ :
...
+      is-positive-le-zero-ℤ y
+        ( transitive-le-ℤ
+          ( zero-ℤ)
+          ( x)
+          ( y)
+          ( I)
+          ( le-zero-is-positive-ℤ x H))
+```
+
+### An integer `x` is negative if and only if `le-ℤ x zero-ℤ
```
fredrik-bakke, line 217:
```suggestion
### An integer `x` is negative if and only if `x < 0`
```

src/elementary-number-theory/strict-inequality-integers.lagda.md
```diff
@@ -179,3 +179,74 @@ reflects-le-right-add-ℤ :
...
+    le-zero-is-negative-ℤ = is-positive-neg-is-negative-ℤ
+
+    is-negative-le-zero-ℤ : le-ℤ x zero-ℤ → is-negative-ℤ x
+    is-negative-le-zero-ℤ H =
+      is-negative-eq-ℤ
+        ( neg-neg-ℤ x)
+        ( is-negative-neg-is-positive-ℤ H)
+```
+
+### An integer lower than a negative integer is negative
```
fredrik-bakke, line 235:
```suggestion
### If an integer is lesser than a negative integer it is negative
```

src/elementary-number-theory/strict-inequality-integers.lagda.md
```diff
@@ -179,3 +179,74 @@ reflects-le-right-add-ℤ :
...
+
+  abstract
+    le-zero-is-positive-ℤ : is-positive-ℤ x → le-ℤ zero-ℤ x
+    le-zero-is-positive-ℤ = is-positive-eq-ℤ (inv (right-zero-law-diff-ℤ x))
+
+    is-positive-le-zero-ℤ : le-ℤ zero-ℤ x → is-positive-ℤ x
+    is-positive-le-zero-ℤ = is-positive-eq-ℤ (right-zero-law-diff-ℤ x)
+```
+
+### An integer greater than a positive integer is positive
```
fredrik-bakke, line 198:
```suggestion
### If an integer is greater than a positive integer it is positive
```

fredrik-bakke:
I wrote an issue about abstracting stuff 👍 It would probably be nice to just have that done so we can move on with other stuff, but it's your choice if you want to do it or not.

fredrik-bakke:
I have a slightly bad feeling about certain naming schemes that are still used in `elementary-number-theory`. In general, if a property records a coherence, or a distributivity property, or some such thing, it should have a descriptor like `coherence`, `triangle`, `distributive`, `compute`, `preserves` in its name to emphasize that and to make it more readable. I'm writing this without pointing to a particular entry because I don't want to go through all that is changed in this PR one more time to find an example, but it is something to keep in mind going forward.

fredrik-bakke:
Also, regarding the canonical inclusions, I think the current names are good enough for now. They can always be changed again in the future

src/elementary-number-theory/strict-inequality-integers.lagda.md
malarbol: I do prefer this. I was not sure it was ok, or if I should have written `<-ℤ` , which didn't look as nice; so I went with `le-ℤ`.

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
malarbol: yes, they are more negative-unit laws (instead of negative unit-laws); you have a negative function, a unit, and you prove that multiplication by the "negative unit" is the same as taking the negative. But we can change it if you want. And I guess we should change the corresponding `integers` header too then.

src/elementary-number-theory/rational-numbers.lagda.md
malarbol: ok. Thanks for the input. I must confess I didn't think too much about the names. I'll try to give it more attention.
I guess there could be a whole family of `preserves-[op]-[inclusion]` for the hierarchy `N - Z - Q` (like `preserves-add-int-N`, etc.)

malarbol:
> I wrote an issue about abstracting stuff 👍 It would probably be nice to just have that done so we can move on with other stuff, but it's your choice if you want to do it or not.

since you mention you want to do it "as an isolated pull request", I understand you mean "get this done right after we merge this"? What would be the plan then? (e.g. what should be abstracted or not, etc.)

>I have a slightly bad feeling about certain naming schemes that are still used in elementary-number-theory. In general, if a property records a coherence, or a distributivity property, or some such thing, it should have a descriptor like coherence, triangle, distributive, compute, preserves in its name to emphasize that and to make it more readable. I'm writing this without pointing to a particular entry because I don't want to go through all that is changed in this PR one more time to find an example, but it is something to keep in mind going forward.

Ok. Sorry. I'll keep that in mind for the future. I'll try to give it another read and see if I spot any name that could be improved this way.

src/elementary-number-theory/addition-integers.lagda.md
```diff
@@ -415,141 +419,152 @@ pr2 is-binary-equiv-left-add-ℤ = is-equiv-left-add-ℤ
...
-    ＝ int-ℕ x +ℤ succ-ℤ (int-ℕ y)
-      by ap ((int-ℕ x) +ℤ_) (inv (succ-int-ℕ y))
-    ＝ succ-ℤ (int-ℕ x +ℤ int-ℕ y)
-      by right-successor-law-add-ℤ (int-ℕ x) (int-ℕ y)
-    ＝ succ-ℤ (int-ℕ (x +ℕ y))
-      by ap succ-ℤ (add-int-ℕ x y)
-    ＝ int-ℕ (succ-ℕ (x +ℕ y))
-      by succ-int-ℕ (x +ℕ y)
+abstract
+  add-int-ℕ : (x y : ℕ) → (int-ℕ x) +ℤ (int-ℕ y) ＝ int-ℕ (x +ℕ y)
```
malarbol, line 524: Should this be `preserves-add-int-ℕ` then?

src/elementary-number-theory/integer-fractions.lagda.md
```diff
@@ -191,21 +201,62 @@ pr1 (pr2 (pr2 equivalence-relation-sim-fraction-ℤ)) = symmetric-sim-fraction-
...
 
+### The negatives of two similar integer fractions are similar
+
+```agda
+module _
+  (x y : fraction-ℤ)
+  where
+
+  abstract
+    sim-neg-sim-fraction-ℤ :
```
malarbol, line 212: Rename --> `preserves-sim-neg-fraction-ℤ`?

src/elementary-number-theory/reduced-integer-fractions.lagda.md
```diff
@@ -64,6 +64,22 @@ is-prop-is-reduced-fraction-ℤ x =
     ( denominator-fraction-ℤ x)
 ```
 
+### The negative of a reduced integer fraction is reduced
+
+```agda
+is-reduced-neg-is-reduced-fraction-ℤ :
```
malarbol, line 70: I guess this could be renamed `preserves-is-reduced-neg-fraction-ℤ` but I think it's also fine like this. As you wish.

src/elementary-number-theory/group-of-rational-numbers.lagda.md
malarbol: Any update on this? I still think `group-of-rational-numbers` is ok.

src/elementary-number-theory/reduced-integer-fractions.lagda.md
fredrik-bakke: I think it is fine as it is currently. I don't think we usually add a verb before `is-` or `has`-predicates.

src/elementary-number-theory/addition-integers.lagda.md
fredrik-bakke: So, an issue, with going around and adding `preserves` and similar verbs to these definitions is that it is likely to introduce inconsistencies. Sure, `preserves-add-int-ℕ` may be preferred, but then what about `add-in-fraction-ℤ` and `mul-real-rational`? (I'm just making up names) I think what we should do is try to be consistent first, and then when a clear better naming scheme comes up, we try to port larger parts of the module at the same time instead, maintaining broad consistency.

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
fredrik-bakke: So, a big difference between how we approach terminology use in agda-unimath as opposed to, say, an article or a small formalization project, or even the TypeTopology library, is that it has to be "globally consistent". Terminology introduced and used on one page has to be consistent with that terminology on any other page, and if two concepts are the same, they should probably also use the same terminology. In contrast, the author of an article may choose to introduce special terminology and notation for the duration of an article for convenience or because it fits some particular scheme for that article.

Our approach to terminology use also means that we avoid getting creative with names, and try to stick to tried-and-tested terminology that one may find in a textbook account or that is otherwise in wide use.

src/elementary-number-theory/reduced-integer-fractions.lagda.md
fredrik-bakke: Indeed. The current name is good for this entry since the conclusion is reads literally `is-reduced (neg-...`. One could also consider calling it `is-reduced-neg-fraction-ℤ`, since the hypothesis is fairly canonical. That's your call though.

src/elementary-number-theory/group-of-rational-numbers.lagda.md
fredrik-bakke: Yep, let's keep it for now.

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
malarbol: > Ah, I see. Your reading of the term "negative unit laws" is different from mine. I'm still unsure if this phrasing is precedented in the literature, and we do try our best to stick to well-established terminology. Do you have a reference for where the term "negative unit laws" is used in this way perhaps?

Ok. Just so you know, I was just writing some headers for the `multiplication-integers` module and, after `### Unit laws` for `left-unit-law-mul-ℤ` and `right-unit-law-mul-ℤ` it just felt natural to write `### Negative unit laws` for `left-neg-unit-law-mul-ℤ` and `right-neg-unit-law-mul-ℤ` (which where already there). But I agree it doesn't seem to be any well-established concept so it was probably a bad idea. Sorry.

Should I go with
```
### Computing multiplication by -1
```
then? Or something like
```
### Multiplication by -1 is the same as the negative
```
to be even more descriptive?

malarbol: >So, a big difference between how we approach terminology use in agda-unimath as opposed to, say, an article or a small formalization project, or even the TypeTopology library, is that it has to be "globally consistent". Terminology introduced and used on one page has to be consistent with that terminology on any other page, and if two concepts are the same, they should probably also use the same terminology. In contrast, the author of an article may choose to introduce special terminology and notation for the duration of an article for convenience or because it fits some particular scheme for that article.

>Our approach to terminology use also means that we avoid getting creative with names, and try to stick to tried-and-tested terminology that one may find in a textbook account or that is otherwise in wide use.

ok. I'll do my best to keep that in mind.

src/elementary-number-theory/addition-integers.lagda.md
malarbol: yes, I think I mentioned in another comment what this would open the gate to a whole `preserves-[op]-[incl]` family of properties. The way I saw it `[incl]` was a placeholder for the different canonical inclusions from `N -> Z -> fraction-Z -> Q -> R` and `[op]` and operation that exists in both types (so for exemple, there wouldn't be any `preserves-neg-int-N` because negatives is not defined on `N`).

So, for your exemples, there would be some propositions:
```agda
preserves-add-rational-fraction-ℤ :
  (x y : fraction-ℤ) →
  Id
    ((rational-fraction-ℤ x) +ℚ (rational-fraction-ℤ y))
    (rational-fraction-ℤ (x +fraction-ℤ y))

preserves-mul-real-ℚ :
  (x y : ℚ) → (real-ℚ x) *ℝ (real-ℚ y) ＝ real-ℚ (x *ℚ y)

```
with there headers `### The canonical inclusion of [X] into [Y] preserves [op]`.

Anyway, this was just a suggestion based on your [other comment](https://github.com/UniMath/agda-unimath/pull/1100#issuecomment-2041193384)  but I don't want to rush any changes. So let's just keep `add-int-N` until you settle on a nice naming scheme.

malarbol: I think I understand now what you meant : this would conflict with the `preserves-sim-*` naming scheme since now, `preserves-add-xx` and `preserves-sim-xx` wouldn't really do the same thing. Naming things is really difficult.

fredrik-bakke: Oh, no, `preserves` is a verb that we use in many different contexts. I don't think they would conflict here

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
fredrik-bakke: I like the suggestion

```md
### Multiplication by -1 is equal to the negative
```

src/elementary-number-theory/addition-integers.lagda.md
malarbol: ok. I'll give it a few more thoughts then 😅

fredrik-bakke: actually, I think `compute-add-int-ℕ` is the correct name in this instance

fredrik-bakke: If we were to view e.g. `rational-fraction-ℤ` as some sort of algebra homomorphism, I would say `preserves` is the correct verb. But it seems to me that we are rather viewing them as primitive maps on some specific types, and with this interpretation `compute` seems more appropriate.

src/elementary-number-theory/multiplication-integers.lagda.md
```diff
@@ -132,7 +140,11 @@ right-neg-unit-law-mul-ℤ (inr (inl star)) = refl
 right-neg-unit-law-mul-ℤ (inr (inr zero-ℕ)) = refl
 right-neg-unit-law-mul-ℤ (inr (inr (succ-ℕ n))) =
   ap (neg-one-ℤ +ℤ_) (right-neg-unit-law-mul-ℤ (inr (inr n)))
```
fredrik-bakke, line 142: Let's call this `compute-mul-right-neg-one-ℤ`?

src/elementary-number-theory/multiplication-integers.lagda.md
```diff
@@ -120,7 +124,11 @@ right-unit-law-mul-ℤ (inr (inl star)) = refl
 right-unit-law-mul-ℤ (inr (inr zero-ℕ)) = refl
 right-unit-law-mul-ℤ (inr (inr (succ-ℕ n))) =
   ap (one-ℤ +ℤ_) (right-unit-law-mul-ℤ (inr (inr n)))
+```
+
+### Multiplication of an integer by `-1` is equal to the negative
 
+```agda
 left-neg-unit-law-mul-ℤ : (k : ℤ) → neg-one-ℤ *ℤ k ＝ neg-ℤ k
 left-neg-unit-law-mul-ℤ k = refl
```
fredrik-bakke, line 133: Let's call this `compute-mul-left-neg-one-ℤ`?

src/elementary-number-theory/rational-numbers.lagda.md
```diff
@@ -135,32 +147,116 @@ eq-ℚ-sim-fraction-ℤ x y H =
...
+### The negative function on the rational numbers is an involution
+
+```agda
+abstract
+  neg-neg-ℚ : (x : ℚ) → neg-ℚ (neg-ℚ x) ＝ x
+  neg-neg-ℚ x = eq-ℚ (neg-ℚ (neg-ℚ x)) x (neg-neg-ℤ (numerator-ℚ x)) refl
 ```
 
-### The reflecting map from ℤ to ℚ
+### The reflecting map from fraction-ℤ to ℚ
```
fredrik-bakke, line 255:
```suggestion
### The reflecting map from `fraction-ℤ` to `ℚ`
```

or

```suggestion
### The reflecting map from integer fractions to rational numbers
```

src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -49,13 +59,124 @@ left-unit-law-mul-ℚ x =
...
+
+right-neg-unit-law-mul-ℚ : (x : ℚ) → x *ℚ neg-one-ℚ ＝ neg-ℚ x
+right-neg-unit-law-mul-ℚ x =
+  ( eq-ℚ-sim-fraction-ℤ
+    ( mul-fraction-ℤ (fraction-ℚ x) (fraction-ℚ neg-one-ℚ))
+    ( neg-fraction-ℤ (fraction-ℚ x))
+    ( ap-mul-ℤ
+      ( right-neg-unit-law-mul-ℤ (numerator-ℚ x))
+      ( inv (right-unit-law-mul-ℤ (denominator-ℚ x))))) ∙
+  ( is-retraction-rational-fraction-ℚ (neg-ℚ x))
```
fredrik-bakke, line 94: how about `compute-mul-(left|right)-neg-one-ℚ`?

fredrik-bakke (approved):

src/elementary-number-theory/addition-integers.lagda.md
malarbol: It's all a matter of perspective but `compute` seems appropriate.
There's also
```agda
mul-int-ℕ : (x y : ℕ) → (int-ℕ x) *ℤ (int-ℕ y) ＝ int-ℕ (x *ℕ y)
```
much could(should?) be renamed  `compute-mul-int-ℕ` for consistency? I'm not sure how many others there are.
Also, beware there's a `compute-mul-ℤ` (in the `multiplication-integers` module) that does quite a different thing.

fredrik-bakke: > Also, beware there's a `compute-mul-ℤ` (in the multiplication-integers module) that does quite a different thing.

Indeed. Seeing as they're categorized in the same section, it seems appropriate that they also share similar names. Note that `compute` is again one of those verbs that are used in many unrelated contexts in the library.

fredrik-bakke: Just to reiterate, please don't start implementing any suggestions we discuss here in this PR

src/elementary-number-theory/multiplication-integers.lagda.md
malarbol: I'm ok to discuss changes in the code I've written, but I'm not sure I'm fit to decide on renaming existing stuff in the library. Maybe we should discuss this with the person who wrote it in the first place?

For what it's worth, I personally like these names, and the symmetry with `(right|left)-unit-laws`. 
On the one hand, I understand the whole philosophy for "well-accepted named scheme" but, on the other hand, if someone asked me to name the property "Multiplication by the negative unit is equal to the negative" I would surely name it "Negative unit laws".

fredrik-bakke:
Okay, I think it's time to merge this one. Thanks again for the amazing effort!

malarbol:
Thank you for your time and help and all!

fredrik-bakke:
Looking forward to your next contributions! And apologies for the complacency in handling this one; I'm still learning how to take on a more executive role as a maintainer and sometimes it takes a bit of time to come back with good answers.

Issue #1103, opened 2024-03-30
Idempotents in Intensional Type Theory


fredrik-bakke:
## Goals
**Overarching goal:** formalize the positive results from _Idempotents in Intensional Type Theory_.
- [ ] Infrastructure for the total type of retracts of a type
- [X] Infrastructure for sequential limits #839
- [X] The splitting type of a split idempotent map is essentially unique
- [ ] Split idempotents are quasiidempotent
- [X] Preidempotents on sets split
- [X] Weakly constant preidempotents split
- [ ] Quasiidempotents split
- [ ] Split idempotents are a retract of quasiidempotent maps
- [ ] The type of (fully coherent) idempotents as the splitting type of the above retract
- [ ] Write paper summary file
- [ ] Include explanations regarding the negative results, although this issue does not aim to have them formalized.

## References

- Michael Shulman, _Idempotents in Intensional Type Theory_, [arXiv:1507.03634](https://arxiv.org/abs/1507.03634)
- Michael Shulman, _Splitting Idempotents_, [HoTT blog post](https://homotopytypetheory.org/2014/12/08/splitting-idempotents/)
- Kraus, _Notions of Anonymous Existence in Martin-Löf Type Theory_, [arXiv:1610.03346](https://arxiv.org/abs/1610.03346)

PR #1105, opened 2024-03-30
Splitting idempotents

fredrik-bakke:
### Summary
- Retracts of a type
  - Define retracts of a type 
  - Characterize equality of retracts
- Weakly constant maps
  - The type of fixed points of weakly constant maps is a proposition
- Idempotent maps
  - Rename "preidempotent maps" to "idempotent maps". This mirrors how we treat "invertible maps" vs "coherently invertible maps", although there is an essential difference between the two concepts: idempotent maps are not "coherentifiable" like invertible maps. That role is taken by "quasicoherently idempotent maps" instead.
- Quasicoherently idempotent maps (called "quasiidempotent maps" in [Shu17])
  - Define quasicoherently idempotent maps
  - Quasicoherently idempotent maps are closed under homotopy (without funext)
- Split idempotent maps
  - Define split idempotent maps
  - Split idempotent maps are quasicoherently idempotent
  - Idempotent maps on sets split
  - Weakly constant idempotent maps split
  - Quasicoherently idempotent maps split
- Retracts of small types are small

Work towards #1103.


src/foundation/homotopy-algebra.lagda.md
```diff
@@ -90,3 +90,59 @@ module _
...
+  assoc-htpy???² : ((p ∙h q) ∙h r) ∙h s ~ p ∙h (q ∙h (r ∙h s))
+  assoc-htpy???² = assoc-htpy (p ∙h q) r s ∙h assoc-htpy p q (r ∙h s)
+
+  assoc-htpy????² : (p ∙h q) ∙h (r ∙h s) ~ p ∙h ((q ∙h r) ∙h s)
+  assoc-htpy????² =
+    assoc-htpy p q (r ∙h s) ∙h ap-concat-htpy p (inv-htpy (assoc-htpy q r s))
+
+  assoc-htpy?????² : (p ∙h q) ∙h (r ∙h s) ~ (p ∙h (q ∙h r)) ∙h s
+  assoc-htpy?????² =
+    inv-htpy (assoc-htpy (p ∙h q) r s) ∙h ap-concat-htpy' s (assoc-htpy p q r)
```
fredrik-bakke, line 147: Please don't misunderstand these as intentionally bad to elicit a reaction. I opened a discussion asking for guidance on how to name these

src/foundation-core/retracts-of-types.lagda.md
```diff
@@ -0,0 +1,137 @@
...
+The predicate `retract B` is used to assert that a type is a retract of `B`,
+i.e., the type `retract B A` is the type of maps from `A` to `B` that come
+equipped with a retraction.
+
+We also introduce more intuitive infix notation `A retract-of B` to assert that
+`A` is a retract of `B`.
+
+```agda
+retract : {l1 l2 : Level} → UU l1 → UU l2 → UU (l1 ⊔ l2)
+retract B A = Σ (A → B) retraction
```
fredrik-bakke, line 55: I realize that this should have been named `is-retract` or `is-retract-of`

src/foundation/homotopy-algebra.lagda.md
fredrik-bakke: I'm gonna remove these again because they are not needed for this PR.

fredrik-bakke:
Since I won't be able to finish this PR before I leave tomorrow, let me write down some thoughts here for when I pick this up again.
- It remains to show that `is-split-idempotent-map f` is a retract of `is-quasiidempotent-map f`
- The splitting of this retract gives us an infinitely coherent definition of idempotent maps.

It's sensible to make a new file for infinitely coherent idempotent maps, but then one may ask if the proof that quasiidempotents split fits better in that file. There is also a question of whether we want more agda-unimath like names for `is-preidempotent` and `is-quasiidempontent`. My initial suggestions would be `is-idempotent` for the former, mirroring e.g. `is-involution` which is not coherent either, and `is-quasicoherent-idempotent` for the latter.

- I should also write informal proofs for the main results, and this shouldn't be difficult.
- Remark on the explicit counter-examples given in the paper.

Any other nice and easy consequences of this formalization?

fredrik-bakke:
I also need to verify that the diagrams render properly

fredrik-bakke:
To partition this work up a bit, I may aim to formalize fully coherent idempotents in a subsequent PR. It still remains to tidy up this PR before it is ready for review, however.

fredrik-bakke:
A couple of questions remain:
- where should `is-small-retract` be placed? It cannot be placed in either `retracts-of-types` or `small-types` due to cyclic dependency issues.
- am I right to change the terminology of "preidempotent map" to "idempotent map", "quasiidempotent map" to "quasicoherently idempotent map", and "idempotent map" to "coherently idempotent map"?
- What are the best names for `right-right-inv` and its cousins?

fredrik-bakke:
This PR is now open to reviews. @tomdjong said he was interested in reviewing this, so I'll tag him :)

src/foundation-core/retracts-of-types.lagda.md
```diff
@@ -0,0 +1,190 @@
...
+maps
+
+```text
+      i        r
+  A -----> B -----> A
+```
+
+and a [homotopy](foundation-core.homotopies.md) `r ∘ i ~ id`. The map `i` is
+called the **inclusion** of the retract data, and the map `r` is called the
+**underlying map of the retraction**.
```
EgbertRijke, line 41: Somewhere in this section we should probably link to `retractions`.

src/foundation/idempotent-maps.lagda.md
```diff
@@ -0,0 +1,162 @@
...
+coherence `J` is replaced as part of this construction. On the other hand, in
+contrast to invertible maps, not every idempotent map can be made to be fully
+coherent or even quasicoherent. For a counterexample see Section 4 of
+{{#cite Shu17}}.
+
+**Terminology.** Our definition of an _idempotent map_ corresponds to the
+definition of a _preidempotent map_ in {{#reference Shu17}} and
+{{#reference Shu14SplittingIdempotents}}, while their definition of an
+_idempotent map_ corresponds in our terminology to a _coherently idempotent
+map_.
```
EgbertRijke, line 66: Hm... I feel like we should consider following Shulman in his choice of terminology, but we can put that off until another pull request.

src/foundation/idempotent-maps.lagda.md
```diff
@@ -0,0 +1,162 @@
...
+  map-idempotent-map = pr1 f
+
+  is-idempotent-idempotent-map :
+    is-idempotent map-idempotent-map
+  is-idempotent-idempotent-map = pr2 f
+```
+
+## Properties
+
+### Being idempotent over a set is a property
```
EgbertRijke, line 97: I think you're mixing on and over sometimes

```suggestion
### Being an idempotent operation on a set is a property
```

src/foundation/path-algebra.lagda.md
```diff
@@ -131,6 +128,26 @@ module _
...
+
+  left-right-inv : (p : x ＝ y) (q : x ＝ y) → p ∙ (inv p ∙ q) ＝ q
+  left-right-inv p q = inv (assoc p (inv p) q) ∙ ap (_∙ q) (right-inv p)
+
+  right-left-inv : (p : x ＝ y) (q : z ＝ y) → (p ∙ inv q) ∙ q ＝ p
+  right-left-inv p q = assoc p (inv q) q ∙ ap (p ∙_) (left-inv q) ∙ right-unit
+
+  left-left-inv : (p : x ＝ y) (q : y ＝ z) → inv p ∙ (p ∙ q) ＝ q
+  left-left-inv p q = inv (assoc (inv p) p q) ∙ ap (_∙ q) (left-inv p)
+```
```
EgbertRijke, line 149: These were already defined here:

https://unimath.github.io/agda-unimath/foundation-core.identity-types.html#concatenating-with-an-inverse-is-inverse-to-concatenating

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+
+  idempotent-map-quasicoherently-idempotent-map : idempotent-map A
+  idempotent-map-quasicoherently-idempotent-map =
+    ( map-quasicoherently-idempotent-map ,
+      is-idempotent-quasicoherently-idempotent-map)
+```
+
+## Properties
+
+### Being quasicoherently idempotent on a set is a property
```
EgbertRijke, line 112: I wonder if this can be strengthened to groupoids, since we have the extra coherence now

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,350 @@
...
+        ( f ·l (g ·l H ∙h H ·r f ∙h I ∙h inv-htpy H)) ∙h
+        ( inv-htpy (H ·r g))
+      by
+      ( right-transpose-htpy-concat
+        ( g ·l is-idempotent-htpy I H)
+        ( H ·r g)
+        ( H ·r (g ∘ g) ∙h (f ·l (g ·l H ∙h H ·r f ∙h I ∙h inv-htpy H)))
+        ( inv-htpy (nat-htpy H ∘ (g ·l H ∙h H ·r f ∙h I ∙h inv-htpy H))))
+      ~ g ·l H ·r g ∙h H ·r (f ∘ g) ∙h I ·r g ∙h inv-htpy (H ·r g)
+      by
```
EgbertRijke, line 223: I wouldn't be surprised if the long proof you programmed here can be significantly shortened using all the infrastructure for whiskering homotopopies. When I was writing that part, I was writing it in such a way that I could eliminate exactly this sort of proof.

Either way, that's just a remark. This doesn't have to be addressed now.

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+open import foundation-core.torsorial-type-families
+```
+
+</details>
+
+## Idea
+
+An endomap `f : A → A` is
+{{#concept "split idempotent" Disambiguation="map of types" Agda=is-split-idempotent}}
+if there exists a type `B` and an inclusion-retraction pair `i , r` displaying
```
EgbertRijke, line 55: Is this with existential quantification, or would it be more accurate to say "... if it comes equipped with"?

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+           f
+       A ----> A
+      ∧ \     ∧
+   i /   \r  / i
+    /     ∨ /
+  B ====== B.
+```
+
+Observe that split idempotent maps are indeed
+[idempotent](foundation.idempotent-maps.md), as witnesses by the composite
```
EgbertRijke, line 69:
```suggestion
[idempotent](foundation.idempotent-maps.md), as witnessed by the composite
```

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+  where
+
+  is-quasicoherently-idempotent-split-idempotent-map :
+    is-quasicoherently-idempotent (map-split-idempotent-map H)
+  is-quasicoherently-idempotent-split-idempotent-map =
+    is-quasicoherently-idempotent-is-split-idempotent
+      ( is-split-idempotent-split-idempotent-map H)
+```
+
+### Every idempotent on a set splits
```
EgbertRijke, line 560:
```suggestion
### Every idempotent map on a set splits
```

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+```text
+              ξₙ₊₁               I aₙ₊₁             f (αₙ₊₁)⁻¹
+    f a₀ ------------> f (f aₙ₊₁) ---> f aₙ₊₁ --------------------> f (f aₙ₊₂)
+     |                    |                  [nat-htpy]  ___refl___/   |
+  (I⁻¹ a₀)    [Ξₙ]        |       I (f aₙ₊₂)            /         (f ∘ f)(αₙ₊₂)
+     ∨                    ∨         ------>           /                ∨
+  f (f a₀) --------> f (f (f aₙ₊₂))   [J]   f (f aₙ₊₂) ----------> f (f (f aₙ₊₃))
+     (f (ξₙ ∙ I aₙ₊₁ ∙ f αₙ₊₁))     ------>           (f ∘ f) (αₙ₊₂)
+                                  f (I aₙ₊₂)
+```
```
EgbertRijke, line 952: funky!

EgbertRijke (approved):
This is a very nice pull request, with many excellent new additions. Thank you so much for making this effort Fredrik!

src/foundation/quasicoherently-idempotent-maps.lagda.md
fredrik-bakke: Thanks for the suggestion! I'll have a look

src/foundation/split-idempotent-maps.lagda.md
fredrik-bakke: I even made it fit within 80 characters!

src/foundation/idempotent-maps.lagda.md
fredrik-bakke: I'd prefer to resolve terminology use issue before this PR is merged and I'm happy to implement any changes we conclude with. In fact, I originally formalized using Shulmans terminology, but then switched as it seems to conflict somewhat with the library's design choices. I've compared the term "idempotent" with "invertible" in the pull request message:

> Rename "preidempotent maps" to "idempotent maps". This mirrors how we treat "invertible maps" vs "coherently invertible maps", although there is an essential difference between the two concepts: idempotent maps are not "coherentifiable" like invertible maps. That role is taken by "quasicoherently idempotent maps" instead (which are called quasiidempotent maps in the paper)

Note, "quasiidempotence" isnt really about idempotence, just whether it is coherent in a good way or not.

fredrik-bakke: Another term we can compare with is perhaps "involution". We don't call them preinvolutions

fredrik-bakke: Let me know your thoughts! I am not as convinced about this change as my messages may sound

fredrik-bakke: > Another term we can compare with is perhaps "involution". We don't call them preinvolutions

Just to expand on this. A coherent involution would be an involution with the following coherence:

$$HH \sim fHf ∙ H$$
which is equivalent to having a coherence

$$Hf \sim fH$$

as given [at nLab](https://ncatlab.org/nlab/show/involution#in_dependent_type_theory).

src/foundation/path-algebra.lagda.md
fredrik-bakke: Oh, it's those again. The way they're organized is so unintuitive to me 🤦‍♂️

src/foundation/split-idempotent-maps.lagda.md
fredrik-bakke: May I suggest "if there is" instead?

src/foundation/idempotent-maps.lagda.md
EgbertRijke: The question seems to be whether it should be a global design decision to make implementations of set-level concepts refer to their direct translation where set-level objects would be replaced by type-level objects, potentially resulting in an incoherent definition, or whether it should refer to a more complicated homotopically correct translation, which may be more difficult to come up with.

Generally, my first intuition tells me that we should make the homotopically correct translation the "main" implementation, but we should definitely allow for some exceptions. But that is not my conclusion :)

To me, the case of invertibility is an exception. I don't really love it that invertibility refers to the incoherent notion, but this is what the community calls it. Here we are just accommodating communication with the rest of the world. Another exception is the notion of injective map. For involutive operations we actually have both a coherent and an incoherent version, and the coherent version also uses the unmodified name "involutive".

These examples of invertibility and injectivity show that incoherent implementations of concepts are often quite useful and relevant concepts on their own. This makes me think that it should not be a global design decision that we should always try to do mathematics one particular way.

Note that examples where we have already decided to make coherent definitions the main definitions are also prominent in the library. H-spaces, unordered pairs, undirected graphs, involutive types, and many other definitions are all fully coherent. So I think we cannot make the decision one way or another by only talking about examples where one of the two approaches is taken. It should be noted, however, that in all of these examples we have convenient names for the fully coherent notion.

Coming to a conclusion, I think it is probably best to remain flexible about whether or not implementations of concepts should be fully coherent. I don't think it should be a global decision. With that conclusion I am quite happy if we leave your choice of terminology as it is, at least for now.

EgbertRijke: > > Another term we can compare with is perhaps "involution". We don't call them preinvolutions
> 
> Just to expand on this. A coherent involution would be an involution with homotopy H:f∘f∼id and the following coherence:
> 
> HH∼fHf∙H which is equivalent to having a coherence
> 
> Hf∼fH
> 
> as given [at nLab](https://ncatlab.org/nlab/show/involution#in_dependent_type_theory).

See here for our existing work on involutions:

https://unimath.github.io/agda-unimath/structured-types.involutive-types.html

src/foundation/split-idempotent-maps.lagda.md
EgbertRijke: That's fine too, although I like using "comes equipped with"

src/foundation/path-algebra.lagda.md
EgbertRijke: That is interesting, because I would not expect anyone to be able to guess what an entry `left-left-inv` is supposed to do, but I would expect people to make a better guess at what `is-section-concat-inv` does.

src/foundation/idempotent-maps.lagda.md
fredrik-bakke: > To me, the case of invertibility is an exception. I don't really love it that invertibility refers to the incoherent notion, but this is what the community calls it.

This is surprising to me. I thought the community usually called it "quasiivertibility" and that agda-unimath was taking a contrarian stance by calling it something else

src/foundation/path-algebra.lagda.md
fredrik-bakke: I can't really be bothered to argue about these, but if I recall correctly, I actually contributed them once before with the same names. And that time you suggested changing the names to what they currently are. That at least attests something to how intuitive they are.

src/foundation/quasicoherently-idempotent-maps.lagda.md
fredrik-bakke: I find your advice a little hard to follow, I thought I was already using your infrastructure for whiskerings. While I agree this proof can likely be significantly shortened, I don't see how that particular infrastructure will help. Don't hesitate to show me if there is a better way of doing things!

fredrik-bakke: Lines 239 to 250 could be shortened with a specific and ugly naturality lemma I think

fredrik-bakke: Lines 226 to 235 could be shortened with an iterated distributivity lemma for left whiskering, which we don't have I think

fredrik-bakke: 256-272 could probably be simplified with an iterated ap-concat-htpy indexed in binary trees

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,352 @@
...
+  where
+
+  is-quasicoherently-idempotent-is-idempotent-is-set :
+    is-idempotent f → is-quasicoherently-idempotent f
+  pr1 (is-quasicoherently-idempotent-is-idempotent-is-set I) = I
+  pr2 (is-quasicoherently-idempotent-is-idempotent-is-set I) x =
+    eq-is-prop (is-set-A ((f ∘ f ∘ f) x) ((f ∘ f) x))
+```
+
+### If `i` and `r` is an inclusion-retraction pair, then `i ∘ r` is quasicoherently idempotent
```
tomdjong, line 170: The mixing of `is` and `and` is awkward. Maybe *form* an inclusion-retraction pair?

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,352 @@
...
+      ( is-idempotent-is-quasicoherently-idempotent F)
+      ( inv-htpy H)
+  pr2 (is-quasicoherently-idempotent-inv-htpy H) =
+    coherence-is-quasicoherently-idempotent-htpy (inv-htpy H)
+```
+
+### Realigning the coherence of a quasicoherent idempotence proof
+
+Given a quasicoherently idempotent map `f`, any other idempotence homotopy
+`H : f ∘ f ~ f` that is homotopic to the coherent one is also coherent.
```
tomdjong, line 299: Rename `H` to `I` since `I` is used in the code?

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,352 @@
...
+  is-quasicoherently-idempotent-is-idempotent-inv-htpy α =
+    ( I , coherence-is-quasicoherently-idempotent-is-idempotent-inv-htpy α)
+```
+
+### Not every idempotent map is quasicoherently idempotent
+
+A counterexample can be constructed using the
+[cantor space](set-theory.cantor-space.md). See Section 4 of {{#cite Shu17}} for
+more details. Note that the statement does not ask for the idempotence witness
+`I : f ∘ f ~ f` to be preserved.
```
tomdjong, line 342: > Note that the statement does not ask for the idempotence witness `I : f ∘ f ~ f` to be preserved.

Can you elaborate on what this means?

src/foundation/quasicoherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,352 @@
...
+
+## See also
+
+- In [`foundation.split-idempotent-maps`](foundation.split-idempotent-maps.md)
+  we show that every quasicoherently idempotent map splits. Moreover, it is true
+  that split idempotent maps are a retract of quasicoherent idempotent maps.
+
+## References
+
+{{#bibliography}} {{#reference Shu17}} {{#reference Shu14SplittingIdempotents}}
```
tomdjong, line 352: I'm not sure what the agda-unimath convention is, but I would probably make it clearer that all results are due to Shulman (or, in case some results aren't, to make clear which results are new). I find that just having a reference at the end is a bit meagre.

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+  essentially-unique-splitting-type-is-split-idempotent H H' =
+    ( map-essentially-unique-splitting-type-is-split-idempotent H H' ,
+      is-equiv-map-essentially-unique-splitting-type-is-split-idempotent
+        ( H)
+        ( H'))
+```
+
+### The type of split idempotent maps on `A` is equivalent to retracts of `A`
+
+Note that the proof relies on the function extensionality axiom.
```
tomdjong, line 327: On line 730 you give a reference to function extensionality.

src/foundation/split-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,1205 @@
...
+  where
+
+  is-small-split-idempotent-is-split-idempotent :
+    is-split-idempotent l1 f
+  is-small-split-idempotent-is-split-idempotent =
+    is-split-idempotent-is-quasicoherently-idempotent
+      ( is-quasicoherently-idempotent-is-split-idempotent S)
+```
+
+### Small types are closed under retracts
```
tomdjong, line 1155: Many thanks for formalizing this! By the way, you could easily add the analogous result for small maps.

tomdjong (commented):
Very nice to see this formalized, @fredrik-bakke!

src/foundation/quasicoherently-idempotent-maps.lagda.md
fredrik-bakke: Oh, actually, since you ask, I'm not sure any of the results on this page are in the paper. I'll have a more thorough look

fredrik-bakke: Oh, yep, there's at least one result that was missing a citation

fredrik-bakke: So the statement "If `i` and `r` form an inclusion-retraction pair, then `i ∘ r` is quasicoherently idempotent" appears in [Shu17], but as a statement in a proof of another lemma, and we write a shorter proof of this fact. I wrote the following:

> This statement can be found as part of the proof of Lemma 3.6 in {{#cite Shu17}}.

src/foundation/quasicoherently-idempotent-maps.lagda.md
fredrik-bakke: I wrote

> A counterexample can be constructed using the
> [cantor space](set-theory.cantor-space.md), see Section 4 of {{#cite Shu17}} for
> more details. Note that the statement does not ask for the idempotence witness
> `I : f ∘ f ~ f` to be preserved, meaning that the counterexample involves an
> idempotent map `f`, such that _no_ idempotence homotopy `f ∘ f ~ f` is
> quasicoherent.

Let me know if this is satisfactory.

src/foundation/split-idempotent-maps.lagda.md
fredrik-bakke: Done :)

src/foundation/quasicoherently-idempotent-maps.lagda.md
tomdjong: Sorry, I just lazily assumed without checking that this file contained a formalization of parts of Shu17. My bad!

fredrik-bakke: no worries!

src/foundation/quasicoherently-idempotent-maps.lagda.md
tomdjong: Maybe it would help to clarify what `the statement` refers to. In any case, I find your addition very clear and unambiguous even though I still don't understand the first half of the sentence :upside_down_face:

fredrik-bakke: Aha! How about this:

> To be clear, what we are asking for is an idempotent map `f`, such that _no_
idempotence homotopy `f ∘ f ~ f` is quasicoherent. A counterexample can be
constructed using the [cantor space](set-theory.cantor-space.md), see Section 4
of {{#cite Shu17}} for more details.

src/foundation/quasicoherently-idempotent-maps.lagda.md
fredrik-bakke: Here's a counterexample.

Consider the identity function on the circle. Two witnesses that it is idempotent are given by `λx.refl` and `λx.loop`. Both of these are quasicoherent. For the latter, we have
```text
  is-quasicoherent (λx.loop) ≐ (id ·l (λx.loop) ~ (λx.loop) ·r id) ≃ ((λx.loop) ~ (λx.loop)),
```
so `λx.refl` is an appropriate witness (under this equivalence) that `λx.loop` is quasicoherent. However, the two idempotence homotopies are clearly not the same.

fredrik-bakke: I guess this counterexample also disproves that "is fully coherent idempotent" is a proposition.

src/foundation/quasicoherently-idempotent-maps.lagda.md
tomdjong: Perfect!

src/foundation/quasicoherently-idempotent-maps.lagda.md
EgbertRijke: It is probably better if you keep this proof as it is, and not replace it with things that you say are ugly:) In my remark I was just speaking from intuition. Seeing all the path operations you had to do, I just thought that these were the typical things I've been cleaning up in earlier pull request.

However, it is impossible for me to just "see" how to make these calculations more efficiently. I'd have to try it for myself, which I don't have time for right now.

My remark doesn't have to be addressed.

fredrik-bakke: Maybe there is a trivial proof of this statement. After all, Mike seems to have thought it so trivial he didn't even need to state it, and my proof of the fact is pretty involved.

src/foundation/idempotent-maps.lagda.md
fredrik-bakke: Alright, I guess I'm okay with leaving this discussion for the future, but let's perhaps consider fixing our crediting system before that time?

fredrik-bakke:
Is this PR ready to merge? I can formalize the counter-example in a subsequent PR thanks to #1115

src/synthetic-homotopy-theory/circle.lagda.md
```diff
@@ -58,15 +58,13 @@ postulate
   base-𝕊¹ : 𝕊¹
 
 postulate
-  loop-𝕊¹ : Id base-𝕊¹ base-𝕊¹
+  loop-𝕊¹ : base-𝕊¹ ＝ base-𝕊¹
```
EgbertRijke, line 61: These changes are good, but can I say that it is bad practice to throw in more refactoring if a PR is already reviewed?

EgbertRijke:
> Is this PR ready to merge? I can formalize the counter-example in a subsequent PR thanks to #1115

Great work, I'll turn on the auto-merge

src/synthetic-homotopy-theory/circle.lagda.md
fredrik-bakke: Yes, you're right to say that

fredrik-bakke: Oh, and sorry for it. This is only due to laziness and I have no excuses.

EgbertRijke: No worries

PR #1111, opened 2024-04-12
The discrete field of rational numbers

malarbol:
The main goal of this pull request is to introduce the discrete field structure on the rational numbers.
In order to do so, we introduce a few basic results in elementary number theory and ring theory.

### Elementary number theory
- strict inequality on the rational numbers is invariant by translation;
- the multiplicative monoid of rational numbers;
- the nonzero rational numbers:
  - standard definition via `is-nonzero-ℚ`;
  - basic properties;
  - nonzero rational numbers are a multiplicative submonoid;
- the positive integer fractions:
  - fractions with positive numerator;
  - basic properties;
- positive integer fractions are invertible up to the similarity relation;
- the positive rational numbers:
  - rational numbers with positive underlying fractions, or, equivalently, with positive numerator;
  - basic properties and notations;
  - positive rational numbers are an additive subsemigroup;
  - positive rational numbers are a multiplicative submonoid;
  - positive rational numbers are invertible;
- the multiplicative group of positive rational numbers;
- the ring of rational numbers is a division ring;
- the discrete field of the rational numbers;
- the multiplicative group of rational numbers:
  - the group of units of the ring of rational numbers;
  - basic properties and notations.
### Ring theory
- invertible elements of rings are stable under negatives;
- invertible elements of nontrivial rings are nonzero.


malarbol:
Hey again!
Still working towards the field structure on rational numbers. This PR introduces new notations and types for nonzero and positive rational numbers, which will be used to define the inverses.
There should be enough do define the commutative multiplicative monoid of (nonzero|positive) rational numbers but I didn't try to reproduce all the properties we wrote for positive etc. integers.

EgbertRijke:
Hi Malarbol!

Thank you so much for your excellent work recently. It is very nice to have all that new infrastructure for rational numbers. It's great to see that you've been making so much progress on it.

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+### The nonzero rational numbers
+
+```agda
+nonzero-ℚ : UU lzero
+nonzero-ℚ = type-subtype is-nonzero-prop-ℚ
+
+ℚˣ : UU lzero
+ℚˣ = nonzero-ℚ
```
EgbertRijke, line 57: Since we have the ring of rational numbers now, wouldn't it make more sense to define ℚˣ as the [group of units](https://unimath.github.io/agda-unimath/ring-theory.groups-of-units-rings.html) of the ring of rational numbers, instead of introducing it as the nonzero rationals and then reproving that it is a group?

malarbol:
> Hi Malarbol!
> 
> Thank you so much for your excellent work recently. It is very nice to have all that new infrastructure for rational numbers. It's great to see that you've been making so much progress on it.

It's a real (pleasure|honor) to contribute to the library. I know my contributions are a bit _basic_  and I hope they don't interfere with your global view of the project.

EgbertRijke:
> > Hi Malarbol!
> > Thank you so much for your excellent work recently. It is very nice to have all that new infrastructure for rational numbers. It's great to see that you've been making so much progress on it.
> 
> It's a real (pleasure|honor) to contribute to the library. I know my contributions are a bit _basic_ and I hope they don't interfere with your global view of the project.

The most basic contributions are the most impactful. Especially important stuff such as the ring of rational numbers! I can only applaud you for taking on the task of developing this part of the library. It is a big help!

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
malarbol: Do you mean to define ℚˣ as the subtype on invertible rational numbers rather than nonzero rational numbers? 
e.g.
```agda
ℚˣ : UU lzero
ℚˣ = type-subtype (subtype-group-of-units-Ring ℚ-Ring)
```
At some point we'll have to prove that nonzero rationals are invertible so it's essentially the same, but I understand if you want to reserve the notation `ˣ` for invertible elements instead of nonzero elements.

EgbertRijke: Yes, I think that is a better definition, because in ring theory the notation Rˣ is used for the group of invertible elements. This means that in the case of ℚ we should follow the general pattern. One obvious advantage is that you get to reuse the existing infrastructure in the `commutative-algebra` and `ring-theory` folders.

However, this doesn't mean that your code needs to be removed. Replacing `ℚˣ` with `nonzero-ℚ` throughout suffices.

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+### The type of positive rational numbers
+
+```agda
+positive-ℚ : UU lzero
+positive-ℚ = type-subtype is-positive-prop-ℚ
+
+ℚ⁺ : UU lzero
+ℚ⁺ = positive-ℚ
```
malarbol, line 79: Are you ok with this definition of `ℚ⁺`? I guess if there are plans for metric spaces, Cauchy sequences, epsilon-delta proofs, etc. this type will be quite important and I don't want to mess things up.
From within the `elementary-number-theory` module this seems sensible but maybe I'm missing some bigger scheme where this notation is reserved for more general constructions on totally ordered discrete fields or something like that?

malarbol: For exemple, at some point, someone might want to write something like this:
```agda
module _
  {l1 l2 : Level} (X : UU l1) (B : ℚ⁺ → X → X → UU l2)
  where

  is-triangular : UU (l1 ⊔ l2)
  is-triangular =
    (δ δ' : ℚ⁺) (x y z : X) →
    B δ' y z →
    B δ x y →
    B (δ +ℚ⁺ δ') x z
```
So it's important to get the _right_ definitions for `ℚ⁺` or `+ℚ⁺`.

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
malarbol: Would you like an `invertible-rational-numbers` module where we'd define 
```agda
invertible-ℚ : UU lzero
invertible-ℚ = type-subtype (subtype-group-of-units-Ring ℚ-Ring)

ℚˣ : UU lzero
ℚˣ = invertible-ℚ
```
and then `ℚˣ-Group`, `ℚˣ-Ab`, `mul-ℚˣ`, `inv-ℚˣ` etc. using the `ring-theory` infrastructure?

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+              ( inv (swap-neg-mul-ℚ y x) ∙_))) ∘
+        ( is-invertible-is-positive-ℚ (neg-ℚ x)))
+      ( is-invertible-is-positive-ℚ x)
+      ( decide-is-negative-is-positive-is-nonzero-ℚ H)
+
+  is-nonzero-is-invertible-ℚ :
+    Σ ℚ (λ y → (x *ℚ y ＝ one-ℚ) × (y *ℚ x ＝ one-ℚ)) →
+    is-nonzero-ℚ x
+  is-nonzero-is-invertible-ℚ (y , L , _) =
+    is-nonzero-left-factor-mul-ℚ x y (inv-tr is-nonzero-ℚ L is-nonzero-one-ℚ)
```
malarbol, line 63: I put this to show that they are the same propositions but I'm not sure it's really relevant.
There could also be a more general proposition on nontrivial rings, e.g. "Invertible elements of a nontrivial ring are nonzero".

src/elementary-number-theory/cross-multiplication-difference-integer-fractions.lagda.md
```diff
@@ -59,6 +59,37 @@ skew-commutative-cross-mul-diff-fraction-ℤ x y =
...
+```agda
+module _
+  (x : fraction-ℤ)
+  where
+
+  cross-mul-diff-zero-fraction-ℤ :
+    cross-mul-diff-fraction-ℤ zero-fraction-ℤ x ＝ numerator-fraction-ℤ x
+  cross-mul-diff-zero-fraction-ℤ =
+    ( right-unit-law-add-ℤ (numerator-fraction-ℤ x *ℤ one-ℤ)) ∙
+    ( right-unit-law-mul-ℤ (numerator-fraction-ℤ x))
```
fredrik-bakke, line 73: abstract

src/elementary-number-theory/cross-multiplication-difference-integer-fractions.lagda.md
```diff
@@ -59,6 +59,37 @@ skew-commutative-cross-mul-diff-fraction-ℤ x y =
...
+  where
+
+  cross-mul-diff-one-fraction-ℤ :
+    Id
+      (cross-mul-diff-fraction-ℤ one-fraction-ℤ x)
+      ((numerator-fraction-ℤ x) -ℤ (denominator-fraction-ℤ x))
+  cross-mul-diff-one-fraction-ℤ =
+    ap-diff-ℤ
+      ( right-unit-law-mul-ℤ (numerator-fraction-ℤ x))
+      ( left-unit-law-mul-ℤ (denominator-fraction-ℤ x))
```
fredrik-bakke, line 90: abstract

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
+# Inverse of nonzero rational numbers
```
fredrik-bakke, line 1: This title doesn't sound sufficiently specific to me. How about "Multiplicative inverses of nonzero rational numbers"? Likewise, the file can be called `multiplicative-inverses-nonzero-rational-numbers`

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+### A rational number is invertible if and only if it is nonzero
+
+```agda
+module _
+  (x : ℚ)
+  where
+
+  is-invertible-is-nonzero-ℚ :
+    is-nonzero-ℚ x →
+    Σ ℚ (λ y → (x *ℚ y ＝ one-ℚ) × (y *ℚ x ＝ one-ℚ))
```
fredrik-bakke, line 46: This should use the definition of `is-invertible-element-Ring`

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
fredrik-bakke: This sounds like a reasonable result to add, it would be very welcome!

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+          ( λ y →
+            map-product
+              ( swap-neg-mul-ℚ x y ∙_)
+              ( inv (swap-neg-mul-ℚ y x) ∙_))) ∘
+        ( is-invertible-is-positive-ℚ (neg-ℚ x)))
+      ( is-invertible-is-positive-ℚ x)
+      ( decide-is-negative-is-positive-is-nonzero-ℚ H)
+
+  is-nonzero-is-invertible-ℚ :
+    Σ ℚ (λ y → (x *ℚ y ＝ one-ℚ) × (y *ℚ x ＝ one-ℚ)) →
```
fredrik-bakke, line 60: This should use the definition of `is-invertible-element-Ring`

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+module _
+  (x : ℚ) (H : is-nonzero-ℚ x)
+  where
+
+  left-inverse-law-mul-is-nonzero-ℚ : (inv-is-nonzero-ℚ H *ℚ x) ＝ one-ℚ
+  left-inverse-law-mul-is-nonzero-ℚ = pr2 (pr2 (is-invertible-is-nonzero-ℚ x H))
+
+  right-inverse-law-mul-is-nonzero-ℚ : (x *ℚ inv-is-nonzero-ℚ H) ＝ one-ℚ
+  right-inverse-law-mul-is-nonzero-ℚ =
+    pr1 (pr2 (is-invertible-is-nonzero-ℚ x H))
```
fredrik-bakke, line 93: `abstract`

src/elementary-number-theory/inverse-positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,116 @@
+# Inverse of positive rational numbers
```
fredrik-bakke, line 1: Multiplicative inverses*, and same for the module name

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
```diff
@@ -0,0 +1,95 @@
+# Inverse of positive integer fractions
```
fredrik-bakke, line 1: Multiplicative inverses* and same for module name

src/elementary-number-theory/inverse-positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,116 @@
...
+open import foundation.dependent-pair-types
+open import foundation.identity-types
+```
+
+</details>
+
+## Idea
+
+[Positive rational numbers](elementary-number-theory.positive-rational-numbers.md)
+have a multiplicative inverse.
```
fredrik-bakke, line 30: link invertible elements rings. If you want, it would also be useful to spell out in a second sentence what a multiplicative inverse is.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,94 @@
...
+open import foundation.identity-types
+open import foundation.transport-along-identifications
+```
+
+</details>
+
+## Idea
+
+[Nonzero rational numbers](elementary-number-theory.nonzero-rational-numbers.md)
+have a multiplicative inverse.
```
fredrik-bakke, line 33: link invertible elements rings. If you want, it would also be useful to spell out in a second sentence what a multiplicative inverse is.

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
```diff
@@ -0,0 +1,95 @@
...
+open import foundation.identity-types
+open import foundation.transport-along-identifications
+```
+
+</details>
+
+## Idea
+
+[Positive integer fractions](elementary-number-theory.positive-integer-fractions.md)
+have a multiplicative inverse up to the canonical similarity relation on
```
fredrik-bakke, line 28: link invertible elements rings. If you want, it would also be useful to spell out in a second sentence what a multiplicative inverse is.

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,168 @@
...
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+The type of
+{{#concept "nonzero" Disambiguation="rational number" Agda=nonzero-ℚ}}
+[rational numbers](elementary-number-theory.rational-numbers.md).
```
fredrik-bakke, line 39: This is not a complete sentence. Please explain briefly what specifically this file achieves.

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,168 @@
...
+neg-nonzero-ℚ (x , H) = (neg-ℚ x , is-nonzero-neg-ℚ x H)
+```
+
+### The product of two nonzero rational numbers is nonzero
+
+```agda
+is-nonzero-mul-ℚ :
+  {x y : ℚ} → is-nonzero-ℚ x → is-nonzero-ℚ y → is-nonzero-ℚ (x *ℚ y)
+is-nonzero-mul-ℚ {x} {y} H K =
+  rec-coproduct H K ∘ (decide-is-zero-factor-is-zero-mul-ℚ x y)
```
fredrik-bakke, line 129: seems like most of these proofs that some rational is nonzero should be abstracted

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,268 @@
...
+```agda
+not-is-negative-is-positive-ℚ :
+  (x : ℚ) → ¬ (is-positive-ℚ (neg-ℚ x) × is-positive-ℚ x)
+not-is-negative-is-positive-ℚ x (N , P) =
+  is-not-negative-and-positive-ℤ
+    ( numerator-ℚ x)
+    ( ( is-negative-eq-ℤ
+        (neg-neg-ℤ (numerator-ℚ x))
+        (is-negative-neg-is-positive-ℤ {numerator-ℚ (neg-ℚ x)} N)) ,
+      ( P))
```
fredrik-bakke, line 184: `abstract`

fredrik-bakke (commented):
Hey @malarbol, I hope you're doing well. Thanks for another awesome contribution!

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
malarbol: same for `positive-ℚ`?

fredrik-bakke: hmm, presumably yes.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
malarbol: I don't really want to import `ring-of-rational-numbers` here, since I need the import the other way round, to prove that `ℚ-Ring` is a division ring.

Maybe we should go with a `multiplicative-monoid-of-rational-numbers` module that would be imported both here and in the `ring-[xxx]` module and use [`is-invertible-element-Monoid`](https://unimath.github.io/agda-unimath/group-theory.invertible-elements-monoids.html#3265) here?

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
malarbol: ok. I can do it later in a separate PR to keep this one focused on the `elementary-number-theory` module.

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
malarbol: There isn't really any ring nor monoid at this point. Would you be ok with this?
```
The integer fraction obtained by swapping the numerator and the denominator of a
[positive integer fraction](elementary-number-theory.positive-integer-fractions.md)
is a
[multiplicative](elementary-number-theory.multiplication-integer-fractions.md)
inverse of the original fraction up to the canonical similarity relation on
[integer fractions](elementary-number-theory.integer-fractions.md).
```
I can try to spell out what a multiplicative inverse is, but it seems a bit too verbose.

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
malarbol: I have a new suggestion there: https://github.com/UniMath/agda-unimath/pull/1111/commits/d21620b8b484c49782a7e18fe8b0ba9b4b8789bd

If we go with the multiplicative monoid route, there's a [submonoid](https://unimath.github.io/agda-unimath/group-theory.submonoids.html) here that could also be mentioned.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
malarbol: I guess this really calls for a `multiplicative-monoid-of-rational-numbers` module.

src/elementary-number-theory/inverse-positive-rational-numbers.lagda.md
malarbol: I guess this really calls for a `multiplicative-monoid-of-rational-numbers` module (again) and there's another submonoid in play if we want.

malarbol:
> Hey @malarbol, I hope you're doing well. Thanks for another awesome contribution!

Hey @fredrik-bakke. Thanks for your feedback. It looks like it's time to address your concerns from https://github.com/UniMath/agda-unimath/pull/1100#discussion_r1543590103

I'll go with the `mutliplicative-monoid-of-rational-numbers` (renaming `group-of-rational-numbers` to `additive-group-of-rational-numbers` on the way, as you foresaw) and try to make this cleaner.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
malarbol: > This sounds like a reasonable result to add, it would be very welcome!

I hope you don't mind but I ended up adding it here: https://github.com/UniMath/agda-unimath/pull/1111/commits/c30afc3a4b0acc45a9ad86e9c0ea7d20aa6a650c. I also included another lemma proving "invertible elements of rings are stable under negatives" which I used to cleanup a bit some definitions.

malarbol:
Hello again.
Introducing the `multiplicative-monoid-of-rational-numbers` module lead me to rewrite a bit the "inverse structure" on rational numbers, focusing more on the algebraic structures than the operations. I think it looks quite cleaner this way.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
fredrik-bakke: I think the issue is that you want to define `is-division-ring-ℚ` in `ring-of-rational-numbers`. That definition should rather go in `field-of-rational-numbers` or some such file.

fredrik-bakke: I haven't looked at your recent changes yet, but from the discussion here, I am not convinced this conversation is resolved

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,117 @@
...
+inv-ℚ⁺ : ℚ⁺ → ℚ⁺
+pr1 (inv-ℚ⁺ (x , P)) = inv-is-positive-ℚ x P
+pr2 (inv-ℚ⁺ (x , P)) = is-positive-denominator-ℚ x
+```
+
+### The multiplicative group of positive rational numbers
+
+```agda
+ℚ⁺-mul-Group : Group lzero
+pr1 ℚ⁺-mul-Group = semigroup-Submonoid ℚ-mul-Monoid ℚ⁺-mul-Submonoid
```
fredrik-bakke, line 94: The word ordering here doesn't quite make sense to me, but I'm not sure how to reconcile it with other established names. I'm tempted to suggest `multiplicative-group-ℚ⁺`, but that would imply that, for instance, the abelian group of integers should also be called `group-ℤ` and I don't want you to rename that as part of this PR. ~To keep things consistent for now, maybe `mul-ℚ⁺-Group` will do? Similarly `ℚ-mul-Group`should be renamed to `mul-ℚ-Group`.  Keep in mind that `ℚ` should be understood as a namespace.~

Actually, after seeing how many new definitions you add in this PR, I want to go back on this and suggest that you use the wording `multiplicative-group-ℚ⁺` from now on.

src/elementary-number-theory/multiplicative-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## Properties
+
+### The multiplicative group of rational numbers is commutative
+
+```agda
+ℚˣ-mul-Ab : Ab lzero
+pr1 ℚˣ-mul-Ab = ℚˣ-mul-Group
+pr2 ℚˣ-mul-Ab =
```
fredrik-bakke, line 76: `multiplicative-abelian-group-ℚ`

src/elementary-number-theory/multiplicative-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,111 @@
...
+the [core](group-theory.core-monoids.md) of the
+[multiplicative monoid of rational numbers](elementary-number-theory.multiplicative-monoid-of-rational-numbers.md).
+
+## Definitions
+
+### The multiplicative group of rational numbers
+
+```agda
+ℚˣ-mul-Group : Group lzero
+ℚˣ-mul-Group = group-of-units-Ring ℚ-Ring
```
fredrik-bakke, line 49: `multiplicative-group-ℚ`

src/elementary-number-theory/multiplicative-monoid-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,61 @@
...
+equipped with
+[multiplication](elementary-number-theory.addition-rational-numbers.md) is a
+[commutative monoid](group-theory.commutative-monoids.md) with unit `one-ℚ`.
+
+## Definitions
+
+### The multiplicative monoid of rational numbers
+
+```agda
+ℚ-mul-Semigroup : Semigroup lzero
```
fredrik-bakke, line 38: `multiplicative-semigroup-ℚ`

src/elementary-number-theory/nonzero-rational-numbers.lagda.md
```diff
@@ -0,0 +1,168 @@
...
+
+### The nonzero rational numbers are a multiplicative submonoid of the rational numbers
+
+```agda
+is-submonoid-ℚ-mul-nonzero-ℚ :
+  is-submonoid-subset-Monoid ℚ-mul-Monoid is-nonzero-prop-ℚ
+pr1 is-submonoid-ℚ-mul-nonzero-ℚ = is-nonzero-one-ℚ
+pr2 is-submonoid-ℚ-mul-nonzero-ℚ x y = is-nonzero-mul-ℚ {x} {y}
+
+nonzero-ℚ-mul-Submonoid : Submonoid lzero ℚ-mul-Monoid
```
fredrik-bakke, line 150: `multiplicative-submonoid-nonzero-ℚ`

src/elementary-number-theory/multiplicative-monoid-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,61 @@
...
+pr1 ℚ-mul-Monoid = ℚ-mul-Semigroup
+pr2 ℚ-mul-Monoid = is-unital-mul-ℚ
+```
+
+## Properties
+
+### The multiplicative monoid of rational numbers is commutative
+
+```agda
+ℚ-mul-Commutative-Monoid : Commutative-Monoid lzero
```
fredrik-bakke, line 58: `multiplicative-commutative-monoid-ℚ`

src/elementary-number-theory/multiplicative-monoid-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,61 @@
...
+pr1 ℚ-mul-Semigroup = ℚ-Set
+pr1 (pr2 ℚ-mul-Semigroup) = mul-ℚ
+pr2 (pr2 ℚ-mul-Semigroup) = associative-mul-ℚ
+
+is-unital-mul-ℚ : is-unital mul-ℚ
+pr1 is-unital-mul-ℚ = one-ℚ
+pr1 (pr2 is-unital-mul-ℚ) = left-unit-law-mul-ℚ
+pr2 (pr2 is-unital-mul-ℚ) = right-unit-law-mul-ℚ
+
+ℚ-mul-Monoid : Monoid lzero
```
fredrik-bakke, line 48: `multiplicative-monoid-ℚ`

src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,69 @@
...
+pr1 (pr2 (pr2 (pr2 ℚ-add-Group))) = left-inverse-law-add-ℚ
+pr2 (pr2 (pr2 (pr2 ℚ-add-Group))) = right-inverse-law-add-ℚ
+```
+
+## Properties
+
+### Tha additive group of rational numbers is commutative
+
+```agda
+ℚ-add-Ab : Ab lzero
```
fredrik-bakke, line 66: `additive-abelian-group-Q`

src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Definitions
+
+### The additive abelian group of rational numbers
+
+```agda
+ℚ-add-Semigroup : Semigroup lzero
+pr1 ℚ-add-Semigroup = ℚ-Set
+pr1 (pr2 ℚ-add-Semigroup) = add-ℚ
+pr2 (pr2 ℚ-add-Semigroup) = associative-add-ℚ
```
fredrik-bakke, line 42: `additive-semigroup-Q`

src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,69 @@
...
+
+is-unital-add-ℚ : is-unital add-ℚ
+pr1 is-unital-add-ℚ = zero-ℚ
+pr1 (pr2 is-unital-add-ℚ) = left-unit-law-add-ℚ
+pr2 (pr2 is-unital-add-ℚ) = right-unit-law-add-ℚ
+
+ℚ-add-Monoid : Monoid lzero
+pr1 ℚ-add-Monoid = ℚ-add-Semigroup
+pr2 ℚ-add-Monoid = is-unital-add-ℚ
```
fredrik-bakke, line 51: `additive-monoid-Q`

src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,69 @@
...
+
+ℚ-add-Group : Group lzero
+pr1 ℚ-add-Group = ℚ-add-Semigroup
+pr1 (pr2 ℚ-add-Group) = is-unital-add-ℚ
+pr1 (pr2 (pr2 ℚ-add-Group)) = neg-ℚ
```
fredrik-bakke, line 56: `additive-group-Q`

src/elementary-number-theory/positive-rational-numbers.lagda.md
fredrik-bakke: I am okay with `ℚ⁺` since it also makes sense to define `ℝ⁺`, `ℤ⁺` and `ℕ⁺`, and I think the notation is fairly standard and intuitive. Similarly, I think your notation for the group of units is also pretty recognizable and standard. My main concern would be that the `⁺` symbol is pretty small, so may not always be so intelligible.

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,296 @@
...
+### The positive rational numbers are an additive subsemigroup of the rational numbers
+
+```agda
+ℚ⁺-add-Subsemigroup : Subsemigroup lzero ℚ-add-Semigroup
+pr1 ℚ⁺-add-Subsemigroup = is-positive-prop-ℚ
+pr2 ℚ⁺-add-Subsemigroup {x} {y} = is-positive-add-ℚ {x} {y}
+
+ℚ⁺-add-Semigroup : Semigroup lzero
+ℚ⁺-add-Semigroup =
+  semigroup-Subsemigroup ℚ-add-Semigroup ℚ⁺-add-Subsemigroup
```
fredrik-bakke, line 238: `additive-semigroup-ℚ⁺`

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,296 @@
...
+pr2 ℚ⁺-mul-Submonoid = is-submonoid-ℚ-mul-ℚ⁺
+
+ℚ⁺-mul-Monoid : Monoid lzero
+ℚ⁺-mul-Monoid = monoid-Submonoid ℚ-mul-Monoid ℚ⁺-mul-Submonoid
+
+ℚ⁺-mul-Commutative-Monoid : Commutative-Monoid lzero
+ℚ⁺-mul-Commutative-Monoid =
+  commutative-monoid-Commutative-Submonoid
+    ℚ-mul-Commutative-Monoid
+    ℚ⁺-mul-Submonoid
```
fredrik-bakke, line 285: multiplicative-monoid-ℚ⁺ and so on

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
```diff
@@ -32,44 +41,50 @@ open import ring-theory.rings
...
+```agda
+is-division-Ring-ℚ-Ring : is-division-Ring ℚ-Ring
+pr1 is-division-Ring-ℚ-Ring = is-nonzero-one-ℚ ∘ inv
+pr2 is-division-Ring-ℚ-Ring x H =
+  rec-coproduct
+    ( ( is-invertible-element-neg-Ring' ℚ-Ring x) ∘
+      ( is-invertible-is-positive-ℚ (neg-ℚ x)))
+    ( is-invertible-is-positive-ℚ x)
+    ( decide-is-negative-is-positive-is-nonzero-ℚ (H ∘ inv))
+```
```
fredrik-bakke, line 90: This should be placed in the file `discrete-field-of-rational-numbers`

src/ring-theory/trivial-rings.lagda.md
```diff
@@ -71,3 +74,19 @@ pr2 (is-contr-is-trivial-Ring R p) x =
...
+  {l : Level} (R : Ring l) (H : is-nontrivial-Ring R) (x : type-Ring R)
+  where
+
+  is-nonzero-is-invertible-element-nontrivial-Ring :
+    is-invertible-element-Ring R x → zero-Ring R ≠ x
+  is-nonzero-is-invertible-element-nontrivial-Ring (y , P , _) K =
+    H
+      ( ( inv (left-zero-law-mul-Ring R y)) ∙
+        ( ap (mul-Ring' R y) K) ∙
+        ( P))
```
fredrik-bakke, line 91:
```suggestion
    H ( ( inv (left-zero-law-mul-Ring R y)) ∙
        ( ap (mul-Ring' R y) K) ∙
        ( P))
```

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
```diff
@@ -32,44 +41,50 @@ open import ring-theory.rings
...
-pr1 is-unital-mul-ℚ = one-ℚ
-pr1 (pr2 is-unital-mul-ℚ) = left-unit-law-mul-ℚ
-pr2 (pr2 is-unital-mul-ℚ) = right-unit-law-mul-ℚ
-
-has-mul-ℚ-Ab : has-mul-Ab ℚ-Ab
-pr1 has-mul-ℚ-Ab = has-associative-mul-ℚ
-pr1 (pr2 has-mul-ℚ-Ab) = is-unital-mul-ℚ
-pr1 (pr2 (pr2 has-mul-ℚ-Ab)) = left-distributive-mul-add-ℚ
-pr2 (pr2 (pr2 has-mul-ℚ-Ab)) = right-distributive-mul-add-ℚ
+has-mul-ℚ-add-Ab : has-mul-Ab ℚ-add-Ab
```
fredrik-bakke, line 54: `has-mul-additive-abelian-group-ℚ` :)

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
fredrik-bakke: Also, I think it should be called `is-division-ring-ℚ`

fredrik-bakke (commented):
Hey again! I noticed a bad naming pattern emerging with this PR. To be clear, I think you're doing a phenomenal job! As we can see, you are already pushing the limits of the old naming schemes, so I'm suggesting some more robust ones for you

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
fredrik-bakke: thirdly, it seems fitting to factor out the second projection into a second lemma, call it `is-invertible-is-nonzero-Q` or `is-invertible-mul-is-nonzero-Q`. Didn't you prove this in one of the files about nonzero rationals already?

malarbol: Are you ok with `is-invertible-element-ring-is-nonzero-ℚ`?
I use to prove it in the `multiplicative-inverses-nonzero-rational-numbers` module but I got rid of it.
I moved it to a lemma in the `discrete-field-of-rational-numbers` but maybe there's a better place for it.

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
malarbol: > I haven't looked at your recent changes yet, but from the discussion here I am not convinced this conversation is resolved

Sorry if I rushed to "resolution". This module doesn't exist and neither this proposition (at the time I _resolved_ the conversation) so it didn't really felt like any issue any more.

After your other comments, it's now
```agda
module _
  (x : ℚ) (H : is-nonzero-ℚ x)
  where

  is-invertible-element-ring-is-nonzero-ℚ : is-invertible-element-Ring ℚ-Ring x
  is-invertible-element-ring-is-nonzero-ℚ =
    rec-coproduct
      ( ( is-invertible-element-neg-Ring' ℚ-Ring x) ∘
        ( is-invertible-is-positive-ℚ (neg-ℚ x)))
      ( is-invertible-is-positive-ℚ x)
      ( decide-is-negative-is-positive-is-nonzero-ℚ H)
```
as a lemma in the `discrete-field-of-rational-numbers` module.

malarbol:
Thanks for your review and your patience. I suspected that the main challenge of this PR would be to find good names and a good hierarchy between modules, more that about the maths itself. I hope it's not a problem for you; maybe you didn't plan to address these naming issues "right now"?

I'm not 100% convinced about your naming scheme `(multiplicative|additive)-[structure]-[carrier-type]`. For structures where the operation is neither `add` nor `mul` (like [elementary-number-theory.monoid-of-natural-numbers-with-maximum](https://unimath.github.io/agda-unimath/elementary-number-theory.monoid-of-natural-numbers-with-maximum.html), or if the multiplication is the composition of maps, or a convolution product, I think it will be difficult to find good adjectives to extend this scheme. 

What do you think of the template `[structure]-[op]-[carrier-type]`, like
- `group-add-ℚ`;
- `monoid-mul-ℚ`;
- `abelian-group-mul-ℚˣ`;
- `monoid-max-ℕ`;
- etc.
Would this conflict with other established notation?

From a user perspective, I think it can also help to have the name of the operator spelled out in the name of object; like if you see `abelian-group-mul-ℚ⁺` you know that the operation of the group is `mul-ℚ⁺`, etc.

malarbol:
To follow up with my suggestion of `[structure]-[op]-[carrier-type]`, I think it could fit well with the existing name schemes. For exemple, we'd have things like
- `type-[Structure] [structure]-[op]-[carrier-type] = [carrier-type]`:
  - `type-Group group-mul-ℚ⁺ = ℚ⁺`;
  - `type-Monoid monoid-mul-ℚ = ℚ`;
  - etc.
- `[structure-op]-[Structure] [structure]-[op]-[carrier-type] = [op]-[carrier-type]`
  - `mul-Group group-add-ℚ = add-ℚ`;
  - `add-Ab abelian-group-mul-ℚ⁺ = mul-ℚ⁺`;
  - etc.

src/elementary-number-theory/multiplicative-group-of-rational-numbers.lagda.md
malarbol: `abelian-group-mul-ℚˣ`?

src/elementary-number-theory/multiplicative-group-of-rational-numbers.lagda.md
malarbol: `group-mul-ℚˣ`?

src/elementary-number-theory/multiplicative-monoid-of-rational-numbers.lagda.md
malarbol: `semigroup-mul-ℚ`?

fredrik-bakke:
> Thanks for your review and your patience. I suspected that the main challenge of this PR would be to find good names and a good hierarchy between modules, more that about the maths itself. I hope it's not a problem for you; maybe you didn't plan to address these naming issues "right now"?
> 
> I'm not 100% convinced about your naming scheme `(multiplicative|additive)-[structure]-[carrier-type]`. For structures where the operation is neither `add` nor `mul` (like [elementary-number-theory.monoid-of-natural-numbers-with-maximum](https://unimath.github.io/agda-unimath/elementary-number-theory.monoid-of-natural-numbers-with-maximum.html), or if the multiplication is the composition of maps, or a convolution product, I think it will be difficult to find good adjectives to extend this scheme.
> 
> What do you think of the template `[structure]-[op]-[carrier-type]`, like
> 
> * `group-add-ℚ`;
> * `monoid-mul-ℚ`;
> * `abelian-group-mul-ℚˣ`;
> * `monoid-max-ℕ`;
> * etc.
>   Would this conflict with other established notation?
> 
> From a user perspective, I think it can also help to have the name of the operator spelled out in the name of object; like if you see `abelian-group-mul-ℚ⁺` you know that the operation of the group is `mul-ℚ⁺`, etc.

Ah, indeed your suggestion seems to extend better than what I have suggested!

malarbol:
> Ah, indeed your suggestion seems to extend better than what I have suggested!

I'll try this name scheme for these modules. If it's convincing enough we can extend it to the rest of the `elementary-number-theory` module (rename `Z-Group`, `monoid-ℕ-Max`, etc.) in a separate PR.

fredrik-bakke:
> > Ah, indeed your suggestion seems to extend better than what I have suggested!
> 
> I'll try this name scheme for these modules. If it's convincing enough we can extend it to the rest of the `elementary-number-theory` module (rename `Z-Group`, `monoid-ℕ-Max`, etc.) in a separate PR.

Yeah, that sounds good!

src/elementary-number-theory/inverse-nonzero-rational-numbers.lagda.md
fredrik-bakke: Ah, then I understand!

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
fredrik-bakke: I mean, you're dedicating an entire file to the concept, so I don't see how a second sentence will make this too verbose. Moreover, I don't expect every reader to know how we have defined inverses.

In general, I think it is good practice to give concrete definitions, even though some things may have already been defined elsewhere. Otherwise, the reader may have to go on a chicken hunt in order to know exactly what a definition means.

malarbol: Maybe it doesn't need to have its own module. Would you rather have it as a few more properties of `positive-integer-fractions`? (note that it's now [multiplicative-inverses-positive-integer-fractions.lagda](src/elementary-number-theory/multiplicative-inverses-positive-integer-fractions.lagda.md))

malarbol:
This new name scheme looks pretty good to me. Let's see how you like it.

src/elementary-number-theory/discrete-field-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,51 @@
...
+
+### The ring of rational numbers is a division ring
+
+```agda
+is-division-ring-ℚ : is-division-Ring ring-ℚ
+pr1 is-division-ring-ℚ = is-nonzero-one-ℚ ∘ inv
+pr2 is-division-ring-ℚ x H = is-invertible-element-ring-is-nonzero-ℚ x (H ∘ inv)
+```
+
+### The rational numbers are a discrete field
```
fredrik-bakke, line 46:
```suggestion
### The rational numbers form a discrete field
```

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,85 @@
...
+
+</details>
+
+## Idea
+
+The [submonoid](group-theory.submonoids.md) of
+[positive rational numbers](elementary-number-theory.positive-rational-numbers.md)
+in the
+[multiplicative monoid of rational numbers](elementary-number-theory.multiplicative-monoid-of-rational-numbers.md)
+is a [commutative group](group-theory.abelian-groups.md).
```
fredrik-bakke, line 36: I don't think this is the idea of the contents of the file. Rather, I'd say the idea is that the positive rational numbers form a submonoid of the multiplicative monoid of rational numbers.

src/elementary-number-theory/multiplicative-group-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,147 @@
...
+
+</details>
+
+## Idea
+
+The multiplicative group of rational numbers is the
+[group of units](ring-theory.groups-of-units-rings.md) of the
+[ring of rational numbers](elementary-number-theory.ring-of-rational-numbers.md),
+the [core](group-theory.cores-monoids.md) of the
+[multiplicative monoid of rational numbers](elementary-number-theory.multiplicative-monoid-of-rational-numbers.md).
```
fredrik-bakke, line 43:
```suggestion
in other words, it is the [core](group-theory.cores-monoids.md) of the
[multiplicative monoid of rational numbers](elementary-number-theory.multiplicative-monoid-of-rational-numbers.md).
```

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,433 @@
...
+  is-positive-denominator-ℚ⁺ = is-positive-denominator-ℚ rational-ℚ⁺
+
+abstract
+  eq-ℚ⁺ : {x y : ℚ⁺} → rational-ℚ⁺ x ＝ rational-ℚ⁺ y → x ＝ y
+  eq-ℚ⁺ {x} {y} = eq-type-subtype is-positive-prop-ℚ
+```
+
+## Properties
+
+### The positive rational numbers are a Set
```
fredrik-bakke, line 132:
```suggestion
### The positive rational numbers form a set
```

src/ring-theory/trivial-rings.lagda.md
```diff
@@ -71,3 +74,18 @@ pr2 (is-contr-is-trivial-Ring R p) x =
       ＝ x
         by left-unit-law-mul-Ring R x
 ```
+
+### Invertible elements of nontrivial rings are different from zero
```
fredrik-bakke, line 78:
```suggestion
### Invertible elements of nontrivial rings are not equal to zero
```

The suggested wording is clearer, as we tend to use the wording "establish a difference" when we speak of elements that are "apart", and apartness is a stronger condition on elements than nonequality.

You can read a little about apartness relations at the following link
- https://unimath.github.io/agda-unimath/foundation.apartness-relations.html

It's a useful concept in constructive algebra, which reminds me that the page is missing a reference to the book I believe the formalization follows, A Course in Constructive Algebra.

fredrik-bakke (approved):
Your work looks excellent! I have only some very minor comments. Let's merge this PR after they have been resolved :)

src/ring-theory/trivial-rings.lagda.md
fredrik-bakke: Regardless, this book may be of interest to you if you're not already familiar with it :)

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
malarbol: I think it is. The submonoid of positive rational numbers in the multiplicative monoid of rational numbers is actually defined in the `positive-rational-numbers` module. This module shows that this submodule is in fact a group. Note that the fact that positive rational elements are invertible in the multiplicative monoid is also proved in the `positive-rational-numbers` module so this module is basically about showing that the inverse is positive; maybe I should emphasize this in the header?

malarbol:
> Your work looks excellent! I have only some very minor comments. Let's merge this PR after they have been resolved :)

Thanks! If you're satisfied with the new names, could you resolve the conversations you opened, or do you want me to do it?

Maybe we should add some more `## See also` to cross-reference between the modules? You can make me a short list of which module should point where and I'll do it; or you can add them directly if you want. (or maybe it's ok for you as it is?)

src/ring-theory/trivial-rings.lagda.md
malarbol: Thanks for the tip and the feedback. Maybe it'll give me idea for my next PR.

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
malarbol: `is-invertible-element-ring-is-nonzero-ℚ` is now proved in the `multiplicative-group-of-rational-numbers`.
(sorry, if it's hard to follow 😅 )

fredrik-bakke: sounds good, thanks for telling me :)

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
fredrik-bakke: Ah, my bad, I thought it said "commutative monoid", and I didn't realize which file I was looking at. I still think the idea section is a little off, maybe something along the following lines is better:

> The positive rational numbers form a subgroup of the multiplicative group of nonzero rational numbers.

malarbol: mmh. I don't really construct it as a subgroup of the multiplicative group of rational numbers so it would feel weird. But maybe I should.

malarbol: > But maybe I should?

After a few more thoughts, I'm not sure if it's a good idea. The carrier type of this group would now be a subtype of `ℚˣ` instead of `ℚ` and I'm not sure how well it will go along with other operations like `mul-ℚ⁺`.

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
malarbol: Are you ok with the new phrasing and having these properties in their own module?

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
malarbol: Maybe we should get rid of this module? It used to contain `is-invertible-is-positive-ℚ` but now that I've moved it to the `positive-rational-numbers` module it's pretty much independent from the rest.

src/ring-theory/trivial-rings.lagda.md
```diff
@@ -71,3 +74,18 @@ pr2 (is-contr-is-trivial-Ring R p) x =
...
+  {l : Level} (R : Ring l) (H : is-nontrivial-Ring R) (x : type-Ring R)
+  where
+
+  is-nonzero-is-invertible-element-nontrivial-Ring :
+    is-invertible-element-Ring R x → zero-Ring R ≠ x
+  is-nonzero-is-invertible-element-nontrivial-Ring (y , P , _) K =
+    H ( ( inv (left-zero-law-mul-Ring R y)) ∙
+        ( ap (mul-Ring' R y) K) ∙
+        ( P))
+```
```
malarbol, line 91: With this wording for the header, shouldn't this return `x ≠ zero-Ring R` instead?

malarbol: Maybe this calls for some `nonzero-elements-rings` module some time in the future?

fredrik-bakke: > With this wording for the header, shouldn't this return `x ≠ zero-Ring R` instead?

That's hardly important imo

fredrik-bakke: > Maybe this calls for some `nonzero-elements-rings` module some time in the future?

It may be that it is more productive to consider elements that are apart from zero, but I don't know.

malarbol: Ok. This wording is more coherent with the current definition of division rings, so let's keep it this way for now.

src/elementary-number-theory/discrete-field-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,51 @@
+# The discrete field of rational numbers
```
fredrik-bakke, line 1: I wonder if this file should be called `field-of-rational-numbers`. The "discrete" qualifier seems like an implementation detail to me, because of how one has to define fields in constructive mathematics I presume.

See here for some discussion
- https://ncatlab.org/nlab/show/field#constructive

What is your opinion?

fredrik-bakke: Oh, I just noticed, our definition of a discrete field does not agree with nLab's definition. It does agree with _A Course in Constructive Algebra_, however.

fredrik-bakke: One rationale for changing the name as I suggested, is that when we inevitably get more definitions of "fields", this file is a natural place to formalize that the rationals form a field in those senses.

src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -0,0 +1,433 @@
...
+          ( fraction-ℚ x))
+        ( one-fraction-ℤ)
+        ( left-inverse-law-mul-is-positive-fraction-ℤ (fraction-ℚ x) P)) ∙
+      ( is-retraction-rational-fraction-ℚ one-ℚ)
+
+    right-inverse-law-mul-is-positive-ℚ : x *ℚ inv-is-positive-ℚ ＝ one-ℚ
+    right-inverse-law-mul-is-positive-ℚ =
+      (commutative-mul-ℚ x _) ∙ (left-inverse-law-mul-is-positive-ℚ)
+
+  is-invertible-is-positive-ℚ : is-invertible-element-Monoid monoid-mul-ℚ x
```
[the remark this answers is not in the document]
fredrik-bakke: I think `is-invertible-is-positive-ℚ` is fine. The disambiguation is only necessary when talking about semirings that are not rings. If you were to talk about a predicate `is-inverse` however, I would probably disambiguate with something like `is-mul-inverse` or `is-add-inverse`

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
fredrik-bakke: Where do you want to place the contents of this file?

src/elementary-number-theory/positive-rational-numbers.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: oh, lol, I just realized, `Q+` is a semiring that is not a ring haha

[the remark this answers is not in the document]
fredrik-bakke: then `is-mul-invertible` seems like an okay alternative

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
malarbol: > Where do you want to place the contents of this file?

I was thinking of removing it entirely 😅 . Maybe it's not _that_ relevant for now?

src/elementary-number-theory/discrete-field-of-rational-numbers.lagda.md
malarbol: Ok. I thought the "discrete" in the name came from the `zero-Ring R ≠ x` part in the definition of `division-Ring` instead of, e.g. an apartness relation or something like that. 
If you are planning to also define other "fields" structure on the rational numbers and put them in this file I think it is indeed a good idea to rename it. As you wish.

malarbol: Should I rename this module or do you prefer to wait for when new field structures are introduced?

fredrik-bakke: Yeah, I think that's a good idea. 👍
Indeed, "discrete" does stem from using negated equality as the apartness relation, but not every field satisfies this property. For instance, the Dedekind reals are not a field in the discrete sense I believe.

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
fredrik-bakke: Oh, okay, that's up to you. Whatever you think is the most practical

src/elementary-number-theory/discrete-field-of-rational-numbers.lagda.md
malarbol: > For instance, the Dedekind reals are not a field in the discrete sense I believe.

yep, I don't believe they are.

> Yeah, I think that's a good idea

Ok, should I change the header too? or the `## Idea`?

src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
malarbol: I'm not sure. It's certainly a "subgroup of the multiplicative group of nonzero rational numbers" but I don't really see how to express it. And I don't really see how to rephrase the header. Do you have any more suggestion?

fredrik-bakke: I think this file can stay as it is now. As pointed out before, in my original comment I hadn't realized that this file was about group structure yet.

src/elementary-number-theory/discrete-field-of-rational-numbers.lagda.md
fredrik-bakke: The main header should be changed. Until the formalizations are changed, however, I think the current idea section is perfectly good

src/elementary-number-theory/inverse-positive-integer-fractions.lagda.md
malarbol: This is the last open conversation. If it's ok with you, this would be ready to merge.

PR #1110, opened 2024-04-12
Plane trees

EgbertRijke:
This is a short pull request introducing plane trees and proves that the type of plane trees is equivalent to the type of full binary trees.

There are some related old files about binary trees which I will clean up in another pull request, but not here.

PR #1112, opened 2024-04-13
Hereditary W-types

EgbertRijke:
In this PR I generalized the equivalence constructed in #1110, to something that I called "hereditary W-types". This PR is independent of #1110.


src/trees/hereditary-w-types.lagda.md
```diff
@@ -0,0 +1,351 @@
...
+
+```text
+  nil : left-𝕎
+  snoc : left-𝕎 → Y → left-𝕎,
+```
+
+which is equivalent to the type `list` of [lists](lists.lists.md). The left
+hereditary W-type `left-hereditary-𝕎` is then equivalent to the type of plane
+trees. Furthermore, in this case the binary W-type associated to `B` and `C` is
+equivalent the inductive generated by
```
fredrik-bakke, line 113:
```suggestion
trees. Furthermore, in this case the binary W-type associated to `B` and `C` is
equivalent to the inductive type generated by
```

src/trees/hereditary-w-types.lagda.md
```diff
@@ -0,0 +1,351 @@
...
+      ( make-left-hereditary-𝕎)
+      ( ap-binary
+        ( make-left-𝕎 a)
+        ( eq-htpy
+          ( λ b →
+            ap
+              ( unpack-left-hereditary-𝕎 B C)
+              ( is-retraction-left-hereditary-binary-𝕎
+                ( make-left-hereditary-𝕎 (S b)))))
+        ( eq-htpy (λ c → is-retraction-left-hereditary-binary-𝕎 (T c))))
```
fredrik-bakke, line 272: this smells like there should be characterizations of identity types

src/trees/hereditary-w-types.lagda.md
```diff
@@ -0,0 +1,351 @@
...
+      ( make-right-hereditary-𝕎)
+      ( ap-binary
+        ( make-right-𝕎 a)
+        ( eq-htpy (λ b → is-retraction-right-hereditary-binary-𝕎 (S b)))
+        ( eq-htpy
+          ( λ c →
+            ap
+              ( unpack-right-hereditary-𝕎 B C)
+              ( is-retraction-right-hereditary-binary-𝕎
+                ( make-right-hereditary-𝕎 (T c))))))
```
fredrik-bakke, line 337: These lemmas also smell like there should've been characterizations of identity types

src/trees/indexed-w-types.lagda.md
```diff
@@ -14,21 +14,44 @@ open import foundation.universe-levels
 
 ## Idea
 
-**Indexed W-types** are a generalization of ordinary W-types using indexed
-containers. The main idea is that indexed W-types are initial algebras for the
-polynomial endofunctor
+An {{#concept "indexed W-type" Agda=indexed-𝕎}} are a generalization of ordinary
```
fredrik-bakke, line 17: singular "An indexed W-type" vs. plural "are"

src/trees/indexed-w-types.lagda.md
```diff
@@ -14,21 +14,44 @@ open import foundation.universe-levels
...
 data
-  i𝕎
-    {l1 l2 l3 : Level} (I : UU l1) (A : I → UU l2) (B : (i : I) → A i → UU l3)
-    (f : (i : I) (x : A i) → B i x → I) (i : I) :
-    UU (l2 ⊔ l3) where
-  tree-i𝕎 : (x : A i) (α : (y : B i x) → i𝕎 I A B f (f i x y)) → i𝕎 I A B f i
+  indexed-𝕎
+    {l1 l2 l3 : Level} (I : UU l1) (A : I → UU l2)
+    (B : (i j : I) → A j → UU l3) (j : I) :
+    UU (l1 ⊔ l2 ⊔ l3) where
```
fredrik-bakke, line 53: don't we usually put `where` on a separate line?

fredrik-bakke (approved):
Looks good to me, although I cannot comment much on the mathematical content without a little more context. Also, feel free to disregard the comments about characterizing identity types if you deem it too much effort for what it's worth.

EgbertRijke:
If you're ok with it I'd prefer to do the characterizations of identity types another time. You're right that they are probably good to have around.

EgbertRijke:
Also, thank you for the review!

fredrik-bakke:
> If you're ok with it I'd prefer to do the characterizations of identity types another time. You're right that they are probably good to have around.

Yep, totally fine

PR #1113, opened 2024-04-14
Acyclic maps are closed under retracts

tomdjong:
Easy corollary as @fredrik-bakke added retracts of maps to the library


fredrik-bakke (approved):
Flawless!

fredrik-bakke:
Thanks for the contribution :)

PR #1118, opened 2024-04-18
Simplicial Type Theory

fredrik-bakke:


fredrik-bakke:
Closing this until work is resumed.

PR #1130, opened 2024-04-26
Functoriality of morphisms of arrows

fredrik-bakke:
Establishes some properties related to that the `pullback-hom` is functorial. I aim to show that orthogonal maps are closed under retracts of either map.


fredrik-bakke:
So, it turns out this hole goes a lot deeper than I thought, so I will have to split this PR into four parts.

1.  The current PR in fact concerns the bifunctoriality of forming morphisms of arrows.
2. In a second PR, I will formalize bifunctoriality of bicomposition, which requires a lot of new homotopy algebra infrastructure.
3. In the third PR, I will formalize the functorial action of pullbacks on cones. (this is independent of point 2).
4. In the fourth PR, I will formalize the full functoriality of the pullback-hom, from which the statement that orthogonal maps are closed under retracts should follow almost immediately.

fredrik-bakke:
Although this PR doesn't achieve everything it set out to achieve, I don't want to lose these changes on the cutting room floor and am marking it as ready for review.

src/foundation/functoriality-morphisms-arrows.lagda.md
```diff
@@ -0,0 +1,650 @@
...
+                         [morphisms of arrows]
+```
+
+that maps pairs of morphisms of arrows to the bicomposition morphism of arrows.
+We may refer to either of the vertical functors as the functorial action of
+morphisms of arrows.
+
+## Definitions
+
+### The type of morphisms of arrows as a pullback
```
EgbertRijke, line 65: It would be nice to have the pullback square as a diagram here

src/foundation/functoriality-pullbacks.lagda.md
```diff
@@ -33,25 +40,235 @@ functorial.
...
+      ( map-standard-pullback 𝒮 𝒯 h' t)
+      ( htpy-left-map-htpy-hom-cospan-diagram 𝒮 𝒯 h h' H
+        ( vertical-map-standard-pullback t))
+      ( htpy-right-map-htpy-hom-cospan-diagram 𝒮 𝒯 h h' H
+        ( horizontal-map-standard-pullback t))
+  coherence-preserves-htpy-map-standard-pullback
+    𝒮@(A , B , X , f , g)
+    𝒯@(A' , B' , X' , f' , g')
+    h@(hA , hB , hX , HA , HB)
+    h'@(hA' , hB' , hX' , HA' , HB')
```
EgbertRijke, line 202: What does this notation mean? I don't think I've ever seen it before

src/orthogonal-factorization-systems/functoriality-pullback-hom.lagda.md
```diff
@@ -24,42 +39,126 @@ open import orthogonal-factorization-systems.pullback-hom
 
 The construction of the
 [pullback-hom](orthogonal-factorization-systems.pullback-hom.md) is functorial.
+There is a commuting triangle of ∞-categories
 
-## Definition
+```text
+  [pairs of arrows of types] ---> [exponentiated cospan diagrams]
```
EgbertRijke, line 45: I'm not sure if it is self-evident what the infinity category of exponentiated cospan diagrams is. I find it hard to understand what you want to tell us with this diagram, which is probably due to it being under explained.

EgbertRijke (approved):
This is a nice pull request, but sometimes a bit hard to understand.

src/foundation/functoriality-pullbacks.lagda.md
fredrik-bakke: the notation `𝒮@(A , B , X , f , g)` gives aliases to the components of the full tuple `𝒮`. So`A ≐ pr1 𝒮` and so on. This maintains more readable goals in Agda and lets you refer to the entire tuple without reconstructing it from its parts.

EgbertRijke: Thanks for the explanation!

src/orthogonal-factorization-systems/functoriality-pullback-hom.lagda.md
fredrik-bakke: Let me know if the updated explanation is sufficiently clear!

EgbertRijke: Great, thank you!

PR #1152, opened 2024-06-03
Fiberwise orthogonal maps and closure properties of the right class

fredrik-bakke:
This is the replacement for #1032.

### DIFF
--- a/src/orthogonal-factorization-systems/fiberwise-orthogonal-maps.lagda.md
+++ b/src/orthogonal-factorization-systems/fiberwise-orthogonal-maps.lagda.md
@@ -1,0 +1,422 @@
+# Fiberwise orthogonal maps
+
+```agda
+module orthogonal-factorization-systems.fiberwise-orthogonal-maps where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.action-on-identifications-functions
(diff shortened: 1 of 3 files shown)


src/orthogonal-factorization-systems/fiberwise-orthogonal-maps.lagda.md
```diff
@@ -0,0 +1,422 @@
...
+             - ∘ f'
+     B' → X -------> A' → X
+        |               |
+  g ∘ - |               | g ∘ -
+        ∨               ∨
+     B' → Y -------> A' → Y
+             - ∘ f'
+```
+
+is also a pullback.
```
EgbertRijke, line 86: Later you use the phrase "fiberwise `f`-orthogonal". Could you introduce that phrase here?

EgbertRijke (commented):
Great PR! I have only one minor remark.

src/orthogonal-factorization-systems/fiberwise-orthogonal-maps.lagda.md
fredrik-bakke: I removed the phrasing "fiberwise `f`-orthogonal" all together

PR #1155, opened 2024-06-06
Use box drawing characters in diagrams

### LINKED ISSUE #1141: Use box-drawing characters for character diagrams
I just realized this afternoon: why are we not using [box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_characters) for drawing our character diagrams? Agda-input: `---`.

Here's a couple of rendered examples:

image

<details><summary>markdown</summary>

```text
          B
      h  ∧| \  g
       /  |   \
     /  f | ⇑   ∨
    A ---------> C
    |     | hB   |
    | ⇗   ∨   ⇗  |
 hA |     B'     | hC
    | h' ∧  \ g' |
    |  /  ⇑   \  |
    ∨/          ∨∨
    A' --------> C'
          f'
```

```text
          B
      h  ∧│ ╲  g
       ╱  │   ╲
     ╱  f │ ⇑   ∨
    A ─────────> C
    │     │ hB   │
    │ ⇗   ∨   ⇗  │
 hA │     B'     │ hC
    │ h' ∧  ╲ g' │
    │  ╱  ⇑   ╲  │
    ∨╱          ∨∨
    A' ────────> C'
          f'
```

```text
       ∙
     ╱   ╲
    ╱     ╲
   ∨       ∨
  ∙ ──────> ∙
```
</details>

The first diagram is our current version of a commuting prism diagram, while the second diagram employs horizontal, vertical, and diagonal box-drawing characters. The third diagram demonstrates what the diagonals can look like with the right horizontal offset.

**Note.** It looks like there is a line spacing issue since the characters don't align perfectly, which they should by definition. This should also be fixed.

It seems to me that using box-drawing characters is objectively an improvement over our current convention, so I propose that we switch to using them. I am happy to back-implement this change if we agree it is better.

fredrik-bakke:
Resolves #1141.


fredrik-bakke (commented):
I've looked through all the changes twice, so I'll invite other reviews now

EgbertRijke:
I have been thinking about this pull request and I've tried my hardest see this as an improvement, but in truth I just don't love it. I have always quite liked the ASCII diagrams and they're easier to typeset than the diagrams with box-drawing symbols, which have an extra layer of unicode input to them. There are also questions whether the diagrams will display correctly across all the various fonts that people use. So I haven't been able to conclude that this is going to be a clear improvement over the old ASCII diagrams.

I'm inclined to suggest to close this PR without merging.

PR #1156, opened 2024-06-20
Fix typos

pitmonticone:


fredrik-bakke (approved):
Thank you for the corrections!

PR #1157, opened 2024-07-06
Continuation modalities and Lawvere–Tierney topologies

fredrik-bakke:
Defines continuation modalities as a generalization of the double negation modality, and shows they define Lavwere–Tierney topologies on types. It also improves term usage in the file about reflective subuniverses.


fredrik-bakke:
With the addition of Lawvere–Tierney topologies, I wonder if this PR is relevant to #930.

src/foundation/continuations.lagda.md
```diff
@@ -0,0 +1,197 @@
...
+## Idea
+
+Given a type `R`,
+{{#concept "continuations" Disambiguation="on a type" Agda=continuation}} on `R`
+
+```text
+  A ↦ ((A → R) → R)
+```
+
+defines a monad on types.
```
EgbertRijke, line 54: I think I'm missing the idea here, i.e., from this description I get no sense of why continuations should have anything to do with double exponents.

The nLab has a very vague description of the idea behind this definition, but I didn't find that entirely satisfying either. Perhaps a more solid idea section would just describe the context in which the continuation monad comes up, as a motivation for introducing it here. I.e., something along the lines of:

"The continuation monad is a monad that comes up in computability theory (cite papers) in the following way... It is defined as..."

src/foundation/continuations.lagda.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Definitions
+
+### The continuation operator
+
+```agda
+continuation :
+  {l1 l2 : Level} (R : UU l1) (A : UU l2) → UU (l1 ⊔ l2)
+continuation R A = ((A → R) → R)
```
EgbertRijke, line 63:
```suggestion
continuation R A = (A → R) → R
```

src/foundation/continuations.lagda.md
```diff
@@ -0,0 +1,197 @@
...
+open import orthogonal-factorization-systems.modal-operators
+open import orthogonal-factorization-systems.uniquely-eliminating-modalities
+```
+
+</details>
+
+## Idea
+
+Given a type `R`,
+{{#concept "continuations" Disambiguation="on a type" Agda=continuation}} on `R`
```
EgbertRijke, line 48: I think the preposition in "continuation on R" confuses me here, and it seems that the nLab doesn't ever say "continuation on R". Is the phrase "continuation on" used in other sources?

One reason why it sounds so strange to me is that if we extrapolate this to the double negation monad, then according to this phrasing a double negation would be a "continuation on the empty type", and I can't see how that phrase makes sense.

src/orthogonal-factorization-systems/continuation-modalities.lagda.md
```diff
@@ -0,0 +1,212 @@
...
+preserves-unit-continuation-modality :
+  {l1 : Level} (R : Prop l1) → continuation (type-Prop R) unit ≃ unit
+preserves-unit-continuation-modality R =
+  equiv-iff'
+    ( continuation-Prop R unit)
+    ( unit-Prop)
+    ( preserves-unit-continuation-modality')
+```
+
+We must verify that continuations distribute over products. Notice that for a
```
EgbertRijke, line 115:
```suggestion
We must verify that continuations distribute over [products](foundation.cartesian-product-types.md). Notice that for a
```

src/orthogonal-factorization-systems/large-lawvere-tierney-topologies.lagda.md
```diff
@@ -0,0 +1,152 @@
...
+      type-Prop (j unit-Prop ⇔ unit-Prop)
+
+    preserves-conjunction-is-large-lawvere-tierney-topology :
+      {l1 l2 : Level} (P : Prop l1) (Q : Prop l2) →
+      type-Prop (j (P ∧ Q) ⇔ j P ∧ j Q)
+
+open is-large-lawvere-tierney-topology public
+```
+
+### The large set of Lawvere-Tierney topologies
```
EgbertRijke, line 66:
```suggestion
### The large type of Lawvere-Tierney topologies
```

EgbertRijke (approved):
Awesome pull request! Sorry that it took so long to get this PR reviewed, but the Lawvere-Tierney stuff is great! 

I had some minor comments.

fredrik-bakke:
Thanks for the review, and for catching the less-than-ideal phrasing!

After looking closer, I cannot find a good source for how to speak about the continuation monad either. If I understand correctly, it is the `A → R` in `(A → R) → R` that is the "continuation". The total expression should be read as "if we can compute R from A, then we can compute R", or "from a continuation on A we have a computation of R". And the act of using the continuation monad in programming can be referred to as "programming in continuation passing style".

In the context of the double negation modality, we should read it as "if A is empty, then we can deduce a contradiction." But, I am not an apprentice in the field of computability theory, and I don't know enough about the subject to give a satisfactory expository explanation. If I recall correctly, the reason I formalized continuations was mainly for the spurious connection I saw to the double negation modality.

I've reworded most of what I can find to refer to "the continuation monad" rather than "continuations". Let me know if this is satisfactory.

EgbertRijke:
Thank you!

Issue #1055, opened 2024-03-04
List of publications


fredrik-bakke:
We should have a list of publications that are formalized in the library. This would be good for exposition, would give new readers some pointers for what to read in the library, and would allow us to brag a little about what the library contains.

The way I see it, the list should be split into two sections. One where the original publication was made with our library in mind, and one where the formalization appeared after the publication, and/or where it may not have been formalized by any of the original authors of the publication. We can also consider subdividing it further into more and less "serious" publications. E.g. published papers in high-standing journals vs. pre-prints & blog posts.

A list of desiderata are
- [ ] Have lists of publications
- [ ] Add references to main theorems where they are formalized
- [ ] Have guidelines for when this list can be updated
- [ ] Some sort of credit-giving mechanism to the formalizers

Depends on #957.

fredrik-bakke:
I'll try to compile a list of such publications in this comment.

_**WIP**_

VojtechStep:
I like what the cubical people do with https://agda.github.io/cubical/Cubical.Papers.Everything.html, where they have a module per publication, and then import the constructions/lemmas/results from the actual library.

fredrik-bakke:
Ah, that's a bit similar to how we organize the OEIS page, that could work!

PR #1160, opened 2024-08-02
Literature – Idempotents in Intensional Type Theory

fredrik-bakke:
Adds a literature file for the article _Idempotents in Intensional Type Theory_ by Shulman. Most of sections 1-3, 5, and 9 are formalized.

#1103 #1055

### DIFF
--- a/src/foundation/coherently-idempotent-maps.lagda.md
+++ b/src/foundation/coherently-idempotent-maps.lagda.md
@@ -41,7 +41,7 @@ the splitting construction on
 given in {{#cite Shu17}}:
 
 ```text
-  is-coherently-idempotent f :≐
+  is-coherently-idempotent f :=
     Σ (a : ℕ → is-quasicoherently-idempotent f), (Π (n : ℕ), split(aₙ₊₁) ~ aₙ)
 ```
 
--- a/src/foundation/coherently-idempotent-maps.lagda.md
+++ b/src/foundation/coherently-idempotent-maps.lagda.md
@@ -1,0 +1,65 @@
+# Coherently idempotent maps
+
+```agda
+module foundation.coherently-idempotent-maps where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import elementary-number-theory.natural-numbers
+
+open import foundation.dependent-pair-types
+open import foundation.homotopy-algebra
+open import foundation.quasicoherently-idempotent-maps
+open import foundation.split-idempotent-maps
+open import foundation.universe-levels
+open import foundation.whiskering-homotopies-composition
+
+open import foundation-core.function-types
+open import foundation-core.homotopies
+open import foundation-core.propositions
+open import foundation-core.retractions
+open import foundation-core.sets
+```
+
+</details>
+
+## Idea
+
+A
+{{#concept "coherently idempotent map" Disambiguation="of types" Agda=is-coherently-idempotent}}
+is an [idempotent](foundation.idempotent-maps.md) map `f : A → A`
+[equipped](foundation.structure.md) with an infinitely coherent hierarchy of
+[homotopies](foundation-core.homotopies.md) making it a "homotopy-correct"
+definition of an idempotent map in Homotopy Type Theory.
+
+**Terminology.** Our definition of a _coherently idempotent map_ corresponds to
+the definition of a _(fully coherent) idempotent map_ in {{#reference Shu17}}
+and {{#reference Shu14SplittingIdempotents}}. Our definition of an _idempotent
+map_ corresponds in their terminology to a _pre-idempotent map_.
+
+## Definitions
+
+### The structure on a map of coherent idempotence
+
+```agda
+is-coherently-idempotent : {l : Level} {A : UU l} → (A → A) → UU l
+is-coherently-idempotent f =
+  Σ ( ℕ → is-quasicoherently-idempotent f)
+    ( λ a →
+      (n : ℕ) →
+      htpy-is-quasicoherently-idempotent
+        ( is-quasicoherently-idempotent-is-split-idempotent
+          ( is-split-idempotent-is-quasicoherently-idempotent
+            ( a (succ-ℕ n))))
+        ( a n))
+```
+
+## See also
+
+- [Split idempotent maps](foundation.split-idempotent-maps.md)
+
+## References
+
+{{#bibliography}} {{#reference Shu17}} {{#reference Shu14SplittingIdempotents}}
--- a/src/foundation/coherently-idempotent-maps.lagda.md
+++ b/src/foundation/coherently-idempotent-maps.lagda.md
@@ -34,6 +34,17 @@ is an [idempotent](foundation.idempotent-maps.md) map `f : A → A`
 [homotopies](foundation-core.homotopies.md) making it a "homotopy-correct"
 definition of an idempotent map in Homotopy Type Theory.
 
+The infinite coherence condition is given by taking the
+[sequential limit](foundation.sequential-limits.md) of iterated application of
+the splitting construction on
+[quasicoherently idempotent maps](foundation.quasicoherently-idempotent-maps.md)
+given in {{#cite Shu17}}:
+
+```text
+  is-coherently-idempotent f :≐
+    Σ (a : ℕ → is-quasicoherently-idempotent f), (Π (n : ℕ), split(aₙ₊₁) ~ aₙ)
+```
+
 **Terminology.** Our definition of a _coherently idempotent map_ corresponds to
 the definition of a _(fully coherent) idempotent map_ in {{#reference Shu17}}
 and {{#reference Shu14SplittingIdempotents}}. Our definition of an _idempotent
--- a/src/literature.lagda.md
+++ b/src/literature.lagda.md
@@ -2,12 +2,13 @@
 
 ## References
 
-{{#bibliography}} {{#reference SvDR20}}
+{{#bibliography}} {{#reference SvDR20}} {{#reference Shu17}}
 
 ## Files in the namespace
 
 ```agda
 module literature where
 
+open import literature.idempotents-in-intensional-type-theory public
 open import literature.sequential-colimits-in-homotopy-type-theory public
 ```
--- a/src/literature.lagda.md
+++ b/src/literature.lagda.md
@@ -1,5 +1,9 @@
 # Formalization of results from the literature
 
+This page is a work in early progress. To see what's happening behind the
+scenes, you can have a look at the associated GitHub issue
+[#1055](https://github.com/UniMath/agda-unimath/issues/1055).
+
 ## References
 
 {{#bibliography}} {{#reference SvDR20}} {{#reference Shu17}}
--- a/src/literature.lagda.md
+++ b/src/literature.lagda.md
@@ -1,8 +1,8 @@
 # Formalization of results from the literature
 
-This page is a work in early progress. To see what's happening behind the
-scenes, you can have a look at the associated GitHub issue
-[#1055](https://github.com/UniMath/agda-unimath/issues/1055).
+> This page is a work in early progress. To see what's happening behind the
+> scenes, you can have a look at the associated GitHub issue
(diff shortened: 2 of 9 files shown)


src/literature.lagda.md
```diff
@@ -1,13 +1,18 @@
 # Formalization of results from the literature
 
+This page is a work in early progress. To see what's happening behind the
+scenes, you can have a look at the associated GitHub issue
+[#1055](https://github.com/UniMath/agda-unimath/issues/1055).
```
fredrik-bakke, line 5: This very much feels like it should be in an infobox. Do we have functionality for that?

src/literature/idempotents-in-intensional-type-theory.lagda.md
```diff
@@ -0,0 +1,423 @@
...
+
+open import foundation.univalence-implies-function-extensionality using
+  ( funext-univalence -- "univalence implies function extensionality"
+  )
+
+-- MISSING: propositional truncations imply function extensionality
+
+open import foundation.equality-cartesian-product-types using
+  ( Eq-product -- observational equality on pairs
+  ; equiv-pair-eq -- "characteriation of the identity types of cartesian product type formation"
```
VojtechStep, line 114:
```suggestion
  ; equiv-pair-eq -- "characterization of the identity types of cartesian product type formation"
```
Also "of cartesian product type formation" doesn't sound right — should it be "of cartesian product types"?

src/literature/idempotents-in-intensional-type-theory.lagda.md
```diff
@@ -0,0 +1,423 @@
...
+
+**Theorem 4.6.** There exists a pre-idempotent on $B\operatorname{Aut}(C)$ that
+does not split.
+
+```agda
+-- TODO
+```
+
+**Corollary 4.7.** It is impossible to prove in MLTT that all pre-idempotents
+split, or even that all pre-idempotents are quasi-idempotent. □
```
VojtechStep, line 284: Why the QED here?

VojtechStep (approved):
👌

src/foundation/coherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,76 @@
...
+definition of an idempotent map in Homotopy Type Theory.
+
+The infinite coherence condition is given by taking the
+[sequential limit](foundation.sequential-limits.md) of iterated application of
+the splitting construction on
+[quasicoherently idempotent maps](foundation.quasicoherently-idempotent-maps.md)
+given in {{#cite Shu17}}:
+
+```text
+  is-coherently-idempotent f :≐
```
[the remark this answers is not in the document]
fredrik-bakke: Oh, I guess I'm a little confused about the notation. In some cases, we define things only up to propositional equality where I would think `:=` was more appropriate. For instance, take the recursion principle of the circle.

src/foundation/coherently-idempotent-maps.lagda.md
```diff
@@ -0,0 +1,76 @@
...
+        ( a n))
+```
+
+## See also
+
+- [Split idempotent maps](foundation.split-idempotent-maps.md)
+
+## References
+
+{{#bibliography}} {{#reference Shu17}} {{#reference Shu14SplittingIdempotents}}
```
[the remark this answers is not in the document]
fredrik-bakke: No worries! So, I've been playing with the idea that every reference should be recorded with the `bibliography`-macro so that if one coincidentally removes the last citation in a text the reference doesn't disappear from the bibliography unless one removes it explicitly. Likewise, it gives a description in the code of what the bibliography will look like without having to scan through the document for `cite`-invocations. I'm still unsure if this is the best approach, though.

src/literature/idempotents-in-intensional-type-theory.lagda.md
fredrik-bakke: Just to signalize that there is no accompanying formalization I suppse

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Contrapositively, if a reference is to be removed you can start by removing the entry by the bibliography. Then you can be sure if the build process is happy that there is no lingering citation of that reference.

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: So, yes, it was intentional; but probably because I've misunderstood the notation.

src/literature/idempotents-in-intensional-type-theory.lagda.md
fredrik-bakke: What's your opinion?

VojtechStep: Depends on why it's not formalized. Here I would guess that the proof is pretty trivial, but the statement is not very useful to have in the library itself. Then I'd still write a one-liner formalization here, as I've done over at Sequential colomits in HoTT, where one of the lemmas was just calling a theorem with the constant `1`.

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
VojtechStep: I think I only ever use `:=` for honest strict "alias" "this is the literal definition" definitions in the prose. For "definitions" holding only up to propositional equality I would use regular `=` with a comment that the left-hand side is determined by the right-hand side or something like that.

src/literature/idempotents-in-intensional-type-theory.lagda.md
fredrik-bakke: Isn't the statement kind of unformalizable?

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
VojtechStep: It feels more natural to me to have the references determined by the citations. What's the use case for knowing what the section looks like without looking at the `cite` calls? I don't think I should care — if I want to cite a paper somewhere I just do it, and the references take care of themselves. Similarly with removing references from a page — when does that happen? The only situation I can imagine right now is when we're decomposing a file into multiple files, and even then it would be more ergonomic to move the formalization along with the prose that contains the citations.

src/literature/idempotents-in-intensional-type-theory.lagda.md
VojtechStep: I was thinking that I'd be happy with showing "not all pre-idempotents split" in HoTT (that would be the oneliner here) and commenting that then "all pre-idempotents split" can't be provable in the weaker MLTT.

fredrik-bakke: But that's what this corollary is! It's the comment after giving a pre-idempotent that doesn't split.

fredrik-bakke: But I'd curious to see if you have a one-liner counterexample, as the article spends an entire section to produce one

fredrik-bakke: Oh but you probably mean just formalizing a weakening of the same statement

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Okay, that's reasonable, thanks for clearing it up for me.

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Hmm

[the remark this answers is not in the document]
fredrik-bakke: If the bibliography is empty, would you say a `bibliography`-invocation should give an error?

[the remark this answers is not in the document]
fredrik-bakke: I suppose in my mind the references serve two purposes. One is for citations and one is for further reading. Sometimes I add references without citing them, but that doesn't mean the reference has a special meaning compared to the other references. By maintaining a separate list of references the two purposes are kept separate and both are organized.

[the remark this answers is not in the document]
VojtechStep: > If the bibliography is empty, would you say a bibliography-invocation should give an error?

I guess, yeah.

Re: using bibliography for both citations and further reading — what about having a "Further reading" section or paragraph in Idea, explaining why the suggested material is relevant? I think that would be more useful than "sneaking in" uncited references.

src/literature/idempotents-in-intensional-type-theory.lagda.md
VojtechStep: > I'd be curious to see if you have a one-liner counterexample

> Oh but you probably mean just formalizing a weakening of the same statement

Yeah, I meant that after formalizing Theorem 4.6 `(g , g-idemoptent , g-does-not-split)`, the oneliner would be
```agda
corollary : not ((f : A -> A) -> is-idempotent f -> is-split-idempotent f)
corollary all-idempotents-split = g-does-not-split (all-idempotents-split g g-idempotent)
```

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: I agree that the way we display/handle references is not the best way. It's probably a good idea to take some inspiration from $n$Lab. While they're probably not entirely consistent on how they organize references, we can shop a bit around and see what we like. For instance, at [inaccessible cardinal](https://ncatlab.org/nlab/show/inaccessible+cardinal) (which is just the first page stumbled upon) they use the references section both for citations and further reading, but they include short explanations for context and summary of the references.

In any case this is clearly outside the scope of this PR so I'll promote this thread to an issue.

src/literature/idempotents-in-intensional-type-theory.lagda.md
fredrik-bakke: It still feels strange to me to have to comment on why there's no accompanying formalization. The statement is clearly a metastatement and so cannot be formalized without having access to the metatheory. I can remove the QED symbol if you prefer?

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: Let me know if I can resolve this conversation or if there's something you'd like me to change first.

src/literature/idempotents-in-intensional-type-theory.lagda.md
fredrik-bakke: Notice that the statement isn't about univalent foundations as a counterexample, but a metastatement about MLTT

src/foundation/coherently-idempotent-maps.lagda.md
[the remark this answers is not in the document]
VojtechStep: I'm resolving the thread, I agree that it's very much out of scope for this PR

src/literature/idempotents-in-intensional-type-theory.lagda.md
VojtechStep: My thinking is that this file is supposed to be about formalization of results from Shu17 in agda-unimath. It says so at the top of the file. So if it is to include statements that are not formalized, it should also include a comment why it's relevant to have it here, even without a formalization.

fredrik-bakke: I added an explanation. Let me know if this is what you were after

VojtechStep: Perfect, thank you 👍

fredrik-bakke:
merging time? 🥹👉👈

VojtechStep:
👉😎👉

PR #1173, opened 2024-09-02
fix typo in binary-relations.lagda.md

bjarki781:
It said before that a relation is irreflexive if there exists a function (x : A) -> neg (R x y).
Should be R x x


fredrik-bakke:
Indeed. Thanks for the correction!

PR #1181, opened 2024-09-09
Equivalences of synthetic categories

ivankobe:
Defined sections, retractions and equivalences of synthetic categories and proved lemma 1.1.6. from the book.

### DIFF
--- a/src/anel-surjections.lagda
+++ b/src/anel-surjections.lagda
@@ -1,0 +1,1 @@
+module anel-surjections where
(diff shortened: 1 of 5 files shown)


EgbertRijke:
Awesome!

We have some pre-commit functionality, which must be run using `make pre-commit` from the terminal. 

`pre-commit` is a python package that can be installed separately, see the agda-unimath installation guide if you don't have it yet.

PR #1182, opened 2024-09-09
defined sections, retractions and equivalences and proved lemma 1.1.6.

ivankobe:


EgbertRijke:
Great! I'm merging this one so that I can also continue working on it, even though I would have some suggestions for improvements.

PR #1189, opened 2024-10-06
Idea text `set-theory`

fredrik-bakke:
Adds an idea text to `set-theory`. Comments are very welcome!

Also, moves Russel's paradox to the `set-theory` namespace, fixes some typographical things, and adds a few number sequences for some reason.

### DIFF
--- a/src/set-theory.lagda.md
+++ b/src/set-theory.lagda.md
@@ -1,5 +1,40 @@
 # Set theory
 
+## Idea
+
+In univalent type theory, what we refer to formally as a _set_ is only in one
+sense what is clasically understood to be a "set". Namely, we say a set is a
+type whose [equality relation](foundation-core.identity-types.md) is a
+[proposition](foundation-core.propositions.md). I.e., any two elements can be
+equal in [at most one](subterminal-types.md) way.
+
+However, both historicallly and in contemporary mathematics, what is commonly
+meant by set theory is also the study of a collection of related foundational
+theories whose basic notions include a _universe_ of _sets_ and a
+propositionally valued _elementhood relation_ on the universe of sets.
+
+While this elementhood relation is not built into type theory as a fundamental
+construct, there is a clear substitute for it — namely, the smallness predicate:
+
+```text
+  is-small l A := Σ (X : UU l), (A ≃ X).
+```
+
+We can say precisely that a type `A` _is an element_ of `UU l` if `A` is
+`UU l`-small in this sense. Indeed, that `is-small l` is a predicate is
+equivalent to the [univalence axiom](foundation-core.univalence.md). This
+highlights a second connection between set theory and univalent type theory
+which is incompatible with the assumption that "set theory must be a study of
+set-level mathematics".
+
+In this module, we assemble ideas historically related to the study of set
+theories both as foundations of set-level mathematics, but also as a study of
+hierarchies in mathematics. This includes ideas such as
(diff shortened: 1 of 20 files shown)


fredrik-bakke:
It seems to me that the inclusion of Russel's paradox in `set-theory` changes the scope of the namespace, which has been partly unclear to me. Hence why I believe an idea text is needed.

src/set-theory.lagda.md
```diff
@@ -1,5 +1,41 @@
...
+In univalent type theory, what we refer to formally as a _set_ is only in one
+sense what is clasically understood to be a "set". Namely, we say a set is a
+type whose [equality relation](foundation-core.identity-types.md) is a
+[proposition](foundation-core.propositions.md). I.e., any two elements can be
+equal in [at most one](foundation.subterminal-types.md) way.
+
+However, both historically and in contemporary mathematics, what is commonly
+meant by set theory is also the study of a collection of related foundational
+theories whose building blocks include a _universe_ of _sets_ and a
+propositionally valued _elementhood relation_ on the universe of sets.
```
EgbertRijke, line 14: It would be great if you have a citation for this claim

src/set-theory.lagda.md
```diff
@@ -1,5 +1,41 @@
...
+```text
+  is-small l A := Σ (X : UU l), (A ≃ X).
+```
+
+We can say that a type `A` _is an element_ of `UU l` if `A` is `UU l`-small in
+this sense. Indeed, that `is-small l` is a predicate is equivalent to the
+[univalence axiom](foundation-core.univalence.md). This highlights a second
+connection between set theory and univalent type theory that is not directly
+compatible with the preconception that "set theory is a study of set-level
+mathematics".
```
EgbertRijke, line 29: Can you elaborate?

fredrik-bakke: to you or in the prose? the idea I suppose is that the universe of sets need not be set-level

src/set-theory.lagda.md
fredrik-bakke: I don't have one off-hand, but I suppose I should go look

EgbertRijke: I am not entirely sure if ZFC is a set theory according to this point of view, because it doesn't include large cardinal axioms and hence not universes.

fredrik-bakke: ZFC is a formal language with a sort, namely the sort of sets, which would be the universe in this picture.

fredrik-bakke: In _Set Theory, an introduction to independence proofs_, Kenneth Kunen uses the word "universe" in the context of ZFC to refer to the collective concept of all sets.

fredrik-bakke:
let me know what you think of the current draft. I am also looking for input on whether this indeed is a good/the correct scope for the namespace.

src/set-theory.lagda.md
EgbertRijke: I think it would be help to clarify that there are several possible meanings of the word universe here. It seems that Kunen uses it to mean "domain of discourse in the theory of sets"/"the sort of sets in the first order theory of sets" whereas in the context of type theory we use it to mean a universe object. I.e., there is an internal and external perspective that currently gets conflated in your sentence.

EgbertRijke: I see that you rephrased it a little, and it got improved. Thanks so much!

PR #1191, opened 2024-10-07
Add citation — computational identity types

fredrik-bakke:
Adds explicit citations to Martín Escardó for the constructions of the involutive and Yoneda identity types.


fredrik-bakke:
I've confirmed with Martín that the attribution is correct, so this PR is ready for review/merging.

VojtechStep:
🚀

PR #1201, opened 2024-10-17
100 Theorems

fredrik-bakke:
Adds a page about our progress on [_Formalizing 100 Theorems_](https://www.cs.ru.nl/~freek/100/).

### DIFF
--- a/src/elementary-number-theory/positive-integers.lagda.md
+++ b/src/elementary-number-theory/positive-integers.lagda.md
@@ -127,8 +127,8 @@ is-set-positive-ℤ : is-set positive-ℤ
 is-set-positive-ℤ =
   is-set-type-subtype subtype-positive-ℤ is-set-ℤ
 
-set-positive-ℤ : Set lzero
-set-positive-ℤ = positive-ℤ , is-set-positive-ℤ
+positive-ℤ-Set : Set lzero
+positive-ℤ-Set = positive-ℤ , is-set-positive-ℤ
 ```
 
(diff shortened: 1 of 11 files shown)


src/elementary-number-theory/positive-integers.lagda.md
```diff
@@ -122,6 +126,9 @@ is-nonzero-is-positive-ℤ {inr (inr x)} H ()
 is-set-positive-ℤ : is-set positive-ℤ
 is-set-positive-ℤ =
   is-set-type-subtype subtype-positive-ℤ is-set-ℤ
+
+set-positive-ℤ : Set lzero
+set-positive-ℤ = positive-ℤ , is-set-positive-ℤ
```
EgbertRijke, line 131:
```suggestion
positive-ℤ-Set : Set lzero
positive-ℤ-Set = positive-ℤ , is-set-positive-ℤ
```

EgbertRijke (approved):
Very nice addition to the library! It's great to see a record of the theorems from Freek's list that we already have in the library.

EgbertRijke:
This PR is already on auto-merge, so I don't want to make any suggestions for changes in this PR. But would it be nice to add credit to the original formalizers of those theorems, and a GitHub tag for pull requests that complete a new theorem?

fredrik-bakke:
Oh, yeah, that would be good!

EgbertRijke:
Maybe we can do that with the OEIS sequences as well, although some of those sequences are quite silly.

PR #1202, opened 2024-10-17
Bezout's lemma – 100 theorems

fredrik-bakke:
Turns out I missed a few results.

### DIFF
--- a/src/literature/100-theorems.lagda.md
+++ b/src/literature/100-theorems.lagda.md
@@ -54,14 +54,15 @@ open import elementary-number-theory.binomial-theorem-natural-numbers using
 
 > This is not yet formalized.
 
-### [60. Bezout's Theorem](https://www.cs.ru.nl/~freek/100/#60) {#60}
+### [60. Bezout's Lemma](https://www.cs.ru.nl/~freek/100/#60) {#60}
+
+Note that the 60th theorem in Freek's list is listed as "Bezout's Theorem", while the linked theorems are formalizations of Bezout's lemma, even though these are different statements.
 
 ```agda
 open import elementary-number-theory.bezouts-lemma-integers using
   ( bezouts-lemma-ℤ)
 open import elementary-number-theory.bezouts-lemma-natural-numbers using
   ( bezouts-lemma-ℕ)
-```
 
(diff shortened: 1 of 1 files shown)


fredrik-bakke:
@EgbertRijke I'm double-checking Freek's list now and there are some theorems I'm a little unsure about. For instance, do we have the "[The Factor and Remainder Theorems](https://www.cs.ru.nl/~freek/100/#89)"?

EgbertRijke:
No we have nothing about polynomial rings

src/literature/100-theorems.lagda.md
```diff
@@ -54,6 +54,15 @@ open import elementary-number-theory.binomial-theorem-natural-numbers using
...
 > This is not yet formalized.
 
+### [60. Bezout's Theorem](https://www.cs.ru.nl/~freek/100/#60) {#60}
+
+```agda
+open import elementary-number-theory.bezouts-lemma-integers using
+  ( bezouts-lemma-ℤ)
+open import elementary-number-theory.bezouts-lemma-natural-numbers using
+  ( bezouts-lemma-ℕ)
+```
```
EgbertRijke, line 64: Careful, Bezout's theorem and Bezout's lemma are different things

fredrik-bakke: Exactly! That's why I missed it before. Judging from the formalizations that are linked from Freek's list, they really mean Bezout's lemma.

EgbertRijke: That's wild:/

EgbertRijke: Maybe we should ask Freek about this, because I find it kind of unsettling to list Bezout's lemma as Bezout's theorem, even if all other libraries seem to do it.

fredrik-bakke:
Would you say we have the "Formula for the Number of Combinations"? 😬

src/literature/100-theorems.lagda.md
fredrik-bakke: Alright, I'll do it! I'm already in the process of writing him an e-mail anyways. I'll CC you

fredrik-bakke:
Alright, I guess this PR is ready for merging. Any further additions can be made as they are found/written.

PR #1206, opened 2024-10-20
WIP: a constructive Cantor–Schröder–Bernstein–Escardó theorem?

fredrik-bakke:


EgbertRijke:
This is very cool!

EgbertRijke:
I keep going back and forth in my thoughts about whether the Escardó postfix is appropriate here. The Escardó postfix is supposed to be for the version that Escardó proved, which generalizes CSB from sets and injective maps to arbitrary types and embeddings. The generalization to types is still true here, but the setup is a bit different, with double-negation stable embeddings.

You're still proving it for arbitrary types, aren't you, or are you doing CSB for double negation sheaves?

fredrik-bakke:
Fairly certain it works for arbitrary types; conditions are only added to the maps. Still, I won't promise any results before they are formalized.

EgbertRijke:
Good luck! It's a great project!

fredrik-bakke:
Thanks! Hopefully I'll have it done before my seminar next Friday :)

EgbertRijke:
Let me know if you need a fast review, so that you have something to show off on your seminar:)

fredrik-bakke:
After a bit more thinking I've realized there is still a non-trivial amount of logic to add before this PR is done. Would it be reasonable to start writing in a `logic` namespace, and then we can move other files about logic to that namespace in a later PR? @EgbertRijke

EgbertRijke:
> After a bit more thinking I've realized there is still a non-trivial amount of logic to add before this PR is done. Would it be reasonable to start writing in a `logic` namespace, and then we can move other files about logic to that namespace in a later PR? @EgbertRijke

Yes, that sounds fair. The idea of a logic namespace came up before, and I agree that it is a good idea to start making one.

(In a parenthetical remark, though, I am a bit concerned that all credit to the original content creators of a file is removed when a file is moved. We obviously had this problem before, but I wasn't so aware of it)

fredrik-bakke:
That's even more of a reason for placing files in the correct place to begin with, although that doesn't resolve the issue of moving files. Obviously, we can't just stop moving files... right?

EgbertRijke:
Yes, we should not stop moving files either. We should probably make an effort to hard-code credit at the beginning of each file, so that people can add their own names and they will stay when files are moved.

That still leaves the problem when files are split up, but at least that problem is smaller than the problem we're having now. I feel quite bad that Brian Lu (blu-bird) isn't credited anymore for his formalization work

fredrik-bakke:
We should discuss using the GitHub attribution script to add initial hardcoded credits that we can then edit later. Sadly, I don't think I have the time to engage in something like this at the moment, but I fully support it.

fredrik-bakke:
I see that we usually phrase small decidable predicates in terms of functions into `Fin 2`, such as for the omniscience principles. However, I'd like to instead use `bool` because I'd like to introduce `is-true := Eq-bool true` and `is-false := Eq-bool false` predicates, and find the fact that one has to pattern match on the elements of `Fin 2` twice a bit hassling. Is it okay if I continue using `bool` over `Fin 2` in my developments in logic @EgbertRijke?

src/domain-theory/directed-complete-posets.lagda.md
```diff
@@ -0,0 +1,174 @@
+# Directed complete posets
```
EgbertRijke, line 1: Would it be possible to move existing files in a separate pull request that only moves files, like I did in #1223 to preserve your authorship of some files about globular types?

fredrik-bakke: Sure! I'll see about how to solve this

fredrik-bakke: Would you mind answering my question above in the PR, btw?

fredrik-bakke:
Turns out the approach I had in mind for this PR doesn't quite work and needs to be rethought a little. Since this PR still covers some additions not covered by other PRs I'll leave it open, but ultimately this PR should be closed until a correct approach is found.

fredrik-bakke:
Whelp, I finally pieced together an incredibly crude formalization of the theorem! Turns out WLPO + decidable embeddings is enough. I'll clean up the code one of the coming days so it can be reviewed. The relevant code is in `cantor-schroder-bernstein-wlpo`

fredrik-bakke:
Alright, I'm opening this PR up for review now. Sorry if parts of it appear less polished. I've sort of lost track of the contents of this PR.

fredrik-bakke:
Never mind. I'd like to let this stew last for a little while longer first.

fredrik-bakke:
Invoking our self-review policy, I will review and merge this PR in a couple of days.

fredrik-bakke:
Just to be clear, this PR does not affect the licensing issues laid out in #1635 and should be considered independently of that.

PR #1208, opened 2024-10-23
Move OEIS to literature

EgbertRijke:
This PR moves the OEIS module to the literature

### DIFF
--- a/src/literature/oeis.lagda.md
+++ b/src/literature/oeis.lagda.md
@@ -10,7 +10,6 @@ module literature.oeis where
 <details><summary>Imports</summary>
 
 ```agda
-open import elementary-number-theory.ackermann-function
 open import elementary-number-theory.exponentiation-natural-numbers
 open import elementary-number-theory.natural-numbers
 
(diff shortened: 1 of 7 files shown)


src/literature/oeis.lagda.md
```diff
@@ -188,3 +175,7 @@ A006369 = map-collatz-bijection
 A046859 : ℕ → ℕ
 A046859 n = ackermann n n
```
fredrik-bakke, line 176: Do we perhaps want to list this as
```suggestion
A046859 n = ackermann n n
  where open import elementary-number-theory.ackermann-function
```

to make the connection to where the function is coming from? On the other hand, `ackermann` is still clickable...

EgbertRijke: The better way is just to introduce the diagonal of the Ackermann function as its own function in the file about the Ackermann function, and then introduce it here by importing it like the other entries.

fredrik-bakke: I suppose so. And then just leave the obvious functions as is?

fredrik-bakke: are you leaving this for a future PR?

EgbertRijke: I just did it. I was in the middle of something in my globular pull request so I couldn't immediately switch branches.

fredrik-bakke (approved):
Looks good to me. Feel free to merge when ready

PR #1209, opened 2024-10-25
Fix euclid-mullin

EgbertRijke:
It turns out that I made the recursive call incorrectly in the original PR that defined the Euclid-Mullin sequence. This PR corrects the mistake.


EgbertRijke:
Since this is a simple bug-fix, I'm gonna go ahead and merge this

PR #1212, opened 2024-10-27
The number of decidable subtypes of a finite type

EgbertRijke:
This PR does some bureaucracy to compute the number of decidable subsets of a finite type, mostly based on already existing formalization.


fredrik-bakke (approved):
Very nice! 😁

PR #1217, opened 2024-10-29
chore: some fixes to links

fredrik-bakke:
Noticed a couple of issues after the recent changes.

### DIFF
--- a/CONTRIBUTORS.toml
+++ b/CONTRIBUTORS.toml
@@ -35,7 +35,7 @@ in Trondheim. His research is on homotopy type theory and higher category theory
 displayName = "Elisabeth Stenholm"
 maintainer = true
 usernames = [ "Elisabeth Stenholm", "Elisabeth Bonnevier" ]
-homepage = "https://elisabeth.bonnevier.one"
+homepage = "https://elisabeth.stenholm.one"
 github = "elisabethstenholm"
 bio = '''
 Elisabeth is a PhD student at the University of Bergen. Her research is on
--- a/src/graph-theory/directed-graph-structures-on-standard-finite-sets.lagda.md
+++ b/src/graph-theory/directed-graph-structures-on-standard-finite-sets.lagda.md
@@ -60,7 +60,7 @@ Labeled-Finite-Directed-Graph = Σ ℕ (λ n → Fin n → Fin n → ℕ)
 
 - [Digraph](https://ncatlab.org/nlab/show/digraph) at $n$Lab
 - [Directed graph](https://ncatlab.org/nlab/show/directed+graph) at $n$Lab
-- [Directed graph](https://www.wikidata.org/entity/Q1137726) on Wikdiata
+- [Directed graph](https://www.wikidata.org/entity/Q1137726) on Wikidata
 - [Directed graph](https://en.wikipedia.org/wiki/Directed_graph) at Wikipedia
 - [Directed graph](https://mathworld.wolfram.com/DirectedGraph.html) at Wolfram
   MathWorld
--- a/src/literature/100-theorems.lagda.md
+++ b/src/literature/100-theorems.lagda.md
@@ -1,4 +1,4 @@
-# Formalizing 100 Theorems
+# Freek Wiedijk's 100 Theorems
 
 This file records formalized results from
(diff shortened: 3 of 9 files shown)


CONTRIBUTORS.toml
```diff
@@ -35,7 +35,7 @@ in Trondheim. His research is on homotopy type theory and higher category theory
 displayName = "Elisabeth Stenholm"
 maintainer = true
 usernames = [ "Elisabeth Stenholm", "Elisabeth Bonnevier" ]
-homepage = "https://elisabeth.bonnevier.one"
+homepage = "https://elisabeth.stenholm.one"
```
EgbertRijke, line 38: Oh my! Did I introduce a bad mistake? Sorry about that

src/literature/100-theorems.lagda.md
```diff
@@ -1,4 +1,4 @@
-# Formalizing 100 Theorems
+# Freek Wiedijk's 100 Theorems
```
EgbertRijke, line 1: I liked the old title because it is the title of the webpage "Formalizing 100 theorems"

EgbertRijke:
Thanks for catching these! I noticed a couple of others, but didn't make a PR yet.

1. Jonathan isn't linked in the readme. I wonder if I didn't save my search-replace on his name and left it out in other places too.
2. Freek's name is sometimes split over multiple sentences, and I didn't link those occurrences of his name. Perhaps some other names aren't linked for the same reason.

src/literature/100-theorems.lagda.md
fredrik-bakke: I don't know. I find the current title is hard to contextualize without opening the file. Without the added context, I don't think it is that easy for a reader to understand the difference between a list of "notable theorems" and "100 theorems", so the added context seems important to me.
Screenshot 2024-10-29 at 17 18 36

CONTRIBUTORS.toml
fredrik-bakke: not bad, just one that should be corrected :)

fredrik-bakke:
> Thanks for catching these! I noticed a couple of others, but didn't make a PR yet.
> 
> 1. Jonathan isn't linked in the readme. I wonder if I didn't save my search-replace on his name and left it out in other places too.
> 2. Freek's name is sometimes split over multiple sentences, and I didn't link those occurrences of his name. Perhaps some other names aren't linked for the same reason.

Do you want me to correct these as well?

src/literature/100-theorems.lagda.md
EgbertRijke: Ok, I agree with your change.

Another thing is that this and the OEIS aren't really literature. These lists would fit better under Library Contents, as in

```
6. Library contents
  6.1 100 Theorems
  6.2 OEIS
  6.3 Full list
```

But this doesn't have to be addressed in this PR. I'll mark this comment as resolved now

src/graph-theory/directed-graph-structures-on-standard-finite-sets.lagda.md
```diff
@@ -60,7 +60,7 @@ Labeled-Finite-Directed-Graph = Σ ℕ (λ n → Fin n → Fin n → ℕ)
 
 - [Digraph](https://ncatlab.org/nlab/show/digraph) at $n$Lab
 - [Directed graph](https://ncatlab.org/nlab/show/directed+graph) at $n$Lab
-- [Directed graph](https://www.wikidata.org/entity/Q1137726) on Wikdiata
+- [Directed graph](https://www.wikidata.org/entity/Q1137726) on Wikidata
```
EgbertRijke, line 63: Lol, how did you catch that one?

EgbertRijke:
Thanks for the fixes! Feel free to merge when you're ready

src/graph-theory/directed-graph-structures-on-standard-finite-sets.lagda.md
fredrik-bakke: Well... I started textsearching for every line in the library that ended with a capitalized word to identify first names. Then I started eliminating the most common words, one of them was `Wikidata`. However, this line still matched which caught my eye :)

EgbertRijke: Awesome!

PR #1222, opened 2024-11-14
Correcting an incorrect definition of discrete relations and discrete graphs

EgbertRijke:
This PR corrects an incorrect definition.


src/foundation/discrete-relations.lagda.md
```diff
@@ -22,12 +22,14 @@ open import foundation-core.propositions
...
+A [reflexive relation](foundation.binary-relations.md) `R` on `A` is said to be
+{{#concept "discrete" Disambiguation="reflexive relations valued in types" Agda=is-discrete-Reflexive-Relation}}
 if, for every element `x : A`, the type family `R x` is
 [torsorial](foundation-core.torsorial-type-families.md). In other words, the
 [dependent sum](foundation.dependent-pair-types.md) `Σ (y : A), (R x y)` is
-[contractible](foundation-core.contractible-types.md) for every `x`. The
+[contractible](foundation-core.contractible-types.md) for every `x`.
+
+The
 {{#concept "standard discrete relation" Disambiguation="binary relations valued in types"}}
```
fredrik-bakke, line 33:
```suggestion
{{#concept "standard discrete relation" Disambiguation="reflexive relations valued in types"}}
```

EgbertRijke:
Maybe an even more correct definition of discrete directed graph is that it is (0 -> 1)-null. This is a proposition again.

EgbertRijke:
> Maybe an even more correct definition of discrete directed graph is that it is (0 -> 1)-null. This is a proposition again.

Nevermind, if a graph is (0 -> 1)-null then it has a set of vertices. That's too restrictive.

EgbertRijke:
I think the definition in this pull request is the way to go.

fredrik-bakke:
I'm confused by why discreteness can't be a property, and I'm too tired to think at the moment, but thanks for fixing the definitions. I'll complete a review when I've rested :)

EgbertRijke:
It appears that I still had it wrong: A directed graph should be discrete if its edge relation is empty. This makes it so that the inclusion of discrete directed graphs is left adjoint to the forgetful functor `(V , E) |-> V` from directed graphs to types. Correcting it now.

src/foundation/discrete-reflexive-relations.lagda.md
```diff
@@ -22,41 +22,22 @@ open import foundation-core.propositions
...
+A [reflexive relation](foundation.binary-relations.md) `R` on `A` is said to be
+{{#concept "discrete" Disambiguation="reflexive relations valued in types" Agda=is-discrete-Reflexive-Relation}}
 if, for every element `x : A`, the type family `R x` is
 [torsorial](foundation-core.torsorial-type-families.md). In other words, the
 [dependent sum](foundation.dependent-pair-types.md) `Σ (y : A), (R x y)` is
-[contractible](foundation-core.contractible-types.md) for every `x`. The
-{{#concept "standard discrete relation" Disambiguation="binary relations valued in types"}}
-on a type `X` is the relation defined by
-[identifications](foundation-core.identity-types.md),
+[contractible](foundation-core.contractible-types.md) for every `x`.
```
fredrik-bakke, line 30: Could you add the explanation that this is an initiality condition here as well?

fredrik-bakke (approved):
Looks good!

fredrik-bakke:
Enabling auto-merge now 👍

PR #1225, opened 2024-11-18
Move DCPOs and directed families to new `domain-theory` namespace

fredrik-bakke:


EgbertRijke:
Hm... so it is worse than I thought. Even if a file is merely moved, GitHub doesn't track who edited it:(

If this doesn't help to preserve authorship of files, is it still worth it to merge this PR?

fredrik-bakke:
Perhaps the issue is with the `chore:` prefix of the PR? Let's try merging this one without the prefix. At least it doesn't hurt to try.

VojtechStep:
The issue isn't with the prefix. That's our internal thing. Git straight up doesn't pick up history from the file's previous location

VojtechStep:
Apparently `git log` has a `--follow` option, but I don't see it available for `git shortlog`. If we want to fix this behavior, then I'd rather discontinue the Git authorship tracking than try to fix its current state

fredrik-bakke:
That's unfortunate; then it sounds like it's meaningless to merge this PR. It's still the same for me though, so if @EgbertRijke wants this merged that's totally fine with me.

PR #1231, opened 2025-01-03
Composition operations on spans of spans

fredrik-bakke:
This PR defines
- composition of spans
- spans of spans
- vertical composition of spans of spans
- horizontal composition of spans of spans


src/foundation/type-arithmetic-standard-pullbacks.lagda.md
```diff
@@ -0,0 +1,417 @@
+# Type arithmetic with standard pullbacks
```
EgbertRijke, line 1: Great idea to separate out these results!

EgbertRijke:
Very nice pull request!

PR #1233, opened 2025-01-04
Rename wild higher categories

fredrik-bakke:
This PR renames "noncoherent wild higher categories" to "noncoherent ω-categories". The latter is both shorter and more descriptive. Additionally, this PR renames "isomorphisms in ω-categories" to "coinductive isomorphisms in ω-categories", as the concept that was defined is strictly weaker than the proper definition of an isomorphism in an ω-category, as discussed with @VojtechStep.

Co-authored-by: VojtechStep [vojtechstepancik@outlook.com](mailto:vojtechstepancik@outlook.com)


EgbertRijke:
What is the precise proposal for the use of the word "noncoherent"? Is it for omega-categories without any laws at all?

PR #1234, opened 2025-01-06
Fix terminology for π-finite types

fredrik-bakke:
Implements the naming change
- "π-finite" ↦ "untruncated π-finite"
- "truncated π-finite" ↦ "π-finite"

and formalizes the missing results where it is stated that a type is π-finite, but it is only formalized that it is untruncated π-finite

Co-authored-by: Egbert Rijke <e.m.rijke@gmail.com>

### DIFF
--- a/src/univalent-combinatorics.lagda.md
+++ b/src/univalent-combinatorics.lagda.md
@@ -15,9 +15,10 @@ the isomorphic objects are identified by the univalence axiom. Univalence can
 therefore help with counting finite structures up to isomorphism. The main piece
 of machinery that helps in this task is the general notion of π-finiteness. A
 level `k` π-finite type is a type that has finitely many connected components,
-such that all its homotopy groups up to level `k` are finite. The π-finite types
-enjoy useful closure properties, such as closedness under Σ, cartesian products,
-coproducts, and closedness under Π under a mild condition.
+such that all its homotopy groups up to level `k` are finite and all its
+homotopy groups above level `k` are trivial. The π-finite types enjoy useful
+closure properties, such as closedness under Σ, cartesian products, coproducts,
+and closedness under Π under a mild condition.
 
 ## Modules in the univalent combinatorics namespace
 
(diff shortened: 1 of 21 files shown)


src/univalent-combinatorics.lagda.md
```diff
@@ -15,9 +15,10 @@ the isomorphic objects are identified by the univalence axiom. Univalence can
 therefore help with counting finite structures up to isomorphism. The main piece
 of machinery that helps in this task is the general notion of π-finiteness. A
 level `k` π-finite type is a type that has finitely many connected components,
-such that all its homotopy groups up to level `k` are finite. The π-finite types
-enjoy useful closure properties, such as closedness under Σ, cartesian products,
-coproducts, and closedness under Π under a mild condition.
+such that all its homotopy groups up to level `k` are finite and all its
+homotopy groups above level `k` are trivial. The π-finite types enjoy useful
+closure properties, such as closedness under Σ, cartesian products, coproducts,
+and closedness under Π under a mild condition.
```
EgbertRijke, line 21: This last sentence could have a citation to Anel

src/univalent-combinatorics/pi-finite-types.lagda.md
```diff
@@ -9,654 +9,120 @@ module univalent-combinatorics.pi-finite-types where
...
-{{#concept "πₙ-finite" Disambiguation="type" Agda=is-π-finite Agda=π-Finite}} if
-it has [finitely](univalent-combinatorics.finite-types.md) many
-[connected components](foundation.connected-components.md) and all of its
-homotopy groups up to level `n` at all base points are finite.
+{{#concept "πₙ-finite" Disambiguation="type" Agda=is-π-finite Agda=π-Finite-Type}}
+if it has [finitely](univalent-combinatorics.finite-types.md) many
+[connected components](foundation.connected-components.md), all of its homotopy
+groups up to level `n` at all base points are finite, and all higher homotopy
+groups are [trivial](group-theory.trivial-groups.md). A type is
+{{#concept "π-finite"}} if it is πₙ-finite for some `n`.
```
EgbertRijke, line 35: Thank you so much for taking this on!

I like the definition of pi-finiteness with an explicit truncation level. We should keep in mind that the notion of pi-finiteness in Anel is with an unspecified truncation level, however. I think it is useful to have a specified truncation level, so I'd like to keep the definition as it is for now, but perhaps we should think of a name for the unspecified version.

fredrik-bakke: Indeed!

PR #1238, opened 2025-01-18
Agda version 2.7.0

fredrik-bakke:
I think it's about time we consider updating the library to Agda 2.7.0. It's very unfortunate that the newest version is still not hosted with homebrew, but considering it's been over 5 months now and [nothing's happening](https://github.com/Homebrew/homebrew-core/pull/193315), we should consider moving away from depending on it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,8 +3,12 @@
 
   inputs = {
     # Stable 24.11 has Agda 2.7.0.1
-    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-24.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
     flake-utils.url = "github:numtide/flake-utils";
+    # We aim to support Python 3.8 as long as Ubuntu 20.24 has LTS,
+    # since it ships with that version. Python 3.8 itself is already
+    # EOL, so it was dropped from nixpkgs 24.05
+    nixpkgs-python.url = "github:NixOS/nixpkgs/nixos-23.11";
     # Nixpkgs with tested versions of mdbook crates;
     # may be removed once we backport new mdbook assets to our
     # modified versions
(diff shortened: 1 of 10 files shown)


flake.nix
```diff
@@ -2,8 +2,8 @@
   description = "agda-unimath";
 
   inputs = {
-    # Unstable is needed for Agda 2.6.4, latest stable 23.05 only has 2.6.3
-    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    # Stable 24.11 has Agda 2.7.0.1
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-24.11";
```
fredrik-bakke, line 6: @VojtechStep Please advise if this change is correct

fredrik-bakke: Oh, and I suppose the `flake.nix` file must be updated also. While Agda 2.7.0 is not available for Arch yet, I don't think we should wait for it to be this time.

VojtechStep: Hm, so there's a minor-ish problem — since Python 3.8 was EOLd in October (it doesn't even receive security updates anymore), it was dropped from nixpkgs 24.11. We can either
- update the flake to 24.11, and provide Python 3.9 in the environment. The downside being that Python 3.8 is still the version provided in Ubuntu 20.24, which has LTS until April, and I'd really like to support it, which will be harder to check if none of us is running that version of Python (I assume I'm the last one, since I'm using the flake)
- use an older version of nixpkgs just for Python. The downside is that it will make the first evaluation of the flake take ε longer

Now that I wrote it down I think the second approach is the better one, since we're already doing it for mdbook (though I'd like to update that at some point in the future), and it's kind of one of the reasons why flakes exist.

I'll open a PR to your branch shortly.

VojtechStep: Oops, I didn't know I can push to your branch directly

fredrik-bakke: That's alright. It's easier this way anyways

VojtechStep (commented):
The changes themselves LGTM, though I'm not sure if we want to move forward with this right now, since Agda 2.7.0 is only available through nix/guix and cabal, according to [repology](https://repology.org/project/agda/versions)

fredrik-bakke:
Alright, everything seems to be ready then. @EgbertRijke, you can merge this PR when you've updated your installation of Agda.

fredrik-bakke:
> The changes themselves LGTM, though I'm not sure if we want to move forward with this right now, since Agda 2.7.0 is only available through nix/guix and cabal, according to [repology](https://repology.org/project/agda/versions)

Oh. Why isn't the new version available on more platforms? It's not like there hasn't been enough time.

fredrik-bakke:
And can't people always install the Agda release using cabal, nix, or from source?

VojtechStep:
> Why isn't the new version available on more platforms?

Your guess is as good as mine, but it's weird that e.g. on Arch, Agda 2.6.4 got two patch updates this month, and it's been flagged out of date since September.

> And can't people always install the Agda release using cabal, nix, or from source?

They can, and apparently installing through cabal is the recommended method, which I didn't know. So I guess that's not a concern.

fredrik-bakke:
So then you support proceeding with a merge when Egbert is ready?

VojtechStep:
Yes

EgbertRijke:
Thanks for the PR! I'll switch to `cabal` too and fetch the latest stable version.

PR #1240, opened 2025-01-26
No-postfix-projections

djspacewhale:
Agda-2.7.0 uses postfix projections by default when casing, as opposed to previous versions of Agda used by the library. This is a very minor PR adding a `--no-postfix-projections` flag to agda-unimath.agda-lib to restore the prior behavior.


EgbertRijke:
Awesome, thank you for the fix!

PR #1241, opened 2025-01-26
Fix definition of ordinals

fredrik-bakke:
- Corrects the extensionality assumption for ordinals.
- Renames well-founded orders to well-founded trees.


src/order-theory/well-founded-relations.lagda.md
```diff
@@ -9,29 +9,34 @@ module order-theory.well-founded-relations where
...
 Given a type `X` equipped with a
-[binary relation](foundation.binary-relations.md) `_ϵ_ : X → X → Type` we say
-that the relation `_ϵ_` is **well-founded** if all elements of `X` are
+[binary relation](foundation.binary-relations.md) `_∈_ : X → X → Type` we say
+that the relation `_∈_` is
+{{#concept "well-founded" Disambiguation="binary relation" WDID=Q338021 WD="well-founded relation" Agda=is-well-founded-Relation Agda=Well-Founded-Relation}}
+if all elements of `X` are
 [accessible](order-theory.accessible-elements-relations.md) with respect to
-`_ϵ_`.
+`_∈_`.
```
EgbertRijke, line 31: Good catch on the notation! `\in` is indeed more convenient.

src/trees/well-founded-trees.lagda.md
```diff
@@ -0,0 +1,118 @@
+# Well-founded trees
```
EgbertRijke, line 1: Hm.. I am now worried that this is going to create a mess, because there aren't actually any trees in sight here. I'd expect a well-founded tree to have an underlying tree, and in trees, the edge relation is not transitive.

Given that there are no underlying trees here (in the current PR), shouldn't we just call these `transitive-well-ordered-relations`?

fredrik-bakke: Ah, so a well-ordered relation is what actually should be called a well-ordered tree?

EgbertRijke: Ay, I misspoke. When I said `well-ordered` I actually meant `well-founded`. 

But my point was that since there are no underlying trees in sight here, we shouldn't use the terminology of trees. We should just stick to the terminology of relations. So my suggestion was to move this back to order-theory, and call this file `transitive-well-founded-relations`. I apologize for confusion caused by my typo.

fredrik-bakke: Oh, I also meant `well-founded` haha. No worries, no confusion was caused!

fredrik-bakke: okay. implemented your suggestion now

EgbertRijke:
Thank you so much for making these improvements!

PR #1242, opened 2025-01-27
Uniqueness of split essential surjectivity of functors

fredrik-bakke:
This short PR formalizes that split essential surjectivity witnesses of fully faithful functors are unique. It also gives some generalizations of this.


src/category-theory/split-essentially-surjective-functors-precategories.lagda.md
```diff
@@ -166,14 +177,110 @@ module _
...
 
-### Being split essentially surjective is a property of fully faithful functors when the codomain is a category
+### Being split essentially surjective is a property of fully faithful functors when the domain is a category
 
-This remains to be shown. This is Lemma 6.8 of _Univalent Categories and the
-Rezk completion_.
+This is Lemma 6.8 of {{#cite AKS15}}, although we give a different proof.
+
+**Proof.** Let `F : 𝒞 → 𝒟` be a functor of precategories, where `𝒞` is
+univalent. It suffices to assume `F` is fully faithful on isomorphism-sets.
```
EgbertRijke, line 185: Usually we don't say "F is fully faithful on some subset of the morphisms". Do you mean that `core F` is a fully faithful functor, where `core` is the grouped core?

src/category-theory/split-essentially-surjective-functors-precategories.lagda.md
```diff
@@ -166,14 +177,110 @@ module _
...
+
+```text
+  (d : 𝒟₀) → Σ (x : 𝒞₀), (Fx ≅ d)
+```
+
+is a proposition it is equivalent to show that if it has an element it is
+contractible, so assume `F` is split essentially surjective. Then, it suffices
+to show that for every `d : 𝒟₀`, the type `Σ (x : 𝒞₀), (Fx ≅ d)` is
+contractible. By split essential surjectivity there is an element `y : 𝒞₀` such
+that `Fy ≅ d` and since postcompositing by an isomorphism is an equivalence of
```
EgbertRijke, line 197: Postcompositing isn't a word

```suggestion
that `Fy ≅ d` and since postcomposing by an isomorphism is an equivalence of
```

PR #1243, opened 2025-01-27
Some extensions of the fundamental theorem of identity types

fredrik-bakke:
So my recent work on π₀-trivial types got me looking for other applications of this concept, and here's one. Essentially, this PR gives an alternative phrasing of the extended fundamental theorem of identity types such that the assumption of inhabitedness/pointedness on the base type `A` is not needed.

Edit: the scope of this PR has grown since the above description was written.

### Summary

- Unbased version of the extended fundamental theorem of identity types
- Structured equality duality
- Strong preunivalence
- Strengthen definition of preunivalent categories
- Weaken assumptions on the type theoretic Yoneda lemma


fredrik-bakke:
Wait a second, I'm doing a stupid

EgbertRijke:
Do you use "structure" in the same way here as in the structure identity principle?

EgbertRijke:
> No, here structure is used differently. Here structure is what the theorem is used to construct, not to characterize.

Probably you should be very careful with terminology then. The title "fundamental theorem of identity types for structures" is suggestive of a theorem related to the structure identity principle. Could you think about phrasing your work in a less ambiguous way?

src/foundation/fundamental-theorem-of-identity-types-structures.lagda.md
```diff
@@ -0,0 +1,121 @@
...
+> TODO
+
+## Theorem
+
+### The unbased fundamental theorem of identity types for structures
+
+Given a structure `𝒫` on types, then the following are logically equivalent:
+
+1. Every unbased map out of the identity types of `A` into the type family
+   `B : A → 𝒰` is `𝒫`-structured.
```
EgbertRijke, line 44: It's very hard for me to make sense of this sentence. I'm afraid I haven't succeeded. I assume you are talking about maps

```text
  f : (x y : A) -> (x = y) -> B ?
```

here, but I have two choices of elements that could go on the question mark, `x` and `y`. I might also be misunderstanding that I should be thinking about one map, instead of a family of maps?

Also, A and B aren't introduced yet before this sentence.

fredrik-bakke: This PR is unfinished

fredrik-bakke: It should be a family of families of maps indexed by x and y. y goes in the questionmark in the formalization, but the opposite choice is just as valid

EgbertRijke: Sorry that I bothered you. I was excited to see what you came up with:) I'll let you finish the PR before commenting further.

fredrik-bakke:
> > No, here structure is used differently. Here structure is what the theorem is used to construct, not to characterize.
> 
> Probably you should be very careful with terminology then. The title "fundamental theorem of identity types for structures" is suggestive of a theorem related to the structure identity principle. Could you think about phrasing your work in a less ambiguous way?

Thanks, that is good to keep in mind! Sorry, this PR was originally something different, but then I realized I was thinking about things wrong so I need to think a little more before it is ready

fredrik-bakke:
As an aside, this PR made me realize there's a canonical strengthening of preunivalence one might ought to consider, so stay tuned for that :)

fredrik-bakke:
I figured out what the proper statement for the extended fundamental theorem should be, so we're back in business 😁

fredrik-bakke:
Okay, _now_ I think this PR is ready to be looked at. I renamed the previous "fundamental theorem of identity types for structures" to "structured equality duality", but that's just a name I'm making up so let me know what you feel about it. I also did some other things, see the PR main body.

src/category-theory/categories.lagda.md
```diff
@@ -170,8 +171,11 @@ module _
 
   is-preunivalent-category-Category :
     is-preunivalent-Precategory (precategory-Category C)
-  is-preunivalent-category-Category x y =
-    is-emb-is-equiv (is-category-Category C x y)
+  is-preunivalent-category-Category x =
+    is-set-is-contr
+      ( fundamental-theorem-id'
+        ( iso-eq-Precategory (precategory-Category C) x)
+        ( is-category-Category C x))
```
EgbertRijke, line 178: Just wondering, what was the issue with the old proof? Did it induce a cyclic dependency?

src/foundation/structured-equality-duality.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+open import foundation-core.retracts-of-types
+open import foundation-core.sections
+open import foundation-core.torsorial-type-families
+```
+
+</details>
+
+## Idea
+
+Given a [structure](foundation.structure.md) `𝒫` on types that transfers along
```
EgbertRijke, line 36: Can you explain in this idea section what it means to "transfer along equivalences"? Is it related to transport along equivalences, for which we have a file?

src/category-theory/categories.lagda.md
fredrik-bakke: The definition of preunivalent category has changed

src/foundation/structured-equality-duality.lagda.md
fredrik-bakke: Yes, it's the same, just that we add it as a hypothesis to avoid univalence, since usually this property is independent of univalence

src/foundation/structured-equality-duality.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+    tr-𝒫
+      ( inv-compute-fiber-map-out-of-identity-type f y b)
+      ( K (x , f x refl) (y , b))
+
+  structured-equality-duality :
+    ( (x : A) (f : (y : A) → (x ＝ y) → B y) (y : A) → structure-map 𝒫 (f y)) ↔
+    ( structure-equality 𝒫 (Σ A B))
+  structured-equality-duality =
+    ( forward-implication-structured-equality-duality ,
+      backward-implication-structured-equality-duality)
```
EgbertRijke, line 76: I'm definitely open to using the name "equality duality", but could you explain your reasoning behind this name a bit more (in the idea section)? Why can we think of this as a duality principle? Many duality principles establish a correspondence between something covariant and something contravariant. Is that the case here too?

src/foundation/structured-equality-duality.lagda.md
```diff
@@ -0,0 +1,120 @@
...
+        ( is-in-subuniverse-equiv 𝒫)
+
+  abstract
+    subuniverse-equality-duality :
+      ( (x : A) (f : (y : A) → (x ＝ y) → B y)
+        (y : A) → is-in-subuniverse-map 𝒫 (f y)) ↔
+      is-separated 𝒫 (Σ A B)
+    subuniverse-equality-duality =
+      structured-equality-duality (is-in-subuniverse-equiv 𝒫)
+```
```
EgbertRijke, line 120: Would you mind giving some "See also" entries to pages where this principle is applies (if any)?

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -38,9 +38,9 @@ open import foundation-core.torsorial-type-families
 
 ## Idea
 
-The **universal property of identity types** characterizes families of maps out
-of the [identity type](foundation-core.identity-types.md). This universal
-property is also known as the **type theoretic Yoneda lemma**.
+The {{#concept "universal property of identity types"}} characterizes families
```
EgbertRijke, line 41:
```suggestion
The {{#concept "universal property of identity types" Agda=...}} characterizes families
```

src/foundation/universal-property-identity-types.lagda.md
```diff
@@ -113,81 +113,126 @@ module _
...
-We first show that [the preunivalence axiom](foundation.preunivalence.md)
-implies that the map `Id : A → (A → 𝒰)` is an
-[embedding](foundation.embeddings.md). Since the
-[univalence axiom](foundation.univalence.md) implies preunivalence, it follows
-that `Id : A → (A → 𝒰)` is an embedding under the postulates of agda-unimath.
+We first show that [injectivity](foundation-core.injective-maps.md) of the map
 
-#### Preunivalence implies that `Id : A → (A → 𝒰)` is an embedding
+```text
+  equiv-eq : {X Y : 𝓤} → (X ＝ Y) → (X ≃ Y)
```
EgbertRijke, line 119: This looks like it's in bold face in my browser. Not sure if that's intended, but below I see universes in regular curly math.

src/foundation/strong-preunivalence.lagda.md
```diff
@@ -0,0 +1,144 @@
+# The strong preunivalence axiom
```
EgbertRijke, line 1: This is very cool!

src/foundation/universal-property-identity-types.lagda.md
EgbertRijke: It looks like the characterization of maps out of the identity types is the equivalence in [equiv-ev-refl](https://unimath.github.io/agda-unimath/foundation.universal-property-identity-types.html#2209). This would be something reasonable to link to.

src/foundation/universal-property-identity-types.lagda.md
fredrik-bakke: I'm assuming you mean the universe `𝓤`/`𝒰`

fredrik-bakke: If so, nice catch! It wasn't intentional.

src/foundation/structured-equality-duality.lagda.md
fredrik-bakke: Alright, so let me try to justify the use of the word duality here. I'd like to try and formulate that in some sense both sides are straightened _and_ unstraightened, but in opposite ways.

On the left-hand side, we have binary families of maps into `B` `f : (x y : A) -> (x = y) -> B y`, something one might view as a straightened thing, and these families of maps have `P`-structures on their fibers, something unstraightened. On the right-hand side, we're considering `Σ A B`, the literal unstraightening of `B`, but we are considering structures on the binary family of identity types over it, something more straightened in my mind. So we're doing a trade to have straightening in one direction for unstraightening in the other. This feels very duality-e to me.

I would like to say though, that this is also just a generalization of the Regensburg extension of the fundamental theorem, and therefore it might very well be more appropriate to give it a name closer to it. On the other hand, the fundamental theorem of identity types is about based identity types, while "structured equality duality" is not.

fredrik-bakke: Sorry that I'm being very loose in my arguments, I think I'll need to let the question rest for a little while until I find a good explanation or a better idea for a name.

src/foundation/structured-equality-duality.lagda.md
fredrik-bakke: Thanks for the clarification! I thought my previous explanation would be sufficient, but realize now that it clearly wasn't. It should be fixed now.

src/category-theory/preunivalent-categories.lagda.md
```diff
@@ -15,45 +15,46 @@ open import foundation.1-types
...
+A {{#concept "preunivalent category" Agda=Preunivalent-Category}} `𝒞` is a
+[precategory](category-theory.precategories.md) for which every mapping of the
+concrete groupoid of objects into the groupoid of
+[isomorphisms](category-theory.isomorphisms-in-precategories.md) is an
+[embedding](foundation-core.embeddings.md). Equivalently, by
+[subuniverse equality duality](foundation.structured-equality-duality.md), a
+preunivalent category is a precategory whose based isomorphism types
+`Σ (x : 𝒞₀), (* ≅ x)` are [sets](foundation-core.sets.md).
+
+The main function of _preunivalence_ is to serve as a common generalization of
```
EgbertRijke, line 41:
```suggestion
The main purpose of _preunivalence_ is to serve as a common generalization of
```

src/category-theory/preunivalent-categories.lagda.md
```diff
@@ -15,45 +15,46 @@ open import foundation.1-types
...
+[strengthening](foundation.strong-preunivalence.md) of the
+[preunivalence axiom](foundation.preunivalence.md).
 
 ## Definitions
 
 ### The predicate on precategories of being a preunivalent category
 
+We define preunivalence of a precategory `𝒞` to be the condition that for every
+`x : 𝒞₀`, the type `Σ (y : 𝒞₀), (x ≅ y)` is a set.
+
```
EgbertRijke, line 57: In situations like this, where you want to redefine something with a different axiom, it is a question whether it is a good idea to replace the old definition, or whether to keep a version of it around while also defining the new version.

I think it could be a good idea to keep the old version around under some name, because it is clearly a condition that could come up sometimes, and under the current proposal of this PR we won't have that old definition in the library anymore.

fredrik-bakke: While I'd currently like to think of the previous definition as "deprecated", I don't have a good rebuttal for the proposal to keep it. However, I'd like to give it a name that makes it clear it is an inferior condition. Perhaps we can call it "quasipreunivalent" or "weakly preunivalent"?

EgbertRijke: Calling these categories "weakly preunivalent" creates a strange situation though, because weak preunivalent categories would then satisfy the preunivalence axiom, and preunivalent categories would satisfy the strong preunivalence axiom. This can only create confusion.

I'd propose keeping the name "Preunivalent categories" for the old concept, and indeed not changing the name "preunivalence" in foundation either. Preunivalence is a somewhat established concept, that other people like Martin Escardo, Evan Cavallo, and Peter Lumsdaine are also interested in. Then we could name the new concept "Strongly preunivalent categories", which is in line with the name for the new axiom that you came up with.

That's quite a mouthful, but it could be an advantage. If people don't like the name "strongly preunivalent category" they have an obvious alternative: Bakke categories :)

fredrik-bakke: Fiiiiine, we'll call them strongly preunivalent. Is it okay if we remove most mentions of preunivalent categories over strongly preunivalent categories though? The preunivalence axiom (for all types) has the advantage over preunivalence for categories, AFAIK, that one can metatheoretically prove that `id-equiv : {l : Level} {X : UU l} -> X ≃ X` is the only type-polymorphic function of this signature, while for categories it is possible that one indeed has multiple families of morphisms of type `(x : obj C) -> iso C x x`.

EgbertRijke: Yes, that's fine. Thank you so much for doing that. 

Could you also include some discussion of why you think the stronger version has advantages? You've clearly thought deeply about these conditions and I could definitely learn from it if you'd write it down carefully.

fredrik-bakke: I've restored the previous definition now, with a note on the caveat

fredrik-bakke:
I've tried writing some explanations as you asked for. If you are not happy with them, I would prefer to change terminology and or remove formalizations until the explanations are no longer needed.

EgbertRijke:
Thank you so much! I think your extra note is very informative, I like it!

I think (or presume) that more people are familiar with the ordinary preunivalence principle than with the stronger version. Having the preunivalence principle in the library with this note that there is also a stronger common generalization is very worthwhile! 

So, thank you for going through the extra effort, and for this cool new insight about identity types! I'll merge this PR now.

PR #1250, opened 2025-01-30
Bump pipeline dependencies

fredrik-bakke:
Due to deprecations with the GitHub API, some of our pipeline dependencies have started failing today, and more will tomorrow. I'm taking the opportunity do to some midwinter cleaning and updating all of our dependencies. Here's the homework I've done, from most to least important:

- `actions/upload-pages-artifact@v3` updates to `upload-artifact@v4` (broke today) and removes a builtin `chmod` command we don't use
- `actions/deploy-pages@v4` updates to `upload-artifact@v4` (broke today), updates to Node20
- `actions/cargo-install@v>=3.2` update to new GitHub API (otherwise breaks tomorrow)
- `actions/configure-pages@v5` swaps to Next.js 13.3 and newer, updates to Node20
- `actions/cache@v4` updates to Node20
- `actions/checkout@v4` updates to Node20
- `actions/actions-mdbook@v3` updates to Node20
- `actions/setup-python@v5` updates to Node20
- `pre-commit/action@v3.0.1` updates dependency to `actions/cache@v4`

~Some input parameters have also been deprecated and removed, I don't know if this will affect us yet.~ Since there are no warnings in our CI logs that we are using deprecated input parameters, I'm assuming we're good on this front.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
         agda: ['2.7.0']
     steps:
       - name: Checkout our repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           path: master
       - name: Setup Agda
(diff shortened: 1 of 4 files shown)


VojtechStep:
The typechecking CI passes, so that looks good, and the only way to test deployment is to merge this, so go for it.

There were actual breaking changes that affect us in newer versions of mdbook, which IIRC some of the newer modules depend on, so I'd separate updating actions from updating mdbook.

PR #1256, opened 2025-01-31
The Archimedean property of `ℕ`

lowasser:
This is an instance where I'm not certain if this is the ideal type for the result.  

In the longer term, these PRs are building up to the Archimedean property for the rationals en route to addition on the reals.  The choice is whether to do an increasingly long chain of embeddings of `ℕ` in the integers, integer fractions, and rationals -- one of my drafts has `rational-fraction-Z (in-fraction-Z (int-N n))`, or to use [`multiple-Ab`](https://unimath.github.io/agda-unimath/group-theory.multiples-of-elements-abelian-groups.html), which defines natural multiplication for all contexts, at which point we might just define the Archimedean property in greater generality [on linearly ordered groups](https://en.wikipedia.org/wiki/Archimedean_property#Definition_for_linearly_ordered_groups).  That approach would also require lots of lemmas proving `multiple-Ab` works equivalently to other kinds of multiplication, but it's plausible.  (There also feels like a bit of tension about linear ordering, because the usual total orders are defined on normal inequality, but most sources I see on the Archimedean property use strict inequality.  I don't feel like I have a strong sense of whether that makes a difference.)

My sense is that the right choice is to do the embeddings, and then separately prove that they are equivalent to `multiple-Ab`, but I'm open to pushback on that.


fredrik-bakke:
While I'm not an expert on constructive real analysis, I would consider the following references to be among the most relevant for your development, since we work in univalent foundations:

- [1]: Homotopy Type Theory: Univalent Foundations of Mathematics. Chapter 11. https://hott.github.io/book/hott-online.pdf.html
- [2]: Analysis in Univalent Type Theory, Auke Booij. In particular, he derives an Archimedean principle for the reals in Lemma 6.7.3. https://etheses.bham.ac.uk/id/eprint/10411/7/Booij2020PhD.pdf

Both of these phrase the archimedian property in terms of strict inequality, so that seems like the safest bet.

Starting with the rational numbers, you can follow Definition 11.2.7 of an _ordered field_ in [1] and go from there. If you want to start with the natural numbers then you _could_ try to work out the proper definition and relations to an _ordered semiring_, but I don't know if that will take you all the way to where you want to go. Please note in particular, that working with interactions between strict and nonstrict inequalities constructively is more complicated and many times requires the structure of an [apartness relation](https://unimath.github.io/agda-unimath/foundation.apartness-relations.html) on the underlying type, as it does for the reals.

src/elementary-number-theory/archimedean-natural-numbers.lagda.md
```diff
@@ -0,0 +1,47 @@
...
+open import foundation.dependent-pair-types
+open import foundation.transport-along-identifications
+```
+
+</details>
+
+## Definition
+
+The Archimedean property of the natural numbers is that for any nonzero natural
+number `x` and natural number `y`, there is an `n : ℕ` such that `y < n *ℕ x`.
```
EgbertRijke, line 26: Note: this suggestion will break pre-commit rules, so after implementing this suggestion you'll need to pull this branch and run `make pre-commit` again.

```suggestion
The {{#concept "Archimedean property" Disambiguation="natural numbers" Agda=archimedean-property-ℕ}} of the [natural numbers](elementary-number-theory.natural-numbers.md) is that for any nonzero natural
number `x` and any natural number `y`, there is an `n : ℕ` such that `y < n *ℕ x`.
```

src/elementary-number-theory/archimedean-natural-numbers.lagda.md
```diff
@@ -0,0 +1,47 @@
+# The Archimedean property of the natural numbers
+
+```agda
+module elementary-number-theory.archimedean-natural-numbers where
```
EgbertRijke, line 4: I'd say that the file name would be slightly improved if it were `archimedian-property-natural-numbers.lagda.md`

EgbertRijke: (I made a typo, but edited the suggestion now)

src/elementary-number-theory/archimedean-natural-numbers.lagda.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Definition
+
+The Archimedean property of the natural numbers is that for any nonzero natural
+number `x` and natural number `y`, there is an `n : ℕ` such that `y < n *ℕ x`.
+
+```agda
+abstract
+  archimedean-property-ℕ :
+    (x y : ℕ) → is-nonzero-ℕ x → Σ ℕ (λ n → le-ℕ y (n *ℕ x))
```
EgbertRijke, line 31:
```suggestion
    (x y : ℕ) → is-nonzero-ℕ x → Σ ℕ (λ n → y <-ℕ n *ℕ x)
```

src/elementary-number-theory/archimedean-natural-numbers.lagda.md
```diff
@@ -0,0 +1,47 @@
...
+  archimedean-property-ℕ x y nonzero-x =
+    (succ-ℕ (quotient-euclidean-division-ℕ x y) ,
+      tr
+        ( λ z → le-ℕ z (succ-ℕ (quotient-euclidean-division-ℕ x y) *ℕ x))
```
EgbertRijke, line 35:
```suggestion
        ( λ z → z <-ℕ succ-ℕ (quotient-euclidean-division-ℕ x y) *ℕ x)
```

EgbertRijke:
Awesome, thank you so much! This will be merged when the checks are complete

fredrik-bakke:
The $n$Lab suggests that you can make sense of the notion of an "archimedean strictly ordered group", so if that's the route you want to go then it seems viable. Again, there is some nuance to the strict vs. non-strict ordering, so I'd advise you to follow the established approach of using a strict ordering.

PR #1258, opened 2025-01-31
Archimedean property of `ℤ`

lowasser:


src/elementary-number-theory/archimedean-property-integers.lagda.md
```diff
@@ -0,0 +1,71 @@
...
+open import foundation.transport-along-identifications
+open import foundation.unit-type
+```
+
+</details>
+
+## Definition
+
+The Archimedean property of the integers is that for any positive integer `x`
+and integer `y`, there is an `n : ℕ` such that `y < int-ℕ n *ℤ x`.
```
EgbertRijke, line 35:
```suggestion
The {{#concept "Archimedean property" Disambiguation="integers" Agda=archimedean-property-ℤ}} of the [integers](elementary-number-theory.integers.md) is that for any [positive integer](elementary-number-theory.positive-integers.md) `x`
and integer `y`, there is an `n : ℕ` such that `y < int-ℕ n *ℤ x`.
```

EgbertRijke:
Great pull request! We just need to take care of links in the idea section, but other than that it's good. Make sure to run `make pre-commit` again after the suggestion is merged.

PR #1261, opened 2025-02-02
Archimedean property of `fraction-ℤ`

lowasser:


src/elementary-number-theory/archimedean-property-integer-fractions.lagda.md
```diff
@@ -0,0 +1,52 @@
...
+The Archimedean property of the integer fractions is that for any
+`x y : fraction-ℤ`, with positive `x`, there is an `n : ℕ` such that `y` is less
+than `n` as an integer fraction times `x`.
+
+```agda
+abstract
+  archimedean-property-fraction-ℤ :
+    (x y : fraction-ℤ) →
+      is-positive-fraction-ℤ x →
+      Σ ℕ (λ n → le-fraction-ℤ y (in-fraction-ℤ (int-ℕ n) *fraction-ℤ x))
```
fredrik-bakke, line 39: Shouldn't the Archimedean property be expressed with existential quantification `∃`? Under your current formulation, the choice of `n` matters, and being Archimedean is a structure rather than a property. If you agree, then I propose you rename this entry to something like `bound-archimedean-property-fraction-ℤ` (up for discussion) and add the entry

```agda
abstract
  archimedean-property-fraction-ℤ :
    (x y : fraction-ℤ) →
    is-positive-fraction-ℤ x →
    ∃ ℕ (λ n → le-fraction-ℤ y (in-fraction-ℤ (int-ℕ n) *fraction-ℤ x))
  archimedean-property-fraction-ℤ x y u =
    unit-trunc-Prop (bound-archimedean-property-fraction-ℤ x y u)
```

below.

lowasser: I don't feel the need to preserve this formulation.

lowasser: #1262 corrects this for the previous versions; once that's merged, this one can follow.

fredrik-bakke: Is there a particular reason why you want to have it deleted? It's in the spirit of the library to record every piece of reusable logic, and this intermediate step seems like it has potential to be useful. From our [coding style document](https://unimath.github.io/agda-unimath/CODINGSTYLE.html):

> [...] we aim to factor out and encapsulate even the tiniest bits of reusable logic or computation in their own definitions.

lowasser: That philosophy makes sense to me, but I'm not sure it makes sense to codify the particular way in which the natural number is chosen unless it has some other useful properties.  If we wanted to show that this produces the lowest such viable `n`, for example, that makes sense to me.  I don't think it's the case that all existence proofs in this library that _could_ be bound in this way -- that could have a `Σ` version and not just a `∃` version -- actually do have both versions.

(In fairness, I think this approach might actually have that property that it produces the least such natural number, but I'd prefer to come back and prove that later than block this PR on that.)

lowasser: Put another way: it makes sense to me that coming back and adding the bound version might make sense in future, with a proof that there's something useful about the particular value being chosen.  But if `n` is the lowest natural number with the desired property, `bound-archimedean-property-fraction-ℤ` might return a pair with `n`, or with `succ-ℕ n`, or whatever.  If we don't want to allow consumers of `bound-archimedean-property-fraction-ℤ` to be sensitive to the exact choice of `n`, we could mark it `abstract`, but at that point it seems like you can't do anything with that function that you couldn't do with the existentially quantified variant.

fredrik-bakke: Forgive me, I don't have the excess capacity to discuss the finer details of this point today. Seeing as Egbert has already made the decision to merge #1262 I'm okay with proceeding with your plan. This is after all of very minor consequence in the short term.

lowasser: Sounds good -- and to reiterate, I'm happy to see this changed to the other approach in the near future, just with some extra theorems I want to save for another PR.

PR #1263, opened 2025-02-04
Lossy unification on arithmetically located cuts

lowasser:
This mostly just makes the compilation significantly faster.


fredrik-bakke (approved):
Thanks a lot!

PR #1264, opened 2025-02-04
Logic and compactness

fredrik-bakke:
A lot of results in logic and foundations.

### DIFF
--- a/src/foundation/apartness-relations.lagda.md
+++ b/src/foundation/apartness-relations.lagda.md
@@ -21,14 +21,16 @@ open import foundation-core.empty-types
 open import foundation-core.identity-types
 open import foundation-core.negation
 open import foundation-core.propositions
+open import foundation-core.truncation-levels
 ```
 
 </details>
 
 ## Idea
 
-An **apartness relation** on a type `A` is a
-[relation](foundation.binary-relations.md) `R` which is
+An
+{{#concept "apartness relation" Disambiguation="types" WD="apartness relation" WDID=Q4779193 Agda=Apartness-Relation Agda=is-apartness-relation}}
+on a type `A` is a [relation](foundation.binary-relations.md) `R` which is
 
 - **Antireflexive:** For any `a : A` we have `¬ (R a a)`
 - **Symmetric:** For any `a b : A` we have `R a b → R b a`
@@ -56,7 +58,7 @@ module _
 
   is-cotransitive-Prop : Prop (l1 ⊔ l2)
   is-cotransitive-Prop =
-    ∀' A (λ a → ∀' A (λ b → ∀' A (λ c → R a b ⇒ (R a c ∨ R b c))))
+    ∀' A (λ a → ∀' A (λ b → ∀' A (λ c → R a b ⇒ (R a c) ∨ (R b c))))
 
   is-cotransitive : UU (l1 ⊔ l2)
   is-cotransitive = type-Prop is-cotransitive-Prop
@@ -80,6 +82,9 @@ module _
 
   apart-Apartness-Relation : A → A → UU l2
   apart-Apartness-Relation x y = type-Prop (rel-Apartness-Relation x y)
+
+  is-apartness-Apartness-Relation : is-apartness-relation rel-Apartness-Relation
+  is-apartness-Apartness-Relation = pr2 R
 
   antirefl-Apartness-Relation : is-antireflexive rel-Apartness-Relation
   antirefl-Apartness-Relation = pr1 (pr2 R)
@@ -123,6 +128,11 @@ module _
   apart-Type-With-Apartness =
     apart-Apartness-Relation apartness-relation-Type-With-Apartness
 
+  is-apartness-rel-apart-Type-With-Apartness :
+    is-apartness-relation rel-apart-Type-With-Apartness
+  is-apartness-rel-apart-Type-With-Apartness =
+    is-apartness-Apartness-Relation apartness-relation-Type-With-Apartness
+
   antirefl-apart-Type-With-Apartness :
     is-antireflexive rel-apart-Type-With-Apartness
   antirefl-apart-Type-With-Apartness =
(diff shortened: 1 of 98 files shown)


src/foundation/apartness-relations.lagda.md
```diff
@@ -144,81 +154,206 @@ module _
...
+  Π-Type-With-Apartness : Type-With-Apartness (l1 ⊔ l2) (l1 ⊔ l3)
+  Π-Type-With-Apartness =
+    ((x : X) → type-Type-With-Apartness (Y x)) ,
+    rel-apart-Π-Type-With-Apartness X Y ,
+    is-antireflexive-apart-Π-Type-With-Apartness ,
+    is-symmetric-apart-Π-Type-With-Apartness ,
+    is-cotransitive-apart-Π-Type-With-Apartness
+```
+
+### Apartness on the type of functions into a type with an apartness relation
```
EgbertRijke, line 301: I could see a world where eventually we'd want to create a separate file `function-types-with-apartness-relation` for this, but let's just keep that in mind for another day:)

src/foundation/decidability-search-universal-quantification.lagda.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+Having ∀-decidability search is
+[logically equivalent](foundation.logical-equivalences.md) to having
+Π-decidability search, but the latter is not a
+[proposition](foundation-core.propositions.md).
+
+**Terminology.** In the terminology of Martín Escardó, a type that has
+∀-decidability search is referred to as _weakly compact_, or _Π-compact_, or
+_satisfying the weak principle of omniscience_. {{#cite TypeTopology}}
```
EgbertRijke, line 90: It would be useful here to mention also whether the terminology "∀-decidability" is our own name for the concept, or whether it is more widely used in other literature (I couldn't tell the answer from what has been written so far in this file). Same for all the related decidability searches.

src/foundation/decidability-search-untruncated-existential-quantification.lagda.md
```diff
@@ -0,0 +1,672 @@
...
+```text
+  (P : decidable-family X) → is-decidable (Σ x. P x).
+```
+
+**Terminology.** In the terminology of Martín Escardó, a type that has
+decidability search on decidable _subtypes_ is referred to as _searchable_,
+_exhaustively searchable_, _exhaustibly searchable_, _exhaustible_,
+_omniscient_, _satisfying the principle of omniscience_, _compact_, or
+_Σ-compact_. {{#cite Esc08}} {{#cite TypeTopology}} We reserve the term
+_Σ-decidability searchable_ for types for which there (merely)
```
EgbertRijke, line 86: The phrase "Σ-decidability searchable" grammatically questionable, because it combines a noun "Σ-decidability" with an adjective "searchable" in the wrong order. Please find better terminology.

src/foundation/decidability-search-universal-quantification.lagda.md
fredrik-bakke: Right, yes. This is new terminology

src/foundation/decidability-search-untruncated-existential-quantification.lagda.md
```diff
@@ -0,0 +1,672 @@
...
+
+## Definitions
+
+### The predicate of having Σ-decidability search
+
+```agda
+has-Σ-decidability-search-Level :
+  {l1 : Level} (l2 : Level) → UU l1 → UU (l1 ⊔ lsuc l2)
+has-Σ-decidability-search-Level l2 X =
+  (P : decidable-family l2 X) → is-decidable (Σ X (family-decidable-family P))
```
EgbertRijke, line 98: If I read this concept correctly, the scheme here is that for families of a certain kind, the total space is again of that kind. (This concept could also be interesting for localizations: local types such that the domain of every map with local fibers into it has a local domain.) So wouldn't it be better for the name to have that kind be postfixed to the name, rather than prefixed as an adjective?

For example, `has-pair-search-decidable-family` would seem more fitting with our naming scheme, or `has-pairing-search-decidable-family`. I'd read this name as follows: There is a function which, given a decidable family searches for a pair or reports that there is no such pair.

Another idea would be just to name this concept "has-Σ-decidability" without postfixing the word "search", because the concept is just about decidability of Σ-types.

fredrik-bakke: Indeed! I suppose for instance projectivity of a type is exactly "∀-inhabitedness"

fredrik-bakke: I tried implementing the terminology "having decidable Σ-types" in the most recent commit

fredrik-bakke:
Martín just explained to me that the "types with double negation dense equality have decidable sigmas" result is a special case of "types are compact iff their totally separated reflections are compact", so that should be remarked in the appropriate places before merging.

src/foundation/weakly-constant-maps.lagda.md
```diff
@@ -125,7 +125,7 @@ module _
     ( fixed-point f , is-prop-fixed-point-is-weakly-constant-map)
 ```
 
-### The action on identifications of a weakly constant map is weakly constant
+### The action on identifications of weakly constant maps are weakly constant
```
VojtechStep:
```suggestion
### The action on identifications of weakly constant maps is weakly constant
```

fredrik-bakke:
I'm going to do a self-review on this one in line with the discussed guidelines for self-review and then merge it so I can move on with my life.

VojtechStep (commented):
I got through about a third of the files, and so far this is the only thing I found, so I'll leave the rest to your self-review

src/foundation/weakly-constant-maps.lagda.md
fredrik-bakke: Thank you! I already found this one myself in my self-review :)

PR #1265, opened 2025-02-04
Inclusion of `ℤ` in `ℚ` preserves addition

lowasser:


src/elementary-number-theory/addition-rational-numbers.lagda.md
```diff
@@ -182,6 +184,32 @@ abstract
           ( sim-reduced-fraction-ℤ y)))
 ```
 
+### The inclusion of integers preserves addition
+
+```agda
+abstract
+  add-rational-ℤ :
```
EgbertRijke, line 191: I'd call this `preserves-add-rational-ℤ`, following a more general naming scheme.

lowasser: (The naming I used in this file is also consistent with what's used everywhere else I could find, e.g. here integer addition on embedded naturals [here](https://unimath.github.io/agda-unimath/elementary-number-theory.addition-integers.html#the-inclusion-of-%E2%84%95-into-%E2%84%A4-preserves-addition))

EgbertRijke: Thanks for pointing that out! Then I will merge this pull request.

I might make the naming more consistent across the library in some future pull request.

PR #1269, opened 2025-02-04
For `p : ℚ⁺`, there is `r : ℚ⁺` with `r + r < p`

lowasser:


src/elementary-number-theory/positive-rational-numbers.lagda.md
```diff
@@ -611,6 +611,34 @@ module _
   le-right-min-ℚ⁺ = pr2 (pr2 strict-min-law-ℚ⁺)
 ```
 
+### Any positive rational number `p` has a `q` with `q + q < p`
```
EgbertRijke, line 614: I'd be surprised if this is the most useful form of this kind of proposition. Something that looks more generally applicable, because it doesn't depend on the exact form of the rational number that satisfies a strict bound, is simply the for any positive rational number `p`, there is a positive rational number strictly smaller than it. In particular, then you get a rational number strictly smaller than half of `p`.

I suppose you have a purpose for the lemma in this form. Could you say a bit more about what you're planning to do with it, and why you chose this form?

lowasser: A cut on the rationals `L, U` is [arithmetically located](https://github.com/UniMath/agda-unimath/blob/master/src/real-numbers/arithmetically-located-cuts.lagda.md#L57) if for any positive rational `epsilon`, there are `L p, U q` with `q - p < epsilon`.

In the course of proving all Dedekind real numbers are arithmetically located, there is a step that takes any positive `epsilon` and shows that there are `L p, U q` with `q - p` equal to `epsilon + epsilon`.  (I am still cleaning it up, so it's not ready for a PR, but that lemma in my current drafts is [here](https://github.com/lowasser/agda-unimath/blob/ccc8e35906fd14c1398287dce0775fb855259544/src/real-numbers/arithmetically-located-cuts.lagda.md#L138).)

To get from that result to showing all reals are arithmetically located, you naturally need to find an `epsilon'` such that `epsilon' + epsilon' < epsilon`.

EgbertRijke: Ok, thank you so much! Did you want to use existential quantification here, or does it need to be a Sigma-type?

lowasser: > In particular, then you get a rational number strictly smaller than half of p.

Currently I'm not aware of any convenient way to express "half of p" -- reciprocals seem pretty painful generally right now.  That's really the hang-up and why I took this approach.  That might be something to improve on down the line, but getting to a reciprocal of `2` as a rational is pretty difficult right now.

lowasser: >  for any positive rational number p, there is a positive rational number strictly smaller than it.

This result is already present [here](https://unimath.github.io/agda-unimath/elementary-number-theory.positive-rational-numbers.html#the-positive-mediant-between-zero-and-a-positive-rational-number).

lowasser: It's on my list, somewhere!

lowasser: > Did you want to use existential quantification here

I'm pretty happy to, let me do that.

lowasser: I have spent the last hour and a half trying to get the existential quantification variant to compile, at least in a plausible amount of time, and I am still failing.  Given that the prior function strict-min is written with a dependent pair, and the discussion [here](https://github.com/UniMath/agda-unimath/pull/1261#discussion_r1938559289), I'm inclined to give up and leave it as-is.

lowasser: I finally made it work reasonably by just having the dependent pair version in a `where`.

PR #1270, opened 2025-02-04
Reassociated conjugation is the identity

lowasser:
There is already a result showing conjugation in an Abelian group is the identity, but I find myself repeatedly needing a differently associated version: `x(yx⁻¹)`, not `(xy)x⁻¹`.

### DIFF
--- a/src/group-theory/abelian-groups.lagda.md
+++ b/src/group-theory/abelian-groups.lagda.md
@@ -601,6 +601,11 @@ module _
   is-identity-conjugation-Ab x y =
     ( ap (add-Ab' A (neg-Ab A x)) (commutative-add-Ab A x y)) ∙
     ( is-retraction-right-subtraction-Ab A x y)
+
+  is-identity-conjugation-assoc-Ab' :
+    (x : type-Ab A) → (add-Ab A x ∘ add-Ab' A (neg-Ab A x)) ~ id
+  is-identity-conjugation-assoc-Ab' x y =
+    inv (associative-add-Ab A x y (neg-Ab A x)) ∙ is-identity-conjugation-Ab x y
 ```
 
 ### Laws for conjugation and addition


src/group-theory/abelian-groups.lagda.md
```diff
@@ -601,6 +601,11 @@ module _
   is-identity-conjugation-Ab x y =
     ( ap (add-Ab' A (neg-Ab A x)) (commutative-add-Ab A x y)) ∙
     ( is-retraction-right-subtraction-Ab A x y)
+
+  is-identity-conjugation-assoc-Ab :
+    (x : type-Ab A) → (add-Ab A x ∘ add-Ab' A (neg-Ab A x)) ~ id
+  is-identity-conjugation-assoc-Ab x y =
+    inv (associative-add-Ab A x y (neg-Ab A x)) ∙ is-identity-conjugation-Ab x y
```
EgbertRijke, line 608: This name, featuring `assoc` right in the middle, doesn't really follow the agda-unimath naming scheme.

Probably the best way to set this up and maintain a consistent naming scheme is to introduce left and right associated conjugation operations, and declare the left associated conjugation operation to be the standard one (the way it is now). This would give operations

`left-conjugation-Group` for the left associated version
`right-conjugation-Group` for the right associated version
`conjugation-Group = left-conjugation-Group` for the standard version

and then we can have

`is-identity-left-conjugation-Ab`
`is-identity-right-conjugation-Ab`
`is-identity-conjugation-Ab = is-identity-left-conjugation-Ab`.

This maintains a consistent naming scheme. Would you be willing to do this? It would be better if we merge a PR with these small couple of changes, than introducing some entries with names that aren't quite right and hoping that someone will eventually refactor it down the line. Refactoring is a lot of work.

lowasser: Done.

PR #1272, opened 2025-02-05
Strict inequality on the reals

lowasser:


src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -0,0 +1,149 @@
...
+between them. This is the definition used in {{#cite UF13}}, section 11.2.1.
+
+```agda
+module _
+  {l1 l2 : Level}
+  (x : ℝ l1)
+  (y : ℝ l2)
+  where
+
+  le-ℝ-Prop : Prop (l1 ⊔ l2)
```
VojtechStep, line 45: Strict inequality should probably be called `lt-R` rather than `le-R`, unless I'm missing something

EgbertRijke: We've been using `le` as a signifier for strict inequality throughout the library. I don't think `lt` is in use anywhere. So I don't think the name needs to change in this pull request.

lowasser: What Egbert said: `le` is used everywhere, I kept using it here.

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -0,0 +1,149 @@
...
+  (x : ℝ l1)
+  (y : ℝ l2)
+  (z : ℝ l3)
+  where
+
+  transitive-le-ℝ : le-ℝ y z → le-ℝ x y → le-ℝ x z
+  transitive-le-ℝ y<z =
+    elim-exists
+      (le-ℝ-Prop x z)
+      (λ p (p-in-ux , p-in-ly) →
```
EgbertRijke, line 121:
```suggestion
    elim-exists
      ( le-ℝ-Prop x z)
      ( λ p (p-in-ux , p-in-ly) →
```

src/real-numbers/strict-inequality-real-numbers.lagda.md
VojtechStep: Sorry, my bad

PR #1276, opened 2025-02-05
Negation reverses strict inequality on `ℝ`

lowasser:


lowasser:
> Ay, it looks like some changes here conflicted with some changes in your other pull request. Would you mind fixing that?

Done.

PR #1277, opened 2025-02-05
Raise subtypes to a universe level

lowasser:


lowasser:
For reference, this is a step towards raising the universe levels of real numbers, to get a zero that can be the additive identity for all levels of reals.

PR #1211, opened 2024-10-25
Refactor elementary number theory

EgbertRijke:
The biggest changes come from a change in the definition of the quotient of a divisible natural number. The updated definition defines the quotient of `n` by `d`, provided that `d | n` is the unique natural number `q ≤ n` such that `q * d = n`. This change prevents some case analyses around the situation where the Curry-Howard interpretation of divisibility is not a proposition. An alternative definition of divisibility is also given: bounded divisibility. This is always a proposition. The poset of the natural numbers by divisibility is updated to use bounded divisibility in favor of propositional truncation.

The change in the definition of the quotient by divisibility triggered changes in the fundamental theorem of arithmetic, which is thoroughly revised after discovering that the strong induction that we previously used was unnecessarily complicated.

New files include:
- Bounded divisibility
- Irrationality of the square root of 2
- Lists of prime numbers
- Multiplicative decompositions of natural numbers
- Nicomachus's Theorem
- Nontrivial divisors of natural numbers
- Number of divisors
- Parity of integers
- Pronic numbers
- Square pyramidal numbers
- Unit integers
- Unit similarity integers
- Well-ordering principles of the integers
- Dependent lists
- Elementhood relation on lists
- Universal quantification on lists

In `commutative-algebra` and `ring-theory` we developed the theory of ideals.

We have defined the following variations of ideals:
- left ideals of semirings
- right ideals of semirings
- ideals of semirings
- left subtractive ideals of semirings
- right subtractive ideals of semirings
- subtractive ideals of semirings
- left ideals of rings
- right ideals of rings
- ideals of rings
- ideals of commutative semirings
- subtractive ideals of commutative semirings
- ideals of commutative rings

Each of these 12 variations is developed according to the following set of modules
- full ideals
- ideals
- ideals generated by elements
- ideals generated by families of elements
- ideals generated by sub(semi)rings
- ideals generated by subsets
- identity kernels
- image ideals (contains the image-pullback Galois connection)
- intersections of ideals (binary)
- joins of families of ideals
- maximal ideals
- nil ideals
- poset of ideals
- principal ideals
- products of ideals
- pullback ideals
- trivial ideals

This accounts for 12x17 modules. In addition, we defined subtractive closure, which is another Galois connection. All of this was in order to have a strong enough development of ideals of commutative semirings, which is used in the refactoring of elementary number theory.

PR #1279, opened 2025-02-05
Decide for natural numbers whether `x < y` or `y ≤ x`

lowasser:
The rationals have it, but the naturals don't.


EgbertRijke:
Some of these things are also addressed in #1211, but your PR is certainly very welcome!

EgbertRijke:
To be a bit more specific, #1211 doesn't touch on the rational numbers, but it vastly expands everything about the naturals. The revisions I am making there have triggered some wider changes across the library, which is why it hasn't converged to a finished PR yet (I admit, that this is bad style on my part).

PR #1280, opened 2025-02-05
Make naturals, integers, integer fractions, and rationals consistent about some inequality theorems

lowasser:
This is an attempt to standardize on the naming scheme established by [`decide-le-leq-ℚ`](https://unimath.github.io/agda-unimath/elementary-number-theory.strict-inequality-rational-numbers.html#11197) and [`not-leq-le-ℚ`](https://unimath.github.io/agda-unimath/elementary-number-theory.strict-inequality-rational-numbers.html#11534) across elementary number types.


EgbertRijke:
Hm... this is the kind of pull request that is probably going to be problematic with #1211, because I'm already doing a lot of the standardization work there.

lowasser:
Would skipping the parts of this PR that rename existing things help?  I would certainly recommend breaking #1211 into more incremental changes like this one.

EgbertRijke:
I think I'd rather improve the naming in #1211 than here in this pull request, because I can make more comprehensive naming changes there that align everything in a coherent way.

PR #1282, opened 2025-02-05
For `p q : ℚ`, `succ-ℚ p * q = q + (p * q)`

lowasser:
Factoring out a lemma from #1273 as threatened.

### DIFF
--- a/src/elementary-number-theory/multiplication-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/multiplication-rational-numbers.lagda.md
@@ -348,22 +348,18 @@ abstract
           ( sim-reduced-fraction-ℤ y)))
 ```
 
-### `succ-ℤ x * p = p + (x * p)`
+### `succ-ℚ p * q = q + (p * q)`
 
 ```agda
-left-mul-succ-rational-ℤ :
-  (x : ℤ) →
-  (p : ℚ) →
-  rational-ℤ (succ-ℤ x) *ℚ p ＝ p +ℚ (rational-ℤ x *ℚ p)
-left-mul-succ-rational-ℤ x p =
+left-mul-succ-ℚ :
+  (p q : ℚ) →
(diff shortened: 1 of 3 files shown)


src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -348,6 +348,24 @@ abstract
           ( sim-reduced-fraction-ℤ y)))
 ```
 
+### `succ-ℤ x * p = p + (x * p)`
+
+```agda
+left-mul-succ-rational-ℤ :
```
EgbertRijke, line 354: For consistency with similar laws of the natural numbers and the integers, it would be better to implement `left-successor-law-mul-ℚ` which should apply to all rational numbers, and a law `compute-rational-succ-ℤ` that establishes that `rational-ℤ (succ-ℤ x) = succ-ℚ (rational-ℤ x)`. Then this function is just a simple application of the previous two, which look more generally useful to me.

When factoring out little lemmas like this one, we still keep in mind their general applicability. To improve the applicability of a lemma, it is best that the elements to appear in general form, if possible. Your lemma here looks somehow too situational, because it involves a term of the form `rational-ℤ (succ-ℤ x)` which is very specific, and having the combination of a general successor law for multiplication of rational numbers and a law that the inclusion of the integers into the rationals preserve successor looks much more useful.

lowasser: I mostly did it this way because there isn't yet a `succ-ℚ`.  I can add that too.

lowasser: Pulled that out into #1283, and refactored this to be in terms of that.

lowasser:
Closing this until #1283 is merged.

lowasser:
Reopened now!

fredrik-bakke (approved):

PR #1283, opened 2025-02-05
Successor and predecessor for `ℚ`

lowasser:


EgbertRijke:
Awesome, this is a much more solid approach!

PR #1285, opened 2025-02-06
Truncated law of excluded middle

lowasser:
I found myself wanting the slightly weaker version described [here](https://ncatlab.org/nlab/show/excluded+middle#truncated_excluded_middle).


EgbertRijke:
This is not weaker, it is equivalent to regular LEM

PR #1286, opened 2025-02-06
Raising a real number's universe level

lowasser:
There's certainly still more to prove about the equivalences between a real and its raised version, but defining it is a start.


src/real-numbers/raising-real-numbers.lagda.md
fredrik-bakke, line 1: The module name should probably be disambiguated further. My first expectation was that this file would be about raising real numbers to powers. I suggest `raising-universe-levels-dedekind-real-numbers`. (as an aside, I don't know why we don't disambiguate all of the files with `dedekind-real-numbers`, we could equally have files on cauchy reals or the likes, and that definitely shouldn't go in the same files.

lowasser: As far as `dedekind-real-numbers` goes, I was taking my cue from the line in that file: "The Dedekind real numbers will be taken as the standard definition of the real numbers in the agda-unimath library."  Based on that sentence, I assumed that if we want results on non-Dedekind reals, they have to disambiguate themselves, but Dedekind reals get to be assumed as default.

lowasser: (the line here: https://unimath.github.io/agda-unimath/real-numbers.dedekind-real-numbers.html#idea)

fredrik-bakke: Ah, yes, I see where you get it from, that makes sense. You can forget my suggestion to disambiguate for Dedekind reals then.

EgbertRijke: I actually agree with Fredrik, that it would probably be better if we disambiguated the file names to specify that they are about Dedekind real numbers. That's for another pull request though.

I also like the suggestion `raising-universe-levels-dedekind-real-numbers` (whether to include `dedekind` in this name I leave up to you, but `raising-universe-levels` reads nicer than `raising`.

lowasser: I left out `dedekind` but have changed the name.

src/real-numbers/raising-universe-levels-real-numbers.lagda.md
```diff
@@ -0,0 +1,129 @@
...
+
+open import real-numbers.dedekind-real-numbers
+```
+
+</details>
+
+## Idea
+
+Real numbers have designated universe levels. For any real number, we can
+construct an equivalent real number in any higher universe.
```
EgbertRijke, line 33: Some hyperlinks are needed in this section. Could you also say in a sentence what it means for two real numbers of different universe levels to be equivalent?

```suggestion
[Real numbers](real-numbers.dedekind-real-numbers.md) have designated [universe levels](foundation.universe-levels.md). For any real number, we can
construct an equivalent real number in any higher universe. This means that ...
```

lowasser: I agree.  A step towards this is in #1295.

lowasser:
Closing this until we have settled on similarity concepts for reals.

lowasser:
This will get redone after the upper/lower Dedekind split, but I'm resurrecting this as a draft to remind myself it'll eventually be relevant again.

src/real-numbers/raising-universe-levels-real-numbers.lagda.md
```diff
@@ -0,0 +1,179 @@
...
+</details>
+
+## Idea
+
+[Real numbers](real-numbers.dedekind-real-numbers.md) are defined at specific
+[universe levels](foundation.universe-levels.md), which do not overlap.
+
+For any real number, we can construct a
+[similar](real-numbers.similarity-real-numbers.md) real number in any higher
+universe.
```
fredrik-bakke, line 45: It's unclear to me what you mean by the first sentence here. Is the point that we have a hierarchy of real numbers indexed by universe levels, so a "small real number" is not syntactically the same as a "large real number" relative to some universe?

```suggestion
For every [universe](foundation.universe-levels.md) `𝒰` there is a type of
[real numbers](real-numbers.dedekind-real-numbers.md) `ℝ` relative to `𝒰`,
`ℝ 𝒰`. Given a larger universe `𝒱`, then we may
{{#concept "raise" Disambiguation="a dedekind real number" Agda=raise-ℝ}} a real
number `x` from the universe `𝒰` to a
[similar](real-numbers.similarity-real-numbers.md) real number in the universe
`𝒱`.
```

fredrik-bakke (approved):

PR #1290, opened 2025-02-07
Minimum operation on rational numbers

lowasser:


EgbertRijke:
Your methods here are immediately generalizable to [decidable total preorders](https://unimath.github.io/agda-unimath/order-theory.decidable-total-preorders.html). In terms of library development, it would be great if you could do so. It would unlock your work in other settings as well, and it would make it more obvious for people to recognize that something more general is used here.

Would you be willing to:
- Create a `## Properties` section on the Decidable total preorders page, with a subsection `### Any two elements in a decidable total preorder have a minimum`
- Similarly create a `## Properties` section on the Decidable total orders page, with a similar subsection and its entry invoking the entry from the Decidable total preorders page?
- Define the minimum of two rational numbers in terms of the minimum of two elements of a decidable total order?

src/elementary-number-theory/minimum-rational-numbers.lagda.md
```diff
@@ -0,0 +1,67 @@
...
+open import foundation.coproduct-types
+open import foundation.function-types
+```
+
+</details>
+
+## Idea
+
+We define the operation of minimum (greatest lower bound) for the rational
+numbers.
```
EgbertRijke, line 23: Saying "We define X" doesn't tell a reader much about what is being defined. It would be better to say

```suggestion
The {{#concept "minimum" Disambiguation="rational numbers" Agda=min-ℚ}} of two rational numbers `p` and `q` is a rational number `r` such that the [logical equivalence](foundation.logical-equivalences.md)

```text
(x ≤ p) ∧ (x ≤ q) ↔ (x ≤ r)
```

holds for every rational number `x`. In other words, the minimum of two rational numbers `p` and `q` is the [greatest binary lower bound](order-theory.greatest-lower-bounds-posets.md) of `p` and `q` in the [linearly ordered set of rational numbers](elementary-number-theory.inequality-natural-numbers.md).
```

This is much more informative, and links the concept with the existing concept of greatest binary lower bound in a poset.

Issue #1235, opened 2025-01-13
Inline code should scale with font size


fredrik-bakke:
This issue is particularly noticeable in the main module headers: 
<img width="355" alt="image" src="https://github.com/user-attachments/assets/fa7dee47-fc4f-48e0-a9b6-3723d20276eb" />

VojtechStep:
I'll just comment on this for now, if anyone wants to play with this — I unknowingly introduced this in #1154, where I changed the code size from `0.875em` to `1.3rem`, to fix #1053. The `em` units are relative to the current element font size, while the `rem` units are relative to the root element size. This fixed the size difference between Agda code blocks and other code blocks, but as a result all code, inline or not, has the same size.

I don't know if we noticed this before, but the reason why non-Agda code blocks were smaller is because they are in a `<code>` element nested in a `<pre>` element, so the scaling down used to be applied twice.

Note that Agda code blocks are just `<pre>` elements, and inline code blocks are just `<code>` elements, so we can't very well just remove the rule from one or the other.

VojtechStep:
I'm pretty sure, before that PR the code blocks were relative to the element font size. It's possible that they were still noticeably smaller than the surrounding text in headers, but they were still bigger than normal text.

VojtechStep:
I know that there were size issues with using the Unicode infinity vs LaTeX infinity, but I think that's because the Unicode one is just small.

fredrik-bakke:
> I'm pretty sure, before that PR the code blocks were relative to the element font size. It's possible that they were still noticeably smaller than the surrounding text in headers, but they were still bigger than normal text.

Right, this might be it.

PR #1292, opened 2025-02-07
Rename min and max operations for decidable total orders

lowasser:
I left out the word `Decidable`.  Oops.


fredrik-bakke (approved):
I'm not a big fan of the use of escaped code in headers, even if we didn't have a font issue with them (#1235). It is better to use words, and only use escaped code if it is necessary to refer to a specific symbol. However, seeing as none of the headers were introduced in this PR I say it's good to merge.

lowasser:
I'll follow up on that.

PR #1293, opened 2025-02-07
For `x y : ℚ`, `x < y` if and only if `real-ℚ x < real-ℚ y`

lowasser:
Adding the reverse implication.


src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -142,11 +142,25 @@ module _
 ### The canonical map from rationals to reals preserves strict inequality
```
fredrik-bakke, line 142:
```suggestion
### The canonical map from rationals to reals preserves and reflects strict inequality
```

PR #1294, opened 2025-02-07
If `q : ℚ` is in the lower cut of a real, `real-ℚ q` is less than that real, and analogously for the upper cut

lowasser:
And the analogous statement for upper cuts.

### DIFF
--- a/src/real-numbers/strict-inequality-real-numbers.lagda.md
+++ b/src/real-numbers/strict-inequality-real-numbers.lagda.md
@@ -235,13 +235,11 @@ module _
             tr (is-in-upper-cut-ℝ x) (inv (neg-neg-ℚ p)) p-in-ux))
 ```
 
-### If a rational `q` is in the lower cut of real `x`, `real-ℚ q < x`
+### If a rational is in the lower cut of `x`, then it is strictly less than `x`
 
 ```agda
 module _
-  {l : Level}
-  (q : ℚ)
-  (x : ℝ l)
+  {l : Level} (q : ℚ) (x : ℝ l)
   where
 
   le-lower-cut-real-ℚ : is-in-lower-cut-ℝ x q → le-ℝ (real-ℚ q) x
--- a/src/real-numbers/strict-inequality-real-numbers.lagda.md
+++ b/src/real-numbers/strict-inequality-real-numbers.lagda.md
@@ -235,6 +235,27 @@ module _
             tr (is-in-upper-cut-ℝ x) (inv (neg-neg-ℚ p)) p-in-ux))
 ```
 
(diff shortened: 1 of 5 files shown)


fredrik-bakke (commented):
It is good practice to also include the dual statement in the same PR, as it seems just as likely that that lemma would be useful. Would you mind adding that to this PR? I.e., if a rational is in the upper cut of `x`, then it is strictly greater than `x`.

lowasser:
> It is good practice to also include the dual statement in the same PR, as it seems just as likely that that lemma would be useful. Would you mind adding that to this PR? I.e., if a rational is in the upper cut of x, then it is strictly greater than x.

That statement *is* already there on line 251, I just didn't include it in the name of the PR because it felt like it'd make it too long.

There is a dual statement that it's an if-and-only-if, though, and I can add that.

lowasser:
Actually, I'm going to hold off on that one for a follow-up.  The dual you want is already there, though.

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -235,6 +235,25 @@ module _
...
+  le-lower-cut-real-ℚ : is-in-lower-cut-ℝ x q → le-ℝ (real-ℚ q) x
+  le-lower-cut-real-ℚ =
+    forward-implication (is-rounded-lower-cut-ℝ x q)
+
+  le-upper-cut-real-ℚ : is-in-upper-cut-ℝ x q → le-ℝ x (real-ℚ q)
+  le-upper-cut-real-ℚ H =
+    elim-exists
+      ( le-ℝ-Prop x (real-ℚ q))
+      ( λ p (p<q , p∈ux) → intro-exists p (p∈ux , p<q))
+      ( forward-implication (is-rounded-upper-cut-ℝ x q) H)
```
fredrik-bakke, line 254: My bad for not noticing – let's add the appropriate header then. As an aside, although this PR is good to merge as-is after this final change, the below proof would be a one-liner with `map-tot-exists` from #1298.

````suggestion
```

### If a rational is in the upper cut of `x`, then it is strictly greater than `x`

```agda
  le-upper-cut-real-ℚ : is-in-upper-cut-ℝ x q → le-ℝ x (real-ℚ q)
  le-upper-cut-real-ℚ H =
    elim-exists
      ( le-ℝ-Prop x (real-ℚ q))
      ( λ p (p<q , p∈ux) → intro-exists p (p∈ux , p<q))
      ( forward-implication (is-rounded-upper-cut-ℝ x q) H)
````

lowasser: I'm looking forward to it.

PR #1295, opened 2025-02-07
Similarity of subtypes is reflexive, transitive, and antisymmetric at the same universe level

lowasser:
Maybe there's a name for this sort of thing?  "Large equivalence relation" or something?

### DIFF
--- a/src/foundation/subtypes.lagda.md
+++ b/src/foundation/subtypes.lagda.md
@@ -137,65 +137,6 @@ module _
     eq-has-same-elements-subtype P Q (λ x → (H x , K x))
 ```
 
-### Similarity of subtypes
-
-```agda
-module _
-  {l1 : Level} {A : UU l1}
-  where
-
-  sim-subtype :
-    {l2 l3 : Level} → subtype l2 A → subtype l3 A → UU (l1 ⊔ l2 ⊔ l3)
-  sim-subtype P Q = (P ⊆ Q) × (Q ⊆ P)
-
-  has-same-elements-sim-subtype :
-    {l2 l3 : Level} (P : subtype l2 A) (Q : subtype l3 A) →
-    sim-subtype P Q → has-same-elements-subtype P Q
-  pr1 (has-same-elements-sim-subtype P Q s x) = pr1 s x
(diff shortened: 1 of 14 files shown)


src/foundation/subtypes.lagda.md
```diff
@@ -138,27 +161,39 @@ module _
   pr2 (sim-has-same-elements-subtype P Q s) x = backward-implication (s x)
 ```
 
-### The containment relation is antisymmetric
```
lowasser, line 141: I did some rearranging to keep symmetry and its properties together, but moved it after this section because I need antisymmetry.

fredrik-bakke:
> Maybe there's a name for this sort of thing? "Large equivalence relation" or something?

That certainly sounds like a good concept. I don't believe we have this formalized, but it is a welcome addition. We already have large binary relations, this should build neatly on top of that.

fredrik-bakke:
Wait, actually, we have similarity in large posets. That's what you want

lowasser:
Right.  And we can't necessarily embed that in this file because of circular dependencies, but maybe [this](https://unimath.github.io/agda-unimath/foundation.large-locale-of-subtypes.html#1471) is already the large poset we're looking for?

I still think this is worth merging just to have it available right there, though, specialized for subsets.

fredrik-bakke:
Hmmm, I think this fits better on the [page on powersets](https://unimath.github.io/agda-unimath/foundation.powersets.html), where the large poset is also formalized.

fredrik-bakke:
Does that cause circular dependencies in your work?

lowasser:
Would you like to move the entire notion of similarity there?  Or leave it defined where it is, and define its properties on the power set page?  It feels a little weird to have it defined separately from the useful properties.

fredrik-bakke:
> Would you like to move the entire notion of similarity there? Or leave it defined where it is, and define its properties on the power set page? It feels a little weird to have it defined separately from the useful properties.

It seems reasonable to me to move everything to the page on powersets 👍

lowasser:
Done.

fredrik-bakke (approved):

PR #1299, opened 2025-02-08
Avoid escaped code in headers for min and max

lowasser:
As discussed in https://github.com/UniMath/agda-unimath/pull/1292#pullrequestreview-2603875335 .

### DIFF
--- a/src/order-theory/decidable-total-orders.lagda.md
+++ b/src/order-theory/decidable-total-orders.lagda.md
@@ -255,7 +255,7 @@ module _
   ... | inr y<x = pr2 y<x
 ```
 
-### `min` is commutative
+### The minimum operation is commutative
 
 ```agda
 module _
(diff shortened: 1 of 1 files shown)


src/order-theory/decidable-total-orders.lagda.md
```diff
@@ -222,7 +222,7 @@ module _
   ... | inr y<x = pr2 y<x
 ```
 
-### `min x y ≤ y`
+### The minimum of two values is less than or equal to the second
 
 ```agda
```
fredrik-bakke, line 227:
```suggestion

```

src/order-theory/decidable-total-orders.lagda.md
```diff
@@ -211,7 +211,7 @@ module _
       ( is-leq-or-strict-greater-Decidable-Total-Order T x y)
 ```
 
-### `min x y ≤ x`
+### The minimum of two values is less than or equal to the first
```
fredrik-bakke, line 214:
```suggestion
### The minimum is a lower bound
```

fredrik-bakke (commented):
How about the following less wordy set-up? :) (and similar for maximum)

PR #1301, opened 2025-02-08
Similarity of real numbers

lowasser:
Nothing very fancy at all here.

### DIFF
--- a/src/real-numbers/similarity-real-numbers.lagda.md
+++ b/src/real-numbers/similarity-real-numbers.lagda.md
@@ -96,9 +96,9 @@ transitive-sim-ℝ :
 transitive-sim-ℝ = transitive-sim-Large-Poset ℝ-Large-Poset
 ```
 
-### Antisymmetry at the same universe level
+### Similar real numbers in the same universe are equal
 
 ```agda
-antisymmetric-sim-ℝ : {l : Level} → (x y : ℝ l) → sim-ℝ x y → x ＝ y
-antisymmetric-sim-ℝ = eq-sim-Large-Poset ℝ-Large-Poset
+eq-sim-ℝ : {l : Level} → (x y : ℝ l) → sim-ℝ x y → x ＝ y
+eq-sim-ℝ = eq-sim-Large-Poset ℝ-Large-Poset
 ```
--- a/src/real-numbers/similarity-real-numbers.lagda.md
+++ b/src/real-numbers/similarity-real-numbers.lagda.md
@@ -1,0 +1,99 @@
+# Similarity on the real numbers
+
+```agda
+module real-numbers.similarity-real-numbers where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.dependent-pair-types
+open import foundation.identity-types
+open import foundation.logical-equivalences
+open import foundation.propositions
+open import foundation.subtypes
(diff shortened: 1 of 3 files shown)


src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -0,0 +1,99 @@
...
+open import real-numbers.inequality-real-numbers
+```
+
+</details>
+
+## Definition
+
+Two [Dedekind real numbers](real-numbers.dedekind-real-numbers.md) are similar
+if they are less than or equal to each other. This is like equality, but two
+reals of different universe levels can be similar.
```
fredrik-bakke, line 30: This text should be moved to an `## Idea` section above

src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -0,0 +1,99 @@
+# Similarity on the real numbers
```
fredrik-bakke, line 1: I find this phrasing a bit odd. Isn't the following better?

```suggestion
# Similarity of real numbers
```

src/real-numbers/similarity-real-numbers.lagda.md
fredrik-bakke, line 30:
```suggestion
## Idea

Two [Dedekind real numbers](real-numbers.dedekind-real-numbers.md) are {{#concept "similar" Disambiguation="Dedekind real numbers" Agda=sim-ℝ}}
if they are [less than or equal](real-numbers.inequality-real-numbers.md) to each other. The similarity relation on real numbers extends the [equality](foundation-core.identity-types.md) relation to real numbers of differing universe levels.

## Definition
```

src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -0,0 +1,104 @@
...
+  sim-ℝ x y →
+  sim-ℝ x z
+transitive-sim-ℝ = transitive-sim-Large-Poset ℝ-Large-Poset
+```
+
+### Antisymmetry at the same universe level
+
+```agda
+antisymmetric-sim-ℝ : {l : Level} → (x y : ℝ l) → sim-ℝ x y → x ＝ y
+antisymmetric-sim-ℝ = eq-sim-Large-Poset ℝ-Large-Poset
```
fredrik-bakke, line 103: This theorem doesn't state that similarity is antisymmetric, it states that similar elements are equal.

```suggestion
eq-sim-ℝ : {l : Level} → (x y : ℝ l) → sim-ℝ x y → x ＝ y
eq-sim-ℝ = eq-sim-Large-Poset ℝ-Large-Poset
```

src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -0,0 +1,104 @@
...
+  (x : ℝ l1) →
+  (y : ℝ l2) →
+  (z : ℝ l3) →
+  sim-ℝ y z →
+  sim-ℝ x y →
+  sim-ℝ x z
+transitive-sim-ℝ = transitive-sim-Large-Poset ℝ-Large-Poset
+```
+
+### Antisymmetry at the same universe level
```
fredrik-bakke, line 99:
```suggestion
### Similar real numbers in the same universe are equal
```

PR #1302, opened 2025-02-08
`q : ℚ` is in the lower cut of a real only if real-ℚ q is less than that real, and analogously for the upper cut

lowasser:
The other direction of the implication we already had.


fredrik-bakke (approved):
Very nice!

PR #1303, opened 2025-02-08
For `x y : ℚ`, `x ≤ y` if and only if `real-ℚ x ≤ real-ℚ y`

lowasser:
We have the analogue for strict inequality; this one's for nonstrict.


lowasser:
Well, we don't actually have it for strict inequality yet, that's #1293.

fredrik-bakke (approved):

PR #1305, opened 2025-02-09
Rational reals are similar to the projection of their rational

lowasser:


src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -102,3 +102,25 @@ transitive-sim-ℝ = transitive-sim-Large-Poset ℝ-Large-Poset
...
 eq-sim-ℝ = eq-sim-Large-Poset ℝ-Large-Poset
 ```
+
+### A rational real is similar to the canonical projection of its rational
+
+```agda
+sim-rational-ℝ :
+  {l : Level} →
+  (x : Rational-ℝ l) →
+  sim-ℝ (real-rational-ℝ x) (real-ℚ (rational-rational-ℝ x))
```
fredrik-bakke, line 112: Do the two reals `real-rational-ℝ x` and `real-ℚ (rational-rational-ℝ x)` live in the same universe? If so, please also record that they are equal.

fredrik-bakke:
Also, for future reference please see my comment [here](https://github.com/UniMath/agda-unimath/pull/1304#issuecomment-2646007436).

src/real-numbers/similarity-real-numbers.lagda.md
lowasser: No, and that's the point; `real-rational-ℝ x` is in universe `l`, and `real-ℚ (rational-rational-ℝ x)` is in `lzero`.

fredrik-bakke: Okay, thanks for the explanation!

PR #1306, opened 2025-02-09
Do syntax for propositional truncations, including application to strict real inequality

lowasser:
Attempting to generalize @VojtechStep's comment [here](https://github.com/UniMath/agda-unimath/pull/1273#issuecomment-2646004000).

I've also used `map-exists` where applicable, but I must admit that despite the use of modules and `where`, which the [code style](https://unimath.github.io/agda-unimath/CODINGSTYLE.html#modules-in-the-agda-unimath-library) argues against, this does seem to improve readability quite a bit.


src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -245,13 +210,15 @@ module _
...
-            tr (is-in-upper-cut-ℝ x) (inv (neg-neg-ℚ p)) p-in-ux))
+    map-exists
+      ( λ p →
+        is-in-subtype (upper-cut-neg-ℝ y) p ×
+        is-in-subtype (lower-cut-neg-ℝ x) p)
+      ( neg-ℚ)
+      ( λ p (x<p , p<y) →
+        tr (is-in-lower-cut-ℝ y) (inv (neg-neg-ℚ p)) p<y ,
+        tr (is-in-upper-cut-ℝ x) (inv (neg-neg-ℚ p)) x<p)
+    where open elim-exists-do (le-ℝ-Prop (neg-ℝ y) (neg-ℝ x))
```
VojtechStep, line 221: The do notation isn't used

src/foundation/existential-quantification.lagda.md
```diff
@@ -331,6 +331,24 @@ module _
     eq-distributive-product-exists-structure P
 ```
 
+## Existential quantification `do` syntax
+
+When eliminating a chain of existential quantifications, which may be
+interdependent, Agda's `do` syntax can eliminate many levels of nesting.
```
VojtechStep, line 337: Could you link to Agda's documentation and include an example? I'm thinking a schematic of a nested proof and its alternative with `do` in a `text` code fence. We don't expect users to know Haskell

VojtechStep (commented):
I'm for this change, but I'll wait for another maintainer to sign off on it. I'm biased because I'm a Haskeller, and I can't say if the new code is more readable to someone not familiar with do notation

fredrik-bakke:
Having syntax modules like the one proposed here is an interesting prospect that has been mulling in the back of my mind for some time, and I would be interested in seeing how we can integrate it well with the agda-unimath design philosophy. (In particular, I almost see syntax modules as a necessity later on for the applications of wild higher category theory, but that's beside the point of the discussion here.) I believe there is a reason that this language feature is not in use currently in the library, but I have not taken part in that discussion before, @EgbertRijke will have to weigh in on that. I expect there may be an issue with how goals display when fancy features like this are used. Moreover, there is a question about readability for people who don't know the ins and outs of Agda. That's another thing that needs to be weighed over before implementing this syntax library-wide.

src/foundation/existential-quantification.lagda.md
```diff
@@ -331,6 +331,24 @@ module _
...
+
+```agda
+module elim-exists-do
+  {l : Level}
+  (P : Prop l)
+  where
+
+  _>>=_ :
```
fredrik-bakke, line 345: I can't see that this symbol, `>>=`, is used anywhere. Does it have some special meaning to the Agda interpreter? If so, is there a reference document somewhere I can read about this?

src/foundation/existential-quantification.lagda.md
```diff
@@ -331,6 +331,24 @@ module _
     eq-distributive-product-exists-structure P
 ```
 
+## Existential quantification `do` syntax
+
+When eliminating a chain of existential quantifications, which may be
+interdependent, Agda's `do` syntax can eliminate many levels of nesting.
+
+```agda
+module elim-exists-do
```
fredrik-bakke, line 340: I believe `do-syntax-elim-exists` is more in line with our naming conventions.

fredrik-bakke:
If we're implementing this syntax for existential quantification, it would be really nice to also have it for propositional truncations and more generally all truncations (later, all localizations).

src/foundation/existential-quantification.lagda.md
VojtechStep: Do notation desugars to calls to whatever `_>>=_` is in scope https://agda.readthedocs.io/en/stable/language/syntactic-sugar.html#do-notation

fredrik-bakke: Ah, then I probably broke a feature in our `reflection` namespace a while back. 😬 Thanks for the link!

lowasser: Added a link to that page.

VojtechStep:
> I expect there may be an issue with how goals display when fancy features like this are used.

I tested it with this code
```agda
data Maybe {l : Level} (A : UU l) : UU l where
  nothing : Maybe A
  just : A → Maybe A

elim-maybe : ∀ {l1 l2} {A : UU l1} {B : UU l2} → (A → Maybe B) → Maybe A → Maybe B
elim-maybe f nothing = nothing
elim-maybe f (just a) = f a

_>>=_ : ∀ {l1 l2} {A : UU l1} {B : UU l2} → Maybe A → (A → Maybe B) → Maybe B
a >>= f = elim-maybe f a

data Test : UU lzero where
  thing : Test

data Bool : UU lzero where true : Bool; false : Bool

not : Bool → Bool
not true = false
not false = true

guard : Bool → Maybe Test
guard true = just thing
guard false = nothing

thing1 : Bool → Maybe Test
thing1 b = do
  _ ← guard b
  _ ← guard (not b)
  just thing

_ : ∀ b → thing1 b ＝ nothing
_ = {! thing1 !}
```
and neither simplifying nor normalizing `thing1` resulted in showing `do` notation (that is both `C-c C-n` to normalize `thing1` in the hole and `C-c C-,` and `C-u C-u C-c C-,` to normalize `thing1` in the goal). Instead I saw `_>>=_` inlined away and got the elimination application
```
Goal: (b : Bool) →
      Id
      (elim-maybe (λ _ → elim-maybe (λ _ → just thing) (guard (not b)))
       (guard b))
      nothing
```

I also tried it with the goal being in another file, and still could get Agda to show me either `do` or `_>>=_`. I'm not that surprised that `do` doesn't show up, since AFAIU that's just surface syntax and there's no reason why the pretty printer should reconstructed, but I'm happy to see `_>>=_` not show up.

As an aside, for the goal-in-another-file test I tried normalizing `asymmetric-le-ℝ`, but gave up after a few minutes. There's _a lot_ of hanging fruit for getting a speed-up by making definitions in `rational-numbers` abstract (e.g. I can reduce the term of `asymmetric-le-ℝ` instantly if I make `eq-ℚ-sim-fraction-ℤ` abstract).

lowasser:
I have gradually been making more and more things abstract for rational numbers, I'll be sure to include that.

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -90,28 +92,20 @@ module _
...
+        ( asymmetric-le-ℚ
+            ( q)
+            ( p)
+            ( le-lower-upper-cut-ℝ x q p q<x x<p))
+        ( not-leq-le-ℚ
+            ( p)
+            ( q)
+            ( le-lower-upper-cut-ℝ y p q p<y y<q))
+        ( decide-le-leq-ℚ p q)
+    where open do-syntax-elim-exists empty-Prop
```
EgbertRijke, line 108: I don't know about this proposal. The code is surely shorter, but I find it very hard to read.

The benefit of having explicit `elim-exists` is that it tells you explicitly what the code is doing, and if you don't know what `elim-exists` is you can just jump to its definition. I feel like some information is lost here, and the code becomes less self-explanatory to an uninitiated person, which I have always felt is extremely important and I don't really want to loose that.

EgbertRijke: We have some `iterate-[operation]`'s in the library. I think an `iterate-elim-exists` would be a reasonable thing to have.

lowasser: I don't see an example of `iterate-` in the library at a glance that looks like it might be analogous.

I do want to claim that understanding this syntax is a constant overhead across the library, whereas understanding many-levels-nested `elim-exists` is pretty painful each time you do it.  My limited progress on multiplication suggests it will be far worse than addition; I would not be shocked by seven or eight nested elim-exists.  I won't have any room for indentation left over, and pulling out lemmas with 10+ arguments doesn't help much either.

As far as "jumping to definitions" is concerned, every use of `do` must be associated with a `where open do-syntax-elim-exists foo`, and `do-syntax-elim-exists` is something you can click on and jump to its definition.

In an attempt to explain a little better what this syntax does, let me have a go at its propositional truncation variant:

```agda
do
  a <- prop-a
  f a
```

is equivalent to `ind-trunc-Prop _ f prop-a`.  (The associated `where` and what goes in the `_` have to indicate what proposition is being satisfied -- though maybe you can get away with the `_`?)  All it does is reverse the argument order for `ind-trunc-Prop` and use a left arrow instead of a lambda with a right arrow.

lowasser: (Possibly relatedly, I find the `elim-exists` harder to read with its current argument order, as standard as it may be.  I would certainly find it easier to follow with its second and third arguments swapped.)

VojtechStep: > if you don't know what elim-exists is you can just jump to its definition

To add to Louis' answer (that you can go-to-definition on the `open do-syntax-elim-exists`), you can also go-to-definition on the `←`, both in an editor and on the website, which takes you to the appropriate definition of `_>>=_`.

> the code becomes less self-explanatory to an uninitiated person

There is always a trade-off between making the code understandable to someone not familiar with the language and actually using the language to help you. If I take this ad absurdum, we might as well avoid pattern matching and recursion, replacing them by explicit eliminators, or completely ban `with` matches; that is to say, I think the line is currently drawn pretty arbitrarily.

> it tells you explicitly what the code is doing [..] I feel like some information is lost here

IMO having more information visible doesn't necessarily mean the proof is easier to understand. Our library design style currently pushes towards sometimes unreasonably long names of definitions (because all identifiers are global), which require specifying more arguments than could be necessary (because we avoid records and instances), and lambdas always increase indentation level. By preferring nested lambdas to a more aligned structure, we create pyramids of doom (right drift of code), even though the reasoning is supposed to be linear. In any case, I don't see where the information is lost — here it's just moved to a less prominent position (the `open do-syntax-elim-exists Motive` is at the end), rather than being duplicated everywhere.

VojtechStep: For general truncations it should be a pretty straightforward application of [map-universal-property-trunc](https://unimath.github.io/agda-unimath/foundation.truncations.html#2876)

EgbertRijke: Thank you both for your further explanations, I thought about it for a bit more, and I do think that the `do` syntax is quite an improvement. I wish I understood it better, but I think you're making a very nice contribution to the library here.

As for my initial reservations, I still think that it is of the highest priority that the code in the library is understandable by the less-initiated. To that end, can you add a clear explanation of how one is supposed to read the `do` syntax? (almost at eli5 level, because I'm really not used to this syntax). What are all the arguments of the do syntax, and what do those arrows mean?

VojtechStep: @lowasser if you don't want to focus on the docs, I can take a crack at them tomorrow

EgbertRijke: Just for context, I've tried to understand the readthedocs page, but I find it terse, and not much of an explanation at all.

fredrik-bakke: I second Egbert's confusion with the `do` syntax

lowasser: I can take care of the documentation work; documenting things like this is part of what I do professionally.

lowasser: I've taken a stab at it.  @VojtechStep , feel free to look at it and offer improvements if you like.

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
...
+showing `Q` requires having a value of `type-Prop P`:
+
+```text
+rec-trunc-Prop
+  ( motive)
+  ( λ q →
+    rec-trunc-Prop
+      ( motive)
+      ( λ q → derive-motive-from p q)
+      ( truncated-prop-Q p))
```
VojtechStep, line 464: I see some unfortunate conflating of a proposition and its witness. Could you rename `truncated-prop-X` to `witness-truncated-prop-X` and maybe `derive-motive-from p q` to `witness-motive-P-Q p q`?

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
...
+[Agda's `do` syntax](https://agda.readthedocs.io/en/v2.7.0/language/syntactic-sugar.html#do-notation)
+is a handy tool to avoid deeply nesting calls to the same lambda-based function.
+For example, consider a case where you are trying to prove a proposition,
+`motive : Prop l`, from a chain of truncated propositions `P` and `Q`, where
+showing `Q` requires having a value of `type-Prop P`:
+
+```text
+rec-trunc-Prop
+  ( motive)
+  ( λ q →
```
VojtechStep, line 460: To emphasize the type of `q` you could use a type annotated lambda `λ (p : P) →`

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
     is-equiv-map-inv-trunc-Prop-diagonal-coproduct
 ```
 
+## `do` syntax for propositional truncation
+
+[Agda's `do` syntax](https://agda.readthedocs.io/en/v2.7.0/language/syntactic-sugar.html#do-notation)
+is a handy tool to avoid deeply nesting calls to the same lambda-based function.
+For example, consider a case where you are trying to prove a proposition,
+`motive : Prop l`, from a chain of truncated propositions `P` and `Q`, where
```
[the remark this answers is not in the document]
VojtechStep: I'd rephrase "from a chain of truncated propositions `P` and `Q`" as "from witnesses of propositional truncations of types `P` and `Q`", to avoid making it sound like `P` and `Q` are propositions

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
...
+do
+  p ← truncated-prop-P
+  q ← truncated-prop-Q p
+  derive-motive-from p q
+where open do-syntax-trunc-Prop motive
+```
+
+Going through each line:
+
+1. You can read the `p ← truncated-prop-P` as an _instruction_ saying, "Get the
```
VojtechStep, line 495: In this context it makes more sense to talk about "extracting the element of `P` out of the witness of `trunc-Prop P`"

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
...
+   value `p` out of its truncation `truncated-prop-Q`." We obviously cannot get
+   values out of truncations in general, but conceptually, we can do it in the
```
VojtechStep, line 497: I'd explicitly mention the universal property: "we cannot extract elements out of witnesses of propositionally truncated types, but since we're eliminating into a proposition, the universal property of the truncation allows us to lift a map from the untruncated type to a map from its truncation".

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,67 @@ module _
...
+   service of proving a proposition, and that's what we're doing here.
+2. `q ← truncated-prop-Q p` says, "Get the value `q` out of the truncation
+   `truncated-prop-Q p`" -- noticing that we can make use of `p` in that line.
```
VojtechStep, line 500: I'd drive home the point that `p` is an element by "we can make use of `p : P` in that line"

VojtechStep (changes requested):
The documentation seems to me to be aimed more towards computer scientists than mathematicians. Could you include an example of a step by step desugaring of the surface syntax? The Agda docs apparently aren't a sufficient explanation, according to https://github.com/UniMath/agda-unimath/pull/1306#discussion_r1953393190, and I agree that it's not immediately visible how their examples relate to truncations.

lowasser:
I've given this a go.  I will certainly admit that the errors here reflect my own limited familiarity and hesitant understanding of univalent type theory, I'm afraid.

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,84 @@ module _
...
+
+```text
+do
+  p ← witness-truncated-prop-P
+  q ← witness-truncated-prop-Q p
+  witness-motive-P-Q p q
+where open do-syntax-trunc-Prop motive
+```
+
+Since Agda's `do` syntax desugars to calls to `>>=`, this is simply syntactic
```
VojtechStep, line 493: We try to avoid using words like "simply", "just", "easily", ...

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,84 @@ module _
...
+  ( λ p → witness-truncated-prop-Q p >>=
+    ( λ q → witness-motive-P-Q p q))
+```
+
+which, inlining the definition of `>>=`, becomes exactly the chain of
+`rec-trunc-Prop` used above.
+
+To read the `do` syntax, it may help to go through each line:
+
+1. `do` simply indicates that we will be using Agda's syntactic sugar for the
```
VojtechStep, line 507: Same here with "simply"

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,84 @@ module _
...
+   `>>=` function defined in the `do-syntax-trunc-Prop` module.
+1. You can read the `p ← witness-truncated-prop-P` as an _instruction_ saying,
+   "Get the value `p` out of the witness of `trunc-Prop P`." We cannot extract
+   elements out of witnesses of propositionally truncated types, but since we're
+   eliminating into a proposition, the universal property of the truncation
+   allows us to lift a map from the untruncated type to a map from its
```
VojtechStep, line 513: On second reading "map from" doesn't sound natural here, maybe "map out of"? I'd defer to a native speaker

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,84 @@ module _
...
+   truncation.
+1. `q ← witness-truncated-prop-Q p` says, "Get the value `q` out of the witness
+   for the truncation `witness-truncated-prop-Q p`" -- noticing that we can make
```
VojtechStep, line 516: We use `---` for em-dashes

VojtechStep (commented):
Up to some nitpicks this LGTM. @EgbertRijke @fredrik-bakke can you parse the application of `do` notation better with this documentation?

src/foundation/propositional-truncations.lagda.md
lowasser: I am a native speaker and do think this is probably the best way to put it, but I'm happy to change it if you feel strongly.

VojtechStep (approved):

src/foundation/existential-quantification.lagda.md
```diff
@@ -202,6 +202,11 @@ module _
         ( elim-exists Q)
 ```
 
+Note that since existential quantification is implemented using propositional
+truncation, the associated
+[do syntax](foundation.propositional-truncations.md#do-syntax) can be used to
```
fredrik-bakke, line 207:
```suggestion
[`do` syntax](foundation.propositional-truncations.md#do-syntax) can be used to
```

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,83 @@ module _
...
+      ( λ (q : Q) → witness-motive-P-Q p q)
+      ( witness-truncated-prop-Q p))
+  ( witness-truncated-prop-P)
+```
+
+The tower of indentation, with many layers of indentation in the innermost
+derivation, is a little awkward even at two levels, let alone more. In
+particular, we have the many duplicated lines of `rec-trunc-Prop motive`, and
+the increasing distance between the `rec-trunc-Prop` and the `trunc-Prop` being
+recursed on. Agda's `do` syntax offers us an alternative.
```
fredrik-bakke, line 472: I don't think this is the best way to phrase the use case for `do` syntax. Rather than phrasing it as mitigating something _bad_ such as the pattern above, which I'm not convinced yet is inherently bad, I would prefer if it was phrased purely as a useful tool for abstraction into the "language of propositional truncations".

src/foundation/propositional-truncations.lagda.md
```diff
@@ -446,6 +446,83 @@ module _
...
+
+```text
+do
+  p ← witness-truncated-prop-P
+  q ← witness-truncated-prop-Q p
+  witness-motive-P-Q p q
+where open do-syntax-trunc-Prop motive
+```
+
+Since Agda's `do` syntax desugars to calls to `>>=`, this is syntactic sugar for
```
fredrik-bakke, line 493: "Desugaring". "syntactic sugar", "inlining", and "internal (representation)" are computer science terms that we probably shouldn't expect the reader to know. Could you try and give a more mathematical explanation of what is going on? I.e., it is a syntax for reasoning in a certain mode of logic, and that unfolds to some primitive thing that we mathematically already understand.

fredrik-bakke: If I'm being a little unclear let me know, I can try and suggest a rephrasing.

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -85,28 +88,20 @@ module _
...
-                  ( p)
-                  ( q)
-                  ( le-lower-upper-cut-ℝ y p q p-in-ly q-in-uy)
-                  ( q≤p))
-              ( decide-le-leq-ℚ p q))
-          ( y<x))
-      ( x<y)
+    do
+      p , x<p , p<y ← x<y
+      q , y<q , q<x ← y<x
```
fredrik-bakke, line 93:
```suggestion
      ( p , x<p , p<y) ← x<y
      ( q , y<q , q<x) ← y<x
```

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -120,23 +115,15 @@ module _
...
-                  ( z)
-                  ( p)
-                  ( q)
-                  ( le-lower-upper-cut-ℝ y p q p-in-ly q-in-uy)
-                  ( q-in-lz)))
-          ( y<z))
+  transitive-le-ℝ y<z x<y =
+    do
+      p , x<p , p<y ← x<y
+      q , y<q , q<z ← y<z
```
fredrik-bakke, line 121:
```suggestion
      ( p , x<p , p<y) ← x<y
      ( q , y<q , q<z) ← y<z
```

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -239,14 +205,14 @@ module _
...
-    elim-exists
-      ( le-ℝ-Prop (neg-ℝ y) (neg-ℝ x))
-      ( λ p (p-in-ux , p-in-ly) →
-        intro-exists
-          ( neg-ℚ p)
-          ( tr (is-in-lower-cut-ℝ y) (inv (neg-neg-ℚ p)) p-in-ly ,
-            tr (is-in-upper-cut-ℝ x) (inv (neg-neg-ℚ p)) p-in-ux))
+  reverses-order-neg-ℝ x<y =
+    do
+      p , x<p , p<y ← x<y
```
fredrik-bakke, line 210:
```suggestion
      ( p , x<p , p<y) ← x<y
```

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -359,17 +325,18 @@ module _
...
-          ( λ _ _ → real-ℚ q)
-          ( λ p r (p<q , p∈ux) (q<r , r∈ly) →
-            intro-exists p (p∈ux , p<q) , intro-exists r (q<r , r∈ly))
-          ( forward-implication (is-rounded-upper-cut-ℝ x q) q∈ux)
-          ( forward-implication (is-rounded-lower-cut-ℝ y q) q∈ly))
+  dense-le-ℝ x<y =
+    do
+      q , x<q , q<y ← x<y
+      p , p<q , x<p ← forward-implication (is-rounded-upper-cut-ℝ x q) x<q
+      r , q<r , r<y ← forward-implication (is-rounded-lower-cut-ℝ y q) q<y
```
fredrik-bakke, line 332:
```suggestion
      ( q , x<q , q<y) ← x<y
      ( p , p<q , x<p) ← forward-implication (is-rounded-upper-cut-ℝ x q) x<q
      ( r , q<r , r<y) ← forward-implication (is-rounded-lower-cut-ℝ y q) q<y
```

fredrik-bakke:
The documentation reads as very computer sciency to me. Keep in mind that the intended audience should include mathematically curious high-schoolers.

src/foundation/propositional-truncations.lagda.md
VojtechStep: The combination of "we shouldn't expect the reader to know what syntactic sugar is" and "we expect the reader to know what a certain mode of logic is" brings to mind the [Average familiarity xkcd](https://xkcd.com/2501/).

I think we'll need to meet somewhere in the middle on this one. I still see this addition as an ergonomics improvement to the usage of Agda-qua-language, so the programming language vocabulary sounds justified. If the docs were stated in terms of e.g. "working in a fragment of logic where we may directly observe witnesses of propositionally truncated types", then I don't think I'd come out of reading it any wiser. Perhaps we should have both explanations?

VojtechStep: Sorry, that was strawman-ish of me. But then I'm not sure what the more mathematical explanation would look like?

lowasser:
I have to admit that I'm not sure how to make this documentation more mathematically framed, and would appreciate help getting it the rest of the way to what you're hoping for.

fredrik-bakke:
@EgbertRijke @VojtechStep

It seems that @lowasser's work is blocked on this PR right now. I'd like to review the documentation properly before it is merged and don't have the time right now. Since it seems we agree on the implementation of the syntax itself, perhaps we could merge this PR without the documentation, and leave that for a second PR so that Louis can continue his work?

VojtechStep:
In this case I'd prefer a CS documentation useful to some contributors over no documentation, but I won't block the PR on a decision about that, any way of merging it is fine with me.

fredrik-bakke:
I'd prefer it if we didn't rush the documentation.

lowasser:
I am disinclined to eliminate the before-and-after example even temporarily, but I've attempted to cut everything else in the hopes we can get this in.

fredrik-bakke:
Okay, I'm merging this so that Louis can continue his work. We can discuss the documentation in further detail or potential alternatives and reverts in a follow-up. Would you be so kind as to open a follow-up draft PR with the documentation, @lowasser?

PR #1312, opened 2025-02-09
Switch from `𝔽` to `Finite-*`

fredrik-bakke:


src/univalent-combinatorics/finite-types.lagda.md
```diff
@@ -273,9 +273,9 @@ abstract
   is-finite-Fin : (k : ℕ) → is-finite (Fin k)
   is-finite-Fin k = is-finite-count (count-Fin k)
 
-Fin-𝔽 : ℕ → 𝔽 lzero
-pr1 (Fin-𝔽 k) = Fin k
-pr2 (Fin-𝔽 k) = is-finite-Fin k
+finite-type-Fin : ℕ → Finite-Type lzero
```
EgbertRijke, line 276: Shouldn't this be called `Fin-Finite-Type`? We also say `unit-Finite-Type` and not `finite-type-unit`, and the convention is similar for the group of integers, and so on.

EgbertRijke: We could also be thinking of the name `standard-Finite-Type`

EgbertRijke:
Are all the extra lines of code just line breaks, or is there some hidden formalization work that I should be looking out for?

fredrik-bakke:
This PR only renames things. The added lines are only because the names got longer.

EgbertRijke:
Makes sense! Just asking what to expect

src/elementary-number-theory/modular-arithmetic.lagda.md
```diff
@@ -130,9 +130,9 @@ abstract
   is-finite-ℤ-Mod {zero-ℕ} H = ex-falso (H refl)
   is-finite-ℤ-Mod {succ-ℕ k} H = is-finite-Fin (succ-ℕ k)
 
-ℤ-Mod-𝔽 : (k : ℕ) → is-nonzero-ℕ k → 𝔽 lzero
-pr1 (ℤ-Mod-𝔽 k H) = ℤ-Mod k
-pr2 (ℤ-Mod-𝔽 k H) = is-finite-ℤ-Mod H
+finite-type-ℤ-Mod : (k : ℕ) → is-nonzero-ℕ k → Finite-Type lzero
```
EgbertRijke, line 133: I think this name should be `ℤ-Mod-Finite-Type`, in line with `ℤ-Mod-Group` and `ℤ-Mod-Ab`

src/univalent-combinatorics/finite-types.lagda.md
fredrik-bakke: I named it `Fin-Finite-Type`

src/foundation/decidable-equivalence-relations.lagda.md
```diff
@@ -65,90 +65,90 @@ is-decidable-equivalence-relation :
 is-decidable-equivalence-relation {A = A} R =
   (x y : A) → is-decidable ( sim-equivalence-relation R x y)
 
-Decidable-equivalence-relation :
+Decidable-Equivalence-Relation :
```
EgbertRijke, line 68: We are somewhat inconsistent with capitalization, which does not have to be addressed in this pull request, but given that `equivalence-relation` is uncapitalized it is a little bit strange that decidable equivalence relations are capitalized.

fredrik-bakke: Ah, I didn't know of this inconsistency. If you want I can capitalize equivalence relations

src/foundation/decidable-propositions.lagda.md
```diff
@@ -265,9 +265,9 @@ number-of-elements-Decidable-Prop =
       ( count-Decidable-Prop)
       ( is-finite-Decidable-Prop))
 
-decidable-Prop-𝔽 : (l : Level) → 𝔽 (lsuc l)
-pr1 (decidable-Prop-𝔽 l) = Decidable-Prop l
-pr2 (decidable-Prop-𝔽 l) = is-finite-Decidable-Prop
+decidable-prop-Finite-Type : (l : Level) → Finite-Type (lsuc l)
+pr1 (decidable-prop-Finite-Type l) = Decidable-Prop l
+pr2 (decidable-prop-Finite-Type l) = is-finite-Decidable-Prop
```
EgbertRijke, line 270: Probably the better capitalization here would be `Decidable-Prop-Finite-Type`, because `Decidable-Prop` is capitalized. Usually, when we write something like `type-Ring` it is to construct *a* type from a ring, and to me `decidable-prop-Finite-Type` would be some construction of a decidable proposition from a finite type.

We really need to sort out our capitalization conventions:)

src/univalent-combinatorics/coproduct-types.lagda.md
```diff
@@ -37,7 +37,7 @@ open import univalent-combinatorics.standard-finite-types
 ## Idea
 
 Coproducts of finite types are finite, giving a coproduct operation on the type
-𝔽 of finite types.
+Finite-Type of finite types.
```
EgbertRijke, line 40: Should this be displayed as code?

```suggestion
`Finite-Type` of finite types.
```

src/univalent-combinatorics/decidable-equivalence-relations.lagda.md
```diff
@@ -50,136 +50,141 @@ each `R x y` is a
 ## Definition
 
 ```agda
-Decidable-equivalence-relation-𝔽 :
-  {l1 : Level} (l2 : Level) (X : 𝔽 l1) → UU (l1 ⊔ lsuc l2)
-Decidable-equivalence-relation-𝔽 l2 X =
-  Decidable-equivalence-relation l2 (type-𝔽 X)
+Decidable-Equivalence-Relation-Finite-Type :
```
EgbertRijke, line 53: I'm wondering if (definitely not asserting that) it is better to name this `decidable-equivalence-relation-Finite-Type`.

My reasoning would be as follows: If we would construct the finite type of decidable equivalence relations on a finite type, that would reasonably be called `Decidable-Equivalence-Relation-Finite-Type`. Also, if we would define a specific decidable equivalence relation on an arbitrary finite type, it would be called `name-decidable-equivalence-relation-Finite-Type`, so having the type of decidable equivalence relations on a finite type half-capitalized seems to work well for that use case.

Like I said, I'm not 100% convinced either way of what a good capitalization guideline is.

src/foundation/decidable-equivalence-relations.lagda.md
EgbertRijke: Relations are capitalized, but subtypes and partitions are also uncapitalized. Instead of arbitrarily starting to capitalize some things that are now uncapitalized, it would be better if we think through a sensible guideline.

fredrik-bakke: Since `Relation` is already capitalized and `Decidable` is already capitalized, and I don't want to change their capitalization as part of this PR, I'm inclined to also capitalize `equivalence-relation`. I'd prefer not to have to sort out capitalizations of other concepts as part of this PR also — so if you want, I can go back and revert the capitalization of `equivalence-relation`.

fredrik-bakke:
By the way, not that I'm going to add further changes to this PR, but should `UU-Fin` rather be called `Type-With-Finite-Cardinality`?

src/univalent-combinatorics/decidable-equivalence-relations.lagda.md
fredrik-bakke: `decidable-equivalence-relation-Finite-Type` reads like "the underlying decidable equivalence relation on a finite type". That would be its equality, right? On the other hand `Decidable-Equivalence-Relation-Finite-Type` reads like "the finite type of decidable equivalence relations on a finite type". When I was renaming this entry I was considering `Decidable-Equivalence-Relation-On-Finite-Type` but decided against it since it seemed unconventional. Perhaps the best name for this entry is actually `type-Decidable-Equivalence-Relation-Finite-Type`, since it is "the underlying type of the finite type of decidable equivalence relations on a finite type".

EgbertRijke: The last suggestion `type-Decidable-Equivalence-Relation-Finite-Type` sounds good. I think it could actually be relevant in some parts of combinatorics to have access to the finite type of decidable equivalence relations, and then `type-...` sounds sensible for its underlying type.

Do you want to go with that?

EgbertRijke:
> By the way, not that I'm going to add further changes to this PR, but should `UU-Fin` rather be called `Type-With-Finite-Cardinality`?

That's a reasonable suggestion

src/foundation/decidable-equivalence-relations.lagda.md
EgbertRijke: Ok, let's leave this as it is in the current state of the pull request. It's not the end of the world to have some small inconsistencies.

src/univalent-combinatorics/decidable-equivalence-relations.lagda.md
fredrik-bakke: Done so now

fredrik-bakke:
Did you finish your review last night or should I be expecting further comments?

PR #1316, opened 2025-02-11
Rename `UU-Fin` to `Type-With-Cardinality-ℕ`

fredrik-bakke:


fredrik-bakke:
This is my last univalent combinatorics PR for the time being.

src/finite-group-theory/finite-type-groups.lagda.md
```diff
@@ -39,12 +39,22 @@ open import univalent-combinatorics.standard-finite-types
...
+pr2 (pr1 (Type-With-Cardinality-ℕ-Group l n)) =
+  is-0-connected-Type-With-Cardinality-ℕ n
+pr2 (Type-With-Cardinality-ℕ-Group l n) =
+  is-1-type-Type-With-Cardinality-ℕ n
+    ( raise-Fin-Type-With-Cardinality-ℕ l n)
+    ( raise-Fin-Type-With-Cardinality-ℕ l n)
+
+Type-With-Cardinality-ℕ-Group' : (l : Level) (n : ℕ) → Group (lsuc l)
+Type-With-Cardinality-ℕ-Group' l n =
+  group-Concrete-Group (Type-With-Cardinality-ℕ-Group l n)
```
EgbertRijke, line 57: Wouldn't it be better to use `Concrete-Group` in the naming scheme here, instead of a prime?

```suggestion
Type-With-Cardinality-ℕ-Concrete-Group :
  (l : Level) (n : ℕ) → Concrete-Group (lsuc l)
pr1 (pr1 (pr1 (Type-With-Cardinality-ℕ-Concrete-Group l n))) =
  Type-With-Cardinality-ℕ l n
pr2 (pr1 (pr1 (Type-With-Cardinality-ℕ-Concrete-Group l n))) =
  raise-Fin-Type-With-Cardinality-ℕ l n
pr2 (pr1 (Type-With-Cardinality-ℕ-Concrete-Group l n)) =
  is-0-connected-Type-With-Cardinality-ℕ n
pr2 (Type-With-Cardinality-ℕ-Concrete-Group l n) =
  is-1-type-Type-With-Cardinality-ℕ n
    ( raise-Fin-Type-With-Cardinality-ℕ l n)
    ( raise-Fin-Type-With-Cardinality-ℕ l n)

Type-With-Cardinality-ℕ-Group : (l : Level) (n : ℕ) → Group (lsuc l)
Type-With-Cardinality-ℕ-Group l n =
  group-Concrete-Group (Type-With-Cardinality-ℕ-Group l n)
```

fredrik-bakke: yep, I agree

fredrik-bakke: Fixed now

EgbertRijke:
This pull request contains some big improvements, so I am ready to merge it.

On the other hand, it also brings to the surface a problem that we will need to address at some point. It is really strange to name a group after the specification of its classifying type. Instead, the group should be named after the symmetries it contains. For example, `Type-Of-Squares-Group` would be a really strange name for the dihedral group. We create this sort of problems when we write `Type-With-Cardinality-N-(Concrete-)Group`. 

Probably the best solution would be to create a separate file for pointed connected types, have the type of types with cardinality n be a pointed connected type, and introduce the _nth symmetric concrete group_ as the pointed connected type of types with cardinality n.

Issue #834, opened 2023-10-12
The long exact sequence of homotopy groups


fredrik-bakke:
It would be very nice to have this sequence formalized in the library.
Some intermediate goals are
- [ ] Define the homotopy sets `πₙ` for `n ≥ 0`
- [ ] Define the homotopy groups `πₙ` for `n ≥ 1`
- [ ] Define the homotopy abelian groups `πₙ` for `n ≥ 2`
- [ ] Define the fiber sequence of a pointed map
- [ ] Show functoriality of taking homotopy sets/groups
- [ ] Define exact sequences of pointed sets

and finally
- [ ] Show that the induced sequence on homotopy groups is exact

### References 
- _Homotopy Type Theory – Univalent Fundations of Mathematics_, sections 8.0 and 8.4
- _An introduction to Homotopy Type Theory_, sections 19.4-19.5

PR #1318, opened 2025-02-12
Fiber sequences

fredrik-bakke:
Some work on #834.


fredrik-bakke:
Since we call maps out of NN sequences, what should we call maps out of ZZ?

PR #1321, opened 2025-02-13
chore: Fix links in `100-theorems`

### LINKED ISSUE #1204: Mdbook doesn't render links in headers in the right-hand bar
I'm guessing this is related to #658 and is a bug on the side of `mdbook`. Still, I wanted to at least record it.
Screenshot 2024-10-19 at 20 15 58

fredrik-bakke:
Fixes #1204.

### DIFF
--- a/src/literature/100-theorems.lagda.md
+++ b/src/literature/100-theorems.lagda.md
@@ -11,7 +11,7 @@ module literature.100-theorems where
 
 ## The list
 
-### 3. The denumerability of the tational numbers {#3}
+### 3. The denumerability of the rational numbers {#3}
 
 **Author:** [Fredrik Bakke](https://www.ntnu.edu/employees/fredrik.bakke)
 
--- a/src/literature/100-theorems.lagda.md
+++ b/src/literature/100-theorems.lagda.md
@@ -11,7 +11,7 @@ module literature.100-theorems where
 
 ## The list
 
-### 3. The Denumerability of the Rational Numbers {#3}
+### 3. The denumerability of the tational numbers {#3}
 
 **Author:** [Fredrik Bakke](https://www.ntnu.edu/employees/fredrik.bakke)
 
@@ -20,7 +20,7 @@ open import elementary-number-theory.rational-numbers using
   ( is-countable-ℚ)
 ```
 
-### 11. The Infinitude of Primes {#11}
+### 11. The infinitude of primes {#11}
 
 **Author:** [Egbert Rijke](https://egbertrijke.github.io)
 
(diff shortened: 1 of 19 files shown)


src/literature/100-theorems.lagda.md
```diff
@@ -11,7 +11,7 @@ module literature.100-theorems where
 
 ## The list
 
-### [3. The Denumerability of the Rational Numbers](https://www.cs.ru.nl/~freek/100/#3) {#3}
+### 3. The Denumerability of the Rational Numbers [(#3)](https://www.cs.ru.nl/~freek/100/#3) {#3}
```
VojtechStep, line 14: I'd expect this to still generate invalid HTML. Could you check the output if it has nested `a` tags?

fredrik-bakke: Ah, hmmm, I guess you're right. Honestly, I don't feel like trying to build the website locally to check. I can remove the links completely instead? Otherwise, I can try to incorporate them without having them in the header.

fredrik-bakke: Alright, I ended up just removing them. They were not particularly useful anyways.

src/literature/100-theorems.lagda.md
```diff
@@ -11,7 +11,7 @@ module literature.100-theorems where
 
 ## The list
 
-### [3. The Denumerability of the Rational Numbers](https://www.cs.ru.nl/~freek/100/#3) {#3}
+### 3. The Denumerability of the Rational Numbers {#3}
```
VojtechStep, line 14: Regarding the `{#n}` header ids: when Agda processes files for HTML output, it creates anchor tags for all tokens in code blocks, which have ids based on their position in the file. That means that as long as the start of the first Agda block is more than 100 characters into the markdown file, ids 1-100 are safe. It's something to keep in mind when using purely numerical ids, but it's fine here, so feel free to resolve this

src/real-numbers/metric-space-of-real-numbers.lagda.md
```diff
@@ -99,7 +99,7 @@ premetric-leq-ℝ l d x y =
 ### The standard premetric on the real numbers is a metric structure
 
 The triangle inequality is the 91st theorem on
-[Freek Wiedijk's](http://www.cs.ru.nl/F.Wiedijk/) list of
+[Freek Wiedijk](http://www.cs.ru.nl/F.Wiedijk/)'s list of
 [100 theorems](literature.100-theorems.md) {{#cite 100theorems}}.
```
VojtechStep, line 103: You might consider linking to the specific theorem on the list, but I don't have strong feelings about it

VojtechStep (approved):
LGTM 👌

src/real-numbers/metric-space-of-real-numbers.lagda.md
fredrik-bakke: Good idea! I'll hyperlink the number in the paragraph to the specific theorem

Issue #1322, opened 2025-02-14
Participate in "1000+ theorems" project


fredrik-bakke:
Properly participating in the 1000+ theorems has a couple of blockers on their side:
- https://github.com/1000-plus/1000-plus.github.io/issues/22
- https://github.com/1000-plus/1000-plus.github.io/issues/23
- https://github.com/1000-plus/1000-plus.github.io/issues/24

This issue is a branch-out from #1214.

## Formalized theorems

- Bezout's lemma, by Bryan Lu
- Binomial theorem, by Egbert Rijke
- Cantor–Schröder–Bernstein theorem, by Elif Uskuplu
- Cantor's theorem, by Egbert Rijke
- Cayley's theorem
- Diaconescu's theorem, by Fredrik Bakke #1226
- Euclid's theorem, by Egbert Rijke
- Fundamental theorem of arithmetic, by Victor Blanchi
- Fundamental theorem of equivalence relations, by Egbert Rijke
- Kleene's fixed point theorem, by Fredrik Bakke
- Knaster–Tarski fixed point theorem, by Fredrik Bakke
- Lawvere's fixed point theorem
- Nicomachus's theorem, by Egbert Rijke #1211
- Yoneda lemma, by Emily Riehl

## Useful links
- [1000-plus.github.io](https://1000-plus.github.io/)
- [1000-plus/1000-plus.github.io](https://github.com/1000-plus/1000-plus.github.io) on GitHub
- [List of theorems](https://en.wikipedia.org/wiki/List_of_theorems) on Wikipedia

VojtechStep:
I see that they have a script to sync with mathlib formalization. We could offer a similar script, since we already expose a machine-readable list of concepts, including wikidata IDs where possible, at https://unimath.github.io/agda-unimath/concept_index.json.

fredrik-bakke:
That would be amazing!

fredrik-bakke:
Note that I posted an issue there about Wikidata identifiers, as it turns out not every theorem on Wikipedia has a Wikidata identifier. There's also a standing issue about switching the identifier system on their side, although it seems the developers decided against it.

fredrik-bakke:
@VojtechStep Do you see a way to link to a row in the 1000+ theorems table, either the `index` one or the `all` one? Am I stupid or do they not have a way to link to an entry??

VojtechStep:
[...]

Issue #1214, opened 2024-10-28
List of notable theorems


EgbertRijke:
Freek's list is weird and arbitrary, and although it is widely used as a benchmark by libraries of formalized mathematics and therefore it is important to track it on our library.

However, it would make sense to me if we have a separate list of "Notable theorems", in which we list theorems in alphabetic order that have been formalized in our library. A reference list could be Wikipedia's [list of notable theorems](https://en.wikipedia.org/wiki/List_of_theorems). There are plenty of notable theorems that are within reach for us, that would be nice to have, but which wouldn't be highlighted in the 100 Theorems list. A benefit of this approach is that what is regarded as important is not determined by one person at one moment in time, but is continuously maintained by the mathematical community and does not have an arbitrary cutoff at 100.

Any thoughts on this?

fredrik-bakke:
Very much agree! I don't think we should restrict ourselves to what Wikipedia lists as notable either, although I expect it's a hairy endeavour to arbit what is and isn't notable.

fredrik-bakke:
If we want to use an external source for the list of theorems however, I can mention again Freek's new project named 1000+ theorems.

https://1000-plus.github.io/

fredrik-bakke:
The reason why I don't find that list sufficient is that the criteria for being on Wikipedia's list don't perfectly align with our interests. To be on Wikipedia's list the theorem needs to have a wikipedia article, meaning the theorem needs to have enough historical precedence or that someone sees a need for it to be accessible to a larger audience in a wikiformat.

EgbertRijke:
> The reason why I don't find that list sufficient is that the criteria for being on Wikipedia's list don't perfectly align with our interests. To be on Wikipedia's list the theorem needs to have a wikipedia article, meaning the theorem needs to have enough historical precedence or that someone sees a need for it to be accessible to a larger audience in a wikiformat (and has enough resources and expertise to write it).

These are all reasons for me why that list is much better than any criteria any individual could come up with.

> don't perfectly align with our interest.

I'd say that this is of critical importance! A list of notable theorems shouldn't be determined by any small group of individual's interests.

fredrik-bakke:
I see, so you wouldn't want a list of "notable theorems... in univalent mathematics"?

EgbertRijke:
[...]

PR #1324, opened 2025-02-14
Page for 1000+ theorems

fredrik-bakke:
Related to #1322 and #1214.


fredrik-bakke:
The authors of Cayley's theorem and Lawvere's fixed point theorem are still unidentified. I've only been able to trace them back to refactoring PRs by Egbert. @EgbertRijke, did you author these theorems' formalizations?

fredrik-bakke:
> The authors of Cayley's theorem and Lawvere's fixed point theorem are still unidentified. I've only been able to trace them back to refactoring PRs by Egbert. @EgbertRijke, did you author these theorems' formalizations?

@EgbertRijke

PR #1327, opened 2025-02-14
Add level constants

fredrik-bakke:
Defines constants for the first 5 universe levels.


fredrik-bakke:
I think this might just be a bad idea, actually.

PR #1330, opened 2025-02-15
Quasigroups

djspacewhale:
Introducing a new folder that, so far, defines quasigroups and loops (unital quasigroups).


src/quasigroups/quasigroups.lagda.md
```diff
@@ -0,0 +1,207 @@
...
+
+{{#concept "Quasigroups" Agda=Quasigroup}} are [sets](foundation-core.sets.md) -
+say, `Q` - equipped with three binary operations - say, `*`, `/`, `\` - such
+that for `x y : type-Set Q`, the following hold:
+
+```text
+  y ＝ (y r/ x) * x
+  y ＝ (y * x) r/ x
+  y ＝ x * (x l/ y)
+  y ＝ x l/ (x * y)
```
fredrik-bakke, line 30: You haven't defined what the symbols `r/` and `l/` mean. Are these supposed to be `/` and `\`?

src/quasigroups.lagda.md
fredrik-bakke, line 1: The justification for creating a new namespace for quasigroups is unclear to me. Why do you not wish to put this in `group-theory`? Notice that we also have semigroups and monoids in `group-theory`. I must admit I'm not familiar with the theory of quasigroups, is the nature of the theory very different from `group-theory`? Given its name, I would guess not, and then it might be appropriate to have it in `group-theory`.

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+  e * x ＝ x
+  x * e ＝ x
+```
+
+Note: we will see that units, when they exist, are unique, and so being a loop
+is actually a property of a quasigroup rather than structure.
+
+## Definitions
+
+### Left units in quasigroups
```
fredrik-bakke, line 46: Units should have a separate file `units-quasigroups`

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  private
+    _*_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
+    _*_ = mul-Quasigroup Q
+
+    _l/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
+    _l/_ = left-div-Quasigroup Q
+
+    _r/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
+    _r/_ = right-div-Quasigroup Q
```
fredrik-bakke, line 61: Private definitions are against our conventions. But... since we're discussing changes to our syntax conventions currently, maybe I can use you as a test dummy? I have an idea for how we might go about having operators like these in the library

src/quasigroups/quasigroups.lagda.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Roughly speaking, quasigroups are nonassociative semigroups, and in fact groups
+are just pointed associative quasigroups.
+
+## Definitions
+
+### Left and right quasigroups
+
+#### Left quasigroups
```
fredrik-bakke, line 40: Left quasigroups can have their own file

src/quasigroups/quasigroups.lagda.md
```diff
@@ -0,0 +1,207 @@
...
+
+left-Quasigroup : (l : Level) → UU (lsuc l)
+left-Quasigroup l =
+  Σ (Set l) λ Q → Σ (type-Set Q → type-Set Q → type-Set Q)
+  (λ mul → Σ (type-Set Q → type-Set Q → type-Set Q) (λ left-div →
+  is-left-cancellative-left-div Q mul left-div ×
+  is-right-cancellative-left-div Q mul left-div))
+```
+
+#### Right quasigroups
```
fredrik-bakke, line 93: Right quasigroups can have their own file

src/quasigroups/quasigroups.lagda.md
```diff
@@ -0,0 +1,207 @@
...
+  is-left-cancellative-right-div-Quasigroup =
+    pr1 (pr2 (pr2 (pr2 (pr2 (pr2 Q)))))
+
+  is-right-cancellative-right-div-Quasigroup :
+    is-right-cancellative-right-div set-Quasigroup
+    mul-Quasigroup right-div-Quasigroup
+  is-right-cancellative-right-div-Quasigroup =
+    pr2 (pr2 (pr2 (pr2 (pr2 (pr2 Q)))))
+```
+
```
fredrik-bakke, line 204: If you add the below module, then you can "`open`" the syntax elsewhere using `open operator-syntax-Quasigroup Q`, to access these operators for your chosen quasigroup `Q`, instead of using private definitions as you do currently.

````suggestion
```

### Quasigroup operator syntax

```agda
module
  operator-syntax-Quasigroup
  {l : Level} (Q : Quasigroup l)
  where

  _*_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _*_ = mul-Quasigroup Q

  _l/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _l/_ = left-div-Quasigroup Q

  _r/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _r/_ = right-div-Quasigroup Q
```

````

Comment (fredrik-bakke) [reply src/quasigroups/quasigroups.lagda.md:204]: We could also consider using the [division slash](https://codepoints.net/U+2215) `∕` and [reverse solidus operator](https://codepoints.net/U+29F5) `⧵` unicode characters for left and right division respectively. These are supposed to be mirrored characters, and semantically they're the right symbols. There's no agda-input sequence for them though, so this requires adding them to our list of special characters. Might still be worth considering though.

File: src/quasigroups/loops.lagda.md [hunk 7]
```diff
@@ -0,0 +1,220 @@
...
+
+  term-has-left-unit-Quasigroup : has-left-unit-Quasigroup → type-Quasigroup Q
+  term-has-left-unit-Quasigroup (e , _) = e
+
+  left-units-agree-Quasigroup :
+    (e f : has-left-unit-Quasigroup) → term-has-left-unit-Quasigroup e
+    ＝ term-has-left-unit-Quasigroup f
+  left-units-agree-Quasigroup (e , e-left-unit) (f , f-left-unit) =
+    equational-reasoning
+    e ＝ (e * f) r/ f
```
fredrik-bakke, line 87:
```suggestion
    e
    ＝ (e * f) r/ f
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+open import quasigroups.quasigroups
+```
+
+</details>
+
+## Idea
+
+{{#concept "Loops" Agda=Loop}} are [quasigroups](quasigroups.quasigroups.md)
+with a designated identity element, that is, `e : type-Quasigroup Q` such that
```
fredrik-bakke, line 33:
```suggestion
with a designated identity element, that is, an element `e : Q` such that
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+      by ap (λ x → x r/ f) (e-left-unit f)
+    ＝ (f * f) r/ f
+      by ap (λ x → x r/ f) (inv (f-left-unit f))
+    ＝ f
+      by inv (is-right-cancellative-right-div-Quasigroup Q f f)
+
+  is-prop-has-left-unit-Quasigroup : is-prop has-left-unit-Quasigroup
+  pr1 (is-prop-has-left-unit-Quasigroup e f) =
+    eq-type-subtype is-left-unit-Quasigroup-Prop
+    (left-units-agree-Quasigroup e f)
```
fredrik-bakke, line 99: When the arguments don't fit on a single line, we break the line at the earliest possible point, and place every new argument on a new line. Here

```suggestion
    eq-type-subtype
      ( is-left-unit-Quasigroup-Prop)
      ( left-units-agree-Quasigroup e f)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+  pr2 (is-prop-has-left-unit-Quasigroup e f) p =
+    is-set-has-uip (is-set-type-subtype is-left-unit-Quasigroup-Prop
+    (is-set-Quasigroup Q)) e f (pr1 (is-prop-has-left-unit-Quasigroup e f)) p
```
fredrik-bakke, line 102:
```suggestion
  pr2 (is-prop-has-left-unit-Quasigroup e f) =
    is-set-has-uip 
      ( is-set-type-subtype is-left-unit-Quasigroup-Prop
        ( is-set-Quasigroup Q)) 
      ( e)
      ( f)
      ( pr1 (is-prop-has-left-unit-Quasigroup e f))
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  term-has-right-unit-Quasigroup : has-right-unit-Quasigroup → type-Quasigroup Q
+  term-has-right-unit-Quasigroup (e , _) = e
+
+  right-units-agree-Quasigroup :
+    (e f : has-right-unit-Quasigroup) → term-has-right-unit-Quasigroup e
+    ＝ term-has-right-unit-Quasigroup f
+  right-units-agree-Quasigroup (e , e-right-unit) (f , f-right-unit) =
+    equational-reasoning
+    e ＝ f l/ (f * e)
```
fredrik-bakke, line 132:
```suggestion
    e
    ＝ f l/ (f * e)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+      by ap (λ x → f l/ x) (e-right-unit f)
+    ＝ f l/ (f * f)
+      by inv ((ap (λ x → f l/ x) (f-right-unit f)))
+    ＝ f
+      by inv (is-right-cancellative-left-div-Quasigroup Q f f)
+
+  is-prop-has-right-unit-Quasigroup : is-prop has-right-unit-Quasigroup
+  pr1 (is-prop-has-right-unit-Quasigroup e f) =
+    eq-type-subtype is-right-unit-Quasigroup-Prop
+    (right-units-agree-Quasigroup e f)
```
fredrik-bakke, line 144:
```suggestion
    eq-type-subtype
      ( is-right-unit-Quasigroup-Prop)
      ( right-units-agree-Quasigroup e f)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+  pr2 (is-prop-has-right-unit-Quasigroup e f) p =
+    is-set-has-uip (is-set-type-subtype is-right-unit-Quasigroup-Prop
+    (is-set-Quasigroup Q)) e f (pr1 (is-prop-has-right-unit-Quasigroup e f)) p
```
fredrik-bakke, line 147:
```suggestion
  pr2 (is-prop-has-right-unit-Quasigroup e f) =
    is-set-has-uip
      ( is-set-type-subtype is-right-unit-Quasigroup-Prop
        ( is-set-Quasigroup Q))
      ( e)
      ( f)
      ( pr1 (is-prop-has-right-unit-Quasigroup e f))
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+A **unit**, as usual, is both a left and right unit. In fact, if `Q` has both a
+left unit `e` and a right unit `f`, then already we have `e ＝ f`, and thus the
+type of units is equivalent to the type of pairs of left and right units.
+
+```agda
+  has-unit-Quasigroup : UU l
+  has-unit-Quasigroup =
+    Σ (type-Quasigroup Q)
+    λ e → is-left-unit-Quasigroup e × is-right-unit-Quasigroup e
```
fredrik-bakke, line 160:
```suggestion
    Σ ( type-Quasigroup Q)
      ( λ e → is-left-unit-Quasigroup e × is-right-unit-Quasigroup e)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  unit-has-unit-Quasigroup : has-unit-Quasigroup → type-Quasigroup Q
+  unit-has-unit-Quasigroup (e , _) = e
+
+  has-unit-has-left-unit-Quasigroup :
+    has-unit-Quasigroup → has-left-unit-Quasigroup
+  has-unit-has-left-unit-Quasigroup (e , e-left-unit , _) = e , e-left-unit
```
fredrik-bakke, line 167:
```suggestion
  has-unit-has-left-unit-Quasigroup (e , e-left-unit , _) = (e , e-left-unit)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  has-unit-has-right-unit-Quasigroup :
+    has-unit-Quasigroup → has-right-unit-Quasigroup
+  has-unit-has-right-unit-Quasigroup (e , _ , e-right-unit) = e , e-right-unit
```
fredrik-bakke, line 171:
```suggestion
  has-unit-has-right-unit-Quasigroup (e , _ , e-right-unit) = (e , e-right-unit)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  has-unit-has-left-and-right-units-Quasigroup :
+    has-unit-Quasigroup → has-left-unit-Quasigroup × has-right-unit-Quasigroup
+  has-unit-has-left-and-right-units-Quasigroup e =
+    (has-unit-has-left-unit-Quasigroup e) ,
+    (has-unit-has-right-unit-Quasigroup e)
```
fredrik-bakke, line 177:
```suggestion
    ( has-unit-has-left-unit-Quasigroup e) ,
    ( has-unit-has-right-unit-Quasigroup e)
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  has-unit-has-left-unit-Quasigroup :
+    has-unit-Quasigroup → has-left-unit-Quasigroup
+  has-unit-has-left-unit-Quasigroup (e , e-left-unit , _) = e , e-left-unit
+
+  has-unit-has-right-unit-Quasigroup :
+    has-unit-Quasigroup → has-right-unit-Quasigroup
+  has-unit-has-right-unit-Quasigroup (e , _ , e-right-unit) = e , e-right-unit
+
+  has-unit-has-left-and-right-units-Quasigroup :
```
fredrik-bakke, line 173: The conclusion of this theorem is that it has left and right units, so that should be the first thing that is mentioned in the name. I.e., this should be called `has-left-and-right-units-has-unit-Quasigroup`

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+  has-unit-has-left-and-right-units-Quasigroup e =
+    (has-unit-has-left-unit-Quasigroup e) ,
+    (has-unit-has-right-unit-Quasigroup e)
+
+  left-and-right-units-agree-Quasigroup :
+    (e : has-left-unit-Quasigroup) → (f : has-right-unit-Quasigroup) →
+    term-has-left-unit-Quasigroup e ＝ term-has-right-unit-Quasigroup f
+  left-and-right-units-agree-Quasigroup (e , e-left-unit) (f , f-right-unit) =
+    equational-reasoning
+    e ＝ (e * f) r/ f
```
fredrik-bakke, line 184:
```suggestion
    e
    ＝ (e * f) r/ f
```

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+      by is-right-cancellative-right-div-Quasigroup Q f e
+    ＝ f r/ f
+      by ap (λ x → x r/ f) (e-left-unit f)
+    ＝ (f * f) r/ f
+      by inv (ap (λ x → x r/ f) (f-right-unit f))
+    ＝ f
+      by inv (is-right-cancellative-right-div-Quasigroup Q f f)
+
+  has-left-and-right-units-is-left-unit-is-right-unit :
```
fredrik-bakke, line 193: `is-right-unit-has-left-unit-Quasigroup` may be a good name for this one.

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+    (e : has-left-unit-Quasigroup) → has-right-unit-Quasigroup →
+    is-right-unit-Quasigroup (term-has-left-unit-Quasigroup e)
+  has-left-and-right-units-is-left-unit-is-right-unit e f x =
+    inv-tr is-right-unit-Quasigroup (left-and-right-units-agree-Quasigroup e f)
+    (pr2 f) x
+
+  has-left-and-right-units-has-unit-Quasigroup :
```
fredrik-bakke, line 200: `has-unit-has-right-unit-has-left-unit-Quasigroup`

Notice the order

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  is-left-unit-Quasigroup-Prop : (e : type-Quasigroup Q) → Prop l
+  is-left-unit-Quasigroup-Prop e =
+    is-left-unit-Quasigroup e , is-prop-is-left-unit-Quasigroup e
+
+  has-left-unit-Quasigroup : UU l
+  has-left-unit-Quasigroup =
+    Σ (type-Quasigroup Q) λ e → is-left-unit-Quasigroup e
+
+  term-has-left-unit-Quasigroup : has-left-unit-Quasigroup → type-Quasigroup Q
```
fredrik-bakke, line 79: We prefer the word `element` over `term`. This projection should therefore be called `element-has-left-unit-Quasigroup`

src/quasigroups/loops.lagda.md
```diff
@@ -0,0 +1,220 @@
...
+
+  is-right-unit-Quasigroup-Prop : (e : type-Quasigroup Q) → Prop l
+  is-right-unit-Quasigroup-Prop e =
+    is-right-unit-Quasigroup e , is-prop-is-right-unit-Quasigroup e
+
+  has-right-unit-Quasigroup : UU l
+  has-right-unit-Quasigroup =
+    Σ (type-Quasigroup Q) λ e → is-right-unit-Quasigroup e
+
+  term-has-right-unit-Quasigroup : has-right-unit-Quasigroup → type-Quasigroup Q
```
fredrik-bakke, line 124: We prefer the term `element` over `term`. This projection should therefore be called `element-has-right-unit-Quasigroup`

src/quasigroups/quasigroups.lagda.md
djspacewhale: I changed the notation from `/ \` to `r/ l/` as the standard forward-slash is apparently reserved for lambda expressions. I see later down you suggest introducing a different Unicode representative so I think this can broadly pend that.

src/quasigroups.lagda.md
djspacewhale: Truthfully I'm approaching this formalization of quasigroups as a way to learn about them, so I'm not an expert on the broad theory. Including them in the `group-theory` namespace could make sense - groups are simply unital associative quasigroups, anyways - but the applications seem rather different; say, finite quasigroups classically correspond to (isotopy types of) latin squares, and their representation theory is much subtler. 

I don't want to step on toes too aggressively being a new contributor but the current namespace convention is a tad frustrating for me. Finding definitions is easier using the website and its search bar, but searching monolithic namespaces by hand is not as elegant as, say, using a hierarchy of namespaces a la cubical. I did think about creating a subnamespace `group-theory/quasigroups` but as this convention does not exist currently I held off. Has this (re)organization been discussed before?

EgbertRijke: @djspacewhale, that is totally fine. I don't think you are stepping on anyone's toes, and your contribution is most welcome! Learning and formalizing at the same time is a great strategy to really get to the bottom of a subject, I use that strategy too. To that end, would you be willing to cite standard references for you definitions, such as

Pflugfelder. Quasigroups and Loops: Introduction?

We have a `references.bib` file in the root folder and there are guidelines about how to include a bibliography at the end of a file. The situation could be improved by now, but it used to be the case that the definition of quasigroups and semigroups was wrong on wikipedia, so it is important to follow a standard reference.

I also think developing quasigroups in the group theory folder could make sense, because some definitions and results about quasigroups will be more general and it would be a natural setting for comparing the two developments. On the other hand, I do understand why you initially chose to create a setup in a separate namespace: The nonassociative case is actually quite different and the `group-theory` name space is quite large, containing many different concepts.

We are aware of the organizational conventions of the cubical library, and other libraries as well, and we have deliberately chosen our current library organization. I personally find other library organizations a bit chaotic, and the one-concept-per-file principle is the cleanest in my view. It is also good to keep in mind that the choices for a library organization are tied to the purpose of the library: In cubical the goal is to facilitate formalization related to current research, which agda-unimath also hopes to facilitate, but agda-unimath has an important secondary purpose, which is to develop an informative resource where ultimately the library webpage could serve as a nlab-style website where you can find precise definitions and infomation about how concepts are interrelated. From that point of view, an organization with `concept.base` and `concept.properties` modules where `concept.properties` becomes a dumping place for all sorts of properties is not helping us at all, and what would be in `concept.base` in cubical could often be spread out over several pages in agda-unimath, each its own concept.

I hope this explanation helps!

djspacewhale: I actually don't have a reference text and have been formalizing roughly what's on Wikipedia and the nlab, cross-referenced with arxiv papers that arise in searches. I'll look for that text though. It's of course not the morally healthiest approach, and I'm no longer a student so sourcing texts may be a bit harder, but I'll find a better primary source!

I also appreciate agda-unimath's goal of being a readable informative source and this knowledge-for-the-people element is one reason I chose to contribute here rather than, say, cubical (long-term I'm also interested in modal type theory and applications to synthetic math, a project that appears at technical odds with cubical). My thought was less around, say, having one `groups.base` defining everything to do with groups, but more around using subfolders to consolidate "subareas" of knowledge, say, putting abelian stuff in `group-theory/abelian-groups/` and so on. That said, I'll respect how the library is currently organized, and Fredrik's notes about separating e.g. left/right quasigroups into their own files is a good note for following the one-concept-per-file principle that I'll incorporate.

.gitignore
```diff
@@ -432,3 +432,6 @@ MAINTAINERS.md
 /website/css/Agda-highlight.css
 /website/images/agda_dependency_graph.svg
 /website/images/agda_dependency_graph_legend.html
+
+# vscode settings
+settings.json
```
VojtechStep, line 437: The library is distributed with recommended VSCode settings, so they are intentionally not ignored

src/foundation/decidable-equality.lagda.md
```diff
@@ -217,6 +218,24 @@ module _
         ( λ x y → eq-Eq-has-decidable-equality d)
 ```
 
+### The [decidable proposition](foundation-core.decidable-propositions.md) of equality in a type with decidable equality
```
VojtechStep: FYI headings get turned into links in the HTML backend, so they may not have links inside. If you want to add links, you can add a paragraph of prose

VojtechStep (commented):
I'm not reviewing this yet, those are just drive-by comments

src/foundation/decidable-equality.lagda.md
djspacewhale: Heard and edited.

PR #1341, opened 2025-02-22
Documentation for do syntax for propositional truncations

lowasser:
As requested in #1306 , starting a new PR for working through the docs.

### DIFF
--- a/src/foundation/propositional-truncations.lagda.md
+++ b/src/foundation/propositional-truncations.lagda.md
@@ -448,9 +448,11 @@ module _
 
 ## `do` syntax for propositional truncation { #do-syntax }
 
-Consider a case where you are trying to prove a proposition, `motive : Prop l`,
-from witnesses of propositional truncations of types `P` and `Q`:
-
+[Agda's `do` syntax](https://agda.readthedocs.io/en/v2.7.0/language/syntactic-sugar.html#do-notation)
+is a handy tool to avoid deeply nesting calls to the same lambda-based function.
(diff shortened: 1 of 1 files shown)


lowasser:
I've learned more about this since February, and maybe this is a better shot?

src/foundation/propositional-truncations.lagda.md
```diff
@@ -469,43 +469,44 @@ module _
...
   p ← witness-truncated-prop-P
   q ← witness-truncated-prop-Q p
-  witness-motive-P-Q p q
-where open do-syntax-trunc-Prop motive
+  motive-P-Q p q
 ```
 
+We can read the line `p ← witness-truncated-prop-P` as "given
+`witness-truncated-prop-P : type-trunc-Prop P`, assume an element `p : P`," and
+we can then use `p` freely on further lines in the `do`. The final line in the
```
fredrik-bakke:
```suggestion
we can then use `p` freely on further lines in the `do` clause. The final line in the
```
or
```suggestion
we can then use `p` freely on further lines in the `do` block. The final line in the
```

src/foundation/propositional-truncations.lagda.md
```diff
@@ -469,43 +469,44 @@ module _
...
+`do` must be a value of `type-Prop motive`.
```
fredrik-bakke:
```suggestion
`do` clause must be a value of `type-Prop motive`.
```
or
```suggestion
`do` block must be a value of `type-Prop motive`.
```

PR #1339, opened 2025-02-21
Transposing addition and subtraction through rational inequalities

lowasser:
I generalized at least to posets in general, but not to strict inequality.  Another small PR that I nevertheless think might require significant feedback (did I name things well?)

PR #1336, opened 2025-02-19
Addition on real numbers

lowasser:
Decided I'm just going to fold it all together.  Depends on #1306.

Issue #1334, opened 2025-02-18
Large groups


lowasser:
Real addition will give us our first large group, but we have to define what that looks like.

The biggest issue I see is the inverse laws: if `x : type-Large-Group G l`, and `neg-Large-Group G x : type-Large-Group G l`, then we have to express that `mul-Large-Group G x (neg-Large-Group G x)` is the identity, but we have a few options:

1. An identity exists at every universe level.
2. The group has to come equipped with a mechanism for raising any element's universe level.
3. We accept some notion of similarity, and bake that into the group.  As a corollary, we need to show that similarity is preserved by the group operation.

I suppose 2 comes with a free 3, but nevertheless, we should probably figure out what this should look like.

EgbertRijke:
I'd keep the definition as simple as possible. A large group is a [large semigroup](https://unimath.github.io/agda-unimath/group-theory.large-semigroups.html) equipped with a unit element at `lzero` and an inverse function that preserves the level, satisfying the usual axioms of a group.

All other stuff is reasonable to put on as extra structure, but you can define structured large group for when you incorporate those.

lowasser:
What axiom is it that the inverse function satisfies?

It can't be that e.g. `mul-Large-Group G x (inv-Large-Group G x) = unit-Large-Group G`, because the left and right sides of that equation are at different universe levels.  That's the problem I'm getting at.

lowasser:
(My suggestions above are all attempts at "what is the simplest way to resolve this issue," which are less simple than I'd like, but I don't yet have better ideas.)

lowasser:
Drilling down to a particular example: you can show `{l : Level} → (x : ℝ l) → sim-ℝ (add-ℝ x (neg-ℝ x)) zero-ℝ`, but they're not equal, because `add-ℝ x (neg-ℝ x) : ℝ l` and `zero-ℝ : ℝ lzero`.  So you either have to bake a notion of similarity into the definition of the group, or...one of the other alternatives, having an identity at every universe level, or a notion of raising universe levels, or something I'm not seeing.

EgbertRijke:
That's a very good point!

One could introduce large monoids first, have a predicate on the elements of large monoids that they are multiplicative units, and then states that xx^-1 is a multiplicative unit. 

This would avoid having to introduce level raising operations and sameness relations across levels.

fredrik-bakke:
> That's a very good point!
> 
[...]

PR #1342, opened 2025-02-22
Addition for lower and upper Dedekind reals

lowasser:
Depends on #1339, but adds ~500LOC beyond that, and seems like a reasonable sub-PR to break away from #1336 .


src/real-numbers/addition-lower-dedekind-real-numbers.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+  semigroup-add-lower-ℝ-lzero ,
+  lower-real-ℚ zero-ℚ ,
+  left-unit-law-add-lower-ℝ ,
+  right-unit-law-add-lower-ℝ
+
+commutative-monoid-add-lower-ℝ-lzero :
+  Commutative-Monoid (lsuc lzero)
+commutative-monoid-add-lower-ℝ-lzero =
+  monoid-lower-ℝ-lzero ,
+  commutative-add-lower-ℝ
```
fredrik-bakke, line 278: Couldn't you just as well make these parametric?

```suggestion
semigroup-add-lower-ℝ : (l : Level) → Semigroup (lsuc l)
semigroup-add-lower-ℝ l =
  (lower-ℝ l , is-set-lower-ℝ l) ,
  add-lower-ℝ ,
  associative-add-lower-ℝ

monoid-lower-ℝ : (l : Level) → Monoid (lsuc l)
monoid-lower-ℝ l =
  semigroup-add-lower-ℝ l ,
  lower-real-ℚ zero-ℚ ,
  left-unit-law-add-lower-ℝ ,
  right-unit-law-add-lower-ℝ

commutative-monoid-add-lower-ℝ :
  (l : Level) → Commutative-Monoid (lsuc l)
commutative-monoid-add-lower-ℝ l =
  monoid-lower-ℝ l ,
  commutative-add-lower-ℝ
```

src/real-numbers/addition-upper-dedekind-real-numbers.lagda.md
```diff
@@ -0,0 +1,268 @@
...
+  semigroup-add-upper-ℝ-lzero ,
+  upper-real-ℚ zero-ℚ ,
+  left-unit-law-add-upper-ℝ ,
+  right-unit-law-add-upper-ℝ
+
+commutative-monoid-add-upper-ℝ-lzero :
+  Commutative-Monoid (lsuc lzero)
+commutative-monoid-add-upper-ℝ-lzero =
+  monoid-upper-ℝ-lzero ,
+  commutative-add-upper-ℝ
```
fredrik-bakke, line 267:
```suggestion
semigroup-add-upper-ℝ : (l : Level) → Semigroup (lsuc l)
semigroup-add-upper-ℝ l =
  (upper-ℝ l , is-set-upper-ℝ l) ,
  add-upper-ℝ ,
  associative-add-upper-ℝ
monoid-upper-ℝ : (l : Level) → Monoid (lsuc l)
monoid-upper-ℝ l =
  semigroup-add-upper-ℝ l ,
  upper-real-ℚ zero-ℚ ,
  left-unit-law-add-upper-ℝ ,
  right-unit-law-add-upper-ℝ
commutative-monoid-add-upper-ℝ :
  (l : Level) → Commutative-Monoid (lsuc l)
commutative-monoid-add-upper-ℝ l =
  monoid-upper-ℝ l ,
  commutative-add-upper-ℝ
```

fredrik-bakke (approved):
Looks good to go aside from the comment on *polymorphism. 👍

src/real-numbers/addition-lower-dedekind-real-numbers.lagda.md
lowasser: Nope.  The identity has to be at the same universe level.

lowasser:
We can either be forced to generalize the additive identity to every universe level, or stick with this as-is.  (This comes back to the problems associated with #1334 .)

We can do that, if we want.  Setting up an additive inverse for every level at least lets us use small groups on every universe level.  Is that your preference?

fredrik-bakke:
Ahh, shoot, that's a bummer. Then I think it is better to leave it as is for now. We shouldn't formalize "raise" operations on real numbers just to get this half-way polymorphism. If we really need polymorphism then the appropriate way to go would be to find the correct definition of a "large monoid" and so on, and since that is an open question for the time being, let's keep it like it is.

lowasser:
FWIW, large monoids are probably reasonably easy; it's groups where it gets really hard.  We can have those if you like.

PR #1344, opened 2025-02-23
Disjoint subtypes

lowasser:
Mostly just a cleanup/refactoring, not part of a larger whole, so it goes in its own PR even if it's small.


VojtechStep:
LGTM, thanks 👌

PR #1345, opened 2025-02-23
Reciprocals of nonzero natural numbers

lowasser:
Many related properties along with it.  Hopefully this will help us do things like simply take half of a rational number and the like.

PR #1347, opened 2025-02-23
Cauchy sequences in metric spaces

lowasser:
Gives us the ability to go back and forth between Cauchy sequences and Cauchy approximations, and their limits.

Naturally, this is going to be necessary for a lot of analysis built on series and sequences.

Depends on #1345.


src/elementary-number-theory/nonzero-natural-numbers.lagda.md
```diff
@@ -86,6 +93,24 @@ pr1 (succ-nonzero-ℕ' n) = succ-ℕ n
...
+pred-nonzero-ℕ : nonzero-ℕ → ℕ
+pred-nonzero-ℕ (succ-ℕ n , _) = n
+pred-nonzero-ℕ (zero-ℕ , H) = ex-falso (H refl)
+
+pred-ℕ⁺ : nonzero-ℕ → ℕ
+pred-ℕ⁺ = pred-nonzero-ℕ
+
+is-section-succ-nonzero-ℕ' : is-section succ-nonzero-ℕ' pred-nonzero-ℕ
+is-section-succ-nonzero-ℕ' (zero-ℕ , H) = ex-falso (H refl)
+is-section-succ-nonzero-ℕ' n⁺@(succ-ℕ n , _) = eq-nonzero-ℕ refl
```
fredrik-bakke, line 108:
```suggestion
is-section-succ-nonzero-ℕ' (succ-ℕ n , _) = eq-nonzero-ℕ refl
```

src/elementary-number-theory/nonzero-natural-numbers.lagda.md
```diff
@@ -57,6 +61,9 @@ nat-ℕ⁺ = nat-nonzero-ℕ
 
 is-nonzero-nat-nonzero-ℕ : (n : nonzero-ℕ) → is-nonzero-ℕ (nat-nonzero-ℕ n)
 is-nonzero-nat-nonzero-ℕ = pr2
+
+eq-nonzero-ℕ : {m n : nonzero-ℕ} → nat-nonzero-ℕ m ＝ nat-nonzero-ℕ n → m ＝ n
+eq-nonzero-ℕ m=n = eq-pair-Σ m=n (eq-is-prop is-prop-nonequal)
```
fredrik-bakke, line 66: This could also be called `is-injective-nat-nonzero-ℕ`

src/metric-spaces/cauchy-sequences-complete-metric-spaces.lagda.md
```diff
@@ -0,0 +1,126 @@
...
+open import metric-spaces.metric-spaces
+```
+
+</details>
+
+## Idea
+
+A [Cauchy sequence](metric-spaces.cauchy-sequences-metric-spaces.md) in a
+[complete metric space](metric-spaces.complete-metric-spaces.md) is a Cauchy
+sequence in the underlying metric space, and always has a limit.
```
fredrik-bakke, line 29:
```suggestion
A [Cauchy sequence](metric-spaces.cauchy-sequences-metric-spaces.md) in a
[complete metric space](metric-spaces.complete-metric-spaces.md) is a Cauchy
sequence in the underlying [metric space](metric-spaces.metric-spaces.md).
Cauchy sequences in complete metric spaces always have a limit.
```

src/metric-spaces/cauchy-sequences-metric-spaces.lagda.md
```diff
@@ -0,0 +1,527 @@
...
+## Idea
+
+A
+{{#concept "Cauchy sequence" Disambiguation="in a metric space" Agda=is-cauchy-approximation-Metric-Space}}
+`x` in a [metric space](metric-spaces.metric-spaces.md) is a mapping from the
+[natural numbers](elementary-number-theory.natural-numbers.md) to the underlying
+type of the metric space such that for any
+[positive rational](elementary-number-theory.positive-rational-numbers.md) `ε`,
+there is a concrete `n : ℕ` such that for any `m, k ≥ n`, `x m` and `x k` are in
+a neighborhood of `ε` of each other.
```
fredrik-bakke, line 49:
```suggestion
an `ε`-neighborhood of each other.
```

src/metric-spaces/cauchy-sequences-metric-spaces.lagda.md
```diff
@@ -0,0 +1,527 @@
...
+              ( x k)
+              ( ε'⁺)
+              ( ε'⁺)
+              ( is-symmetric-structure-Metric-Space
+                ( M)
+                ( ε'⁺)
+                ( x k)
+                ( lim)
+                ( n≤m⇒|xm-l|<ε' k n≤k))
+              ( n≤m⇒|xm-l|<ε' m n≤m))
```
fredrik-bakke, line 211:
```suggestion
        λ m k n≤m n≤k →
        is-monotonic-structure-Metric-Space
          ( M)
          ( x m)
          ( x k)
          ( ε'⁺ +ℚ⁺ ε'⁺)
          ( ε⁺)
          ( 2ε'<ε)
          ( is-triangular-structure-Metric-Space
            ( M)
            ( x m)
            ( lim)
            ( x k)
            ( ε'⁺)
            ( ε'⁺)
            ( is-symmetric-structure-Metric-Space
              ( M)
              ( ε'⁺)
              ( x k)
              ( lim)
              ( n≤m⇒|xm-l|<ε' k n≤k))
            ( n≤m⇒|xm-l|<ε' m n≤m))
```

src/metric-spaces/cauchy-sequences-metric-spaces.lagda.md
```diff
@@ -0,0 +1,527 @@
...
+                ( δ⁺)
+                ( ε⁺ +ℚ⁺ δ⁺)
+                ( le-left-add-rational-ℚ⁺ δ ε⁺)
+                ( pr2
+                  ( is-cauchy-sequence-cauchy-sequence-Metric-Space M x δ⁺)
+                  ( mε)
+                  ( mδ)
+                  ( mδ≤mε)
+                  ( refl-leq-ℕ mδ)))
+            ( linear-leq-ℕ mε mδ)
```
fredrik-bakke, line 272: Does Agda complain if you do this?

```suggestion
      let
        mε = modulus-of-convergence-cauchy-sequence-Metric-Space M x ε⁺
        mδ = modulus-of-convergence-cauchy-sequence-Metric-Space M x δ⁺
        xmε = map-cauchy-sequence-Metric-Space M x mε
        xmδ = map-cauchy-sequence-Metric-Space M x mδ
      in
        rec-coproduct
          ( λ mε≤mδ →
            is-monotonic-structure-Metric-Space
              ( M)
              ( xmε)
              ( xmδ)
              ( ε⁺)
              ( ε⁺ +ℚ⁺ δ⁺)
              ( le-right-add-rational-ℚ⁺ ε δ⁺)
              ( pr2
                ( is-cauchy-sequence-cauchy-sequence-Metric-Space M x ε⁺)
                ( mε)
                ( mδ)
                ( refl-leq-ℕ mε)
                ( mε≤mδ)))
          ( λ mδ≤mε →
            is-monotonic-structure-Metric-Space
              ( M)
              ( xmε)
              ( xmδ)
              ( δ⁺)
              ( ε⁺ +ℚ⁺ δ⁺)
              ( le-left-add-rational-ℚ⁺ δ ε⁺)
              ( pr2
                ( is-cauchy-sequence-cauchy-sequence-Metric-Space M x δ⁺)
                ( mε)
                ( mδ)
                ( mδ≤mε)
                ( refl-leq-ℕ mδ)))
          ( linear-leq-ℕ mε mδ)
```

fredrik-bakke (commented):
I'm not a fan of the prevalent use of `let` bindings here. One thing is to use them for the monad syntax we have been discussing, but here you're using them to formalize lemmas and to create aliases.

src/metric-spaces/cauchy-sequences-metric-spaces.lagda.md
```diff
@@ -0,0 +1,527 @@
...
+
+  abstract
+    is-cauchy-sequence-cauchy-sequence-cauchy-approximation-Metric-Space :
+      is-cauchy-sequence-Metric-Space
+        ( M)
+        ( map-cauchy-sequence-cauchy-approximation-Metric-Space)
+    is-cauchy-sequence-cauchy-sequence-cauchy-approximation-Metric-Space
+      ε⁺@(ε , _) =
+      let
+        (ε'⁺@(ε' , _) , 2ε<ε) = bound-double-le-ℚ⁺ ε⁺
```
lowasser, line 379: should be `2ε'` here

lowasser:
I guess my understanding of the *objective* of the stylistic practice here is to encourage pulling out reusable lemmas and functions rather than reinventing logic repeatedly.  I see and buy that, though as I work in this project I am becoming less convinced over time that that scales all that well.  Many individual steps to proving theorems feel so narrow and specific that generating even a name for them as a lemma feels oppressively difficult.  You can try to make them more general and get a better name as a result, but the work to apply the general case to the specific need often feels like as least as much work as the effort to prove the specific case from scratch without the lemma.

But separately, many of the the aliases in this PR give an easier-to-follow, shorter name to what would otherwise be a long function call -- `let xm = map-cauchy-sequence-cauchy-approximation-Metric-Space M x m`, for example, a case in which there aren't any lemmas or functions to pull out because it's already maximally reduced.  Such aliases make the logic easier for me, at least, to follow and (especially) debug; they make the proof more closely resemble anything I'd write as a "traditional" proof in English.  The case of `let (ε'⁺@(ε' , _) , 2ε'<ε) = bound-double-le-ℚ⁺ ε⁺` could be done as `ind` or `rec` on the dependent pair type, but -- as with the `do` syntax we've discussed and that I think I've successfully convinced you is a positive development -- that'd result in gratuitously deep nesting that additionally separates the newly defined variables visually from their semantic relationship to `ε⁺` (since `ind-Sigma` and `rec-Sigma` put the dependent pair after the entire lambda which uses the components).

I can find more steps to combine into more general lemmas and functions, and I see some examples in this PR where that shouldn't be too difficult.  Combining `map-cauchy-sequence-Metric-Space` and `modulus-of-convergence-cauchy-sequence-Metric-Space` is a good example, though I'd have to think a bit about what to name that new function.  But for many of the more alias-style usages in the `let`s, inlining feels as though it accomplishes nothing more than making the result harder to comprehend.  I certainly claim that I could not have figured out how to write this without at least using the `let` versions as an intermediate step before inlining things.

EgbertRijke:
I understand your sentiment, and I have seen from working with students that sensibly breaking down code into reusable lemmas is one of the hardest things for newcomers to learn. 

On the other hand, it is a skill that is possible to learn, and I don't really believe in your scaling argument. In fact, I think the opposite is true. There are parts of the library that have been coded in less refactorable style, and in practice it has just been about impossible to do anything with it. Refactoring that code is an insurmountable task, probably worse that just formalizing it from scratch all anew. As a results, these parts of the code have been sitting in the library with low maintenance. Furthermore, these are often the parts of the library that take up the most time of the type checker. So I am a firm believer in breaking down everything into small, sensible, reusable lemmas.

On the other hand, if you have the sense that it hampers your progress, I'd rather have that you formalize in your preferred style. According to the rising sea model, eventually the code written in small reusable lemmas will catch up and these parts of the library will also be nicely refactored.

lowasser:
I am willing to go ahead and trust that experience for the lemmas aspect; that comment has successfully moved me there. 
 I'm not sure I'm clear to what extent what you said just applies to the aliasing case, where I don't see a way to break out _any_ lemma from e.g. `let xm = map-cauchy-sequence-cauchy-approximation-Metric-Space M x m`.

EgbertRijke:
I think it is fine. I use `let` expressions too on rare occasions.

I also thought that I should add that I don't mean to say that the code you write, or any code from anyone that gets rewritten including my own, is bad code. It's normal that things get rewritten, and it's normal that after having formalized something once, later a better/new interesting approach comes along. We're not setting our formalization in stone, it changes all the time.

src/metric-spaces/cauchy-sequences-metric-spaces.lagda.md
```diff
@@ -0,0 +1,578 @@
...
+                  ( mε≤mδ)))
+            ( λ mδ≤mε →
+              is-monotonic-structure-Metric-Space
+                ( M)
+                ( xmε)
+                ( xmδ)
+                ( δ⁺)
+                ( ε⁺ +ℚ⁺ δ⁺)
+                ( le-left-add-rational-ℚ⁺ δ ε⁺)
+                ( is-symmetric-structure-Metric-Space
```
lowasser, line 304: Calling this out as an example of where I'm finding pulling out lemmas painful.

In response to the earlier feedback, I added a lemma for when you're at the modulus of convergence exactly, and applied it here.  But I haven't managed to eliminate any `let` aliases, and the amount of work to apply one change to that lemma -- swapping the roles of the points in the metric space -- exceeds the amount of work it took to prove the lemma in the first place; the original expression here was

```
pr2
  ( is-cauchy-sequence-cauchy-sequence-Metric-Space M x δ⁺)
  ( mε)
  ( mδ)
  ( mδ≤mε)
  ( refl-leq-ℕ mδ)))
```

and now we have to apply two functions with very long names to mostly repeated arguments.

Other common cases of this issue are when I need to reassociate and commute expressions from the general case.

Could I add another version of that lemma that applies symmetry built-in?  Sure.  But at that point, we're building alternate versions of each lemma just for every variation on the use case, and I'm not sure that's better either.

I'm going to leave it like this, because I trust your judgement that this sort of thing is for the best long term, but I did at least want to give a specific example where this doesn't feel great to me.

fredrik-bakke:
It seems that a conclusion was reached on the use of `let` bindings here, is that correct? If so, we can move ahead and merge this PR I believe

lowasser:
I believe I've eliminated all the `let` bindings here that seem reasonable: what remains are aliases for concision and pair decompositions.

PR #1348, opened 2025-02-24
An auto-updating dependency graph of the library

fredrik-bakke:
Adds an auto-updating dependency graph of the library á la what I shared on Discord to the art page.


.github/workflows/pages.yaml
```diff
@@ -92,6 +92,12 @@ jobs:
 
       - run: python3 -m pip install -r master/scripts/requirements.txt
 
+      - name: Generate dependency graph rendering
+        run:
+          python3 master/scripts/generate_dependency_graph_rendering.py
```
VojtechStep, line 97: Could this be a Makefile rule? It's best to have as little bespoke logic in CI pipelines as possible.

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+
+if __name__ == "__main__":
+    repo = "unimath/agda-unimath"
+    root_dir = "src"
+    min_rank_imports = 20
+
+    parser = argparse.ArgumentParser(description="Generate Agda dependency graph.")
+    parser.add_argument("output_file", type=str, help="Path to save the output graph.")
+    parser.add_argument("format", type=str, choices=["svg", "png", "pdf"], help="Output format of the graph.")
+    parser.add_argument("--min_rank_imports", type=int, default=20, help="Number of top imported files to exclude from the graph.")
```
VojtechStep, line 118: Is this argument used? I don't know how Python's `argparse` works, but I'd assume that it doesn't change the local `min_rank_imports`

EgbertRijke:
Looks good!

I'd just put newer art on top, so that the first one is always the newest one.

EgbertRijke:
Since this is auto-updating, it is relevant how much time it takes to generate this graph. Is it reasonably fast?

fredrik-bakke:
It takes max 10 seconds

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+
+def count_imports(graph):
+    """Count the number of times each module is imported."""
+    import_counts = defaultdict(int)
+    for dependencies in graph.values():
+        for dep in dependencies:
+            import_counts[dep] += 1
+    return import_counts
+
+def build_dependency_graph(root_dir, min_rank_node=20):
```
VojtechStep, line 44: I don't see how `min_rank_node` is an appropriate name for this argument

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+from utils import find_agda_files, parse_agda_imports, count_lines_in_file
+import argparse
+
+primes_hash_str = (19,2,1,7,5,3,11,13,17)
+def hash_str(s) :
+    return sum(ord(char) * primes_hash_str[i%len(primes_hash_str)] for i,char in enumerate(s))
+
+def fetch_github_labels(repo):
+    """Fetch labels and their colors from a GitHub repository."""
+    url = f"https://api.github.com/repos/{repo}/labels"
```
VojtechStep, line 16: I'd really prefer if the website build didn't have to fetch this data. Even if it means the label colors won't be in sync, but we don't change the labels much, so the colors can be static

EgbertRijke:
I'm happy with this PR in principle, but I can't comment on the code. Perhaps @VojtechStep wants to have a look. Otherwise I'm fine with merging.

scripts/utils/__init__.py
```diff
@@ -242,3 +242,34 @@ def is_file_modified(file_path):
         return False
     except subprocess.CalledProcessError:
         return True
+
+def find_agda_files(root_dir: str) -> List[str]:
```
VojtechStep, line 246: What are you gaining here by not using the functions to list git files and filter agda files, which already exist?

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+                # dep_color = node_colors[dep]
+                # edge_color =  average_color(node_color, dep_color) + "10"
+                edge_color =  node_color + "10"
+                dot.edge(node, dep, color=edge_color, arrowhead="none")
+
+    dot.render(output_file, format=format, cleanup=True)
+    print(f"Graph saved as {output_file}.{format}")
+
+if __name__ == "__main__":
+    repo = "unimath/agda-unimath"
```
VojtechStep, line 111: The canonical spelling would be `UniMath`, and it's exported as a constant from our `utils` module

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+    # Remove top modules from the graph
+    for module in top_imports:
+        if module in graph:
+            del graph[module]
+        for dependencies in graph.values():
+            dependencies.discard(module)
+
+    return graph, file_sizes
+
+def average_color(color1, color2):
```
VojtechStep, line 78: I don't see this function used anywhere, but that might be a result of me reviewing the PR on my phone

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,124 @@
...
+    max_lines = max(file_sizes.values(), default=1)
+    node_sizes = {node: max(0.05, 0.3 * math.sqrt(file_sizes.get(node, 0) / max_lines)) for node in graph}
+    node_colors = {node: module_based_color(node[:node.rfind(".")], label_colors) for node in graph}
+
+    for node, dependencies in graph.items():
+        node_color = node_colors[node]
+        dot.node(node, shape="circle", style="filled", fillcolor=node_color, color="#00000000", width=str(node_sizes[node]), height=str(node_sizes[node]), label="")
+        for dep in dependencies:
+            if dep in graph:  # Ensure we're not linking to removed nodes
+                # dep_color = node_colors[dep]
```
VojtechStep, line 102: Please remove commented code

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: Ah, no, I meant to remove that

.github/workflows/pages.yaml
fredrik-bakke: sure, I can do that. Though, I don't know how to "silently fail" the script in the makefile, and I don't want our CI to fail if this script breaks at some point

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: why not? I thought this would be nice since it is 0 maintenance. How about fetching the data, but falling back to preloaded values?

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: it is not used if I remove the commented code, which I thought about keeping in case I wanted to make stylistic changes. But I'm fine with removing it

scripts/utils/__init__.py
fredrik-bakke: oh, I had forgotten about the other one, and this one skips the main module files

VojtechStep: I brought this up because I'd like to avoid inconsistent behavior in scripts, and adding subtly differently behaving functions to this "miscellaneous" modules just makes finding the right functions to reuse that much harder. Currently everything uses git tracked files, except for `remove_unused_imports`, which rolls its own `get_agda_files` and I'm not sure why. You can implement this `find_agda_files` by composing `get_git_tracked_files`, `get_agda_files` and filtering out files with only one slash in their path (the paths have the `src/` prefix).

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: Turns out the previous code had a bug where it didn't fetch all the labels, so thanks for making me detect that!

scripts/utils/__init__.py
fredrik-bakke: I refactored `find_agda_files` to build on the other methods, and moved it back to `generate_dependency_graph_rendering.py`. Is that a reasonable solution for you?

.github/workflows/pages.yaml
fredrik-bakke: figured it out I think

fredrik-bakke:
I've tried incorporating your feedback @VojtechStep, let me know if issues persist

scripts/generate_dependency_graph_rendering.py
```diff
@@ -0,0 +1,135 @@
...
+    agda_files = get_agda_files(git_tracked_files)
+    return [str(f) for f in agda_files if f.parts[0] == root_dir and len(f.parts) > 2]
+
+
+fallback_label_colors = {'category-theory': 'fbca04', 'commutative-algebra': '3577BB','domain-theory': 'FCBFF5', 'elementary-number-theory': '4A6123', 'finite-algebra': '524F88', 'finite-group-theory': '9A01E2', 'foundation': '284530', 'globular-types': '4A7555', 'graph-theory': 'CCC0C4', 'group-theory': '963872', 'higher-group-theory': 'E79FEE', 'lists': 'C9D999', 'literature': '3C3C0A', 'logic': 'A9CFCE', 'metric-spaces': '923E82', 'modal-logic': 'A4DC70', 'modal-type-theory': 'E45F85', 'oeis': 'F984EC', 'order-theory': '533A22', 'organic-chemistry': '07C851', 'orthogonal-factorization-systems': '578072', 'real-numbers': 'B66877', 'reflection': '000000', 'ring-theory': '1A1277', 'set-theory': '98335C', 'simplicial-type-theory': '8C3F69', 'species': 'EDA55F', 'structured-types': '069F6E', 'synthetic-category-theory': '2FEABE', 'synthetic-homotopy-theory': 'B0D45A', 'trees': '0F1D69', 'type-theories': '610CCA', 'univalent-combinatorics': 'F70D61', 'universal-algebra': '5467C3', 'wild-category-theory': 'E2C12E'}
```
VojtechStep, line 24: Please format this constant and capitalize the name, so that it's clear that it's a constant

scripts/generate_dependency_graph_rendering.py
VojtechStep: You resolved the discussion but there's no change?

VojtechStep (commented):
One formatting edit and it should be good to go 👌

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: it's the minimum rank among "top indegree nodes", I'll elaborate the name a little further

fredrik-bakke: Sorry, the github pullrequests plugin on vscode is a little buggy. I must've closed this one thinking I had already resolved it.

fredrik-bakke:
done

VojtechStep:
My problem with `min_rank_nodes` wasn't `nodes`, but `min` and `rank` 😅. To me, the outcome of passing an argument `min_rank_imports=20` should mean "only count modules imported >=20 times". Instead it drops the 20 most imported modules, so `min` means "minimum to drop", and `rank` means "place in order by number of indegrees", which to me is unintuitive. For the current functionality I propose `most_imported_drop_count`. If the script instead put a hard cap on how many times a file can be included to make it into the graph (which might be more useful, since what you're really trying to avoid is a messy graph?), then I'd call it something like `max_imported_by`

fredrik-bakke:
Oh, I'm sorry. I'll change it in a follow-up

fredrik-bakke:
My reading is: if you rank the nodes by the number of imports, then the minimum rank a node is allowed to have is 20. But I'm happy to change it

PR #1350, opened 2025-02-25
fix: Art/graph

fredrik-bakke:
Forgot to add a dependency for the graphviz backend


Makefile
```diff
@@ -187,6 +187,7 @@ install-website-dev:
 	@cargo install mdbook-katex@0.5.7
 	@cargo install mdbook-pagetoc@0.1.7
 	@cargo install mdbook-catppuccin@1.2.0
+	@sudo apt-get install -y graphviz
```
VojtechStep, line 190: This should go into the dependency list in `HOWTO_INSTALL.md` instead, because system package managers don't have a unified interface. This won't work for me (on NixOS), it won't work for the people on Arch, and I don't think it even works for you and Egbert on MacOS?

VojtechStep (changes requested):

PR #1356, opened 2025-03-01
Rebrand from "1000plus" to "Wikipedia's list of theorems"

fredrik-bakke:


fredrik-bakke:
I still believe we should move ahead with this change since we are not collaborating with 1000plus in any capacity, and their project is both unfinished and seems to have gone somewhat stale.

src/literature/wikipedia-list-of-theorems.lagda.md
fredrik-bakke, line 1: Should the file name include a posessive 's', `wikipedias-list-of-theorems`? This would look a little less alien IMO
@VojtechStep

VojtechStep: I personally don't think it needs one, but feel free to add it

fredrik-bakke: okay, then whatever

fredrik-bakke:
I'm merging this as Egbert gave a pre-emptive approval.

fredrik-bakke:
The checks pass locally, so I'm merging this 👍

PR #1361, opened 2025-03-10
Anodyne maps

fredrik-bakke:
Defines anodyne maps and weakly anodyne maps and establishes some basic closure properties.

### DIFF
--- a/.vscode/agda.code-snippets
+++ b/.vscode/agda.code-snippets
@@ -21,14 +21,19 @@
     "description": "Equational reasoning",
     "prefix": ["equational-reasoning"]
   },
-  "Homotopy-reasoning": {
+  "Homotopy reasoning": {
     "body": ["homotopy-reasoning ? ~ ? by ?"],
-    "description": "Homotopy-reasoning",
+    "description": "Homotopy reasoning",
     "prefix": ["homotopy-reasoning"]
   },
-  "Equivalence-reasoning": {
+  "Equivalence reasoning": {
     "body": ["equivalence-reasoning ? ≃ ? by ?"],
-    "description": "Equivalence-reasoning",
+    "description": "Equivalence reasoning",
     "prefix": ["equivalence-reasoning"]
+  },
+  "Retract reasoning": {
+    "body": ["retract-reasoning ? retract-of ? by ?"],
+    "description": "Retract reasoning",
+    "prefix": ["retract-reasoning"]
   }
 }
--- a/src/category-theory/cores-precategories.lagda.md
+++ b/src/category-theory/cores-precategories.lagda.md
@@ -121,6 +121,10 @@ module _
     iso-Precategory (core-precategory-Precategory C) x y ≃ iso-Precategory C x y
   inv-compute-iso-core-Precategory =
     inv-compute-iso-Pregroupoid (core-pregroupoid-Precategory C)
+
+  inclusion-iso-core-Precategory :
+    iso-Precategory C x y → iso-Precategory (core-precategory-Precategory C) x y
+  inclusion-iso-core-Precategory = map-equiv compute-iso-core-Precategory
 ```
 
 ### The core is replete
--- a/src/category-theory/isomorphisms-in-precategories.lagda.md
+++ b/src/category-theory/isomorphisms-in-precategories.lagda.md
@@ -22,6 +22,7 @@ open import foundation.retractions
 open import foundation.sections
 open import foundation.sets
 open import foundation.subtypes
+open import foundation.transport-along-identifications
 open import foundation.universe-levels
 ```
 
@@ -420,7 +421,7 @@ module _
     is-section-hom-inv-is-iso-Precategory C p
 ```
 
-### Inverses of isomorphisms
+### The inverse operation on isomorphisms
 
 ```agda
 module _
@@ -526,6 +527,49 @@ module _
       ( id-iso-Precategory C)
       ( is-section-hom-inv-iso-Precategory C f)
 ```
+
+### The 3-for-2 property of isomorphisms
+
+```agda
+module _
+  {l1 l2 : Level}
+  (C : Precategory l1 l2)
+  {x y z : obj-Precategory C}
+  (f : hom-Precategory C x y)
+  (g : hom-Precategory C y z)
+  where
+
+  is-iso-left-factor-Precategory :
+    is-iso-Precategory C f →
+    is-iso-Precategory C (comp-hom-Precategory C g f) →
+    is-iso-Precategory C g
+  is-iso-left-factor-Precategory F GF =
+    tr
+      ( is-iso-Precategory C)
+      ( equational-reasoning
+        ( comp-hom-Precategory C
+          ( comp-hom-Precategory C g f)
+          ( hom-inv-is-iso-Precategory C F))
+        ＝ ( comp-hom-Precategory C
+            ( g)
+            ( comp-hom-Precategory C f ( hom-inv-is-iso-Precategory C F)))
+          by
+          associative-comp-hom-Precategory C
+            ( g)
+            ( f)
+            ( hom-inv-is-iso-Precategory C F)
+        ＝ (comp-hom-Precategory C g (id-hom-Precategory C))
+          by
+            ap
+              ( comp-hom-Precategory C g)
+              ( is-section-hom-inv-is-iso-Precategory C F)
+        ＝ g
+          by right-unit-law-comp-hom-Precategory C g)
+      ( is-iso-comp-is-iso-Precategory C GF (is-iso-inv-is-iso-Precategory C F))
+```
+
+> It remains to formalize the implication
+> `is-iso g → is-iso (g ∘ f) → is-iso f`.
 
 ### The inverse operation is a fibered involution on isomorphisms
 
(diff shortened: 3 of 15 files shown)


VojtechStep:
How are the changes to descent related to anodyne maps? I can't see from just reading the changes on GitHub where you need those.

fredrik-bakke:
I've continued work in this branch after the initial work on anodyne maps concluded, since subsequent work depends on this and it tends to take a long while for my work to be reviewed. I'm hoping to formalize Mather's second cube theorem. By the way, how do you feel about renaming "descent data of pushouts" to "reduced descent data of pushouts"? So we can name the thing with an additional family `PS` over `S` and two families of equivalences `PS ≃ PA ∘ f` and `PS ≃ PB ∘ g` as "descent data of pushouts"?

fredrik-bakke:
The changes to `fiber'` are also not needed for anodyne maps. In all, the original part of this PR is about 500 lines.

fredrik-bakke:
I've preemptively moved the new changes to a different branch

VojtechStep:
> By the way, how do you feel about renaming "descent data of pushouts" to "reduced descent data of pushouts"?

I don't see a reason to. The current development uses this definition, and the cube theorems talk about cartesian transformations.

fredrik-bakke:
> > By the way, how do you feel about renaming "descent data of pushouts" to "reduced descent data of pushouts"?
> 
> I don't see a reason to. The current development uses this definition, and the cube theorems talk about cartesian transformations.

The proposed definition seems more useful to me, and mirrors the signature of the underlying diagram. For instance, I'm currently working on formalizing the flattening lemma for (expanded) descent data of pushouts, since that's the formulation I need for Mather's cube theorem.

VojtechStep:
An alternative is to call the type with two equivalences "equifibered spans", precisely because they mirror the shape of a span. For sequential colimits the two notions agree, and the SvDR paper uses "equifibered sequences", so `equifibered-sequential-diagram` is already in the library, in addition to `descent-data-sequential-colimit` (which is defined as an alias in this case).

fredrik-bakke:
Ah, yep, that works, thanks!

fredrik-bakke:
@VojtechStep Hi again. I read a little up on the origin of the terminology, and being "equifibered" is a property of a natural transformation/display/dependent structure, not the structure itself. I.e., a span can't be "equifibered", but a dependent or displayed span can be, i.e., a span can only be equifibered _relative to some other span_. In particular, every span is equifibered over itself. This seems to be inconsistent with how you've implemented the terminology in the library. What's the justification for the current terminology usage?

fredrik-bakke:
If you want a noun for such a structure, one might speak of an  _equifibration over a span_, or _equifibration of spans_.

VojtechStep:
I was suggesting that if you have a span `σ : A <-f- S -g-> B`, then you can call the type `(PA : A -> UU) * (PS : S -> UU) * (PB : B -> UU) * (fam-equiv PS (PA . f)) * (fam-equiv PS (PB . g))` the type of *equifibered spans over `σ`*. That should be consistent with the standard nomenclature, because if you take the total span, then the first projection will be equifibered (i.e. what being equifibered would mean for a displayed structure).

See that `equifibered-sequential-diagram` is essentially a dependent sequential diagram bundled with a predicate saying that the maps are fiberwise equivalences. I was thinking about calling it `equifibered-sequential-diagram-over` or `equifibered-sequential-diagram-over-sequential-diagram`, but I decided there's no reason to be this explicit in the name, precisely because as you say, equifibrancy is a property of a displayed structure.

fredrik-bakke:
What you have formalized is an "equifibered dependent sequential diagram", which is different from an equifibered sequential diagram over a sequential diagram, i.e., a morphism of sequential diagram `f : A -> B` whose fibers are sequential diagrams of equivalences. Can I suggest renaming the formalized concept to `equifibered-dependent-sequential-diagram`?

fredrik-bakke:
Sorry, let me elaborate a little. I think the disambiguation/elaboration matters, as the term "equifibered" is not that common, at least, I had not heard of it before, and it sounds similar to other unrelated concepts like "equivariance" and "isofibration". By omitting key terms like "dependent" or "displayed", or "over", the meaning and usage of the term was not clear to me, and I suspect it might not be to others either.

VojtechStep:
There's precedent for calling dependent structures "fibered"/"over" and their instantiation at a point `a` "fiber at `a`"/"over `a`" , and not reserving those just for the unstraightened versions; both in the library (e.g. fiberwise equivalences) and in type-theoretical literature (e.g. what I formalized is called "a sequence equifibered over `(A, a)`" in the SvDR20 paper).

Unless you envision formalizing fibers of general morphisms of sequential diagrams and spans, I would strongly prefer to keep the current naming scheme. And even if you do, I'm inclined to believe that they would be less useful/convenient to work with, since by talking about fibers you raise the dimension of identifications you need to care about — I experimented recently with sections of unstraightened fibered sequential diagrams, and it was very painful.

src/foundation/postcomposition-pullbacks.lagda.md
```diff
@@ -121,38 +122,33 @@ triangle-map-standard-pullback-postcomp T f g c h =
...
+      is-pullback f g c → (T : UU l5) →
+      is-pullback (postcomp T f) (postcomp T g) (postcomp-cone T f g c)
+    is-pullback-postcomp-is-pullback is-pb-c T =
+      is-equiv-top-map-triangle _ _ _
+        ( triangle-map-standard-pullback-postcomp T f g c)
+        ( is-equiv-map-standard-pullback-postcomp f g T)
+        ( universal-property-pullback-is-pullback f g c is-pb-c T)
+
+  abstract
+    is-pullback-is-pullback-postcomp :
```
VojtechStep, line 141: `l5` from the module telescope isn't referenced here, so I'd expect Agda to complain about not being able to determine its value through unification?

src/synthetic-homotopy-theory/pushouts.lagda.md
```diff
@@ -459,6 +460,27 @@ module _
...
 
+### Pushout cocones satisfy the pullback property of the pushout
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} {S : UU l1} {A : UU l2} {B : UU l3}
+  (f : S → A) (g : S → B) {X : UU l4} (c : cocone f g X)
+  where
+
+  pullback-property-pushout-is-pushout :
```
VojtechStep: IME keeping these conversions between characterizations of universal properties abstract is observably better for performance

src/synthetic-homotopy-theory/cocartesian-morphisms-arrows.lagda.md
```diff
@@ -121,6 +125,46 @@ module _
...
+module _
+  {l1 l2 l3 l4 l : Level} {A : UU l1} {B : UU l2} {X : UU l3} {Y : UU l4}
+  (f : A → B) (g : X → Y) (α : cocartesian-hom-arrow f g) (S : UU l)
+  where
+
+  precomp-cocartesian-hom-arrow : hom-arrow (precomp g S) (precomp f S)
+  precomp-cocartesian-hom-arrow =
+    precomp-hom-arrow f g (hom-arrow-cocartesian-hom-arrow f g α) S
+
+  is-cartesian-precomp-cocartesian-hom-arrow :
```
VojtechStep: Unless you need the specific computation, I'd suggest to make this `abstract`

src/orthogonal-factorization-systems/anodyne-maps.lagda.md
```diff
@@ -0,0 +1,172 @@
...
+  {A : UU l1} {B : UU l2} {C : UU l3} {D : UU l4}
+  (f : A → B) (j : C → D)
+  where
+
+  is-anodyne-Level :
+    (l5 l6 : Level) → UU (l1 ⊔ l2 ⊔ l3 ⊔ l4 ⊔ lsuc l5 ⊔ lsuc l6)
+  is-anodyne-Level l5 l6 =
+    {X : UU l5} {Y : UU l6} (g : X → Y) → is-orthogonal f g → is-orthogonal j g
+
+  is-anodyne : UUω
```
VojtechStep: Suggestion: `is-anodyne-map`? You added the `-map` suffix to constancy predicates in #1383, and I think it makes sense

src/orthogonal-factorization-systems/anodyne-maps.lagda.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+</details>
+
+## Idea
+
+A map $j : C → D$ is said to be
+{{#concept "anodyne" Disambiguation="map of types" Agda=is-anodyne}} with
+respect to a map $f : A → B$, or **$f$-anodyne**, if every map that is
+orthogonal to $f$ is also orthogonal to $j$.
```
VojtechStep: Link to orthogonal maps please

src/orthogonal-factorization-systems/weakly-anodyne-maps.lagda.md
```diff
@@ -0,0 +1,89 @@
...
+  {A : UU l1} {B : UU l2} {C : UU l3} {D : UU l4}
+  (f : A → B) (j : C → D)
+  where
+
+  is-weakly-anodyne-Level :
+    (l5 l6 : Level) → UU (l1 ⊔ l2 ⊔ l3 ⊔ l4 ⊔ lsuc l5 ⊔ lsuc l6)
+  is-weakly-anodyne-Level l5 l6 =
+    {X : UU l5} {Y : UU l6} (g : X → Y) → is-local-map f g → is-local-map j g
+
+  is-weakly-anodyne : UUω
```
VojtechStep: Suggestion: `is-weakly-anodyne-map`

VojtechStep (commented):
LGTM :shipit:

src/foundation/postcomposition-pullbacks.lagda.md
fredrik-bakke: thanks!

PR #1363, opened 2025-03-12
Add graph legend

fredrik-bakke:
Adds a legend explaining the color of the nodes in the agda-unimath dependency graph introduced in #1348.


HOWTO-INSTALL.md
```diff
@@ -21,8 +21,8 @@ In order to contribute to the agda-unimath library you will additionally need:
 1. `git`
 2. `make`
 3. `python` version 3.8 or newer
-4. The python libraries `pre-commit`, `pybtex`, `requests`, `tomli`, and
-   `graphviz`. These can be installed by running
+4. The python libraries `pre-commit`, `pybtex`, `requests`, `tomli`,
+   `setuptools`, and `graphviz`. These can be installed by running
```
VojtechStep, line 25: I know this is a draft, but I'm getting worried what you need setuptools for

fredrik-bakke: I don't need it, but `make serve-website` throws errors when it is not installed (although it seems to work respectively)

VojtechStep: What kind of errors? I don't see anything that should impact `serve-website` in this PR

fredrik-bakke: The errors might be due to the fact that my local installation uses a different version of mdbook

VojtechStep: If it's because of a mdbook version mismatch, then I'd rather not include the setuptools change in this PR

fredrik-bakke:
Here's a preview of the graph legend.
image

VojtechStep:
I'll look at it properly tomorrow, but for now it feels a little sus that you're generating a brand new `<html></html>` file that you're then including in a different html file

scripts/generate_dependency_graph_rendering.py
```diff
@@ -133,6 +133,20 @@ def build_dependency_graph(root_dir, most_imported_drop_count=20):
 
     return graph, file_sizes
 
+def generate_html_legend(label_colors, used_labels, output_file, label_counts):
+    """Generate an HTML legend with used namespaces and their colors."""
+    sorted_labels = sorted(used_labels, key=lambda label: label_counts.get(label, 0), reverse=True)
+    with open(output_file, 'w') as f:
+        f.write('<html>\n<head>\n<style>\n')
```
VojtechStep, line 140: This file is later included with an `{{#include}}`, so it can't be a full document with an `<html>` and `<head>`. Since HTML doesn't have a way to write this kind of "scoped" rules (AFAIK, but it seems to go directly against the spirit of CSS), you do need to put it in a CSS file and preferrably have a bit less ambiguous class names. You can either put the rules in `general.css`, or in another file and add a `<link>` here, but using the general file is fine with me.

theme/css/general.css
```diff
@@ -22,7 +22,8 @@ body {
 }
 
 pre,
-code {
+code,
+.code {
```
VojtechStep, line 26: Wouldn't it be more idiomatic to use `<pre>` for the labels instead of `<span class="code">`?

scripts/generate_dependency_graph_rendering.py
```diff
@@ -133,6 +133,20 @@ def build_dependency_graph(root_dir, most_imported_drop_count=20):
...
+        f.write('.dot { height: 9px; width: 9px; border-radius: 50%; display: inline-block; margin-right: 2.5px; margin-left: 5px }\n')
+        f.write('.label { white-space: nowrap; font-family: monospaced; font-size: 50%; display: inline-block; }\n')
+        f.write('</style>\n</head>\n<body>\n')
+        for label in sorted_labels:
+            if label in label_colors:
+                color = label_colors[label]
+                f.write(f'<span class="code" class="label" id="{label}"><span class="dot" style="background-color: #{color};"></span>{label}</span>\n')
```
VojtechStep, line 147: Duplicate attributes like `class="code" class="label"` are not valid HTML. The way to have multiple classes is separating them with spaces, so `class="code label"`. As it is written the `label` class was never applied.

Makefile
```diff
@@ -148,9 +148,13 @@ website/css/Agda-highlight.css: ./scripts/generate_agda_css.py ./theme/catppucci
 website/images/agda_dependency_graph.svg: ${AGDAFILES}
 	@python3 ./scripts/generate_dependency_graph_rendering.py website/images/agda_dependency_graph svg || true
 
+website/images/agda_dependency_graph_legend.html: ${AGDAFILES}
```
VojtechStep, line 151: Instead of duplicating whole rules, you should put both filenames on the left of the colon, and since they are both produced at once, the dependency should use `&:` instead of `:` ([docs](https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html), "Rules with grouped targets"), so
```make
website/...graph.svg website/...legend.html &: ${AGDAFILES}
	@python3 ...
```

scripts/generate_dependency_graph_rendering.py
fredrik-bakke: I changed it up following your suggestions. However, now the line height is much larger than it should be. Do you know what might be causing that? Changing the css value of `line-height` for the contents of the legend has no effect.

fredrik-bakke: For reference see picture below. I have no understanding of the inner workings of web rendering, but it seems to be that a space character is inserted between the labels, and that character has a much larger line height.

image

fredrik-bakke: nvm fixed it
image

theme/css/general.css
```diff
@@ -253,6 +253,13 @@ span.prefer-nobreak {
   display: inline-block;
 }
 
+.art-fredrik-dependency-graph-legend {
+  line-height: 1;
```
VojtechStep, line 257: Having `line-heigth: 1` in both classes is redundant, otherwise it's ready to be merged 👌

PR #1364, opened 2025-03-13
Preunivalence implies strong preunivalence

fredrik-bakke:
As demonstrated by Evan Cavallo, strong preunivalence follows from preunivalence + function extensionality.
By an analogous proof, preunivalent categories are strongly preunivalent.


fredrik-bakke:
This PR currently contains a postulate for the 3-for-2 property for isomorphisms. @EgbertRijke has formalizations for split epi-/monomorphisms in precategories in an unmerged branch (#1071), so I'll leave that for whenever those are merged.

fredrik-bakke:
I found a simple route to prove the last lemma, so this PR is ready for review.

src/category-theory/isomorphisms-in-precategories.lagda.md
```diff
@@ -527,6 +528,48 @@ module _
...
+          by
+            ap
+              ( comp-hom-Precategory C g)
+              ( is-section-hom-inv-is-iso-Precategory C F)
+        ＝ g
+          by right-unit-law-comp-hom-Precategory C g)
+      ( is-iso-comp-is-iso-Precategory C GF (is-iso-inv-is-iso-Precategory C F))
+```
+
+> It remains to formalize the other 2 cases of the 3-for-2 property.
```
VojtechStep: Just `is-iso g -> is-iso (g . f) -> is-iso f`, right? Since the third case is proven by `is-iso-comp-is-iso-Precategory`?

src/foundation-core/retracts-of-types.lagda.md
```diff
@@ -203,6 +203,37 @@ module _
...
+
+```text
+retract-reasoning
+  X retract-of Y by retract-1
+    retract-of Z by retract-2
+    retract-of V by retract-3
+```
+
+The retract constructed in this way is
+`comp-retract retract-3 (comp-retract retract-2 retract-1)`, i.e., the retract
```
VojtechStep, line 218: This is not true, see #1369 — the innermost composition is `comp-retract retract-1 id-retract`, and `comp-retract` isn't strictly unital on the right. You're left with `ap id p ∙ refl` instead of `p` in the retraction proof.

VojtechStep (commented):
LGTM

src/foundation-core/retracts-of-types.lagda.md
fredrik-bakke: Since this error already existed in other explanations, how about we let this one through and treat every error uniformly in a different pr?

fredrik-bakke: either way they are equal, just not judgmentally so

VojtechStep: In any case it's not worth blocking this PR any longer

PR #1365, opened 2025-03-15
Flattening lemma for equifibered dependent span diagrams

fredrik-bakke:
- [X] Improve universe polymorphism of descent data for pushouts
- [X] Define equifibered dependent span diagrams
- [ ] Prove flattening lemma for equifibered dependent span diagrams


fredrik-bakke:
Note that the naming for equifibered diagrams has not yet been determined; these are only working names.

PR #1372, opened 2025-03-19
Pairs of natural numbers with a fixed sum

lowasser:
This is the last bit I need for associativity of multiplication on formal power series, and it's unexpectedly hard for some reason: I can't manage to fill in the last bits of the `eq-pair-Σ`.  I think the induction principle on identity types might be the core of the solution, but I haven't yet figured out how to apply it.

Posting it here in hopes someone else might have ideas.

### DIFF
--- a/src/elementary-number-theory/pairs-with-natural-sums.lagda.md
+++ b/src/elementary-number-theory/pairs-with-natural-sums.lagda.md
@@ -1,0 +1,93 @@
+# Pairs of natural numbers with a given sum
+
+```agda
+module elementary-number-theory.pairs-with-natural-sums where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import elementary-number-theory.addition-natural-numbers
+open import elementary-number-theory.equality-natural-numbers
+open import elementary-number-theory.natural-numbers
+
+open import foundation.dependent-pair-types
+open import foundation.equality-dependent-pair-types
+open import foundation.equivalences
+open import foundation.function-types
+open import foundation.identity-types
+open import foundation.propositions
+open import foundation.retractions
+open import foundation.sections
+open import foundation.sets
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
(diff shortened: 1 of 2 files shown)


src/elementary-number-theory/pairs-with-natural-sums.lagda.md
```diff
@@ -0,0 +1,93 @@
...
+  pr1 (map-inv-equiv-pair-with-sum-pr1-pr2 ((c , p , p+c=n) , _)) =
+    (p , c , commutative-add-ℕ c p ∙ p+c=n)
+  pr2 (map-inv-equiv-pair-with-sum-pr1-pr2 (_ , y)) = y
+
+  is-section-map-inv-equiv-pair-with-sum-pr1-pr2 :
+    is-section
+      map-equiv-pair-with-sum-pr1-pr2
+      map-inv-equiv-pair-with-sum-pr1-pr2
+  is-section-map-inv-equiv-pair-with-sum-pr1-pr2
+    x@(y@(c , p , p+c=n) , a , b , b+a=p) =
```
fredrik-bakke, line 69: I'm not sure if this is exactly what you were asking in the description of your PR, but in this proof, you can pattern match on the equality type, done by writing the name of your path variable in the active hole, and executing the `C-c C-c` command. Or, alternatively, by writing in `refl` for the corresponding variable

```suggestion
    x@(y@(c , p , refl) , a , b , refl) =
```

lowasser: I tried that, and it didn't help -- but I got it another way.

fredrik-bakke: in this case it will mean you assume that `n` is definitionally equal to `p + c`, and `p` is definitionally equal to `b + a`

lowasser: The current version should show the use of the induction principle I had to resort to.  I couldn't make anything short of that work -- and I've tried for an hour or two.

PR #1373, opened 2025-03-20
Remove dependency between `BUILTIN` and postulates

fredrik-bakke:
This PR is preparation for #1379.

The `BUILTIN` pragma cannot be invoked inside of a parametrized module, which in practice means that we cannot have a dependency on an axiom in a module where a `BUILTIN` is invoked. This PR resolves that by moving dependencies around and splitting a couple of files into smaller parts.


fredrik-bakke:
After some preliminary experimentation, there are some technical issues to point out:
- Named modules with parameters are incompatible with `BUILTIN` pragmas, so I've refactored all modules that have such a pragma to depend only on files that are not (transitively) dependent on any module that uses funext anywhere.
- There's an issue with public imports when using named imports. See the current typechecking error. Essentially, such definitions are duplicated in the scope, leading to "ambiguous names"

fredrik-bakke:
The python scripts will be removed later.

fredrik-bakke:
> I've refactored all modules that have such a pragma to depend only on files that are not (transitively) dependent on any module that uses funext anywhere.

If relevant, I can pull these changes to a different branch for a separate review later.

src/elementary-number-theory/addition-natural-numbers.lagda.md
```diff
@@ -154,19 +152,6 @@ abstract
...
-abstract
-  is-emb-left-add-ℕ : (x : ℕ) → is-emb (x +ℕ_)
-  is-emb-left-add-ℕ x =
-    is-emb-is-injective is-set-ℕ (is-injective-left-add-ℕ x)
-
-  is-emb-right-add-ℕ : (x : ℕ) → is-emb (_+ℕ x)
-  is-emb-right-add-ℕ x =
-    is-emb-is-injective is-set-ℕ (is-injective-right-add-ℕ x)
-```
-
```
EgbertRijke, line 196: I couldn't find where you moved these to. Can you help me out?

src/elementary-number-theory/multiplication-natural-numbers.lagda.md
```diff
@@ -194,14 +192,6 @@ is-injective-left-mul-ℕ k H p with
...
 ... | pair l refl = is-injective-left-mul-succ-ℕ l p
 
-is-emb-left-mul-ℕ : (x : ℕ) → is-nonzero-ℕ x → is-emb (x *ℕ_)
-is-emb-left-mul-ℕ x H =
-  is-emb-is-injective is-set-ℕ (is-injective-left-mul-ℕ x H)
-
-is-emb-right-mul-ℕ : (x : ℕ) → is-nonzero-ℕ x → is-emb (_*ℕ x)
-is-emb-right-mul-ℕ x H =
-  is-emb-is-injective is-set-ℕ (is-injective-right-mul-ℕ x H)
-
```
EgbertRijke, line 248: Where were these moved to?

src/elementary-number-theory/addition-natural-numbers.lagda.md
fredrik-bakke: These were probably removed because I didn't know of a good place to put them at the time, and they couldn't stay here because their dependency files import function extensionality. It seems like a reasonable trade-off to me. They're very easy to rederive when we know of a good place to put them. Same goes for the others you ask for.

fredrik-bakke: Let me be more definite. They are removed, not moved elsewhere.

src/foundation/dependent-products-contractible-types.lagda.md
```diff
@@ -0,0 +1,99 @@
...
+### The type of functions into a contractible type is contractible
+
+```agda
+is-contr-function-type :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  is-contr B → is-contr (A → B)
+is-contr-function-type is-contr-B = is-contr-Π (λ _ → is-contr-B)
+```
+
+### The type of equivalences between contractible types is contractible
```
EgbertRijke: It feels unnatural to have "the type of equivalences between contractible types is contractible" in a file about dependent products of contractible types.

Same for the section below about being contractible being a proposition.

One possibility I could think of is that if we start breaking up the files about contractible types, propositions, and truncated types, then we could make files about equivalences of these things, and the universe of these things as well:

`equivalences-contractible-types`
`subuniverse-of-contractible-types`

src/foundation/dependent-products-propositions.lagda.md
```diff
@@ -0,0 +1,243 @@
...
+
+```agda
+abstract
+  is-contr-endomaps-is-prop :
+    {l : Level} (P : UU l) → is-prop P → is-contr (P → P)
+  is-contr-endomaps-is-prop P is-prop-P =
+    is-proof-irrelevant-is-prop (is-prop-function-type is-prop-P) id
+```
+
+### Being a proposition is a proposition
```
EgbertRijke: same comment here

src/foundation/dependent-products-truncated-types.lagda.md
```diff
@@ -0,0 +1,190 @@
...
+  is-prop-is-trunc (succ-𝕋 k) A =
+    is-trunc-Π neg-one-𝕋
+      ( λ x → is-trunc-Π neg-one-𝕋 (λ y → is-prop-is-trunc k (x ＝ y)))
+
+is-trunc-Prop : {l : Level} (k : 𝕋) (A : UU l) → Σ (UU l) (is-trunc neg-one-𝕋)
+pr1 (is-trunc-Prop k A) = is-trunc k A
+pr2 (is-trunc-Prop k A) = is-prop-is-trunc k A
+```
+
+### The type of equivalences between truncated types is truncated
```
EgbertRijke, line 131: Same comment here

EgbertRijke:
This pull request looks good to me. Whenever you have a bit of time, could you resolve the merge conflicts so that we can merge this?

EgbertRijke:
Disregard the above. I figured out how to fix those issues myself using `gh`. I'll resolve all the merge conflicts and my comments.

Thanks for the good work, and good luck with your thesis!

PR #1367, opened 2025-03-16
Sums and products over arbitrary finite types

lowasser:
In preparation for polynomials and formal power series, most notably multiplication of formal power series (#1357), this starts digging into sums over arbitrary finite types, and generalizes from sums on e.g. semirings to products on monoids -- mostly commutative monoids, since they're the ones where arbitrary finite sums make sense.

There is some tension with naming over whether `group-theory.products-monoids` should indicate the monoid on the Cartesian product of two other monoids, or products of elements in a monoid.  Since monoids (and groups in general) use multiplicative terminology, we can't get away with what rings have done by differentiating sums and products.  (And we don't have any support for products of elements of rings, either, though the generalization to commutative monoids is intended to make that possible down the road.)

This is marked a draft because there's one last thing I want to prove, which is that you can split sums over coproducts of arbitrary finite types.

PR #1374, opened 2025-03-21
Commutative semiring of formal power series on commutative semirings

lowasser:
This is enormous, but about half of it is just #1367 .


src/foundation/complements-decidable-subtypes.lagda.md
fredrik-bakke, line 1: There's already a file `logic.complements-decidable-subtypes` (sorry, the logic namespace is a little fragmented right now while we resolve an issue with author attributions) also, the proof I sent you on Discord is a bit neater, so please use that one :)

lowasser: This takes a surprising amount of work, but is also not actually necessary for the design in this PR.  Deleted the changes here, and factored out the work required to get the desired lemma in `comlements-decidable-subtypes` into #1376 .

lowasser:
Closing this to restart the approach from #1444 instead.

PR #1376, opened 2025-03-23
Equivalence between a type and the coproduct of a decidable subtype and its complement

lowasser:


src/foundation/unions-subtypes.lagda.md
```diff
@@ -101,3 +110,56 @@ module _
...
+    rec-coproduct inl-disjunction inr-disjunction
+
+  equiv-union-subtype-coproduct-disjoint-subtype :
+    type-subtype (union-subtype A B) ≃
+    type-subtype subtype-coproduct-disjoint-subtype
+  equiv-union-subtype-coproduct-disjoint-subtype =
+    equiv-tot
+      ( λ x →
+        equiv-iff'
+          ( A x ∨ B x)
```
fredrik-bakke, line 156: The `∨` symbol can be hard to distinguish from the latin letter `v`, so I recommend using parentheses here

```suggestion
          ( (A x) ∨ (B x))
```

src/foundation/unions-subtypes.lagda.md
```diff
@@ -101,3 +110,56 @@ module _
...
+    (x : X) → is-prop (type-Prop (A x) + type-Prop (B x))
+  is-prop-coproduct-disjoint-prop x =
+    is-prop-all-elements-equal (all-elements-equal-coproduct-disjoint-prop x)
+
+  subtype-coproduct-disjoint-subtype : subtype (l2 ⊔ l3) X
+  subtype-coproduct-disjoint-subtype x =
+    type-Prop (A x) + type-Prop (B x) , is-prop-coproduct-disjoint-prop x
+
+  iff-disjunction-coproduct-disjoint-subtype :
+    (x : X) → type-iff-Prop (A x ∨ B x) (subtype-coproduct-disjoint-subtype x)
```
fredrik-bakke, line 143:
```suggestion
    (x : X) → type-iff-Prop ((A x) ∨ (B x)) (subtype-coproduct-disjoint-subtype x)
```

src/foundation/disjoint-subtypes.lagda.md
```diff
@@ -38,6 +39,20 @@ module _
...
   disjoint-subtype : UU (l1 ⊔ l2 ⊔ l3)
   disjoint-subtype = type-Prop disjoint-subtype-Prop
+
+module _
+  {l1 l2 l3 : Level} {A : UU l1}
+  (B : decidable-subtype l2 A) (C : decidable-subtype l3 A)
+  where
+
+  disjoint-decidable-subtype-Prop : Prop (l1 ⊔ l2 ⊔ l3)
+  disjoint-decidable-subtype-Prop =
```
fredrik-bakke, line 49: This should be called `is-disjoint-prop-decidable-subtype`, correspondinly, `disjoint-subtype-Prop` should be called `is-disjoint-prop-subtype`, sorry that I didn't catch that last time

src/foundation/disjoint-subtypes.lagda.md
```diff
@@ -38,6 +39,20 @@ module _
...
+    disjoint-subtype-Prop
+      ( subtype-decidable-subtype B)
+      ( subtype-decidable-subtype C)
+
+  disjoint-decidable-subtype : UU (l1 ⊔ l2 ⊔ l3)
```
fredrik-bakke, line 54:
```suggestion
  is-disjoint-decidable-subtype : UU (l1 ⊔ l2 ⊔ l3)
```

src/foundation/unions-subtypes.lagda.md
```diff
@@ -101,3 +110,56 @@ module _
...
+  all-elements-equal-coproduct-disjoint-prop :
+    (x : X) → all-elements-equal (type-Prop (A x) + type-Prop (B x))
+  all-elements-equal-coproduct-disjoint-prop x (inl _) (inl _) =
+    ap inl (eq-type-Prop (A x))
+  all-elements-equal-coproduct-disjoint-prop x (inr _) (inr _) =
+    ap inr (eq-type-Prop (B x))
+  all-elements-equal-coproduct-disjoint-prop x (inl ax) (inr bx) =
+    ex-falso (H x (ax , bx))
+  all-elements-equal-coproduct-disjoint-prop x (inr bx) (inl ax) =
+    ex-falso (H x (ax , bx))
```
fredrik-bakke, line 131: This result should be derived from the corresponding result for disjunctions of propositions: if `¬ (A ∧ B)` then `A + B` is a proposition and equivalent to `A ∨ B`

lowasser: Does that exist anywhere?  Is there a good name for that concept?  I looked around and thought about it, but I don't even know what to call e.g. `¬ (A ∧ B)`.  "Incompatible propositions"?  "Contradictory"?

fredrik-bakke: Hmm, good question. I'm not sure I know a name for this. We have exclusive sum and exclusive disjunction, which are about different kinds of disjunctions "without the intersection", but nothing about the negation of the conjunction specifically. We also have files on de Morgan's law, which is about how `¬ (A ∧ B)` relates to `(¬ A) ∨ (¬ B)`, but I can't recall seeing another name for it than "the negation of the conjunction".

fredrik-bakke: All in all, there are a few candidates for "incompatible" or "contradictory" or "mutually exclusive propositions" — I'm not sure what follows from what. What comes to mind is
1.  `¬ (A ∧ B)`
2. `(¬ A) ∨ (¬ B)`
3. `(A ⇒ ¬ B)`
4. `(A ⇒ ¬ B) ∨ (B ⇒ ¬ A)`
5. `(¬ A ⇒ B)`
6. `(¬ A ⇒ B) ∨ (¬ B ⇒ A)`

Maybe some of these are silly conditions, and maybe some of them are wrong. The point is there might be many nonequivalent conditions here, and that's worthwhile to keep in mind when we name one of them.

lowasser: Right.  The reason I didn't write the generalization is because this specific form, at least, has a clear and unambiguous name: disjointness is unambiguously the right term for subtypes.  I agree there's something more general to say, but in the absence of a clear name for the more general thing, I'm hesitant to just throw some name we may not like into this PR.

fredrik-bakke: We could call them _disjoint propositions_ since they are precisely _disjoint subsets_ of the 1-element type? Otherwise, I suggest looking up some references.

fredrik-bakke:
I'm marking this PR as a draft since it does not typecheck.

Issue #1379, opened 2025-03-24
Remove postulates


fredrik-bakke:
With #1373 I began work towards removing axioms as postulates in the library, this has multiple benefits as well as drawbacks.

### Pros
- It enables axiom analysis such as reverse mathematics. For instance, we can formally establish that univalence implies function extensionality. We currently have a "formal" proof in the library, but it relies on an external (nonformal) tool to verify that no part of that proof applies any other axiom than univalence. Alternatively, the reader must verify manually that no part of the proof relies on any other axiom.
- We can have developments of mutually inconsistent mathematical theories in the library.
- Lowers the barrier between TypeTopology formalizations and agda-unimath formalizations.
- Potentially makes it possible to reenable the `--safe` pragma for most of the library.

### Cons
- Requires more bookkeeping. E.g. if a module depends on function extensionality, we must state so explicitly with
  ```agda
  open import foundation.function-extensionality-axiom
  open import foundation.truncations

  module
    namespace.modulename
      ( funext : function-extensionality)
      ( truncations : truncations-exist) 
    where
  ```
  and must pass on these axioms to every dependency that also need these. E.g.

  ```agda
  open import namespace.other-module1 funext
  open import namespace.other-module2 funext truncations
  open import namespace.other-module3
  open import namespace.other-module4 truncations
  ```
  the dependency resolution here could potentially be partially automated.

VojtechStep:
> so that we don't have to maintain funext as a second hypothesis
Coming back to this, univalence gives you `function-extensionality-Level l l`, not `function-extensionality-Level l1 l2`, AFAIK. Is that incorrect, or do you propose removing the second hypothesis from modules which only need extensionality of maps with domain and codomain in the same universe?

fredrik-bakke:
> > so that we don't have to maintain funext as a second hypothesis
> 
> Coming back to this, univalence gives you `function-extensionality-Level l l`, not `function-extensionality-Level l1 l2`, AFAIK. Is that incorrect, or do you propose removing the second hypothesis from modules which only need extensionality of maps with domain and codomain in the same universe?

That's incorrect https://martinescardo.github.io/TypeTopology/MGS.FunExt-from-Univalence.html#786

fredrik-bakke:
Sorry, the cited formalization entry isn't homotopy coherent. This one is though: https://martinescardo.github.io/TypeTopology/MGS.FunExt-from-Univalence.html#5425

fredrik-bakke:
[...]

PR #1380, opened 2025-03-24
Preview: depostulate axioms

fredrik-bakke:
Here's a preview of what the library might look like with depostulated axioms.

#1379


scripts/remove_unused_imports.py
```diff
@@ -115,7 +115,7 @@ def filter_agda_files(f): return utils.is_agda_file(pathlib.Path(f)) and os.path
 
     temp_dir = 'temp'
     status = 0
-    agda_options = '--without-K --exact-split --no-import-sorts --auto-inline --no-caching'
+    agda_options = '--without-K --exact-split --no-import-sorts --auto-inline --no-caching -WnoPatternShadowsConstructor'
```
EgbertRijke, line 118: Why wouldn't we want these warnings?

Also, is the question of including this pragma orthogonal to the purpose of this pull request, or am I missing something?

fredrik-bakke: Ah, I thought this pragma promoted the warning to an error. It is orthogonal to this PR, but not to #1373 on which it depends

fredrik-bakke: (There it caused errors in the file `elementary-number-theory.multiplication-integers`)

PR #1378, opened 2025-03-24
Limits of sequences in metric spaces

malarbol:
This PR formalizes a few definitions and results about limits in (pre|pseudo|)metric spaces.
It introduces the following modules:

- `metric-spaces.sequences-premetric-spaces`:
  - sequences in premetric spaces.

- `metric-spaces.limits-sequences-premetric-spaces`:
  - limits of sequences in premetric spaces;
  - short maps preserve limits.

- `metric-spaces.sequences-pseudometric-spaces`:
  - sequences in pseudometric spaces.

- `metric-spaces.limits-sequences-pseudometric-spaces`:
  - limits of sequences in pseudometric spaces;
  - indistinguishability of limits of a sequence in a pseudometric space.

- `metric-spaces.sequences-metric-spaces`:
  - the metric space of sequences in a metric space.

- ` metric-spaces.limits-sequences-metric-spaces`:
  - limits of sequences in metric spaces;
  - unicity of the limit of a sequence in a metric space.

- `metric-spaces.convergent-sequences-metric-spaces`:
  - sequences in a metric space that have a limit;
  - short maps between metric spaces preserve convergent sequences.

- `metric-spaces.metric-space-of-convergent-sequences-in-a-metric-space`:
  - the metric space of convergent sequences in a metric space.

PR #1382, opened 2025-03-26
Eventually pointed sequences

malarbol:
A lot of properties from https://github.com/UniMath/agda-unimath/pull/1378 follow the pattern
```agda
Σ ℕ (λ N → (n : ℕ) → leq-ℕ N n → X n)
``` 
This PR introduces this concept for sequences of types and a few applications: eventually equal sequences and eventually constant sequences.


src/foundation/eventually-pointed-sequences-types.lagda.md
```diff
@@ -0,0 +1,169 @@
...
+```agda
+module _
+  {l : Level} (A : ℕ → UU l)
+  where
+
+  is-modulus-eventually-pointed-sequence : ℕ → UU l
+  is-modulus-eventually-pointed-sequence N = (n : ℕ) → leq-ℕ N n → A n
+
+  is-eventually-pointed-sequence : UU l
+  is-eventually-pointed-sequence = Σ ℕ is-modulus-eventually-pointed-sequence
```
fredrik-bakke, line 44: This property should be phrased with existential quantification. You can call the untruncated thing something like a `bound-modulus-eventually-pointed-sequence` (open for suggestions)

fredrik-bakke: Actually, wouldn't this just be a `modulus-eventually-pointed-sequence`?

malarbol: > This property(!) should be phrased with existential quantification. You can call the untruncated thing something like a `bound-modulus-eventually-pointed-sequence` (open for suggestions)

As we've seen in https://github.com/UniMath/agda-unimath/pull/1378, there's still some debate (I think) on when to use dependent pairs vs existential quantifications. I'm starting with the _easy_ version. Maybe the truncated things could me called `is-merely-eventually-XXX`?

src/foundation/constant-sequences.lagda.md
```diff
@@ -0,0 +1,121 @@
...
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+A [sequence](foundation.sequences.md) `u` is
+{{#concept "constant" Disambiguation="sequence" Agda=is-constant-sequence}} if
+`u p ＝ u q` for all `p` and `q`.
```
fredrik-bakke, line 27: Constancy conditions on functions into arbitrary types is a bit of a hairy topic (one wants both something that is a property, but also something that satisfies all the expected conditions of a constant function). What you've phrased here is the condition of "weak constancy" (see [`foundation.weakly-constant-maps`](https://unimath.github.io/agda-unimath/foundation.weakly-constant-maps.html)), which is not in general a proposition when the codomain is not a set. Not even the condition that the [image](https://unimath.github.io/agda-unimath/foundation.images.html) of the sequence is a proposition is correct. This issue is studied in some detail in [_Notions of Anonymous Existence in Martin-Löf Type Theory_](https://arxiv.org/abs/1610.03346) by Nicolai Kraus et. al.

See also [`foundation.null-homotopic maps`](https://unimath.github.io/agda-unimath/foundation.null-homotopic-maps.html)

malarbol: ok. let's just forget about constant sequences. I guess I've got a few things to read.

src/foundation/eventually-pointed-sequences-types.lagda.md
fredrik-bakke: You're free to use the easy version, I'm just saying that it does not deserve the name `is-eventually-pointed-sequence`, and that a more fitting name for the thing you're defining is `modulus-eventually-pointed-sequence` since every eventually pointed sequence is eventually pointed in an infinite number of distinct ways with your current definition. Doing arithmetic on the modulus is surely useful, and for that you will need the untruncated version. This is analogous to the distinction between finite types and counting in univalent combinatorics. A finite type with $n$ elements has $n!$ distinct ways of being counted, but you don't for instance distinguish between two 2-element sets by the way the order in which the elements are counted.

src/foundation/constant-sequences.lagda.md
fredrik-bakke: I invite you to define constant sequences in sets instead. There, the theory will work mostly as you expect (aside from possible constructivity considerations)

malarbol: After all, I'm not sure I really need the concept right now, or in the immediate future, so we can just leave it out of this PR.

fredrik-bakke: I can take the opportunity to ask Nicolai tomorrow if he knows of a good choice of notion of constancy, actually. This sort of issue also came up when I worked on #1318.

src/foundation/eventually-pointed-sequences-types.lagda.md
malarbol: ok, what about `has-modulus-eventually-pointed-sequence`?

src/foundation/constant-sequences.lagda.md
fredrik-bakke: After reading up a little, [this article](https://arxiv.org/abs/1411.2682) (also by Nicolai) seems to give a constructive answer to the question. In your setup, a fully coherent notion of _constant sequence_ in an arbitrary type is to ask that the sequence is null-homotopic as a function. So then you don't have to worry about truncation levels after all! 😃

malarbol: Anyway, this PR was just a preamble of the next one on sequences in posets, where I intended to prove a simple version of the squeeze theorem if `u ≤ v ≤ w` with `u` and `w` eventually equal with the same eventual value, then `v` is eventually constant with this value.

I think I have enough in this PR, or I can just include its content in the next one.

fredrik-bakke: Okay :)

fredrik-bakke: I formalized the general concept of coherently constant maps here: #1383 (note that the notion coincides with null-homotopic maps when the domain is pointed, such as for sequences)

src/foundation/eventually-constant-sequences.lagda.md
fredrik-bakke, line 1: This file can be called `moduli-eventually-constant-sequences`, leaving room for the eventual file (no pun intended)`eventually-constant-sequences` which considers the proof theoretic correct notion of eventually constant sequences.

src/foundation/eventually-constant-sequences.lagda.md
```diff
@@ -0,0 +1,98 @@
+# Eventually constant sequences
```
fredrik-bakke, line 1:
```suggestion
# Moduli of eventually constant sequences
```

src/foundation/eventually-constant-sequences.lagda.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Idea
+
+A [sequence](foundation.sequences.md) `u` is
+{{#concept "eventually constant" Disambiguation="sequence" Agda=has-modulus-eventually-constant-sequence}}
+if `u p ＝ u q` for sufficiently large `p` and `q`.
+
+## Definitions
+
+### Eventually constant sequences
```
fredrik-bakke, line 36:
```suggestion
### Moduli of eventually constant sequences
```

src/foundation/eventually-constant-sequences.lagda.md
```diff
@@ -0,0 +1,98 @@
...
+
+```agda
+module _
+  {l : Level} {A : UU l} (u : sequence A)
+  where
+
+  has-modulus-eventually-constant-sequence : UU l
```
fredrik-bakke, line 43: Under your current definitions, you can't deduce that "an eventually constant sequence has a modulus" without a choice principle. By omitting the qualifier `has` you avoid this hairy problem with semantics. "Of course you can't extract a choice of a modulus from an eventually constant sequence without choice!"

```suggestion
  modulus-eventually-constant-sequence : UU l
```

This is analogous to how we say "countings of finite types" as opposed to "finite types have counting".

src/foundation/eventually-constant-sequences.lagda.md
```diff
@@ -0,0 +1,98 @@
...
+  has-modulus-eventually-constant-sequence =
+    has-modulus-eventually-pointed-sequence
+      (λ p → has-modulus-eventually-pointed-sequence (λ q → u p ＝ u q))
+```
+
+### The eventual value of an eventually constant sequence
```
fredrik-bakke, line 49: I believe the standard nomenclature is to call this "eventual value" the "clustering point" or "asymptotic value". Please find a standard reference to follow for the terminology usage here.

src/foundation/eventually-constant-sequences.lagda.md
```diff
@@ -0,0 +1,98 @@
...
+  where
+
+  has-modulus-eventually-constant-const-sequence :
+    has-modulus-eventually-constant-sequence (const ℕ x)
+  pr1 has-modulus-eventually-constant-const-sequence = zero-ℕ
+  pr2 has-modulus-eventually-constant-const-sequence p I =
+    (zero-ℕ , λ _ _ → refl)
+```
+
+### An eventually constant sequence is eventually equal to the constant sequence of its eventual value
```
fredrik-bakke, line 84:
```suggestion
### A sequence with a modulus of eventual constancy has a modulus of eventual equality to the constant sequence of its ??asymptotic value??
```

src/foundation/eventually-equal-sequences.lagda.md
fredrik-bakke, line 1: `moduli-eventually-equal-sequences`

fredrik-bakke (changes requested):
Please implement my suggestions for all of the contributed files

malarbol:
> Please implement my suggestions for all of the contributed files

Ok, I'll do my best. Do you have any suggestion for 
- eventually equal --> modulus of eventual equality
- eventually constant --> modulus of eventual constancy
- eventually pointed --> ??

malarbol:
I'll just stall this PR and go back to sequences in posets. Although it's really tempting to compare values "for sufficiently large `n`" I think I lack the experience to identify the good concepts, etc. For example, maybe the `eventual` stuff should be defined for sequences `ℕ → Prop l`?

fredrik-bakke:
Moduli of eventually constant sequences, moduli of eventually equal sequences, moduli of eventually pointed sequences of types

PR #1387, opened 2025-03-28
More logic

fredrik-bakke:
Pulls logic additions from #1264.

### DIFF
--- a/src/foundation/cantors-theorem.lagda.md
+++ b/src/foundation/cantors-theorem.lagda.md
@@ -16,16 +16,14 @@ open import foundation.function-extensionality
 open import foundation.logical-equivalences
 open import foundation.negation
 open import foundation.powersets
-open import foundation.propositional-truncations
 open import foundation.surjective-maps
 open import foundation.universe-levels
 
-open import foundation-core.empty-types
 open import foundation-core.fibers-of-maps
-open import foundation-core.propositions
 
 open import logic.de-morgan-propositions
 open import logic.de-morgan-subtypes
+open import logic.double-negation-dense-maps
 open import logic.double-negation-stable-subtypes
 ```
 
@@ -74,18 +72,20 @@ module _
   subtype-theorem-Cantor : powerset l2 X
   subtype-theorem-Cantor x = neg-Prop (f x x)
 
-  abstract
-    not-in-image-subtype-theorem-Cantor : ¬ (fiber f subtype-theorem-Cantor)
-    not-in-image-subtype-theorem-Cantor (ξ , α) =
-      no-fixed-points-neg-Prop (f ξ ξ) (iff-eq (htpy-eq α ξ))
+  not-in-image-subtype-theorem-Cantor : ¬ (fiber f subtype-theorem-Cantor)
+  not-in-image-subtype-theorem-Cantor (ξ , α) =
+    no-fixed-points-neg-Prop (f ξ ξ) (iff-eq (htpy-eq α ξ))
+
+  theorem-double-negation-dense-Cantor : ¬ (is-double-negation-dense-map f)
+  theorem-double-negation-dense-Cantor H =
+    H subtype-theorem-Cantor not-in-image-subtype-theorem-Cantor
 
   abstract
     theorem-Cantor : ¬ (is-surjective f)
-    theorem-Cantor H =
-      apply-universal-property-trunc-Prop
-        ( H subtype-theorem-Cantor)
-        ( empty-Prop)
-        ( not-in-image-subtype-theorem-Cantor)
+    theorem-Cantor =
+      map-neg
+        is-double-negation-dense-map-is-surjective
+        theorem-double-negation-dense-Cantor
 ```
 
 ### Cantor's theorem for the set of decidable subtypes
@@ -101,21 +101,24 @@ module _
   map-theorem-decidable-Cantor : decidable-subtype l2 X
   map-theorem-decidable-Cantor x = neg-Decidable-Prop (f x x)
 
-  abstract
-    not-in-image-map-theorem-decidable-Cantor :
-      ¬ (fiber f map-theorem-decidable-Cantor)
-    not-in-image-map-theorem-decidable-Cantor (x , α) =
-      no-fixed-points-neg-Decidable-Prop
-        ( f x x)
-        ( iff-eq (ap prop-Decidable-Prop (htpy-eq α x)))
+  not-in-image-map-theorem-decidable-Cantor :
+    ¬ (fiber f map-theorem-decidable-Cantor)
+  not-in-image-map-theorem-decidable-Cantor (x , α) =
+    no-fixed-points-neg-Decidable-Prop
+      ( f x x)
+      ( iff-eq (ap prop-Decidable-Prop (htpy-eq α x)))
+
+  theorem-double-negation-dense-decidable-Cantor :
+    ¬ (is-double-negation-dense-map f)
+  theorem-double-negation-dense-decidable-Cantor H =
+    H map-theorem-decidable-Cantor not-in-image-map-theorem-decidable-Cantor
 
   abstract
     theorem-decidable-Cantor : ¬ (is-surjective f)
-    theorem-decidable-Cantor H =
-      apply-universal-property-trunc-Prop
-        ( H map-theorem-decidable-Cantor)
-        ( empty-Prop)
-        ( not-in-image-map-theorem-decidable-Cantor)
+    theorem-decidable-Cantor =
+      map-neg
+        is-double-negation-dense-map-is-surjective
+        theorem-double-negation-dense-decidable-Cantor
 ```
 
 ### Cantor's theorem for the set of double negation stable subtypes
@@ -133,21 +136,25 @@ module _
   map-theorem-double-negation-stable-Cantor x =
     neg-Double-Negation-Stable-Prop (f x x)
 
-  abstract
-    not-in-image-map-theorem-double-negation-stable-Cantor :
-      ¬ (fiber f map-theorem-double-negation-stable-Cantor)
-    not-in-image-map-theorem-double-negation-stable-Cantor (x , α) =
-      no-fixed-points-neg-Double-Negation-Stable-Prop
-        ( f x x)
-        ( iff-eq (ap prop-Double-Negation-Stable-Prop (htpy-eq α x)))
+  not-in-image-map-theorem-double-negation-stable-Cantor :
+    ¬ (fiber f map-theorem-double-negation-stable-Cantor)
+  not-in-image-map-theorem-double-negation-stable-Cantor (x , α) =
+    no-fixed-points-neg-Double-Negation-Stable-Prop
+      ( f x x)
+      ( iff-eq (ap prop-Double-Negation-Stable-Prop (htpy-eq α x)))
+
+  theorem-double-negation-dense-double-negation-stable-Cantor :
+    ¬ (is-double-negation-dense-map f)
+  theorem-double-negation-dense-double-negation-stable-Cantor H =
+    H map-theorem-double-negation-stable-Cantor
+      not-in-image-map-theorem-double-negation-stable-Cantor
 
   abstract
     theorem-double-negation-stable-Cantor : ¬ (is-surjective f)
-    theorem-double-negation-stable-Cantor H =
-      apply-universal-property-trunc-Prop
-        ( H map-theorem-double-negation-stable-Cantor)
-        ( empty-Prop)
-        ( not-in-image-map-theorem-double-negation-stable-Cantor)
+    theorem-double-negation-stable-Cantor =
+      map-neg
+        is-double-negation-dense-map-is-surjective
+        theorem-double-negation-dense-double-negation-stable-Cantor
 ```
 
 ### Cantor's theorem for the set of De Morgan subtypes
@@ -163,21 +170,24 @@ module _
   map-theorem-de-morgan-Cantor : de-morgan-subtype l2 X
   map-theorem-de-morgan-Cantor x = neg-De-Morgan-Prop (f x x)
 
-  abstract
-    not-in-image-map-theorem-de-morgan-Cantor :
-      ¬ (fiber f map-theorem-de-morgan-Cantor)
-    not-in-image-map-theorem-de-morgan-Cantor (x , α) =
-      no-fixed-points-neg-De-Morgan-Prop
-        ( f x x)
-        ( iff-eq (ap prop-De-Morgan-Prop (htpy-eq α x)))
+  not-in-image-map-theorem-de-morgan-Cantor :
+    ¬ (fiber f map-theorem-de-morgan-Cantor)
+  not-in-image-map-theorem-de-morgan-Cantor (x , α) =
+    no-fixed-points-neg-De-Morgan-Prop
+      ( f x x)
+      ( iff-eq (ap prop-De-Morgan-Prop (htpy-eq α x)))
+
+  theorem-double-negation-dense-de-morgan-Cantor :
+    ¬ (is-double-negation-dense-map f)
+  theorem-double-negation-dense-de-morgan-Cantor H =
+    H map-theorem-de-morgan-Cantor not-in-image-map-theorem-de-morgan-Cantor
 
   abstract
     theorem-de-morgan-Cantor : ¬ (is-surjective f)
-    theorem-de-morgan-Cantor H =
-      apply-universal-property-trunc-Prop
-        ( H map-theorem-de-morgan-Cantor)
-        ( empty-Prop)
-        ( not-in-image-map-theorem-de-morgan-Cantor)
+    theorem-de-morgan-Cantor =
+      map-neg
+        is-double-negation-dense-map-is-surjective
+        theorem-double-negation-dense-de-morgan-Cantor
 ```
 
 ## References
--- a/src/foundation/decidable-dependent-function-types.lagda.md
+++ b/src/foundation/decidable-dependent-function-types.lagda.md
@@ -7,9 +7,14 @@ module foundation.decidable-dependent-function-types where
 <details><summary>Imports</summary>
 
 ```agda
+open import foundation.decidable-propositions
 open import foundation.decidable-types
+open import foundation.dependent-pair-types
 open import foundation.functoriality-dependent-function-types
+open import foundation.irrefutable-equality
 open import foundation.maybe
+open import foundation.mere-equality
+open import foundation.propositions
 open import foundation.uniformly-decidable-type-families
 open import foundation.universal-property-coproduct-types
 open import foundation.universal-property-maybe
@@ -19,7 +24,8 @@ open import foundation-core.coproduct-types
 open import foundation-core.empty-types
 open import foundation-core.equivalences
 open import foundation-core.function-types
-open import foundation-core.negation
+
+open import logic.propositionally-decidable-types
 ```
 
 </details>
@@ -46,6 +52,28 @@ is-decidable-Π-uniformly-decidable-family (inl a) (inr b) =
   inr (λ f → b a (f a))
 is-decidable-Π-uniformly-decidable-family (inr na) _ =
   inl (ex-falso ∘ na)
+
+is-decidable-prop-Π-uniformly-decidable-family :
+  {l1 l2 : Level} {A : UU l1} {B : A → UU l2} →
+  is-decidable A →
+  is-uniformly-decidable-family B →
+  ((x : A) → is-prop (B x)) →
+  is-decidable-prop ((a : A) → (B a))
+is-decidable-prop-Π-uniformly-decidable-family dA dB H =
+  ( is-prop-Π H , is-decidable-Π-uniformly-decidable-family dA dB)
+
+abstract
+  is-decidable-prop-Π-uniformly-decidable-family' :
+    {l1 l2 : Level} {A : UU l1} {B : A → UU l2} →
+    is-inhabited-or-empty A →
+    is-uniformly-decidable-family B →
+    ((x : A) → is-prop (B x)) →
+    is-decidable-prop ((a : A) → (B a))
+  is-decidable-prop-Π-uniformly-decidable-family' {A = A} {B} dA dB H =
+    elim-is-inhabited-or-empty-Prop'
+      ( is-decidable-prop-Prop ((a : A) → (B a)))
+      ( λ d → is-decidable-prop-Π-uniformly-decidable-family d dB H)
+      ( dA)
 ```
 
 ### Decidablitilty of dependent products over coproducts
@@ -74,6 +102,97 @@ is-decidable-Π-Maybe {B = B} du de =
   is-decidable-equiv
     ( equiv-dependent-universal-property-Maybe B)
     ( is-decidable-product du de)
+```
+
+### Dependent products of decidable propositions over a base with double negation dense equality are decidable propositions
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (B : A → Decidable-Prop l2)
+  where
+
+  is-decidable-Π-has-double-negation-dense-equality-base :
+    has-double-negation-dense-equality A →
+    is-decidable A →
+    is-decidable ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-Π-has-double-negation-dense-equality-base H dA =
+    is-decidable-Π-uniformly-decidable-family dA
+      ( is-uniformly-decidable-family-has-double-negation-dense-equality-base
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA))
+
+  is-decidable-prop-Π-has-double-negation-dense-equality-base :
+    has-double-negation-dense-equality A →
+    is-decidable A →
+    is-decidable-prop ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-prop-Π-has-double-negation-dense-equality-base H dA =
+    is-decidable-prop-Π-uniformly-decidable-family dA
+      ( is-uniformly-decidable-family-has-double-negation-dense-equality-base
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA))
+      ( is-prop-type-Decidable-Prop ∘ B)
+
+  is-decidable-prop-Π-has-double-negation-dense-equality-base' :
+    has-double-negation-dense-equality A →
+    is-inhabited-or-empty A →
+    is-decidable-prop ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-prop-Π-has-double-negation-dense-equality-base' H dA =
+    is-decidable-prop-Π-uniformly-decidable-family' dA
+      ( is-uniformly-decidable-family-has-double-negation-dense-equality-base'
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA)
+        ( is-prop-type-Decidable-Prop ∘ B))
+      ( is-prop-type-Decidable-Prop ∘ B)
+```
+
+### Dependent products of decidable propositions over a base with mere equality are decidable propositions
+
+Assuming the base `A` is empty or 0-connected, a dependent product of decidable
+propositions over `A` is again a decidable proposition.
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (B : A → Decidable-Prop l2)
+  where
+
+  is-decidable-Π-all-elements-merely-equal-base :
+    all-elements-merely-equal A →
+    is-decidable A →
+    is-decidable ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-Π-all-elements-merely-equal-base H dA =
+    is-decidable-Π-uniformly-decidable-family dA
+      ( is-uniformly-decidable-family-all-elements-merely-equal-base
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA))
+
+  is-decidable-prop-Π-all-elements-merely-equal-base :
+    all-elements-merely-equal A →
+    is-decidable A →
+    is-decidable-prop ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-prop-Π-all-elements-merely-equal-base H dA =
+    is-decidable-prop-Π-uniformly-decidable-family dA
+      ( is-uniformly-decidable-family-all-elements-merely-equal-base
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA))
+      ( is-prop-type-Decidable-Prop ∘ B)
+
+  is-decidable-prop-Π-all-elements-merely-equal-base' :
+    all-elements-merely-equal A →
+    is-inhabited-or-empty A →
+    is-decidable-prop ((x : A) → type-Decidable-Prop (B x))
+  is-decidable-prop-Π-all-elements-merely-equal-base' H dA =
+    is-decidable-prop-Π-uniformly-decidable-family' dA
+      ( is-uniformly-decidable-family-all-elements-merely-equal-base'
+        ( H)
+        ( is-decidable-Decidable-Prop ∘ B)
+        ( dA)
+        ( is-prop-type-Decidable-Prop ∘ B))
+      ( is-prop-type-Decidable-Prop ∘ B)
 ```
 
 ### Decidability of dependent products over an equivalence
--- a/src/foundation/decidable-maps.lagda.md
+++ b/src/foundation/decidable-maps.lagda.md
@@ -13,8 +13,10 @@ open import foundation.coproduct-types
 open import foundation.decidable-equality
 open import foundation.decidable-types
 open import foundation.dependent-pair-types
+open import foundation.embeddings
 open import foundation.functoriality-cartesian-product-types
 open import foundation.functoriality-coproduct-types
+open import foundation.hilbert-epsilon-operators-maps
 open import foundation.identity-types
 open import foundation.retracts-of-maps
 open import foundation.universe-levels
@@ -81,14 +83,13 @@ module _
 ### Decidable maps are closed under homotopy
 
 ```agda
-abstract
-  is-decidable-map-htpy :
-    {l1 l2 : Level} {A : UU l1} {B : UU l2} {f g : A → B} →
-    f ~ g → is-decidable-map g → is-decidable-map f
-  is-decidable-map-htpy H K b =
-    is-decidable-equiv
-      ( equiv-tot (λ a → equiv-concat (inv (H a)) b))
-      ( K b)
+is-decidable-map-htpy :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f g : A → B} →
+  f ~ g → is-decidable-map g → is-decidable-map f
+is-decidable-map-htpy H K b =
+  is-decidable-equiv
+    ( equiv-tot (λ a → equiv-concat (inv (H a)) b))
+    ( K b)
 ```
 
 ### Composition of decidable maps
@@ -279,3 +280,32 @@ module _
       ( retract-fiber-retract-map f g R x)
       ( G (map-codomain-inclusion-retract-map f g R x))
 ```
+
+### Decidable maps have Hilbert ε-operators
+
+```agda
+ε-operator-map-is-decidable-map :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f : A → B} →
+  is-decidable-map f → ε-operator-map f
+ε-operator-map-is-decidable-map F = ε-operator-is-decidable ∘ F
+```
+
+### Decidable injective maps are embeddings
+
+**Proof.** Given a decidable map `f : A → B` then `f` decomposes
+`B ≃ (im f) + B∖(im f)`. Restricting to `im f` we have a section given by the
+Hilbert ε-operator on `f`. Now, by injectivity of `f` we know this restriction
+map is an equivalence. Hence, by 3-for-2 `f` is also an embedding.
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f : A → B}
+  where
+
+  is-emb-is-injective-is-decidable-map :
+    is-decidable-map f → is-injective f → is-emb f
+  is-emb-is-injective-is-decidable-map H K =
+    is-emb-is-injective-ε-operator-map (ε-operator-map-is-decidable-map H) K
+```
+
+There is also an analogous proof using the double negation image.
--- a/src/foundation/decidable-type-families.lagda.md
+++ b/src/foundation/decidable-type-families.lagda.md
@@ -1,0 +1,166 @@
+# Decidable families
+
+```agda
+module foundation.decidable-type-families where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.coproduct-types
+open import foundation.decidable-subtypes
+open import foundation.decidable-types
+open import foundation.dependent-pair-types
+open import foundation.double-negation
+open import foundation.irrefutable-equality
+open import foundation.negation
+open import foundation.propositions
+open import foundation.transport-along-identifications
+open import foundation.universe-levels
+
+open import foundation-core.function-types
+open import foundation-core.identity-types
+```
+
+</details>
+
+## Idea
+
+A type family `B : A → 𝒰` is said to be
+{{#concept "decidable" Disambiguation="type family" Agda=is-decidable-family}}
+if we, for every `x : A`, can either construct an element of `B x`, or we can
+prove that it is [empty](foundation-core.empty-types.md). In other words, we
+interpret decidability via the
+[Curry–Howard interpretation](https://en.wikipedia.org/wiki/Curry–Howard_correspondence)
+of logic into type theory. A related concept is that a type family is either
+[inhabited](foundation.inhabited-types.md) or empty, where inhabitedness of a
+type is expressed using the
+[propositional truncation](foundation.propositional-truncations.md).
+
+## Definitions
+
+### The Curry–Howard interpretation of decidability
+
+```agda
+is-decidable-family : {l1 l2 : Level} {A : UU l1} (P : A → UU l2) → UU (l1 ⊔ l2)
+is-decidable-family {A = A} P = (x : A) → is-decidable (P x)
+
+decidable-family : {l1 : Level} (l2 : Level) → UU l1 → UU (l1 ⊔ lsuc l2)
+decidable-family l2 A = Σ (A → UU l2) is-decidable-family
+
+module _
+  {l1 l2 : Level} {A : UU l1} (P : decidable-family l2 A)
+  where
+
+  family-decidable-family : A → UU l2
+  family-decidable-family = pr1 P
+
+  is-decidable-decidable-family : is-decidable-family family-decidable-family
+  is-decidable-decidable-family = pr2 P
+```
+
+### The underlying decidable type family of a decidable subtype
+
+```agda
+decidable-family-decidable-subtype :
+  {l1 l2 : Level} {A : UU l1} → decidable-subtype l2 A → decidable-family l2 A
+decidable-family-decidable-subtype P =
+  ( is-in-decidable-subtype P , is-decidable-decidable-subtype P)
+```
+
+## Properties
+
+### Reindexing decidable type families
+
+```agda
+module _
+  {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} (P : decidable-family l3 B)
+  where
+
+  reindex-decidable-family : (f : A → B) → decidable-family l3 A
+  reindex-decidable-family f =
+    ( family-decidable-family P ∘ f , is-decidable-decidable-family P ∘ f)
+```
+
+### The negation of a decidable family is decidable
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (P : decidable-family l2 A)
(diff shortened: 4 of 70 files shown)


src/foundation/double-negation-images.lagda.md
```diff
@@ -0,0 +1,322 @@
...
+open import logic.double-negation-stable-embeddings
+```
+
+</details>
+
+## Idea
+
+The
+{{#concept "double negation image" Disambiguation="of a map" Agda=double-negation-im}}
+of `f : A → B` is the essentially unique type that factorizes `f` as a
```
VojtechStep, line 46: You write about an essentially unique type, but I see no reference to uniqueness (or universal properties) in this file. Honestly this paragraph is pretty cryptic, and I find it hard to believe it's at all understandable to many people outside Martín's sphere of influence 😅. Would you mind rephrasing it, and at least drawing a diagram?

src/foundation/inequality-booleans.lagda.md
```diff
@@ -0,0 +1,226 @@
...
+The
+{{#concept "standard inequality relation" Disambiguation="on booleans" Agda=}}
+on [booleans](foundation.booleans.md) is the inductively defined
+[relation](foundation.binary-relations.md) given by the four inequalities
+
+```text
+  false ≤ false
+  false ≤ true
+  true ≤ true
+  true ≰ false.
```
VojtechStep, line 43: Why would a non-inequality be a generator of an inductive relation? And since the relation is not literally given by the inequalities, wouldn't it be better to write what it actually is defined as, and then note an alternative way of looking at it? I think I noticed this trend of the prose and Agda matching only up to (sometimes unproven) equivalence

src/foundation/inequality-booleans.lagda.md
```diff
@@ -0,0 +1,226 @@
...
+open import order-theory.preorders
+open import order-theory.total-orders
+```
+
+</details>
+
+## Idea
+
+The
+{{#concept "standard inequality relation" Disambiguation="on booleans" Agda=}}
```
VojtechStep, line 35: Incomplete `Agda=`

src/foundation/inequality-booleans.lagda.md
```diff
@@ -0,0 +1,226 @@
...
+leq-left-or-bool : {x y : bool} → leq-bool x y → leq-bool (or-bool x y) y
+leq-left-or-bool {true} {true} p = star
+leq-left-or-bool {false} {true} p = star
+leq-left-or-bool {false} {false} p = star
+
+leq-right-or-bool : {x y : bool} → leq-bool x y → leq-bool (or-bool y x) y
+leq-right-or-bool {x} {true} p = star
+leq-right-or-bool {false} {false} p = star
+
+leq-or-bool'' :
```
VojtechStep, line 177: These should have better names, but I'm currently too tired to come up with any

VojtechStep (commented):
I managed to go through the first half of files, hopefully to be continued some time next week

src/foundation/decidable-dependent-function-types.lagda.md
```diff
@@ -46,6 +52,28 @@ is-decidable-Π-uniformly-decidable-family (inl a) (inr b) =
   inr (λ f → b a (f a))
 is-decidable-Π-uniformly-decidable-family (inr na) _ =
   inl (ex-falso ∘ na)
+
+is-decidable-prop-Π-uniformly-decidable-family :
```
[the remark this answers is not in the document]
fredrik-bakke: no, thanks for pointing it out

src/foundation/decidable-dependent-function-types.lagda.md
```diff
@@ -76,6 +104,53 @@ is-decidable-Π-Maybe {B = B} du de =
     ( is-decidable-product du de)
 ```
 
+### Dependent products of decidable propositions over a base with mere equality are decidable propositions
+
+Assuming the base `A` is empty or 0-connected, a dependent product of decidable
```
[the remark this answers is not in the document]
fredrik-bakke: The two are equivalent, but that equivalence has not been formalized. I'll add an explanation

src/foundation/decidable-maps.lagda.md
```diff
@@ -81,14 +83,13 @@ module _
...
 ```agda
-abstract
-  is-decidable-map-htpy :
-    {l1 l2 : Level} {A : UU l1} {B : UU l2} {f g : A → B} →
-    f ~ g → is-decidable-map g → is-decidable-map f
-  is-decidable-map-htpy H K b =
-    is-decidable-equiv
-      ( equiv-tot (λ a → equiv-concat (inv (H a)) b))
-      ( K b)
+is-decidable-map-htpy :
```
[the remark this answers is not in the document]
fredrik-bakke: I suppose I unabstracted it because it computes nicely. If you prefer it, I can make it `opaque` instead?

src/foundation/decidable-maps.lagda.md
```diff
@@ -279,3 +280,32 @@ module _
...
+  is-decidable-map f → ε-operator-map f
+ε-operator-map-is-decidable-map F = ε-operator-is-decidable ∘ F
+```
+
+### Decidable injective maps are embeddings
+
+**Proof.** Given a decidable map `f : A → B` then `f` decomposes
+`B ≃ (im f) + B∖(im f)`. Restricting to `im f` we have a section given by the
+Hilbert ε-operator on `f`. Now, by injectivity of `f` we know this restriction
+map is an equivalence. Hence, by 3-for-2 `f` is also an embedding.
```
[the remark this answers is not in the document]
fredrik-bakke: I'll add a diagram. 3-for-2 here is meant in the extended sense that 2 maps in a triangle are embeddings, and they are configured such that we may conclude the third is also an embedding. I'll rewrite

src/foundation/decidable-maps.lagda.md
```diff
@@ -279,3 +280,32 @@ module _
...
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} {f : A → B}
+  where
+
+  is-emb-is-injective-is-decidable-map :
+    is-decidable-map f → is-injective f → is-emb f
+  is-emb-is-injective-is-decidable-map H K =
+    is-emb-is-injective-ε-operator-map (ε-operator-map-is-decidable-map H) K
+```
+
+There is also an analogous proof using the double negation image.
```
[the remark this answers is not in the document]
fredrik-bakke: I couldn't formalize the thing here because of cyclic dependencies. The analogous proof may be preferred since it avoids propositional truncations. I'll update the prose

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+```agda
+decidable-family-decidable-subtype :
+  {l1 l2 : Level} {A : UU l1} → decidable-subtype l2 A → decidable-family l2 A
+decidable-family-decidable-subtype P =
+  ( is-in-decidable-subtype P , is-decidable-decidable-subtype P)
+```
+
+## Properties
+
+### Reindexing decidable type families
```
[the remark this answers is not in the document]
fredrik-bakke: base change?

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+  neg-decidable-family : decidable-family l2 A
+  neg-decidable-family =
+    ( ¬_ ∘ family-decidable-family P , is-decidable-neg-decidable-family)
+```
+
+### Composition of decidable families
+
+Given a decidable family of types with double negation dense equality
+`P : A → 𝒰` and a decidable type family `Q : (x : A) → P x → 𝒰` then we may
+_compose_ `Q` after `P` and obtain a decidabe type family `Q ∘ P : A → 𝒰`.
```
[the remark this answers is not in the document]
fredrik-bakke: this is an application of type duality. I'll add it to the explanation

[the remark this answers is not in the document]
fredrik-bakke: also there's a typo `decidabe`

src/foundation-core/decidable-propositions.lagda.md
```diff
@@ -251,38 +256,71 @@ type-neg-Decidable-Prop :
...
-  {l : Level} (A : UU l) →
-  is-decidable (type-trunc-Prop A) → is-merely-decidable A
-is-merely-decidable-is-decidable-trunc-Prop A (inl x) =
-  apply-universal-property-trunc-Prop x
-    ( is-merely-decidable-Prop A)
-    ( unit-trunc-Prop ∘ inl)
-is-merely-decidable-is-decidable-trunc-Prop A (inr f) =
-  unit-trunc-Prop (inr (f ∘ unit-trunc-Prop))
+
+  is-decidable-prop-function-type' :
```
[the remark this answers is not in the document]
fredrik-bakke: with decidable types there of course is relevant computational content: you may produce an element of your  proof relevant type. Decidable propositions, on the other hand, not so much.

src/foundation-core/decidable-propositions.lagda.md
```diff
@@ -251,38 +256,71 @@ type-neg-Decidable-Prop :
...
+    ( rec-coproduct
+      ( λ p → is-prop-function-type (is-prop-type-is-decidable-prop (K p)))
+      ( λ np → is-prop-is-contr (universal-property-empty-is-empty P np Q))
+      ( H)) ,
+    ( is-decidable-function-type' H (is-decidable-type-is-decidable-prop ∘ K))
+
+  is-decidable-prop-function-type :
+    is-decidable P → is-decidable-prop Q → is-decidable-prop (P → Q)
+  is-decidable-prop-function-type H K =
+    ( is-prop-function-type (is-prop-type-is-decidable-prop K)) ,
```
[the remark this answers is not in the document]
fredrik-bakke: I went with this proof because it is different, but either is fine. If you want I can change it.

src/foundation/cantors-theorem.lagda.md
```diff
@@ -101,21 +101,24 @@ module _
...
-        ( f x x)
-        ( iff-eq (ap prop-Decidable-Prop (htpy-eq α x)))
+  not-in-image-map-theorem-decidable-Cantor :
+    ¬ (fiber f map-theorem-decidable-Cantor)
+  not-in-image-map-theorem-decidable-Cantor (x , α) =
+    no-fixed-points-neg-Decidable-Prop
+      ( f x x)
+      ( iff-eq (ap prop-Decidable-Prop (htpy-eq α x)))
+
+  theorem-double-negation-dense-decidable-Cantor :
```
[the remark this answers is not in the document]
fredrik-bakke: How about adding them below an "alternative statements" main section? I agree the formulation with surjections and powersets is the main phrasing of Cantor's theorem

src/foundation/inequality-booleans.lagda.md
fredrik-bakke: oo shouldn't this give an error?

src/foundation/double-negation-images.lagda.md
```diff
@@ -0,0 +1,322 @@
...
+              is-double-negation-dense-unit-double-negation-im f
+                ( pr1 ip)
+                ( λ ηq →
+                  np
+                  ( pr1 ηq ,
+                    ap (inclusion-double-negation-im f) (pr2 ηq) ∙ pr2 ip))))) ,
+      ( refl))
+
+  double-negation-eliminating-map-inclusion-double-negation-im :
+    double-negation-im f →¬¬ X
```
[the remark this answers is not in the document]
fredrik-bakke: Yeah, you might be right. I could propose `¬¬→`, but maybe it is just better to use their textual name?

src/foundation/inequality-booleans.lagda.md
fredrik-bakke: It is one direction of the universal property of `or`, so something like `is-lower-bound-or-bool` should work

fredrik-bakke: sorry, upper bound*

src/foundation/cantors-theorem.lagda.md
```diff
@@ -109,13 +116,17 @@ module _
...
+      ¬ (is-double-negation-dense-map f)
+    theorem-double-negation-dense-decidable-Cantor H =
+      H map-theorem-decidable-Cantor not-in-image-map-theorem-decidable-Cantor
+
+    theorem-decidable-Cantor :
+      ¬ (is-surjective f)
+    theorem-decidable-Cantor =
+      map-neg
+        is-double-negation-dense-map-is-surjective
+        theorem-double-negation-dense-decidable-Cantor
```
VojtechStep:
```suggestion
      map-neg
        ( is-double-negation-dense-map-is-surjective)
        ( theorem-double-negation-dense-decidable-Cantor)
```

and similar parenthetization issues later in this file

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+
+### The negation of a decidable family is decidable
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} (P : decidable-family l2 A)
+  where
+
+  is-decidable-neg-decidable-family :
+    is-decidable-family (¬_ ∘ family-decidable-family P)
```
VojtechStep: I think this would be more readable with the composition spelled-out

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+  is-decidable-neg-decidable-family =
+    is-decidable-neg ∘ is-decidable-decidable-family P
+
+  neg-decidable-family : decidable-family l2 A
+  neg-decidable-family =
+    ( ¬_ ∘ family-decidable-family P , is-decidable-neg-decidable-family)
```
VojtechStep: Same here

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+### Composition of decidable families
+
+Given a decidable family of types with double negation dense equality
+`P : A → 𝒰` and a decidable type family `Q : (x : A) → P x → 𝒰` then we may
+_compose_ `Q` after `P` and obtain a decidable type family `Q ∘ P : A → 𝒰`.
```
VojtechStep: It would be nice to spell out the fibers of the composition, because it's not immediately clear that you're supposed to form sigma types

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+
+```agda
+module _
+  {l1 l2 l3 : Level} {A : UU l1}
+  where
+
+  is-decidable-comp-decidable-family-decidable-subtype' :
```
VojtechStep: I don't think this and the following definition are related to subtypes specifically?

src/foundation/decidable-type-families.lagda.md
```diff
@@ -0,0 +1,166 @@
...
+      ( decidable-family-decidable-subtype P)
+      ( Q)
+      ( λ x p q →
+        intro-double-negation
+          ( eq-is-prop (is-prop-is-in-decidable-subtype P x)))
+```
+
+### Decidable families on the subuniverse of propositions
+
+> TODO
```
VojtechStep: Could you spell out a bit more what exactly you expect to be here?

src/foundation/inequality-booleans.lagda.md
```diff
@@ -0,0 +1,226 @@
...
+bool-Poset : Poset lzero lzero
+bool-Poset = (bool-Preorder , (λ x y → antisymmetric-leq-bool {x} {y}))
+
+bool-Total-Order : Total-Order lzero lzero
+bool-Total-Order = (bool-Poset , is-total-leq-bool)
+
+bool-Decidable-Total-Order : Decidable-Total-Order lzero lzero
+bool-Decidable-Total-Order =
+  ( bool-Poset ,
+    ( is-total-leq-bool) ,
```
VojtechStep:
```suggestion
    is-total-leq-bool ,
```

This is the most consistent with this file's style, but I think it's still inconsistent with other files in this PR.

src/foundation/irrefutable-propositions.lagda.md
VojtechStep, line 1: I'd expect irrefutable types and irrefutable propositions to live in the same namespace, or types to live in foundation and props to live in logic; what's the rationale here?

src/foundation/irrefutable-propositions.lagda.md
```diff
@@ -9,28 +9,24 @@ module foundation.irrefutable-propositions where
...
 ```
 
 </details>
 
 ## Idea
 
 The [subuniverse](foundation.subuniverses.md) of
-{{#concept "irrefutable propositions" Agda=Irrefutable-Prop}} consists of
+{{#concept "irrefutable propositions" Agda=Irrefutable-Prop}}, or **double
+negation dense propositions**, consists of
```
VojtechStep: "Double negation dense propositions" should be a concept, since you're referring to it elsewhere

src/foundation/iterating-functions.lagda.md
```diff
@@ -2,6 +2,8 @@
 
 ```agda
 module foundation.iterating-functions where
+
+open import foundation-core.iterating-functions public
```
VojtechStep, line 6: From browsing the files, it looks like this split is required because of `logic.propositionally-decidable-maps` (since it's the only file importing the foundation-core module), but what's the dependency cycle there? Propositionally decidable maps look almost like a leaf

src/foundation/mere-equality.lagda.md
```diff
@@ -45,6 +51,13 @@ module _
   is-prop-mere-eq x y = is-prop-type-trunc-Prop
 ```
 
+### Types whose elements are merely equal
```
VojtechStep:
```suggestion
### Types whose elements are all merely equal
```

src/foundation/propositional-truncations.lagda.md
```diff
@@ -246,7 +265,7 @@ module _
 
   abstract
     map-idempotent-trunc-Prop :
-      type-trunc-Prop (type-trunc-Prop A) → type-trunc-Prop A
+      ║ (║ A ║₋₁) ║₋₁ → ║ A ║₋₁
```
VojtechStep, line 268: If nested truncation needs to be written like this, then it should probably be mentioned in the note on notation

src/foundation/uniformly-decidable-type-families.lagda.md
```diff
@@ -173,3 +186,76 @@ module _
...
+        ( is-uniformly-decidable-family-has-double-negation-dense-equality-base ∘
+          inl)
+        ( |a|)
+    is-uniformly-decidable-family-has-double-negation-dense-equality-base'
+      ( inr na) K =
+      is-uniformly-decidable-family-has-double-negation-dense-equality-base
+        ( inr na)
+```
+
+### A family of decidable propositions over a decidable base with mere equality is uniformly decidable
```
VojtechStep, line 235: "base with mere equality"

src/logic/complements-decidable-subtypes.lagda.md
```diff
@@ -98,3 +107,31 @@ module _
...
+        inv-right-unit-law-Σ-is-contr
+          ( λ x →
+            is-proof-irrelevant-is-prop
+              ( is-prop-is-decidable (is-prop-is-in-decidable-subtype P x))
+              ( is-decidable-decidable-subtype P x))
+    ≃ type-decidable-subtype P + type-complement-decidable-subtype P
+      by
+        left-distributive-Σ-coproduct A
+          ( is-in-decidable-subtype P)
+          ( is-in-complement-decidable-subtype P)
```
VojtechStep:
```suggestion
      A
      ≃ Σ A (λ x → is-decidable (is-in-decidable-subtype P x))
        by
        inv-right-unit-law-Σ-is-contr
          ( λ x →
            is-proof-irrelevant-is-prop
              ( is-prop-is-decidable (is-prop-is-in-decidable-subtype P x))
              ( is-decidable-decidable-subtype P x))
      ≃ type-decidable-subtype P + type-complement-decidable-subtype P
        by
        left-distributive-Σ-coproduct A
          ( is-in-decidable-subtype P)
          ( is-in-complement-decidable-subtype P)
```

This conforms to the codestyle guide.

src/logic/dirk-gentlys-principle.lagda.md
```diff
@@ -0,0 +1,74 @@
...
+[De Morgan's law](logic.de-morgans-law.md), Section 8.5 {{#cite Diener18}}.
+
+> The name is based on the guiding principle of the protagonist of Douglas
+> Adam’s novel Dirk Gently’s Holistic Detective Agency who believes in the
+> “fundamental interconnectedness of all things.” {{#cite Diener18}}
+
+## Statement
+
+```agda
+instance-prop-Dirk-Gently's-Principle :
```
VojtechStep: From what I found, we don't use the possessive `'s` in Agda identifiers, e.g. we have `structure-Hatcher-Acyclic-Type` for Hatcher's acyclic type, or `double-negation-Peirces-law` for irrefutability of Peirce's law.
Personally I prefer not using `'s`.

src/logic/dirk-gentlys-principle.lagda.md
```diff
@@ -0,0 +1,74 @@
...
+Dirk-Gently's-Principle =
+  {l1 l2 : Level} → Dirk-Gently's-Principle-Level l1 l2
+```
+
+## Properties
+
+### The law of excluded middle implies Dirk Gently's principle
+
+**Proof.** Assuming the law of excluded middle, then every proposition is either
+true or false, so since false ≤ true, we are done.
```
VojtechStep: I feel like parsing this line of thought is actually more difficult than coming up with a proof on the spot

src/logic/irrefutable-types.lagda.md
```diff
@@ -0,0 +1,167 @@
...
+open import logic.double-negation-elimination
+```
+
+</details>
+
+## Idea
+
+The [subuniverse](foundation.subuniverses.md) of
+{{#concept "irrefutable types" Agda=Irrefutable-Type}}, or **double negation
+dense types**, consists of types `X` for which the
```
VojtechStep: "double negation dense types" should be a concept

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+```
+
+</details>
+
+## Idea
+
+A map `f : A → B` is
+{{#concept "double negation dense" Agda=is-double-negation-dense-map}}, if all
+of its [fibers](foundation-core.fibers-of-maps.md) are
+[irrefutable](foundation.irrefutable-propositions.md). I.e., for every `y : B`,
```
VojtechStep: You're linking to `foundation.irrefutable-propositions`, and defining the predicate in terms of `double-negation-type-Prop`. Wouldn't it be better to link to irrefutable types and then use `is-irrefutable-prop-Type` from that module, to have the prose and formalization synced?

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} (f : A → B) →
+  is-prop (is-double-negation-dense-map f)
+is-prop-is-double-negation-dense-map f =
+  is-prop-type-Prop (is-double-negation-dense-map-Prop f)
+```
+
+### The type of double negation dense maps
+
+```agda
+infix 5 _↠¬¬_
```
VojtechStep, line 94: I think it would be more in line with the "infix operators are mostly notation which shouldn't appear in identifiers" guideline if the primary definition was `double-negation-dense-map`, and then the infix symbol was defined as that. That's the way it's done for e.g. existential quantification and equivalences

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+open import logic.double-negation-stable-embeddings
+open import logic.irrefutable-types
+```
+
+</details>
+
+## Idea
+
+A map `f : A → B` is
+{{#concept "double negation dense" Agda=is-double-negation-dense-map}}, if all
```
VojtechStep: Is this "dense" the same as in "j-dense propositions" from Modalities in HoTT? I spent some time trying to connect the two, but couldn't make it work. If we follow the usual naming where "a map is X if all fibers are X", then it would mean that `is-dense X := ◯ X`, which doesn't look right. If anything, I'd expect this concept to be called "irrefutable maps".
In any case I think it deserves a note, since we already have definitions for double negation stable types and separated types, which follow the established nomenclature (I'm not particularly knowledgeable in this field though, maybe it's not the case that "stable", "dense", "connected" and "separated" have the same level of ubiquity).

src/foundation/irrefutable-equality.lagda.md
```diff
@@ -0,0 +1,162 @@
...
+  is-prop-irrefutable-eq x y = is-prop-double-negation
+
+  irrefutable-eq-Prop : A → A → Prop l
+  irrefutable-eq-Prop x y = (irrefutable-eq x y , is-prop-irrefutable-eq x y)
+```
+
+### Types with double negation dense equality
+
+```agda
+has-double-negation-dense-equality : {l : Level} → UU l → UU l
```
VojtechStep: I comment on this later in `double-negation-dense-maps`, but I'm really unsure where this "dense" adjective comes from. It doesn't seem to be the same "dense" as in the context of modalities (at least to me)

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+
+  is-double-negation-dense-map-double-negation-dense-map :
+    is-double-negation-dense-map map-double-negation-dense-map
+  is-double-negation-dense-map-double-negation-dense-map = pr2 f
+```
+
+### The type of all double negation dense maps out of a type
+
+```agda
+Double-Negation-Dense-Map :
```
VojtechStep, line 113: Why does the default notion bundle the codomain? I don't see references to this definition outside of this file.

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+  {l1 : Level} (l2 : Level) → UU l1 → UU (l1 ⊔ lsuc l2)
+Double-Negation-Dense-Map l2 A = Σ (UU l2) (A ↠¬¬_)
+
+module _
+  {l1 l2 : Level} {A : UU l1} (f : Double-Negation-Dense-Map l2 A)
+  where
+
+  type-Double-Negation-Dense-Map : UU l2
```
VojtechStep, line 121: I don't understand why the codomain has the special status of "the underlying type"

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+    is-double-negation-dense-map g →
+    is-double-negation-dense-map (map-product f g)
+  is-double-negation-dense-map-product {f} {g} F G (c , d) =
+    map-double-negation
+      ( map-inv-compute-fiber-map-product f g (c , d))
+      ( is-irrefutable-product (F c) (G d))
+
+  double-negation-dense-map-product :
+    (A ↠¬¬ C) → (B ↠¬¬ D) → ((A × B) ↠¬¬ (C × D))
+  double-negation-dense-map-product f g =
```
VojtechStep: I think this would be more readable with coprojections

src/logic/double-negation-dense-maps.lagda.md
```diff
@@ -0,0 +1,516 @@
...
+  is-equiv-htpy-eq-double-negation-dense-map =
+    fundamental-theorem-id
+      is-torsorial-htpy-double-negation-dense-map
+      htpy-eq-double-negation-dense-map
+
+  extensionality-double-negation-dense-map :
+    (g : A ↠¬¬ B) → (f ＝ g) ≃ htpy-double-negation-dense-map g
+  extensionality-double-negation-dense-map g =
+    ( htpy-eq-double-negation-dense-map g ,
+    is-equiv-htpy-eq-double-negation-dense-map g)
```
VojtechStep:
```suggestion
      is-equiv-htpy-eq-double-negation-dense-map g)
```

I think?

src/logic/double-negation-dense-subtypes.lagda.md
```diff
@@ -0,0 +1,143 @@
...
+
+open import logic.double-negation-dense-maps
+```
+
+</details>
+
+## Idea
+
+A
+{{#concept "double negationdense" Disambiguation="subtype" Agda=is-double-negation-dense-subtype Agda=double-negation-dense-subtype}}
```
VojtechStep:
```suggestion
{{#concept "double negation dense" Disambiguation="subtype" Agda=is-double-negation-dense-subtype Agda=double-negation-dense-subtype}}
```

src/logic/double-negation-dense-subtypes.lagda.md
```diff
@@ -0,0 +1,143 @@
...
+  is-double-negation-dense-inclusion-is-double-negation-dense-subtype H x =
+    map-double-negation (λ y → ((x , y) , refl)) (H x)
+
+  double-negation-dense-inclusion-is-double-negation-dense-subtype :
+    is-double-negation-dense-subtype P → type-subtype P ↠¬¬ A
+  double-negation-dense-inclusion-is-double-negation-dense-subtype H =
+    ( inclusion-subtype P ,
+      is-double-negation-dense-inclusion-is-double-negation-dense-subtype H)
+
+  is-dense-is-double-negation-dense-inclusion-subtype :
```
VojtechStep: I'm not necessarily opposed to this name, but it's not consistent with the rest of the names in this section

src/logic/double-negation-elimination.lagda.md
```diff
@@ -171,24 +172,97 @@ double-negation-elim-neg :
 double-negation-elim-neg A f p = f (ev p)
 ```
 
+### Double negation elimination for dependent sums of types with double negation elimination over a double negation stable proposition
```
VojtechStep, line 176: The title speaks about "double negation stable propositions", but the definitions are named and implemented in terms of double negation elimination instead.

src/logic/markovian-types.lagda.md
```diff
@@ -91,6 +82,10 @@ module _
 
 > This remains to be formalized.
 
+### Types with decidability search are markovian
```
VojtechStep:
```suggestion
### Types with decidability search are Markovian
```
But also I don't think we have anything about searches formalized, and I'm not sure what a "decidability search" is. Could you elaborate?

src/logic/propositional-double-negation-elimination.lagda.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+### If the negation of a type with double negation elimination is decidable, then the type is merely decidable
+
+```agda
+module _
+  {l1 : Level} {A : UU l1}
+  where
+
+  is-inhabited-or-empty-is-decidable-neg-has-prop-double-negation-elim :
```
VojtechStep, line 137: The header says "merely decidable", but the formalization uses `is-inhabited-or-empty`

src/logic/propositional-double-negation-elimination.lagda.md
```diff
@@ -0,0 +1,269 @@
...
+    ( inl nx) =
+    inr nx
+  is-inhabited-or-empty-is-decidable-neg-has-prop-double-negation-elim f
+    ( inr nnx) =
+    inl (f nnx)
+```
+
+**Remark.** It is an established fact that both the property of satisfying
+double negation elimination, and the property of having decidable negation, are
+strictly weaker conditions than being decidable. Therefore, this result
```
VojtechStep: I don't understand how this proves independence. Also could you include a source for the "established fact"?

src/logic/propositional-double-negation-elimination.lagda.md
```diff
@@ -0,0 +1,269 @@
...
+
+```agda
+has-prop-double-negation-elim-has-double-negation-elim :
+  {l : Level} {A : UU l} →
+  has-double-negation-elim A →
+  has-prop-double-negation-elim A
+has-prop-double-negation-elim-has-double-negation-elim H = unit-trunc-Prop ∘ H
+```
+
+### Propositional double negation elimination for merely decidable types
```
VojtechStep: Heading says "merely decidable", formalization says `is-inhabited-or-empty`

src/logic/propositionally-decidable-maps.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+open import foundation-core.iterating-functions
+
+open import logic.propositionally-decidable-types
+```
+
+</details>
+
+## Idea
+
+A [map](foundation-core.function-types.md) is said to be
```
VojtechStep, line 35: "Propositionally decidable maps/types" don't actually appear to ever be mentioned in prose outside of their files? You usually use "merely decidable", wouldn't it then be better to name the concept that? Especially since the corresponding definition is `is-inhabited-or-empty(-map)?`, not `is-propositionally-decidable(-map)?`

src/logic/propositionally-decidable-maps.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+  is-inhabited-or-empty-map-htpy H K b =
+    is-inhabited-or-empty-equiv
+      ( equiv-tot (λ a → equiv-concat (inv (H a)) b))
+      ( K b)
+```
+
+### Decidable maps are propositionally decidable
+
+```agda
+is-inhabited-or-empty-map-is-decidable-map :
```
VojtechStep, line 91: Is this definition intentionally non-abstract, in contrast with the other functions?

src/logic/propositionally-double-negation-eliminating-maps.lagda.md
```diff
@@ -0,0 +1,124 @@
...
+</details>
+
+## Idea
+
+A [map](foundation-core.function-types.md) is said to be
+{{#concept "propositionally double negation eliminating" Disambiguation="map of types" Agda=is-prop-double-negation-eliminating-map}}
+if its [fibers](foundation-core.fibers-of-maps.md) satisfy
+[propositional double negation elimination](logic.propositional-double-negation-elimination.md).
+I.e., for every `y : B`, if `fiber f y` is
+[irrefutable](foundation.irrefutable-propositions.md), then we have then the
```
VojtechStep:
```suggestion
[irrefutable](foundation.irrefutable-propositions.md), then we have that the
```

src/foundation/decidable-maps.lagda.md
```diff
@@ -279,3 +280,32 @@ module _
       ( retract-fiber-retract-map f g R x)
       ( G (map-codomain-inclusion-retract-map f g R x))
 ```
+
+### Decidable maps have Hilbert ε-operators
```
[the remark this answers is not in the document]
fredrik-bakke: The link can be found right below, at `ε-operator-map f`, but the sentiment of your comment seems to rather be that the concept should be explained here, which I agree with, so I'll do that and in the process insert a link in the prose.

src/logic/propositionally-decidable-maps.lagda.md
fredrik-bakke: While I can't recall my rationale for this terminology anymore, notice that there are multiple concepts one might refer to as "merely decidable":
- `║ is-decidable-map f ║₋₁`
- `(x : A) → ║ is-decidable (fiber f x) ║₋₁`
- `(x : A) → is-inhabited-or-empty (fiber f x)`

while the last two are equivalent, the first one is a stronger assumption which is implied by choice. I don't know if perhaps I reserved "merely decidable map" for the first one, I'll have to think about it

src/foundation/decidable-maps.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: when it comes to decidability of types, I believe it is in general beneficial if they compute, so it is good to have the option to unfold them. For instance, one might want to try and normalize them so the proof assistant can tell you wheter a given case is true or false, even if one does not care about the particular witness of truth.

src/foundation/double-negation-images.lagda.md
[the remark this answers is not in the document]
fredrik-bakke: what do you think?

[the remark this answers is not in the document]
fredrik-bakke: I just find the name `double-negation-eliminating-map` to be a bit verbose, compared to some sort of special notation.

src/foundation/double-negation-images.lagda.md
fredrik-bakke: I expanded a bunch on the idea section, let me know if it is satisfactory now or if I included something new that is confusing.

src/foundation/decidable-maps.lagda.md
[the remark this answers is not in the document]
VojtechStep: Sure, `opaque` works

src/foundation/double-negation-images.lagda.md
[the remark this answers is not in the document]
VojtechStep: I'm on my phone so I can't easily check, but I don't remember seeing this type showing up so often that using the verbose name would be too annoying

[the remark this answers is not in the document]
fredrik-bakke: Well, it's part of a larger notation scheme, and some of the others might actually see some fair use. There is also double negation dense embeddings and double negation covers. These are nice because you can use them at times to avoid prop truncs and by extension funext. Since you don't like this notation we should reconsider the notation for all of them.

For decidable maps we use the superscript d annotation. We could do something similar and use superscript dn for instance.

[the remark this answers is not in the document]
fredrik-bakke: Sorry, I should've thought of this before making my original comment

src/logic/dirk-gentlys-principle.lagda.md
fredrik-bakke: Do you want me to remove it or rephrase?

src/foundation/irrefutable-propositions.lagda.md
fredrik-bakke: These indiscrepancies are largely due to attempting to avoid moving files that were already there, in anticipation of a fix to our accrediting system. I'm on my phone right now so can't check this particular case, but that's my guess.

src/foundation/irrefutable-propositions.lagda.md
fredrik-bakke: Even when it is only an alternative name for the same concept? My convention har been to only use #concept once per concept, and not per term.

src/foundation/propositional-truncations.lagda.md
fredrik-bakke: It probably doesn't need to, actually. Do you want me to remove the parentheses?

fredrik-bakke: Oh, you mean as a convention? I'm not sure, I just happened to prefer it here...

src/logic/propositionally-decidable-maps.lagda.md
fredrik-bakke: Probably yes. Cf my other comment on unfolding

src/foundation/double-negation-images.lagda.md
VojtechStep: Much better, thank you 👌

src/foundation/double-negation-images.lagda.md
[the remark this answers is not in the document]
VojtechStep: I wasn't aware of the superscript notations. In retrospect, I think the current operator is okay as-is, and the problem was rather that I was reviewing on Github. It should be clearer in an editor or on the website.

src/foundation/irrefutable-equality.lagda.md
fredrik-bakke: "Dense" here means "dense with respect to the double negation modality", precisely in the sense you are referring to.

src/foundation/irrefutable-propositions.lagda.md
VojtechStep: I would use it for anything that you use as a defined concept in the prose.

src/logic/markovian-types.lagda.md
fredrik-bakke: It is in the next next PR

src/foundation/propositional-truncations.lagda.md
VojtechStep: No, I thought Agda parsing required it to be written with parentheses, which could surprise others. If it's not necessary, then it's fine the way it is

src/logic/dirk-gentlys-principle.lagda.md
VojtechStep: Removing is fine by me. I think rewriting it would take about the same amount of effort as formalizing it 😃

src/foundation/double-negation-images.lagda.md
VojtechStep: Well, actually the link between "X-dense" and "X-stable" for X being double negation and X being a generic modality is still a little hazy to me, as I asked in https://github.com/UniMath/agda-unimath/pull/1387#discussion_r2126802598, so I can't really tell how this uniqueness condition follows

src/foundation/irrefutable-equality.lagda.md
fredrik-bakke: For proposition-valued modalities ○, "dense" reduces to `○ X` has an element, hence why this makes sense.

fredrik-bakke: I added an explanation

src/logic/dirk-gentlys-principle.lagda.md
fredrik-bakke: Added

src/logic/double-negation-dense-maps.lagda.md
fredrik-bakke: This mirrors the role of `Surjection` in `foundation.surjective-maps`.

src/logic/double-negation-dense-maps.lagda.md
fredrik-bakke: it's what was done in the case of surjections.

src/logic/double-negation-elimination.lagda.md
fredrik-bakke: double negation stability and double negation elimination coincides for propositions.

src/logic/propositionally-decidable-maps.lagda.md
fredrik-bakke: I should also note that `is-inhabited-or-empty (fiber f x)` computes more than `║ is-decidable (fiber f x) ║₋₁`, which is why the former is preferred in the formalizations.

src/logic/double-negation-dense-maps.lagda.md
fredrik-bakke: Yes, this is "dense" in the same sense as "j-dense-propositions". I've added a note on terminology. You can find all of the terms "stable", "dense", "connected" and "separated" defined in _Modalities in Homotopy Type Theory_.

fredrik-bakke: Let me know if further elaboration is required.

src/foundation/double-negation-images.lagda.md
fredrik-bakke: A type Y is X-dense and X-stable if and only if Y is contractible.

fredrik-bakke: Since X defines an orthogonal factorization system, any factorization through the X-dense image is unique. That is the definition of the factorization system being* _orthogonal_. On the other hand, "X-stable" can be understood as satisfying a uniqueness principle with respect to X, and "X-dense" can be understood as satisfying inhabitedness with respect to X.
Hope this clarifies your confusion.

src/logic/propositional-double-negation-elimination.lagda.md
fredrik-bakke: The connection should be clear. Since both conditions are strictly weaker, then if they were dependent of eachother assuming both would also be strictly weaker, but here we demonstrate that if we assume both we get something that is not strictly weaker.

fredrik-bakke: Added a citation for this established fact.

src/foundation/iterating-functions.lagda.md
fredrik-bakke: This file was probably split preemptively so that multiplication can be compared to iterated addition and exponentiation can be compared to iterated multiplication and so on.

fredrik-bakke:
@VojtechStep Thank you for the detailed review! I've tried to address all of your comments now. Let me know if further changes are required.

VojtechStep:
I might not have the time today, but I should be able to look at all the updated PRs tomorrow

VojtechStep:
Eh, sorry, it will have to wait one more day 😬

fredrik-bakke:
No worries!

fredrik-bakke:
FYI I won't have time to work on any additional review before monday next week at the earliest.

src/logic/propositional-double-negation-elimination.lagda.md
VojtechStep: Okay that's clear, thanks

src/logic/double-negation-dense-maps.lagda.md
VojtechStep: Thanks so much for those notes, it makes the connection very clear 💯

VojtechStep:
No additional changes necessary, queued for merging 🚀

PR #1343, opened 2025-02-22
Cauchy completeness of the Dedekind real numbers

lowasser:
Depends on addition (#1336), so inevitably a long ways off, but this certainly feels like a major step towards real analysis, with the ability to define limits.

PR #1391, opened 2025-04-01
Addition on the real numbers is uniformly continuous

lowasser:
Incorporates some components of #1343 .


src/metric-spaces/monotonic-premetric-structures.lagda.md
```diff
@@ -57,4 +62,29 @@ module _
...
+  abstract
+    is-weakly-monotonic-is-monotonic-Premetric :
+      is-monotonic-Premetric → is-weakly-monotonic-Premetric
+    is-weakly-monotonic-is-monotonic-Premetric H x y d₁ d₂ d₁≤d₂ x~y =
+      rec-coproduct
+        ( λ d₁<d₂ → H x y d₁ d₂ d₁<d₂ x~y)
+        ( λ d₂≤d₁ →
+          tr
+            ( λ d → type-Prop (B d x y))
+            ( eq-ℚ⁺
```
lowasser, line 80: turn this into a general lemma about antisymmetry of leq-ℚ⁺

PR #1392, opened 2025-04-02
The maximum function on the real numbers is uniformly continuous

lowasser:
Builds on #1391 .


lowasser:
Rewriting this completely.

PR #1396, opened 2025-04-04
Limits of Cauchy sequences are unique

lowasser:
Just a little PR that will be necessary to prove things like "the limit of this sequence _is_ this value."


lowasser:
Going to wait until I have something bigger to fold this into.

PR #1388, opened 2025-03-28
Monotonic sequences in posets

malarbol:
This PR introduces a few definitions and results to work with sequences is preorders, posets or strictly preordered sets.
As an application, it introduces the concept of `subsequences` of a sequence `u`: the sequences `u ∘ f` for some strictly increasing map `f : ℕ → ℕ`.

PR #1371, opened 2025-03-19
Bernoulli's inequality on the positive rational numbers

malarbol:
This PR implements some of my suggestions from https://github.com/UniMath/agda-unimath/issues/1354#issuecomment-2731327699 

- `group-theory.arithmetic-sequences-semigroups`:
  - sequences in a semigroup with a common difference: `uₙ₊₁ = uₙ + d`;
  - standard arithmetic sequences given by initial value and common difference;
  - homotopy between arithmetic sequences with same initial value and common difference.
  
 - `elementary-number-theory.arithmetic-sequences-positive-rational-numbers`:
   -  arithmetic sequences in the additive semigroup of positive rational numbers;
   - basic properties;
   - computational rule `uₙ = u₀ + n d`;
   - an arithmetic sequence in ℚ⁺ is strictly increasing.

 - `elementary-number-theory.geometric-sequences-positive-rational-numbers`:
   - arithmetic sequences in the multiplicative semigroup of positive rational numbers;
   - basic properties;
   -  computational rule `uₙ = u₀ rⁿ `;
   - constant sequences are geometric;
   - geometric sequences with common ratio `r > 1` are strictly increasing;
   - multiplication of geometric sequences in ℚ⁺;
   - inversion of geometric sequences in ℚ⁺.

- `elementary-number-theory.bernoullis-inequality-positive-rational-numbers`: 
  - `∀ (h : ℚ⁺) (n : ℕ) → 1 + (n + 1)h ≤ (1 + n h)(1 + h)`;
  - Bernoulli's inequality in ℚ⁺: `∀ (h : ℚ⁺) (n : ℕ) → 1 + n h ≤ (1 + h)ⁿ`.

PR #1399, opened 2025-04-07
sequences large preorders/posets

malarbol:
We can discuss how much of https://github.com/UniMath/agda-unimath/pull/1388/ we want in large posets. I guess one of the main example would be sequences of real numbers so I'll try to keep that in mind.  When we merge https://github.com/UniMath/agda-unimath/pull/1371/ I could work on arithmetic sequences of positive real numbers.


malarbol:
Hey @fredrik-bakke. If you see anything reviewing #1388 that you think should be transposed here, feel free to make a list and I'll do my best to fill the blanks.

fredrik-bakke (approved):
Looks good

PR #1400, opened 2025-04-09
Distance between rational numbers

lowasser:


lowasser:
For posterity, this is intended as a step towards multiplication on reals, bounding the width of the interval produced [here](https://unimath.github.io/agda-unimath/elementary-number-theory.closed-intervals-rational-numbers.html#5699).  If we can make that interval smaller than an epsilon of our choice, we can get that the product of reals is arithmetically located, the final step to proving it's a real number.

src/elementary-number-theory/distance-rational-numbers.lagda.md
```diff
@@ -0,0 +1,325 @@
...
+      ( swap-right-diff-leq-ℚ
+        ( p)
+        ( q)
+        ( ε)
+        ( transitive-leq-ℚ
+          ( p -ℚ q)
+          ( rational-dist-ℚ p q)
+          ( ε)
+          ( |p-q|≤ε)
+          ( leq-diff-dist-ℚ p q))))
```
fredrik-bakke, line 216:
```suggestion
    ( leq-transpose-left-diff-ℚ
      ( q)
      ( ε)
      ( p)
      ( swap-right-diff-leq-ℚ
        ( q)
        ( p)
        ( ε)
        ( transitive-leq-ℚ
          ( q -ℚ p)
          ( rational-dist-ℚ p q)
          ( ε)
          ( |p-q|≤ε)
          ( leq-reversed-diff-dist-ℚ p q)))) ,
    ( leq-transpose-left-diff-ℚ
      ( p)
      ( ε)
      ( q)
      ( swap-right-diff-leq-ℚ
        ( p)
        ( q)
        ( ε)
        ( transitive-leq-ℚ
          ( p -ℚ q)
          ( rational-dist-ℚ p q)
          ( ε)
          ( |p-q|≤ε)
          ( leq-diff-dist-ℚ p q))))
```

src/elementary-number-theory/distance-rational-numbers.lagda.md
```diff
@@ -0,0 +1,325 @@
...
+        ( q)
+        ( p)
+        ( ε)
+        ( concatenate-leq-le-ℚ
+          ( q -ℚ p)
+          ( rational-dist-ℚ p q)
+          ( ε)
+          ( leq-reversed-diff-dist-ℚ p q)
+          ( |p-q|<ε))) ,
+      le-transpose-left-diff-ℚ
```
fredrik-bakke, line 261:
```suggestion
          ( |p-q|<ε)))) ,
    ( le-transpose-left-diff-ℚ
```

fredrik-bakke (approved):
Nice work!

PR #1403, opened 2025-04-11
total metric spaces

malarbol:
Some work on **total metric spaces**: metric spaces such that all elements are at bounded distance.
I think it's related to **distance functions**:
-  the existence of a distance function will make a metric space total;
-  maybe there's some kind of converse, where we can use the inhabitedness of the set of upper bound on the distance to define some (lower|upper|)real that would effectively act as a distance? 

I think we'll need this as an hypothesis for [Banach fixed point theorem](https://en.wikipedia.org/wiki/Banach_fixed-point_theorem) (or at least, the proof I have in mind uses it) so that's certainly a concept we might want to add at some point.

### DIFF
--- a/src/metric-spaces.lagda.md
+++ b/src/metric-spaces.lagda.md
@@ -90,6 +90,7 @@ open import metric-spaces.short-functions-metric-spaces public
 open import metric-spaces.short-functions-premetric-spaces public
 open import metric-spaces.subspaces-metric-spaces public
 open import metric-spaces.symmetric-premetric-structures public
+open import metric-spaces.total-metric-spaces public
 open import metric-spaces.triangular-premetric-structures public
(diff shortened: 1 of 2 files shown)


fredrik-bakke:
> We can define an upper real that acts as a distance; if we want to get it into a conventional real number I'm uncertain what conditions are required. I'd expect at least the need for decidability of the premetric.

In case you weren't aware of this resource, here's a paper that considers some arithmetic on real numbers using upper and lower reals. It might have some answers for you, but I don't recall the details

https://www.researchgate.net/publication/362446739_Point-free_Construction_of_Real_Exponentiation
(Someone posted about it in #analysis in the Discord a short while ago)

malarbol:
This is now part of #1417.

PR #1405, opened 2025-04-15
More precise Archimedean properties

malarbol:
Definitions for the tightest Archimedean bound.
Application: the integer part of a rational number (floor function).

### DIFF
--- a/src/elementary-number-theory.lagda.md
+++ b/src/elementary-number-theory.lagda.md
@@ -21,6 +21,7 @@ open import elementary-number-theory.additive-group-of-rational-numbers public
 open import elementary-number-theory.archimedean-property-integer-fractions public
 open import elementary-number-theory.archimedean-property-integers public
 open import elementary-number-theory.archimedean-property-natural-numbers public
+open import elementary-number-theory.archimedean-property-positive-integers public
 open import elementary-number-theory.archimedean-property-positive-rational-numbers public
 open import elementary-number-theory.archimedean-property-rational-numbers public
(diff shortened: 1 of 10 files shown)


malarbol:
I'm working towards the floor function on the rational numbers. @lowasser do you think there's any hope of extending it to the real numbers (e.g. using that it's locally constant on the right or something)? Or to the positive real numbers?

lowasser:
I'm actively certain that it's not possible to extend to the real numbers (without choice principles).

malarbol:
> I'm actively certain that it's not possible to extend to the real numbers (without choice principles).

Ok. I was afraid so. Thanks.

fredrik-bakke:
Unless you consider that a choice principle

malarbol:
> Unless you consider that a choice principle

I think I kind of do... 😅 Anyways, at some point I guess it will be useful to study more precisely the relations between different omniscience / choice principles and their applications to real analysis. But I already  have a lot in mind, so I'll try to take things slowly.

malarbol:
I finally realized it's just a convoluted way to talk about euclidean division of integers. I'll close this and start over 😅.

fredrik-bakke:
Here's one useful fact I'd like to highlight again. One of the fundamental differences between metric spaces and closeness spaces, which I've mentioned a couple of times in the past, is that there is a WLPO that holds constructively for the closeness function on a closeness space, but this principle does not allow you to construct discontinuous functions. Hence closeness spaces behave better constructively compared to metric spaces, at least in this regard.

PR #1409, opened 2025-04-25
Correct typos in set-theory.countable-sets

ecavallo:


fredrik-bakke (approved):
Hi Evan, thank you for the corrections!

PR #1413, opened 2025-04-25
Link styling tweaks

fredrik-bakke:
## Summary

- [X] ~Add dotted underlines under links that are adjacent making it possible to tell them apart, e.g.:
image
image
  The styling does not apply to code, or concept paragraph symbols, and it only applies to links with another link directly adjacent to it.~
- [X] Don't display underline on hover in code. New: 
image
  vs old: 
image
- [X] Add external link indicator, e.g., the symbol after the first link below image


fredrik-bakke:
I'm not a big fan of the aesthetics of the dotted underline for links actually. It is probably better to practice not having adjacent links instead.

theme/css/general.css
```diff
@@ -135,7 +135,62 @@ h6:target::before {
...
+.content h4 > a[href^='https']::after,
+.content h5 > a[href^='http']::after,
+.content h5 > a[href^='https']::after,
+.content h6 > a[href^='http']::after,
+.content h6 > a[href^='https']::after,
+.content dt > a[href^='http']::after,
+.content dt > a[href^='https']::after,
+.content dd > a[href^='http']::after,
+.content dd > a[href^='https']::after {
+  content: '⇱';
```
VojtechStep: All external link icons point to the upper right. Since there's no unicode symbol for this icon specifically, I think it'd be better to use an inline SVG, as explained e.g. here: https://christianoliff.com/blog/styling-external-links-with-an-icon-in-css/

theme/css/general.css
```diff
@@ -135,7 +135,62 @@ h6:target::before {
...
+}
+.content pre a:hover,
+.content code a:hover,
+pre .content a:hover,
+code .content a:hover,
+a.concept:hover {
+  text-decoration: none;
+}
+/* External link indicator */
+.content p > a[href^='http']::after,
```
VojtechStep: What are all those selectors trying to exclude? From my testing only the on-page TOC uses absolute links, so you could do `.content a[href^="https://"]` and remove the icon in `nav.pagetoc a`

VojtechStep (commented):
I don't think adding the dotted underlines with JavaScript is a good move. It adds overhead on every page load, linear in the number of links (which is most of the content), and the resulting separation is still pretty easy to miss IMO. The more scalable solution would be to try and phrase things in such a way that you don't get runs of links (like the original `|transitive| |well-founded relation|` got reworded into `|relation| that is |transitive| and |well-founded|`).

theme/css/general.css
fredrik-bakke: Thanks for the link! Here's the new styling following your cited guide

image

fredrik-bakke:
> I don't think adding the dotted underlines with JavaScript is a good move. It adds overhead on every page load, linear in the number of links (which is most of the content), and the resulting separation is still pretty easy to miss IMO. The more scalable solution would be to try and phrase things in such a way that you don't get runs of links (like the original `|transitive| |well-founded relation|` got reworded into `|relation| that is |transitive| and |well-founded|`).

I have to disagree that it's not an improvement, but let's not fight about this now. The issue is very minor anyways, and I've reverted the relevant changes.

theme/css/general.css
fredrik-bakke: Did I implement your suggestion correctly?

theme/css/general.css
```diff
@@ -135,7 +135,46 @@ h6:target::before {
...
+  content: '';
+  width: 11px;
+  height: 11px;
+  margin-left: 4px;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16' fill='currentColor' viewBox='0 0 16 16'%3E%3Cpath fill-rule='evenodd' d='M8.636 3.5a.5.5 0 0 0-.5-.5H1.5A1.5 1.5 0 0 0 0 4.5v10A1.5 1.5 0 0 0 1.5 16h10a1.5 1.5 0 0 0 1.5-1.5V7.864a.5.5 0 0 0-1 0V14.5a.5.5 0 0 1-.5.5h-10a.5.5 0 0 1-.5-.5v-10a.5.5 0 0 1 .5-.5h6.636a.5.5 0 0 0 .5-.5z'/%3E%3Cpath fill-rule='evenodd' d='M16 .5a.5.5 0 0 0-.5-.5h-5a.5.5 0 0 0 0 1h3.793L6.146 9.146a.5.5 0 1 0 .708.708L15 1.707V5.5a.5.5 0 0 0 1 0v-5z'/%3E%3C/svg%3E");
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: contain;
+  display: inline-block;
+  color: var(--fg);
```
VojtechStep: This doesn't actually affect the color of the symbol, which is always black, even on dark backgrounds, rendering it invisible.

What should work is replacing all the `background-*` properties with `mask-*` properties, and setting the color with `background-color: var(--fg);`. This works at least for me on Firefox, but Safari browsers might need duplicating all those properties with `--webkit-`

theme/css/general.css
```diff
@@ -135,7 +135,46 @@ h6:target::before {
...
+}
+nav.pagetoc a::after {
+  display: none !important;
+}
+a.concept[href^='http']::after,
+a.concept[href^='https']::after {
+  content: '' !important;
 }
```
VojtechStep: This doesn't do anything, since the content is already empty. You want `display: none !important;`.
Same thing in `print.css`

theme/css/general.css
```diff
@@ -135,7 +135,46 @@ h6:target::before {
...
+.content pre a:hover,
+.content code a:hover,
+pre .content a:hover,
+code .content a:hover,
+a.concept:hover {
+  text-decoration: none;
+}
+
+/* External link indicator using inline SVG */
+.content a[href^="http"]::after,
```
VojtechStep: This also matches the image on the homepage, so you get an external link symbol in between paragraphs. You probably also want to exclude image links with something like
```css
a:has(> img)::after {
  display: none !important;
}
```

theme/css/general.css
```diff
@@ -135,7 +135,46 @@ h6:target::before {
...
+pre .content a,
+code .content a,
+.Agda a,
+a.concept {
+  text-decoration: none !important;
+}
+.content pre a:hover,
+.content code a:hover,
+pre .content a:hover,
+code .content a:hover,
```
VojtechStep: What do these two match? I don't think they match anything. Same for the same pair above

theme/css/general.css
```diff
@@ -135,7 +135,46 @@ h6:target::before {
...
-  text-decoration: underline;
+  text-decoration: solid underline;
+}
+
+.content pre a,
+.content code a,
+pre .content a,
+code .content a,
+.Agda a,
+a.concept {
```
VojtechStep: Actually, what other code blocks other than Agda are you trying to exclude here? Wouldn't `pre.Agda a` work just fine?

VojtechStep: Also isn't this covered by L135 already? This whole underlying thing feels like it should just be
```css
pre.Agda a:hover {
  text-decoration: none;
}
```

theme/css/general.css
fredrik-bakke: Turns out the exclusion selector for the navigation pane does not work on firefox 

image

fredrik-bakke: I'm really out of my depth here, but I've implemented your suggestion

VojtechStep: I'll try to take a closer look later this week

VojtechStep: My ass did not take a closer look later that week 💀

What version of Firefox were you testing? The only thing that might cause problems is the `:has` selector, but that's been supported since all browsers, including the mobile ones, since December 2023 (I just checked with Firefox 121).

fredrik-bakke: A newer one than that I'm sure. Unfortunately, it has updated again since I tested it, so I no longer know which version I used at the time. I can test it again if you like

VojtechStep: I tested it with 121, 142, and 115 ESR. The 115 indeed doesn't support `:has`, but it's only supported on Windows 8 and macOS Mojave, and that's until March 2026. In any case, I put the selectors inside an `:is`, which is resilient to unsupported inner selectors, so everything besides the `:has` works. The one thing that's broken on that version is hiding the symbols from image links, but AFAIK we use that only once?

fredrik-bakke: yeah, on the main page...

VojtechStep: I added an extra class to hide the indicator on the one image on the homepage, since it looks bad if there is one

fredrik-bakke: excellent!

fredrik-bakke:
> I'm happy with these changes, but let me know if you manage to reproduce the external link indicators showing up in your Firefox. Otherwise feel free to merge it

I'll test it right now

fredrik-bakke:
> I'll test it right now

Nope, they're gone!

PR #1415, opened 2025-04-27
chore: Spelling corrections by codespell

fredrik-bakke:
Fixes many common spelling mistakes using [codespell](https://github.com/codespell-project/codespell)

PR #1416, opened 2025-04-27
Add codespell pre-commit hook

fredrik-bakke:
- [X] Add a spelling checker pre-commit hook using [codespell](https://github.com/codespell-project/codespell). See #1415 for a list of corrections made by this hook.
- [x] Merge #1415
- [x] Documentation concerning how to use the accompanying `ignore` and `dictionary` files.


fredrik-bakke:
Turns out I wasn't using the newest version of codespell, so this PR now contains 9 additional spelling mistakes.

.pre-commit-config.yaml
```diff
@@ -13,8 +13,53 @@ repos:
...
+        name: Spellcheck with codespell (additional builtin)
+        files: ^src\/
+        types_or: [text, markdown]
+        args:
+          [
+            '--config',
+            '.codespellrc',
+            '--builtin',
+            'informal,usage,rare,code,names,en-GB_to_en-US',
+          ]
```
fredrik-bakke, line 51: I couldn't get codespell to use all three sets of dictionaries in a single run, unfortunately.

.pre-commit-config.yaml
```diff
@@ -13,8 +13,53 @@ repos:
       # - id: check-json # Doesn't accept json with comments
       # - id: requirements-txt-fixer # Can't handle comments
 
+  - repo: https://github.com/codespell-project/codespell
```
VojtechStep, line 16: Instead of three `repo` blocks there should be one, with multiple entries in `hooks`

.codespellrc
```diff
@@ -0,0 +1,8 @@
+[codespell]
+skip = .git,node_modules,*.agda
```
VojtechStep, line 2: All invocations in the pre-commit config are restricted to `^src/`, so why is this necessary?

.pre-commit-config.yaml
VojtechStep: - why are you separating this set of builtins from the default builtins? The default is `clear,rare`, so `rare` is checked twice
- you can use multiple dictionaries by separating them with a comma

So the three hooks can be replaced by one doing `--builtin clear,rare,informal,usage,code,names,en-GB_to_en-US -D codespell-dictionary.txt,-`

.pre-commit-config.yaml
```diff
@@ -13,8 +13,53 @@ repos:
...
   - repo: local
     hooks:
+      - id: sort-codespell-files
+        name: Sort codespell dictionary and ignore files
+        entry: bash -c 'sort -o "$1" "$1" && sort -o "$2" "$2"' --
+        args: [codespell-dictionary.txt, codespell-ignore.txt]
+        language: system
+        files: ^(codespell-dictionary\.txt|codespell-ignore\.txt)$
+        pass_filenames: false
+
```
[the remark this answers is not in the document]
VojtechStep: Sorry, typo, it's supposed to be `$f`; for whatever reason I was retyping the command instead of copy-pasting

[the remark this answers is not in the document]
fredrik-bakke: thanks!

fredrik-bakke:
Thanks for the thorough review. Your comments were very helpful! I changed the applicable files pattern so that files outside of `src` are checked as well, with a few exceptions. I found a bunch more spelling mistakes this way. This tool seems to be incredibly useful, actually; the false positive rate is very low.

VojtechStep:
The reason why I asked about the two places where you can configure patterns to ignore was: either the way the check is executed automatically ignores files not tracked by git (in which case it doesn't make sense to explicitly ignore `.git`), or it checks all files which aren't ignored (in which case we should at least explicitly ignore `book/` and `docs/`). Could you find out what's actually happening?

fredrik-bakke:
Indeed, it ignores untracked files, as is standard for pre-commit hooks. Thanks for reminding me

VojtechStep:
Okay, LGTM, thanks for integrating this 👌

PR #1398, opened 2025-04-07
Metric properties of real negation, absolute value, addition and maximum

malarbol:
This PRs proves the following metric properties of real operations:

- `raise-ℝ` is an isometry;
- `neg-ℝ` is an isometry;
- `abs-ℝ` is a short function;
- `add-ℝ` is an isometry from `ℝ` to the metric space of isometries `ℝ → ℝ`;
- `max-ℝ` is a short function from `ℝ` to the metric space of short functions `ℝ → ℝ`.

It also introduces the following concepts:

- the metric space of functions between metric spaces;
- the metric space of short functions between metric spaces;
- the metric space of isometries between metric spaces.

PR #1417, opened 2025-04-27
Lipschitz-continuous functions between metric spaces

malarbol:
This introduces the concept of *Lipschitz* functions between metric spaces and proves that left/right multiplication of rational numbers is Lipschitz.

More precisely, `metric-spaces.lipschitz-functions-metric-spaces` introduces the following concepts and results:
  - Lipschitz constant for a function between metric spaces;
  - Lipschitz functions between metric spaces (functions with some Lipschitz constant);
  - constant functions are Lipschitz for all `α : ℚ⁺`;
  - short functions are Lipschitz functions with Lipschitz constant equal to 1;
  - Lipschitz functions are uniformly continuous;
  - the composition of Lipschitz maps is Lipschitz;
  - being a Lipschitz map is homotopy invariant;
  - Lipschitz maps preserve elements at bounded distance.

The metric space of Lipschitz functions is introduced  in `metric-spaces.metric-space-of-lipschitz-functions-metric-spaces`.

We also introduce a new concept in the `metric-spaces` module:
- `metric-spaces.elements-at-bounded-distance-metric-spaces`: the equivalence relation of being at bounded distance in a metric space (being in _some_ neighborhood).


src/real-numbers/metric-space-of-real-numbers.lagda.md
```diff
@@ -327,6 +328,54 @@ is-isometry-metric-space-leq-real-ℚ d x y =
     ( map-product
       ( leq-add-positive-le-le-add-positive-ℚ x y d)
       ( leq-add-positive-le-le-add-positive-ℚ y x d))
+
+isometry-metric-space-leq-real-ℚ :
+  isometry-Metric-Space
+    ( metric-space-leq-ℚ)
+    ( metric-space-leq-ℝ lzero)
+isometry-metric-space-leq-real-ℚ =
+  ( real-ℚ , is-isometry-metric-space-leq-real-ℚ)
```
fredrik-bakke, line 337: shouldn't this be universe polymorphic?

fredrik-bakke: Oh, this code was outdated?

malarbol: this and the universe-polymorphic version were actually part of #1398 😅

malarbol: But I guess it would make sense to move all the package `raise-real-ℚ`, `raise-zero-ℝ`, etc. in the `rational-real-numbers` module and then `isometry-metric-space-leq-raise-ℝ` here instead.

malarbol: > But I guess it would make sense to move all the package `raise-real-ℚ`, `raise-zero-ℝ`, etc. in the `rational-real-numbers` module and then `isometry-metric-space-leq-raise-ℝ` here instead.

This really felt cleaner, so I did it in https://github.com/UniMath/agda-unimath/pull/1417/commits/3bdccdbf2bd3f9d3e53b8e8713a964f8dad78467.

src/metric-spaces/lipschitz-functions-metric-spaces.lagda.md
```diff
@@ -0,0 +1,405 @@
...
+
+A positive rational `α : ℚ⁺` is a
+{{#concept "Lipschitz constant" Disambiguation="of a function between metric spaces" Agda=lipschitz-constant-function-Metric-Space}}
+of a [function](metric-spaces.functions-metric-spaces.md) `f : A → B` between
+[metric spaces](metric-spaces.metric-spaces.md) if for any `x y : A`, if `d` is
+an upper bound of the distance between `x` and `y` in `A`, `α * d` is an upper
+bound on the distance between `f x` and `f y` in `B`; in that case, `f` is
+called an **α-Lipschitz** function. A function which admits a Lipschitz constant
+is called a
+{{#concept "Lipschitz function" Disambiguation="between metric spaces" WD="Lipschitz function" WDID=Q652707 Agda=lipschitz-function-Metric-Space}}.
```
fredrik-bakke, line 56:
```suggestion
A positive rational `α : ℚ⁺` is a
{{#concept "Lipschitz constant" Disambiguation="of a function between metric spaces" Agda=lipschitz-constant-function-Metric-Space}}
of a [function](metric-spaces.functions-metric-spaces.md) `f : A → B` between
[metric spaces](metric-spaces.metric-spaces.md) if for any `x y : A`, if `d` is
an upper bound of the distance between `x` and `y` in `A`, `α * d` is an upper
bound on the distance between `f x` and `f y` in `B`. If `α` is a Lipschitz
constant for `f`, then `f` is called an **α-Lipschitz** function. A function
that admits a Lipschitz constant is called a
{{#concept "Lipschitz function" Disambiguation="between metric spaces" WD="Lipschitz function" WDID=Q652707 Agda=lipschitz-function-Metric-Space}}.
```

src/metric-spaces/total-metric-spaces.lagda.md
```diff
@@ -0,0 +1,76 @@
...
+
+open import metric-spaces.metric-spaces
+```
+
+</details>
+
+## Idea
+
+A [metric space](metric-spaces.metric-spaces.md) is
+{{#concept "total" Disambiguation="metric space" Agda=is-total-Metric-Space}} if
```
fredrik-bakke, line 24:
```suggestion
{{#concept "total" Disambiguation="metric space" Agda=is-total-Metric-Space Agda=Total-Metric-Space}} if
```

src/metric-spaces/total-metric-spaces.lagda.md
```diff
@@ -0,0 +1,76 @@
+# Total metric spaces
```
fredrik-bakke, line 1: I just want to point out that the concept of "total metric spaces" could also be understood as dependent sums of metric spaces over a metric space. In weighted graph theory, I believe they use the word "connected weighted graph" to mean that every pair of vertices have a path of finite weight. Although, this word could also be misinterpreted as something else. Could you cite a reference that uses your current terminology perhaps?

malarbol: > Could you cite a reference that uses your current terminology perhaps?

TBH, I just took the terminology from "total (pre)order" but I don't know if there's any reference using this terminology. This is part of the definition of [Richman premetric space](https://ncatlab.org/nlab/show/Richman+premetric+space) but I don't know if they give a name to this property in the original paper. 

I'm ok with changing the name if you want. As you mentioned, I'm not sure "connected metric space" is better since it could be misinterpreted with the topological property (unless there's any relation between the two concepts?). "Totally measurable" doesn't look right either, since it looks too much like "Lebesgue measure".

malarbol: Note that this concept will also be important for Banach fix point theorem. I believe you need the metric space to be total for the unicity of the fix point and total+complete for the existence.

fredrik-bakke: I'm okay with keeping the terminology as is, but it'd be nice if it was explained where it comes from.

malarbol: I'll try to think of something. But, as you know, prose is not really my _forte_ so if you think of something, feel free to edit the header as you please.

fredrik-bakke: Well, I don't know where you've gotten the terminology from. Something like `The terminology comes from {{#cite my-reference}}, and should be understood as totality in the sense that the distance function is defined everywhere. This is not to be confused with total spaces of metric spaces.` should do fine.

malarbol: > I just want to point out that the term "total metric spaces" could also be misinterpreted as dependent sums of metric spaces over a metric space.

Do you mean something like this:
```agda
module _
  {l1 l2 l3 : Level}
  (X : UU l1) (P : X → Metric-Space l2 l3)
  where

  type-Σ-Metric-Space : UU (l1 ⊔ l2)
  type-Σ-Metric-Space = Σ X (λ x → type-Metric-Space (P x))
```
with a metric structure on it?  We could do another module `dependent-sum-metric-spaces` with the corresponding results, and then, in the `total-metric-spaces` module, explain that **total** is to be understood as in  [total preorders](https://unimath.github.io/agda-unimath/order-theory.total-preorders.html) or [total orders](https://unimath.github.io/agda-unimath/order-theory.total-orders.html) while the type dependent sums of over metric spaces is defined in `dependent-sum-metric-spaces`. The name `Σ-Metric-Space` makes it coherent with `Π-Metric-Space`.

fredrik-bakke: Yes, this would be another understanding of total metric space. There's no need to add this as part of the current pr though. Since you don't have a reference for the terminology, we should say something to this effect. E.g. have some disclaimer that this term may not align with the literature.

fredrik-bakke: Hoe exactly is it comparable to total orders?

fredrik-bakke: > Yes, this would be another understanding of total metric space. There's no need to add this as part of the current pr though. Since you don't have a reference for the terminology, we should say something to this effect. E.g. have some disclaimer that this term may not align with the literature.

Yet another understanding of the word would be a metric space as the domain of some notion of fibration or fiber bundle of metric spaces, but these are also known as displayed metric spaces I think.

fredrik-bakke: To be clear, I'm happy with the usage of "total" here, I just want the terminology to be justified to the reader too.

malarbol: > How exactly is it comparable to total orders?

In a metric space, you can construct the relation of "being neighbors / at finite distance":

```agda
module _
  {l1 l2 : Level} (A : Metric-Space l1 l2)
  (x y : type-Metric-Space A)
  where

  is-neighbor-prop-Metric-Space : Prop l2
  is-neighbor-prop-Metric-Space =
    ∃ ℚ⁺ (λ d → structure-Metric-Space A d x y)

  is-neighbor-Metric-Space : UU l2
  is-neighbor-Metric-Space = type-Prop is-neighbor-prop-Metric-Space

  is-prop-is-neighbor-Metric-Space :
    is-prop is-neighbor-Metric-Space
  is-prop-is-neighbor-Metric-Space =
    is-prop-type-Prop is-neighbor-prop-Metric-Space
```

This relation is an equivalence relation, hence a preorder. This preorder is total if and only if the metric space is total. In which case all elements are equivalent under the neighbor relation.

I'll write it down properly; would this be enough to justify the terminology?

src/metric-spaces/total-metric-spaces.lagda.md
malarbol: Did you really mean to keep both `Agda=...` in the concept or is this a typo? I didn't know it was allowed.

src/metric-spaces/total-metric-spaces.lagda.md
malarbol: See https://github.com/UniMath/agda-unimath/pull/1417/commits/a16c956f80a6618ce911a91a473b7d252f000667

malarbol: I added a last property to tie it up together: "Lipschitz maps preserve neighbors". 

I know you only asked for a one-sentence disclaimer and received 300LOC to review but I hope these changes make the terminology more understandable.

src/metric-spaces/lipschitz-functions-metric-spaces.lagda.md
```diff
@@ -0,0 +1,451 @@
...
+</details>
+
+## Idea
+
+A positive rational `α : ℚ⁺` is a
+{{#concept "Lipschitz constant" Disambiguation="of a function between metric spaces" Agda=lipschitz-constant-function-Metric-Space}}
+of a [function](metric-spaces.functions-metric-spaces.md) `f : A → B` between
+[metric spaces](metric-spaces.metric-spaces.md) if for any `x y : A`, if `d` is
+an upper bound of the distance between `x` and `y` in `A`, `α * d` is an upper
+bound on the distance between `f x` and `f y` in `B`. If `α` is a Lipschitz
```
fredrik-bakke, line 54:
```suggestion
an upper bound of the distance between `x` and `y` in `A`, then `α * d` is an upper
bound on the distance between `f x` and `f y` in `B`. If `α` is a Lipschitz
```

src/metric-spaces/neighbors-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+### Being neighbors in a metric space is symmetric
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  where
+
+  symmetric-is-neibhbor-Metric-Space :
```
fredrik-bakke, line 108:
```suggestion
  symmetric-is-neighbor-Metric-Space :
```

src/metric-spaces/neighbors-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+  is-equivalence-relation-is-neighbor-Metric-Space =
+    ( refl-is-neighbor-Metric-Space A) ,
+    ( symmetric-is-neibhbor-Metric-Space A) ,
+    ( transitive-is-neighbor-Metric A)
+```
+
+## See also
+
+- [Total metric spaces](metric-spaces.total-metric-spaces.md): metric spaces
+  where all elements are neighbors of each other.
```
fredrik-bakke, line 174:
```suggestion
- [Total metric spaces](metric-spaces.total-metric-spaces.md) are metric spaces
  where all elements are neighbors of each other.
```

src/metric-spaces/neighbors-metric-spaces.lagda.md
fredrik-bakke, line 1: I find the terminology introduced in this file quite strange. Yes, two elements are neighbors if and only if they share some neighborhood, but the term "neighbor" reads very much like "adjacent" here. In graph theory, however, two elements are not necessarily adjacent of they live in the same path component. It's like, sure, I might be neighbors with someone who is the neighbor of someone who is the neighbor of someone who is the neighbor of someone who is the ... who is the neighbor of someone who lives in another city, but I won't say a person who lives in another city is my neighbor!

src/metric-spaces/neighbors-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+open import order-theory.preorders
+```
+
+</details>
+
+## Idea
+
+Two elements `x y : A` of a [metric-space](metric-spaces.metric-spaces.md) are
+{{#concept "neighbors" Disambiguation="in a metric space" Agda=neighbor-Metric-Space}}
+if they are in some neighborhhod in `A`, i.e. if there
```
fredrik-bakke, line 33:
```suggestion
if they are in some neighborhood in `A`, i.e. if there
```

src/metric-spaces/neighbors-metric-spaces.lagda.md
fredrik-bakke: Would you even say that you are the neighbor of everyone in your neighborhood? I don't think that's even necessarily correct to say.

malarbol: well, this is inherent to the name [neighborhood-Metric-Space](https://unimath.github.io/agda-unimath/metric-spaces.metric-spaces.html#5014) : [ℚ⁺](https://unimath.github.io/agda-unimath/elementary-number-theory.positive-rational-numbers.html#4477) → [Relation](https://unimath.github.io/agda-unimath/foundation.binary-relations.html#1181) [l2](https://unimath.github.io/agda-unimath/metric-spaces.metric-spaces.html#4115) [type-Metric-Space](https://unimath.github.io/agda-unimath/metric-spaces.metric-spaces.html#4246); since ℚ⁺ is unbounded, "neighborhoods" can get arbitrarily large so "neighbors" may live far apart from each other.

> but I wouldn't say a person who lives in another city is my neighbor!

Maybe it depends on the context? If you were talking to someone on the other side of the globe, you might refer to someone living the city next to you a "neighbor"? 

I mean, the sentence "two elements are neighbors if they are in _some_ neighborhood" doesn't seem that strange to me. But you need to contemplate big neighborhoods.

fredrik-bakke: I think a more appropriate name for this notion is simply "elements at bounded distance". Notice also that with your proposed terminology, in a traditional metric space all elements are necessarily neighbors. This just sounds off to me.

fredrik-bakke: > Maybe it depends on the context? If you were talking to someone on the other side of the globe, you might refer to someone living the city next to you a "neighbor"?

This is a fair point.

fredrik-bakke: Alright, maybe I'm reading too much into the meaning of "neighbor" here

src/metric-spaces/lipschitz-functions-metric-spaces.lagda.md
```diff
@@ -0,0 +1,451 @@
...
+          ( H d x y N))
+
+  is-lipschitz-htpy-function-Metric-Space :
+    is-lipschitz-function-Metric-Space A B f →
+    is-lipschitz-function-Metric-Space A B g
+  is-lipschitz-htpy-function-Metric-Space =
+    map-is-inhabited lipschitz-constant-htpy-function-Metric-Space
+```
+
+### Lipschitz maps perserve neighbors in metric spaces
```
fredrik-bakke, line 405: Is this not a property of continuous functions in general?

src/metric-spaces/neighbors-metric-spaces.lagda.md
malarbol: > I think a more appropriate name for this notion is simply "elements at bounded distance".

There's no actual definition of distance so this feels a bit off too. "Constructing a proof of `N d x y` amounts to saying that d is an upper bound on the distance from x to y." But it's more an interpretation than an actual construction.

> Notice also that with your proposed terminology, in a traditional metric space all elements are necessarily neighbors. This just sounds off to me.

I don't know. Once you allow some elements to be at infinite distance, calling "elements at finite distance" neighbors doesn't seem too far-fetched.

src/metric-spaces/total-metric-spaces.lagda.md
fredrik-bakke: This makes me wonder: since you say totality is not enough to construct a distance function `d : A → A → ℝ`, what do you need to actually get such a function? Presumably, the correct definition of the real numbers for this purpose would be the MacNeille reals, but I could be wrong, of course.

fredrik-bakke: Perhaps you could try and show that total metric spaces have a [total](https://unimath.github.io/agda-unimath/foundation.total-partial-functions.html) distance function `d : A → A → ℝ` where `ℝ` are the MacNeille reals? If so, that is great justification for the name.

src/metric-spaces/neighbors-metric-spaces.lagda.md
fredrik-bakke: > There's no actual definition of distance so this feels a bit off too. "Constructing a proof of N d x y amounts to saying that d is an upper bound on the distance from x to y." But it's more an interpretation than an actual construction.

"Bounded distance" is the concept captured by "neighbors", so I don't think this name is inappropriate.

src/metric-spaces/neighbors-metric-spaces.lagda.md
```diff
@@ -110,8 +112,10 @@ module _
     is-neighbor-Metric-Space A x y →
     is-neighbor-Metric-Space A y x
   symmetric-is-neighbor-Metric-Space x y =
-    map-trunc-Prop
-      ( λ (d , Nxy) → d , is-symmetric-structure-Metric-Space A d x y Nxy)
+    map-exists
+      ( is-upper-bound-dist-Metric-Space A y x)
+      ( id)
+      ( λ d → is-symmetric-structure-Metric-Space A d x y)
```
fredrik-bakke, line 118:
```suggestion
    map-tot-exists
      ( is-upper-bound-dist-Metric-Space A y x)
      ( λ d → is-symmetric-structure-Metric-Space A d x y)
```
:)

src/metric-spaces/lipschitz-functions-metric-spaces.lagda.md
malarbol: mmh. I don't think it is (with our current terminology). I understand why you'd expect this. Maybe it's a point against the name "neighbors" in metric spaces.

src/metric-spaces/neighbors-metric-spaces.lagda.md
malarbol: FYI, there seems to be a slight inconsistence on the implicit/explicit arguments of the maps `map-exists`, `map-tot-exists`, `map-binary-exists`, etc.

fredrik-bakke: Ah, that's because in the special case of `map-tot-exists`, Agda is always able to infer the type family, but not in the other instances. Sorry, I forgot that.

src/metric-spaces/neighbors-metric-spaces.lagda.md
malarbol: ok. So what about 
```agda
  is-elem-at-bounded-dist-prop-Metric-Space : Prop l2
  is-elem-at-bounded-dist-prop-Metric-Space =
    ∃ ℚ⁺ (is-upper-bound-dist-prop-Metric-Space A x y)

  is-elem-at-bounded-dist-Metric-Space : UU l2
  is-elem-at-bounded-dist-Metric-Space =
    type-Prop is-elem-at-bounded-dist-prop-Metric-Space

...
```
replace `neighbor` by `elem-at-bounded-dist`?

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+Two elements `x y : A` of a [metric-space](metric-spaces.metric-spaces.md) are
+{{#concept "at bounded distance" Disambiguation="in a metric space" Agda=element-at-bounded-dist-Metric-Space}}
+if there [exists](foundation.existential-quantification.md) some
+[positive rational number](elementary-number-theory.positive-rational-numbers.md)
+`δ : ℚ⁺` such that `x` and `y` are `δ`-neighbors in `A`. Being at bounded
+distance in a metric space is an equivalence relation.
+
+## Definitions
+
+### The relation of being neighbors in a metric space
```
fredrik-bakke, line 42:
```suggestion
### The relation of being at bounded distance in a metric space
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+</details>
+
+## Idea
+
+Two elements `x y : A` of a [metric-space](metric-spaces.metric-spaces.md) are
+{{#concept "at bounded distance" Disambiguation="in a metric space" Agda=element-at-bounded-dist-Metric-Space}}
+if there [exists](foundation.existential-quantification.md) some
+[positive rational number](elementary-number-theory.positive-rational-numbers.md)
+`δ : ℚ⁺` such that `x` and `y` are `δ`-neighbors in `A`. Being at bounded
+distance in a metric space is an equivalence relation.
```
fredrik-bakke, line 38: link equivalence relation

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+
+</details>
+
+## Idea
+
+Two elements `x y : A` of a [metric-space](metric-spaces.metric-spaces.md) are
+{{#concept "at bounded distance" Disambiguation="in a metric space" Agda=element-at-bounded-dist-Metric-Space}}
+if there [exists](foundation.existential-quantification.md) some
+[positive rational number](elementary-number-theory.positive-rational-numbers.md)
+`δ : ℚ⁺` such that `x` and `y` are `δ`-neighbors in `A`. Being at bounded
```
fredrik-bakke, line 37:
```suggestion
`δ : ℚ⁺` such that `x` and `y` share a `δ`-neighborhood in `A`. Being at bounded
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
fredrik-bakke, line 38: Maybe we could include some of our discussion on the alternative term "neighbor" here? I think we both had some good points about the term, and that could be insightful for the reader to see. If you want I can try and cook up something.

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+  is-element-at-bounded-dist-Metric-Space =
+    type-Prop is-element-at-bounded-dist-prop-Metric-Space
+
+  is-prop-is-element-at-bounded-dist-Metric-Space :
+    is-prop is-element-at-bounded-dist-Metric-Space
+  is-prop-is-element-at-bounded-dist-Metric-Space =
+    is-prop-type-Prop is-element-at-bounded-dist-prop-Metric-Space
+```
+
+### Element-at-bounded-dists in metric spaces
```
fredrik-bakke, line 64:
```suggestion
### Elements at bounded distance in metric spaces
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+    is-element-at-bounded-dist-Metric-Space
+      ( A)
+      ( x)
+      ( value-element-at-bounded-dist-Metric-Space H)
+  is-element-at-bounded-dist-elem-element-at-bounded-dist-Metric-Space = pr2
+```
+
+## Properties
+
+### Being at bounded distance in a metric space is reflexive
```
fredrik-bakke, line 91:
```suggestion
### The relation of being at bounded distance in a metric space is reflexive
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+### Being at bounded distance in a metric space is symmetric
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  where
+
+  symmetric-is-element-at-bounded-dist-Metric-Space :
```
fredrik-bakke, line 113:
```suggestion
  symmetric-rel-bounded-dist-Metric-Space :
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+## Properties
+
+### Being at bounded distance in a metric space is reflexive
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  where
+
+  refl-is-element-at-bounded-dist-Metric-Space :
```
fredrik-bakke, line 98:
```suggestion
  refl-bounded-dist-Metric-Space :
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+
+### The relation of being neighbors in a metric space
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  (x y : type-Metric-Space A)
+  where
+
+  is-element-at-bounded-dist-prop-Metric-Space : Prop l2
```
fredrik-bakke, line 50: Actually, the following name seems to work just fine for the relation, no?
```suggestion
  bounded-dist-prop-Metric-Space : Prop l2
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+  is-element-at-bounded-dist-prop-Metric-Space =
+    ∃ ℚ⁺ (λ d → structure-Metric-Space A d x y)
+
+  is-element-at-bounded-dist-Metric-Space : UU l2
```
fredrik-bakke, line 54:
```suggestion
  bounded-dist-Metric-Space : UU l2
```

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
fredrik-bakke: or `bounded-dist-rel-prop-Metric-Space`

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
fredrik-bakke: or `bounded-dist-rel-Metric-Space`

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
malarbol: If you feel like it, go ahead, but I don't think it's really necessary. I agree we both had good points for our names but you actually convinced me that `neighbor` is a bit too generic and could be misleading; this term usually carries some notion of "locality", or "small neighborhoods (like for small δ)" which isn't really reflected by this relation. 

I may come back to this topic sometimes in the future though. I think there still some interesting to say: basically a metric space is total if all elements are equivalent under this relation, so the quotient under this relation should be propositional (and contractible if the space is inhabited). In general, taking the quotient, and then subspaces of elements in a given equivalence class, we should be able to re-construct the original space as a "the total space of a family of total metrics spaces", which I think could be an interesting result, and also connects this with the other interpretation of "_total_ metric space".

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,183 @@
...
+
+open import order-theory.preorders
+```
+
+</details>
+
+## Idea
+
+Two elements `x y : A` of a [metric space](metric-spaces.metric-spaces.md) are
+{{#concept "at bounded distance" Disambiguation="in a metric space" Agda=element-at-bounded-dist-Metric-Space}}
```
malarbol, line 34: I have doubts on whether the `Agda=...` should point to `element-at-bounded-dist-Metric-Space` or `bounded-dist-Metric-Space` or both if its allowed (hence my [question](https://github.com/UniMath/agda-unimath/pull/1417/files/17c74e3ace27df5c99cf66b7dd7af5835a3ac28e#r2089445299))

src/metric-spaces/total-metric-spaces.lagda.md
malarbol: well. I need learn about MacNeille reals and think about it. We can talk about this again.

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
fredrik-bakke: Both is ok

malarbol: > Actually. Element at bounded distance is the collection of all elements at bounded distance of a given (base) point. So that should be a second concept

I replaced the link with the right definition. I'm not sure if you actually want to define a second concept for `element-at-bounded-dist-Metric-Space`  in the `## Idea` section; if so, I believe you'll write it much better than I would, so please go ahead.

src/metric-spaces/total-metric-spaces.lagda.md
fredrik-bakke: It is allowed :)

src/metric-spaces/total-metric-spaces.lagda.md
```diff
@@ -0,0 +1,118 @@
...
+
+</details>
+
+## Idea
+
+A [metric space](metric-spaces.metric-spaces.md) is
+{{#concept "total" Disambiguation="metric space" Agda=is-total-Metric-Space Agda=Total-Metric-Space}}
+if all elements are at bounded distance, i.e., if the
+[preorder of elements at bounded distance in the metric space](metric-spaces.elements-at-bounded-distance-metric-spaces.md)
+is [total](order-theory.total-preorders.md).
```
fredrik-bakke, line 35: Hmmm, my mind has actually started to change on your usage of the term "total metric space". Every justification we have tried to cook up so far hasn't worked out, and one might ought to reserve the name "total" for metric spaces satisfying the property I laid out before as well:

>metric spaces have a [total](https://unimath.github.io/agda-unimath/foundation.total-partial-functions.html) distance function `d : A → A → ℝ` where `ℝ` are the MacNeille reals

Until we find sufficient justification for the name, can you please add a disclaimer that this is nonstandard terminology that might conflict with literature?

fredrik-bakke, line 35: As discussed elsewhere, this property does not correspond to totality of the given preorder, but indiscreteness, and I don't think that's an interesting property to mention here.

fredrik-bakke: For premetrics, the correct phrasing of "totality" would be that either `B d x y` or  `B d y x` for some positive rational `d`, and this seems like a slightly strange property to care about to me.

malarbol: I'll just remove `metric-spaces.total-metric-spaces` from this PR. I only needed it to state `is-constant-all-lipschitz-constant-function-Metric-Space` which seemed like a nice converse to `all-lipschitz-constant-is-constant-function-Metric-Space` and I didn't expect this name to be _that_ controversial. 

I'll add something in #1421 and we can come back to this concept later, when we really need it, and after we find a better name.

fredrik-bakke: That is really unfortunate, why do you want to remove it? It seems like a useful concept so it would be great to have it in the library. All I ask is that we are precise with the terminology

fredrik-bakke: It is okay to have temporary terminology, and it is okay to have disclaimers on the pages. This is just part of good practice to be as informative and accurate as possible. If we figure out that the terminology is good later, we can remove the disclaimer

malarbol: > That is really unfortunate, why do you want to remove it?

- I don't really _need_ this concept for this PR. I think the content on Lipschitz maps is enough;
- the name I chose seems much more controversial than I expected and I think we should give us some more time until we find a good one;
- I prefer to move forward with these results on Lipschitz maps and then #1424 and #1421 than thinking about a disclaimer explaining my bad naming schemes. 

> It seems like a useful concept so it would be great to have it in the library. 

It probably is. If that's so, it will naturally show up again sometimes. Hopefully we'll have found a better name for it, or we'll write a disclaimer. But I don't think anyone (besides me) really asked for this concept, or need it for anything they're working on, so there's no rush to add it now.

> It is okay to have temporary terminology, and it is okay to have disclaimers on the pages. This is just part of good practice to be as informative and accurate as possible. If we figure out that the terminology is good later, we can remove the disclaimer

I'd rather just wait, give it more thoughts, talk about it; maybe other people interested in metric spaces/analysis will join the conversation. In the meantime, let's focus on content that we're all ok with and move forward.

fredrik-bakke: Again the term "connectedness" comes up in this discussion. 🤔 Although I've never heard that term be used about relations.

src/metric-spaces/lipschitz-functions-metric-spaces.lagda.md
```diff
@@ -0,0 +1,417 @@
...
+
+</details>
+
+## Idea
+
+A
+[positive rational number](elementary-number-theory.positive-rational-numbers.md)
+`α : ℚ⁺` is a
+{{#concept "Lipschitz constant" Disambiguation="of a function between metric spaces" Agda=lipschitz-constant-function-Metric-Space}}
+of a [function](metric-spaces.functions-metric-spaces.md) `f : A → B` between
```
fredrik-bakke, line 54:
```suggestion
for a [function](metric-spaces.functions-metric-spaces.md) `f : A → B` between
```

fredrik-bakke (approved):

PR #1419, opened 2025-04-28
Strict orders

fredrik-bakke:
- Defines large strict preorders
- Defines one-sided and two-sided similarity of elements in (large) strict preorders
- Defines (large) strict orders as (large) strict preorders that are extensional with respect to two-sided similarity
- Fixes a name for similarity of Galois connections


src/order-theory/strict-posets.lagda.md
```diff
@@ -0,0 +1,280 @@
...
+  is-irreflexive-le-Strict-Poset : is-irreflexive le-Strict-Poset
+  is-irreflexive-le-Strict-Poset =
+    is-irreflexive-le-Strict-Preorder strict-preorder-Strict-Poset
+
+  is-transitive-le-Strict-Poset : is-transitive le-Strict-Poset
+  is-transitive-le-Strict-Poset =
+    is-transitive-le-Strict-Preorder strict-preorder-Strict-Poset
+```
+
+Similarity of elements in strict posets.
```
lowasser, line 105: Is this supposed to be a section header?

lowasser (approved):

lowasser (approved):

src/order-theory/strict-posets.lagda.md
fredrik-bakke: No, because this way I don't have to reopen the module. The proper format is probably to write a new file called `similarity-of-elements-strict-posets` though, so I should probably do that

src/order-theory/large-strict-posets.lagda.md
```diff
@@ -0,0 +1,338 @@
...
+{{#concept "extensionality principle" Disambiguation="of large strict posets" Agda=extensionality-principle-Large-Strict-Preorder}}
+that
+[similar elements](order-theory.similarity-of-elements-large-strict-preorders.md)
+are [equal](foundation-core.identity-types.md). More concretely, if $x$ and $y$
+are such that for every $z$, we have
+
+- $z < x$ [if and only if](foundation.logical-equivalences.md) $z < y$, and
+- $x < z$ if and only if $y < z$,
+
+then $x = y$.
```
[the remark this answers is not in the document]
fredrik-bakke: Indeed, thanks for pointing it out. Let me know if the new version is clear.

src/order-theory/strict-posets.lagda.md
fredrik-bakke: done now

fredrik-bakke:
Thank you for the review.

While going through the formalization again today, I thought a little more about this concept, and I'm not so sure this structure deserves the name "strict poset", nor if it is the concept you need. Although. It _is_ the "correct" univalence/extensionality principle for strict preorders.

I would expect it to be true of the correct notion of strict poset that the induced strict ordering $(x < y) := (x ≤ y) ∧ (x ≠ y)$, or perhaps more usefully $(x < y) := (x ≤ y) ∧ (x ＃ y)$ when we have an apartness relation $＃$ available, is a strict poset whenever $≤$ defines a poset, but this is not true for the concept at hand. Indeed, consider the discrete poset on two points $0$ and $1$. I.e., $x ≤ y$ iff $x = y$. Then the induced strict ordering is not extensional because $0$ and $1$ are not equal. While this mirrors the story for semicategories, I don't find that explanation satisfactory.

It seems like the present notion of "strict poset" is closer to a notion of strict _linear_ order, but I am not entirely sure where exactly they place in the landscape. Maybe we should call them simply "strict orders"?

fredrik-bakke:
Sorry to re-request a review, but given the larger changes, I figured I should run it past you again before merging.

PR #1423, opened 2025-05-03
Replace symbol for similarity

### LINKED ISSUE #1418: Change symbol for similarity
The library has adopted the symbol `≈` for "similarity" of elements, for instance, for elements in preorders, and ranks of elements in a W-type. However, from my mathematical training, the symbol `≈` was used to mean "approximates", e.g., "`3 ≈ π`" _relative to some externally defined criterion_, and this seems to be how the symbol is used by many mathematicians still today. Part of the point of approximations is that they allow simplifying assumptions that include more solutions that are still satisfactory according to some criterion. However, with our current usage of the symbol, two elements are approximately the same in an extensional structure if and only if they are literally the same. Therefore, I propose using a different symbol for the notion of similarity of elements.
- [Approximation](https://en.wikipedia.org/wiki/Approximation) at Wikipedia
- [Almost equal to `≈`](https://codepoints.net/U+2248) at codepoints.net

Some suggestions for alternatives are
- `~`, tilde, is already used for similarity of functions, and is commonly used for equivalence relation reasoning (note that approximation relations are commonly non-transitive!)
- `∼` the [tilde operator](https://codepoints.net/U+223c). `agda-input: \sim`
- `≍`, `agda-input: \asymp`, is used for similarities/indistinguishabilities in the book [The Univalence Principle](https://arxiv.org/abs/2102.06275)

fredrik-bakke:
Replace the current notation for similarity/indiscernibility `x ≈ y` with `x ≍ y` as proposed in #1418.

Resolves #1418.

### DIFF
--- a/src/foundation/images-subtypes.lagda.md
+++ b/src/foundation/images-subtypes.lagda.md
@@ -205,7 +205,7 @@ module _
 **Proof:** The asserted similarity follows at once from the similarity
 
 ```text
-  pullback-subtype (g ∘ f) ≈ pullback-subtype g ∘ pullback-subtype f
+  pullback-subtype (g ∘ f) ≍ pullback-subtype g ∘ pullback-subtype f
 ```
 
 via the image-pullback Galois connection.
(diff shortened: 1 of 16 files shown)


EgbertRijke:
I think I'll just remove the symbol entirely for now. You were right to point out that the symbol ≈ was not used in its most conventional meaning here. I think we can say similarly that the most conventional meaning of ≍ is asymptotic equivalence, if I understand it correctly, so it shouldn't be used here either.

PR #1428, opened 2025-05-12
Autoformatting for BibTeX

fredrik-bakke:
Sets up autoformatting for BibTeX files using `bibtex-tidy`. This sorts, aligns, and appropriately punctuates`references.bib`. It also adds the convention to have a comma after the last field in an entry. This is to make fields easier to copy-paste and edit.


.pre-commit-config.yaml
```diff
@@ -90,6 +90,40 @@ repos:
...
 
+      - id: format-bibtex
+        name: Format BibTeX references
+        language: system
+        files: \.bib$
+        types: [file]
+        entry: >
+          bash -c ' if ! command -v bibtex-tidy &> /dev/null; then
+            echo "bibtex-tidy not found, installing via npm..."
+            npm install -g bibtex-tidy
```
VojtechStep, line 101: AFAIK so far we managed to avoid explicitly depending on and messing with the user's nodejs installation. I have a strong preference to keep it that way. bibtex-tidy looks like it ships with a [pre-commit hook](https://github.com/FlamingTempura/bibtex-tidy#sec-precommit). Is there a reason for not using it?

fredrik-bakke: Oh, nice!

fredrik-bakke: I switched up the setup to do as you suggest. Thanks for checking!

references.bib
```diff
@@ -1,959 +1,956 @@
...
-  journal     = {{Logical Methods in Computer Science}},
-  volume      = {19},
-  issue       = {2},
-  year        = {2023},
-  month       = {05},
-  keywords    = {Computer Science - Logic in Computer Science ; Mathematics - Logic},
-  eprint      = {2111.00482},
-  eprinttype  = {arxiv},
-  eprintclass = {cs}
+  title         = {On Small Types in Univalent Foundations},
```
VojtechStep, line 281: The double curly braces are removed from this title and journal, and from at least on other title (Awodey22); is that change correct?

fredrik-bakke: Thanks for catching that. Indeed, it was a mistake that crept in during development

fredrik-bakke:
@VojtechStep Just want to check, you are not finished with a review?

VojtechStep:
I am, I was waiting for the checks to finish, but now I see they aren't running for some reason

fredrik-bakke:
Oh, right, strange. The empty commit seems to have fixed it, though.

fredrik-bakke:
So then we're good to merge, @VojtechStep? :)

VojtechStep:
Yup, perfect, thanks 👌

PR #1433, opened 2025-05-19
RFC: Functor currying

arnoudvanderleer:
I thought it would be nice to add some abstract nonsense to the library. Since agda (and in particular, agda-unimath) is new to me, I would like to get some early feedback on my coding style (and other things that would be good to know before I write more code).
If it turns out that this is perfect already, I am fine with merging it immediately, and having the rest of my contributions in a future PR ^^

### DIFF
--- a/src/foundation/cartesian-product-types.lagda.md
+++ b/src/foundation/cartesian-product-types.lagda.md
@@ -31,24 +31,6 @@ tr-product :
 tr-product B C refl u = refl
 ```
 
-### Identities of products
-
-```agda
-identity-product :
-  {l1 l2 : Level}
-  {A : UU l1} {B : UU l2}
-  {a a' : A} {b b' : B}
-  → a ＝ a' → b ＝ b' → (a , b) ＝ (a' , b')
-identity-product refl refl = refl
-
-identity-product' :
-  {l1 l2 : Level}
-  {A : UU l1} {B : UU l2}
-  {ab ab' : A × B}
-  → pr1 ab ＝ pr1 ab' → pr2 ab ＝ pr2 ab' → ab ＝ ab'
-identity-product' refl refl = refl
-```
-
 ### Subuniverses closed under cartesian product types
 
 ```agda
--- a/src/foundation/cartesian-product-types.lagda.md
+++ b/src/foundation/cartesian-product-types.lagda.md
@@ -31,6 +31,24 @@ tr-product :
 tr-product B C refl u = refl
 ```
 
+### Identities of products
+
+```agda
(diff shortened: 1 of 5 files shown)


src/foundation/cartesian-product-types.lagda.md
```diff
@@ -31,6 +31,24 @@ tr-product :
...
+
+```agda
+identity-product :
+  {l1 l2 : Level}
+  {A : UU l1} {B : UU l2}
+  {a a' : A} {b b' : B}
+  → a ＝ a' → b ＝ b' → (a , b) ＝ (a' , b')
+identity-product refl refl = refl
+
+identity-product' :
```
lowasser, line 44: I'm pretty sure this is reinventing `eq-pair` [here](https://unimath.github.io/agda-unimath/foundation.equality-cartesian-product-types.html#1385).

ElifUskuplu:
Hi @arnoudvanderleer thanks for this PR, and sorry that it has been waiting for feedback for so long. I compared it with the current master and wanted to ask if you are still interested in updating it. If so, here are some comments.

Most of the PR is new material. The library does not have a currying functor from [C × D, E] to [C, [D, E]], and it does not have the construction of an equivalence of precategories from a fully faithful and split essentially surjective functor. So both of those are welcome.

The two lemmas you added to fully-faithful-functors-precategories, about the inverse map preserving composition and identities, do not exist as separate lemmas yet, but the same argument is already written out inside the proof that fully faithful functors are essentially injective. It would be nice to keep your lemmas and have that existing proof use them, so the argument is not repeated. In general, whenever something already exists in the library, please use it instead of defining it again, as happened with eq-pair earlier in this PR.

A few smaller things, mostly from docs/CODINGSTYLE.md and docs/FILE-CONVENTIONS.md. Names describe the type of the thing being constructed and end with the structure they live in, so for example curry-functor-Precategory, and the final construction in the equivalence file should get a descriptive name instead of "result". The second file still has "File title" as its title, and the Idea sections should link the concepts they mention to their pages in the library. In the equivalence file the two precategories share the same universe levels, and the style guide asks for separate levels for each assumed type, as in the rest of the category theory folder. The style guide also asks that the module named after the file be the only named module, so CurryFunctor should become an anonymous module. Finally, please run make pre-commit before pushing; it formats the code and regenerates the namespace index for you.

An uncurrying functor in the other direction, and the fact that the two form an equivalence, would be a natural next step, but that can be a separate PR.

If you no longer have time for this, just let us know and someone can continue from your branch, keeping you as a co-author.

PR #1438, opened 2025-06-04
Fix #1437

### LINKED ISSUE #1437: Sorting order for codespell dictionary is platform dependent
The sorting key seems to depend on platform or some other local configuration. My own machine has suddenly started to sort the files case-insensitively.

fredrik-bakke:


.pre-commit-config.yaml
```diff
@@ -33,7 +33,7 @@ repos:
     hooks:
       - id: sort-codespell-files
         name: Sort codespell dictionary and ignore files
-        entry: bash -c 'for f in "$@"; do sort -o "$f" "$f"; done' --
+        entry: bash -c 'for f in "$@"; do LC_ALL=C sort -o "$f" "$f"; done' --
```
VojtechStep, line 36: I don't have a laptop on me, but the syntax should be `LC_ALL=C sort ...`

PR #1440, opened 2025-06-05
Update category theory tables and define category of simplicial sets

fredrik-bakke:


tables/precategories.md
VojtechStep, line 1: Some removals in this table seem a bit arbitrary, e.g. why were the groups and metric spaces precategories removed?

fredrik-bakke: Because theyre in the category list I believe

VojtechStep: But then commutative rings and sets should also go. Though are the tables included together, or is one supposed to know to also look at the table of categories if they want to find precategories?

PR #1444, opened 2025-06-14
Convolution of sequences in commutative semirings and rings

lowasser:
Convolution is also the multiplication operation on formal power series, so this is intended as a step on the road to analysis.


src/elementary-number-theory/pairs-with-natural-sums.lagda.md
```diff
@@ -0,0 +1,290 @@
...
+            ( eq-type-Prop (Id-Prop ℕ-Set _ _)))
+
+  equiv-permute-components-triple-with-sum-pr2 :
+    Aut (Σ (pair-with-sum-ℕ n) (pair-with-sum-ℕ ∘ pr1 ∘ pr2))
+  equiv-permute-components-triple-with-sum-pr2 =
+    map-equiv-permute-components-triple-with-sum-pr2 ,
+    ( ( map-inv-equiv-permute-components-triple-with-sum-pr2 ,
+        is-section-map-inv-equiv-permute-components-triple-with-sum-pr2) ,
+      ( map-inv-equiv-permute-components-triple-with-sum-pr2 ,
+        is-retraction-map-inv-equiv-permute-components-triple-with-sum-pr2))
```
fredrik-bakke: It should be preferred to use the `is-equiv-is-invertible` in this case for two reasons. For one it gives the guarantee that the left and right inverse are definitionally the same. Secondly, it makes the definition implementation independent of `is-equiv`.

src/elementary-number-theory/strict-inequality-natural-numbers.lagda.md
```diff
@@ -404,3 +408,16 @@ preserves-le-add-ℕ {a} {b} {c} {d} H K =
...
+
+```agda
+equiv-le-succ-ℕ-leq-ℕ :
+  (n : ℕ) → Σ ℕ (λ k → le-ℕ k (succ-ℕ n)) ≃ Σ ℕ (λ k → leq-ℕ k n)
+equiv-le-succ-ℕ-leq-ℕ n =
+  ( λ (k , k<sn) → k , leq-le-succ-ℕ k n k<sn) ,
+  ((λ (k , k≤n) → k , le-succ-leq-ℕ k n k≤n) ,
+    λ (k , k≤n) → eq-pair-Σ refl (eq-type-Prop (leq-ℕ-Prop k n))) ,
+  ((λ (k , k≤n) → k , le-succ-leq-ℕ k n k≤n) ,
+    λ (k , k<sn) → eq-pair-Σ refl (eq-type-Prop (le-ℕ-Prop k (succ-ℕ n))))
```
fredrik-bakke: here you should probably also use `is-equiv-is-invertible`

src/elementary-number-theory/pairs-with-natural-sums.lagda.md
```diff
@@ -0,0 +1,290 @@
...
+        let
+          (l' , l'+k=n) =
+            subtraction-leq-ℕ k n (leq-subtraction-ℕ k n l l+k=n)
+        in
+          eq-pair-eq-fiber
+            ( eq-pair-Σ
+              ( is-injective-right-add-ℕ k (l'+k=n ∙ inv l+k=n))
+              ( eq-type-Prop (Id-Prop ℕ-Set (l +ℕ k) n)))) ,
+    (( λ (k , l , l+k=n) → k , leq-subtraction-ℕ k n l l+k=n) ,
+      (λ (k , k≤n) → eq-pair-eq-fiber (eq-type-Prop (leq-ℕ-Prop k n))))
```
fredrik-bakke: Replace this by a `is-equiv-is-invertible`. I'd also consider factoring out the `is-section` and `is-retraction` part, but I won't block the PR on this.

src/commutative-algebra/convolution-sequences-commutative-semirings.lagda.md
```diff
@@ -0,0 +1,548 @@
...
+  pr1
+    ( pr2
+      ( pr2
+        ( has-mul-convolution-additive-commutative-monoid-sequence-Commutative-Semiring))) =
+    left-distributive-convolution-add-sequence-Commutative-Semiring R
+  pr2
+    ( pr2
+      ( pr2
+        ( has-mul-convolution-additive-commutative-monoid-sequence-Commutative-Semiring))) =
+    right-distributive-convolution-add-sequence-Commutative-Semiring R
```
fredrik-bakke: would you not find this more readable if it was defined as an iterated pair?

fredrik-bakke: same for most of the surrounding ones

src/commutative-algebra/convolution-sequences-commutative-semirings.lagda.md
```diff
@@ -0,0 +1,548 @@
...
+                  ( left-distributive-mul-sum-finite-Commutative-Semiring
+                    ( R)
+                    ( _)
+                    ( _)
+                    ( _)))
+      where
+        _*R_ :
+          type-Commutative-Semiring R → type-Commutative-Semiring R →
+          type-Commutative-Semiring R
+        _*R_ = mul-Commutative-Semiring R
```
fredrik-bakke: If you insert this as a `let` binding at the top instead, then the reader does not have to scroll all the way down to find out what the notation means.

src/commutative-algebra/convolution-sequences-commutative-rings.lagda.md
```diff
@@ -0,0 +1,243 @@
...
+    unit-convolution-sequence-Commutative-Semiring
+      ( commutative-semiring-Commutative-Ring R)
+
+  abstract
+    left-unit-law-convolution-sequence-Commutative-Ring :
+      (a : sequence (type-Commutative-Ring R)) →
+      convolution-sequence-Commutative-Ring R
+        ( unit-convolution-sequence-Commutative-Ring)
+        ( a)
+        ＝ a
```
malarbol, line 100: I think the `＝` should go the end of the previous line and `a` aligned with `convolution-sequence-Commutative-Ring`. @fredrik-bakke ?

src/commutative-algebra/convolution-sequences-commutative-rings.lagda.md
```diff
@@ -0,0 +1,243 @@
...
+    left-unit-law-convolution-sequence-Commutative-Ring =
+      left-unit-law-convolution-sequence-Commutative-Semiring
+        ( commutative-semiring-Commutative-Ring R)
+
+    right-unit-law-convolution-sequence-Commutative-Ring :
+      (a : sequence (type-Commutative-Ring R)) →
+      convolution-sequence-Commutative-Ring R
+        ( a)
+        ( unit-convolution-sequence-Commutative-Ring)
+        ＝ a
```
malarbol, line 110: same indentation comment.

src/elementary-number-theory/strict-inequality-natural-numbers.lagda.md
malarbol: I think they're some space/parenthesis missing.

src/elementary-number-theory/strict-inequality-natural-numbers.lagda.md
```diff
@@ -414,10 +414,10 @@ preserves-le-add-ℕ {a} {b} {c} {d} H K =
...
-  ( λ (k , k<sn) → k , leq-le-succ-ℕ k n k<sn) ,
-  ((λ (k , k≤n) → k , le-succ-leq-ℕ k n k≤n) ,
-    λ (k , k≤n) → eq-pair-Σ refl (eq-type-Prop (leq-ℕ-Prop k n))) ,
-  ((λ (k , k≤n) → k , le-succ-leq-ℕ k n k≤n) ,
-    λ (k , k<sn) → eq-pair-Σ refl (eq-type-Prop (le-ℕ-Prop k (succ-ℕ n))))
+pr1 (equiv-le-succ-ℕ-leq-ℕ n) (k , k<sn) = (k , leq-le-succ-ℕ k n k<sn)
+pr2 (equiv-le-succ-ℕ-leq-ℕ n) =
+  is-equiv-is-invertible
+    ( λ (k , k≤n) → k , le-succ-leq-ℕ k n k≤n)
+    ( λ (k , k≤n) → eq-pair-Σ refl (eq-type-Prop (leq-ℕ-Prop k n)))
```
fredrik-bakke: `eq-pair-Σ refl` is the same as `eq-pair-eq-fiber` btw

src/commutative-algebra/convolution-sequences-commutative-rings.lagda.md
fredrik-bakke: Yes, thank you for the comment. I try to spend less time nitpicking to save my time, although I do appreciate it when these things are in order.

src/elementary-number-theory/pairs-with-natural-sums.lagda.md
fredrik-bakke, line 1: Just in case this isn't an alternative you've considered, we could also call this a binary sum decomposition of natural numbers. That might fit better into a general naming scheme idk

lowasser: I'm happy to rename it later if a more general naming scheme materializes, is how I feel about that, I think.

fredrik-bakke, line 1: I am not happy with the name of this file and think it should be renamed before merging. How are we supposed to interpret "pairs with natural sums", that there is some sort of pairing operation on types equipped with a sum that is _natural_... in what?

fredrik-bakke: If you don't have a better proposal I suggest we stick with `binary-sum-decompositions-natural-numbers` for now

fredrik-bakke (approved):

src/univalent-combinatorics/standard-finite-types.lagda.md
```diff
@@ -491,3 +495,39 @@ is-preunivalent-Fin : is-preunivalent Fin
...
+  (n : ℕ) → is-retraction (nat-le-Fin-reverse n) (fin-reverse-le-ℕ n)
+is-retraction-fin-reverse-le-ℕ (succ-ℕ n) (inl x) =
+  ap inl (is-retraction-fin-reverse-le-ℕ n x)
+is-retraction-fin-reverse-le-ℕ (succ-ℕ n) (inr star) = refl
+
+equiv-fin-le-ℕ : (n : ℕ) → Fin n ≃ Σ ℕ (λ m → le-ℕ m n)
+equiv-fin-le-ℕ n =
+  nat-le-Fin-reverse n ,
+  ( fin-reverse-le-ℕ n , is-section-fin-reverse-le-ℕ n) ,
+  ( fin-reverse-le-ℕ n , is-retraction-fin-reverse-le-ℕ n)
```
malarbol, line 529: Out of curiosity, is there any particular reason to use `nat-Fin-reverse` instead of `nat-Fin` (and, say, `nat-le-Fin`, etc.) here ?

lowasser: I had a much harder time proving the section and retraction laws with those.

PR #1445, opened 2025-06-15
Series and convergence

lowasser:


lowasser:
There might need to be additional continuity conditions on the ring operations.

I have to admit I'm leaning more and more towards just naming the backing structure `Analysis-Setting`.

src/analysis/series.lagda.md
```diff
@@ -0,0 +1,75 @@
...
+
+  partial-sums-series :
+    sequence (type-Commutative-Ring-In-Complete-Metric-Space RM)
+  partial-sums-series n =
+    sum-fin-sequence-type-Commutative-Ring
+      ( commutative-ring-Commutative-Ring-In-Complete-Metric-Space RM)
+      ( n)
+      ( λ k → terms-series σ (nat-Fin n k))
+
+  converges-series : UU (l1 ⊔ l2)
```
fredrik-bakke: How do you feel about using `is-convergent` over `converges`? The former makes it easier to conclude at a glance that we are dealing with a property. For instance, one might wrongly anticipate some sort of type of `convergence` that might not be a proposition, and then one has to read the whole word `converges` in order to make sure it is the propositional predicate.

src/analysis/series.lagda.md
```diff
@@ -0,0 +1,75 @@
...
+
+open import univalent-combinatorics.standard-finite-types
+```
+
+</details>
+
+## Idea
+
+A series is an infinite sum taken over the natural numbers. If it has a limit,
+it is said to converge, otherwise it is said to diverge.
```
fredrik-bakke, line 29: Please add links and concept invocation(s)

fredrik-bakke:
> There might need to be additional continuity conditions on the ring operations.
> 
> I have to admit I'm leaning more and more towards just naming the backing structure `Analysis-Setting`.

The concept of a "Commutative ring in a complete metric space" seems closely related to the notion of a [Banach algebra](https://ncatlab.org/nlab/show/Banach+algebra) or some related variant thereof. Perhaps you can use that to come up with a shorter name for the structure you are considering that also doesn't capture the namespace "analysis (setting)", which is rich in structures to consider for different things.

lowasser:
Restarting this.

Issue #1421, opened 2025-05-01
Refactor `metric-spaces` module


malarbol:
Recent contributions in the `metric-spaces` modules raise concerns about the names of some concepts in the `metric-spaces` module. 

- [ ] rename `indistinguishable` --> `indiscernible` (see [comment](https://github.com/UniMath/agda-unimath/pull/1378/#issuecomment-2844877794))
- [ ] refactor `pre`/`pseudo`/`metric-spaces`

fredrik-bakke:
Here's a couple more points:
- [x] The divide metric space/metric structure seems a bit over the top. I suggest we factor these to be contained in the same files (and same for premetrics and pseudometrics and any other such divides in this namespace).
- [ ] The concept of indistinguishability should be factored to separate files, called `similarty-of-elements-metric-spaces` and so on. This is consistent with other notions of similarity in the library.
- [x] The definition of indistinguishability in premetric spaces is not the correct one, it only applies to (reflexive?) symmetric premetric spaces. I propose we forget the non pseudo- case and only first define similarity for pseudometric spaces.

fredrik-bakke:
For future reference, the correct similarity relation on premetric spaces is probably

```text
(x ≍ y) :=
  ((ε : ℚ⁺) (z : A) → (N ε x z ↔ N ε y z)) ×
  ((ε : ℚ⁺) (z : A) → (N ε z x ↔ N ε z y)) ×
  ((ε : ℚ⁺) → (N ε x x ↔ N ε y y))
```

malarbol:
> For future reference, the correct similarity relation on premetric spaces is probably
> 
> ```
> (x ≍ y) :=
>   ((ε : ℚ⁺) → (B ε x x ↔ B ε y y)) ×
>   ((ε : ℚ⁺) (z : A) → (B ε x z ↔ B ε y z)) ×
>   ((ε : ℚ⁺) (z : A) → (B ε z x ↔ B ε z y))
> ```

After a few weeks I finally _understood_ this comment. But I'm pretty sure the first condition is redundant as it's a consequence of the other two.

fredrik-bakke:
> > For future reference, the correct similarity relation on premetric spaces is probably
> > 
> > ```
> > (x ≍ y) :=
> >   ((ε : ℚ⁺) → (B ε x x ↔ B ε y y)) ×
> >   ((ε : ℚ⁺) (z : A) → (B ε x z ↔ B ε y z)) ×
[...]

PR #1447, opened 2025-06-17
Metric semigroups

malarbol:
This PR introduces the notion of `Metric-Semigroup`, metric spaces with a short associative binary operator.

Credit @lowasser for https://github.com/UniMath/agda-unimath/pull/1447/commits/9e36ee4cb60533a860a03b784fda0145abbf31da


src/metric-spaces/metric-semigroups.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  where
+
+  mul-short-mul-Metric-Space :
+    short-function-Metric-Space
+      ( A)
+      ( metric-space-of-short-functions-Metric-Space A A) →
```
fredrik-bakke: Is it formalized somewhere that `metric-space-of-short-functions-Metric-Space A` is right adjoint to "`product-Metric-Space A`", so that this can confidently be used for "binary short functions"?

fredrik-bakke: Otherwise that might be a good sanity-check to formalize

malarbol: > Is it formalized somewhere that metric-space-of-short-functions-Metric-Space A is right adjoint to "product-Metric-Space A", so that this can confidently be used for "binary short functions"?

I don't think so, we didn't do much in the categorical POV of metric spaces. I don't think we even have the concept "`product-Metric-Space A`" Do you think it's true? If so, I guess we could add this result sometimes in the future. I've never worked with right adjoints and I'm still working on #1421 so I'll need some time to think about it.

malarbol:
This is also a suggestion of prelude for some `Analysis-Space` mentioned in #1445. Actually, if we think of the `analysis` module as an intersection of algebraic structures (groups, rings, linear algebra) and convergence structures (metric-spaces, topological-spaces in the future, etc.), this could probably go to `analysis/metric-semigroups.lagda.md`.

src/metric-spaces/metric-semigroups.lagda.md
fredrik-bakke: It might be true, and I think it is a question worth considering since you are defining what it means to be a binary short function here. If it is not true you may be giving a wrong definition. The adjointedness condition I mention is known in the literature as [cartesian closedness](https://ncatlab.org/nlab/show/cartesian+closed+category), and means that we can observe collections of morphisms in the category as _internal objects_ in the category itself. E.g., given that it is true, the metric space of short functions between two metric spaces represent in a concrete sense the collection of short functions _viewed internally_. The concrete way in which it represents this collection is precisely by the equivalence

$$\mathrm{Hom}(A \times B, C) \cong \mathrm{Hom}(A, C^B)$$

Where $\mathrm{Hom}$ is the type of short functions, $A \times B$ is the product metric space, and $C^B$ is the metric space of short functions. This equivalence is what the adjointedness condition I mention would unfold to.

fredrik-bakke: The question I'd like you to ask yourself is: what is the difference between

```agda
    short-function-Metric-Space A (metric-space-of-short-functions-Metric-Space A A)
```

and

```agda
    short-function-Metric-Space (product-Metric-Space A A) A
``` 

Are they the same? If so, there is no problem.

fredrik-bakke:
> This is also a suggestion of prelude for some `Analysis-Space` mentioned in #1445.

I don't see the need for using such a generic term as `Analysis-Space`, when the object at hand seems to be essentially a `Banach-Ring`. 

> Actually, if we think of the `analysis` module as an intersection of algebraic structures (groups, rings, linear algebra) and convergence structures (metric-spaces, topological-spaces in the future, etc.), this could probably go to `analysis/metric-semigroups.lagda.md`.

I like this idea!

malarbol:
> > This is also a suggestion of prelude for some `Analysis-Space` mentioned in #1445.
> 
> I don't see the need for using such a generic term as `Analysis-Space`, when the object at hand seems to be essentially a `Banach-Ring`.

Sure, at the moment `Analysis-Space` is just a placeholder name for the concept we are trying to pin out. It might as well be `Complete-Metric-Ring`. I think `Banach-Space` would fit better for complete metric spaces where all elements are at finite distance (what I formerly called _total_ complete metric space) so that Banach fixed point theorem holds in Banach spaces. 

> > Actually, if we think of the `analysis` module as an intersection of algebraic structures (groups, rings, linear algebra) and convergence structures (metric-spaces, topological-spaces in the future, etc.), this could probably go to `analysis/metric-semigroups.lagda.md`.
> 
> I like this idea!

Thanks. I'll move the module.

src/metric-spaces/metric-semigroups.lagda.md
malarbol: > Are they the same? If so, there is no problem.

I think they are but I'll need some time to think about it and write a proper proof. As I said, we don't really have `product-Metric-Space` yet so there are a lot of missing pieces.

fredrik-bakke:
> Sure, at the moment `Analysis-Space` is just a placeholder name for the concept we are trying to pin out. It might as well be `Complete-Metric-Ring`. I think `Banach-Space` would fit better for complete metric spaces where all elements are at finite distance (what I formerly called total complete metric space) so that Banach fixed point theorem holds in Banach spaces.

Good, just wanted to make sure we are on the same page. Keep in mind that a banach space is also a vector space 👍

src/metric-spaces/metric-semigroups.lagda.md
malarbol: > The question I'd like you to ask yourself is: what is the difference between
> 
> ```agda
>     short-function-Metric-Space A (metric-space-of-short-functions-Metric-Space A A)
> ```
> 
> and
> 
> ```agda
>     short-function-Metric-Space (product-Metric-Space A A) A
> ```
> 
> Are they the same? If so, there is no problem.

I've been working on this in https://github.com/UniMath/agda-unimath/pull/1447/commits/ac05ffd690f5e31c31c60e73235bd7ee2428ffd3. I can't quite finish the proof though; I'm not sure if I'm missing something or if it's simply not true?

malarbol: So, basically, I could construct an isometry 
```agda
  isometry-ev-pair-short-function-product-Metric-Space :
    isometry-Metric-Space
      ( metric-space-of-short-functions-Metric-Space
        ( product-Metric-Space X Y)
        ( Z))
      ( metric-space-of-short-functions-Metric-Space
        ( X)
        ( metric-space-of-short-functions-Metric-Space Y Z))
```
and and inverse **in the space of functions** `product-Metric-Space X Y -> Z`
```agda
  map-ind-short-function-product-Metric-Space :
    short-function-Metric-Space
      ( X)
      ( metric-space-of-short-functions-Metric-Space Y Z) →
    map-type-Metric-Space
      ( product-Metric-Space X Y)
      ( Z)
```
If the latter takes value in the space of **short** functions, then we have 
```agda
    metric-space-of-short-functions-Metric-Space
      ( product-Metric-Space X Y)
      ( Z) ＝
    metric-space-of-short-functions-Metric-Space
      ( X)
      ( metric-space-of-short-functions-Metric-Space Y Z)
```
but I don't know how to prove this last step. Do you have any idea how to complete this proof? Or maybe it's not always true?

malarbol: If anyone is reading this:
- with this setting, I can prove that `map-ind-short-function-product-Metric-Space f` is uniformly continuous, but still not short;
- I'm starting to believe it's not true: `add-ℚ` is a 
```
short-function-Metric-Space
  ( ℚ)
  ( metric-space-of-short-functions-Metric-Space ℚ ℚ)
```
but I don't think it's short as a function `product-Metric-Space ℚ ℚ → ℚ`.

So, regarding https://github.com/UniMath/agda-unimath/pull/1447#discussion_r2153300397, let's assume they're not the same and _we have a problem_. How should we solve it?

NB: whatever the definition of `Metric-Semigroup` ends up being, I think we'd want `ℚ , add-ℚ` to be an instance of it.

src/analysis/metric-semigroups.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+open import metric-spaces.metric-spaces
+open import metric-spaces.short-functions-metric-spaces
+```
+
+</details>
+
+## Idea
+
+A {{#concept "metric semigroup" Agda=Metric-Semigroup}} is a
+[metric-space](metric-spaces.metric-spaces.md) equipped with a
```
fredrik-bakke, line 32:
```suggestion
[metric space](metric-spaces.metric-spaces.md) [equipped](foundation.structure.md) with a
```

src/metric-spaces/metric-semigroups.lagda.md
fredrik-bakke: It seems impossible to me that `add-ℚ` is not a short function viewed as `product-Metric-Space ℚ ℚ → ℚ`, considering it is commutative. Are you sure this is the case?

fredrik-bakke: Based on [this Wikipedia page](https://en.wikipedia.org/wiki/Product_metric), I propose you define the product metric space using the supremum, i.e. maximum distance of the coordinates, and then use `short-function-Metric-Space (product-Metric-Space A A) A` in your definitions. This should always be correct, and then we can later resolve the question of wheter ` short-function-Metric-Space A (metric-space-of-short-functions-Metric-Space A A)` is equivalent or not.

src/metric-spaces/products-metric-spaces.lagda.md
```diff
@@ -0,0 +1,602 @@
...
+open import metric-spaces.uniformly-continuous-functions-metric-spaces
+```
+
+</details>
+
+## Idea
+
+A pair of [metric spaces](metric-spaces.metric-spaces.md) induces a metric space
+over the [Cartesian product](foundation.cartesian-product-types.md) of the
+carrier types of its arguments.
```
fredrik-bakke, line 54: and in fact there is an infinite family of equivalent choices of metric on the product. One for every "real number" in $(0,\infty]$. The canonical choice is $\infty$.

fredrik-bakke: which is what you've chosen below I think

src/metric-spaces/products-metric-spaces.lagda.md
```diff
@@ -0,0 +1,602 @@
...
+    is-triangular-Premetric structure-product-Metric-Space
+  is-triangular-structure-product-Metric-Space
+    (a1 , b1) (a2 , b2) (a3 , b3) d12 d23 (a1~a2 , b1~b2) (a2~a3 , b2~b3) =
+      ( is-triangular-structure-Metric-Space A a1 a2 a3 d12 d23 a1~a2 a2~a3 ,
+        is-triangular-structure-Metric-Space B b1 b2 b3 d12 d23 b1~b2 b2~b3)
+
+  is-local-structure-product-Metric-Space :
+    is-local-Premetric structure-product-Metric-Space
+  is-local-structure-product-Metric-Space =
+    is-local-is-tight-Premetric
```
fredrik-bakke, line 95: looks like you could add `is-tight-structure-product-Metric-Space` as another lemma?

src/metric-spaces/products-metric-spaces.lagda.md
```diff
@@ -0,0 +1,602 @@
...
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} (A : Metric-Space l1 l2) (B : Metric-Space l3 l4)
+  where
+
+  type-product-Metric-Space : UU (l1 ⊔ l3)
+  type-product-Metric-Space = type-Metric-Space A × type-Metric-Space B
+
+  structure-product-Metric-Space : Premetric (l2 ⊔ l4) type-product-Metric-Space
```
fredrik-bakke, line 68: why not call it `premetric-product-Metric-Space`?

src/metric-spaces/products-metric-spaces.lagda.md
```diff
@@ -0,0 +1,602 @@
...
+      g)
+
+  isometry-ev-pair-short-function-product-Metric-Space :
+    isometry-Metric-Space
+      ( metric-space-of-short-functions-Metric-Space
+        ( product-Metric-Space X Y)
+        ( Z))
+      ( metric-space-of-short-functions-Metric-Space
+        ( X)
+        ( metric-space-of-short-functions-Metric-Space Y Z))
```
fredrik-bakke, line 338: Wait, doesn't this show that the category of metric spaces is cartesian closed? 🤩

fredrik-bakke: Oh, I see, isometries don't characterize equality

fredrik-bakke: That's a bummer. So then only the category of "metric spaces up to isometry" is cartesian closed

fredrik-bakke: Still I find it strange if they are not equivalent...

src/metric-spaces/metric-semigroups.lagda.md
malarbol: > Based on [this Wikipedia page](https://en.wikipedia.org/wiki/Product_metric), I propose you define the product metric space using the supremum, i.e. maximum distance of the coordinates,

I think that's what this definition does: upper bounds of the "product distance" are upper bounds on each coordinates, i.e. upper bounds of the max if the distances. This is also coherent with the definition of "dependent product of metric spaces" where the distance is the supremum. 

> It seems impossible to me that `add-ℚ` is not a short function viewed as `product-Metric-Space ℚ ℚ → ℚ`, considering it is commutative. 

I was surprised too, and I don't see how commutativity should help...

> Are you sure this is the case?

Unless I'm mistaken, it basically boils down to proving that 

```
|(x + y) - (x' + y')| ≤ max (|x - x'| , |y - y'|) 
```

and I don't see a way out. It _is_ a `lipschitz-function-Metric-Space (product-Metric-Space A A) A` though. Would that help?

 > then use `short-function-Metric-Space (product-Metric-Space A A) A` in your definitions. This should always be correct, and then we can later resolve the question of wheter ` short-function-Metric-Space A (metric-space-of-short-functions-Metric-Space A A)` is equivalent or not.

So `ℚ , add-ℚ` will _not_ be a metric semigroup?

src/metric-spaces/products-metric-spaces.lagda.md
malarbol: after #1421 there will be no more "premetrics". With #1450 this would probably be something like
```agda
neighborhood-prop-product-Metric-Space : Rational-Neighborhood-Relation (l2 ⊔ l4) type-product-Metric-Space
```

malarbol:
There are sill a lot of open questions on this PR so I'll put in draft-mode until we figure things out. In the meantime I hope we can have a look at #1450 so we can advance on #1421 etc.

PR #1451, opened 2025-07-10
Rational abelian groups

djspacewhale:
A **rational abelian group** is an abelian group that is both divisible and torsion-free. These are special among the abelian groups as they are precisely the left modules over the ring of rational numbers. This pull builds significant infrastructure for formalizing the theory of rational abelian groups, with eyes towards rational homotopy theory and commutative algebra.

Issue #1435, opened 2025-06-03
Analysis: series


lowasser:
I'm getting back to this and trying to build a good definition for _series_.

Wikipedia uses the term "series" to refer only to infinite sums, and I buy that.  I also claim that there's really no useful notion of series without a corresponding notion of convergence.

So what exactly is the setting in which a series is defined?  A commutative monoid with a metric space?  We don't necessarily have to be quite so general, but I'd at least like something general enough to encompass real and complex numbers, and maybe even some other handy spaces.

If we were in the habit of using standard topologies instead of metric spaces, we might define it over a topological group?  I don't know.  Ideas welcome.  I could also just start the ungeneralized version, sticking to real numbers, but that'd feel very out of character for this project.

fredrik-bakke:
I'll likely have some more thought-out input on this later, but for now I can recommend checking out the concept of "closeness spaces" as considered in Todd Ambride's thesis _[Exact Real Search: Formalised Optimisation and Regression in Constructive Univalent Mathematics](https://arxiv.org/abs/2401.09270)_. These seem like a promising setting in which to study convergence of series constructively. In particular, they should at times allow you to avoid principles of omniscience, although I do not have sufficient experience in this area to make a definitive statement. If you should choose to use this concept, I have the compactness of the extended natural numbers formalized in some unmerged branch (currently blocked on #1264 #1387).

lowasser:
I have to admit I'm not yet convinced that we need -- at least, anytime soon -- something specialized to constructive settings, at least for the problem I'm describing here.  The scope of this issue is intended to enclose only "in what settings can you define convergence of a series at all."

(The constructive setting places lots of constraints on how you can _show_ convergence -- no monotone convergence, you have to lean much harder on Cauchy conditions -- but that's a separate problem that we're making lots of progress on with Bernoulli's inequality and the like.)

At this point, my default is to construct, like, `Ab-With-Metric` or some other combined structure.

fredrik-bakke:
Ah, I see, I misunderstood your question the first time around. My bad

lowasser:
Or, reframed: we already have metric spaces and lots of infrastructure for defining convergence there, and I believe that'll suffice for our near term needs.  I'm just starting to skim the thesis you linked, but I think it's strictly more sophisticated than we need to get analysis and some basic functions defined.

lowasser:
[...]

PR #1452, opened 2025-07-12
Hom-ring, ℤ, ℚ, etc.

malarbol:
This PR aims to fill some gaps in ring-theory relations between ℤ and ℚ:

- `rational-ℤ` is the initial ring homomorphism `ℤ → ℚ`;
- elements of `ℤ⁺` are invertible in `ℚ`.

### DIFF
--- a/src/elementary-number-theory/ring-of-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/ring-of-rational-numbers.lagda.md
@@ -14,14 +14,19 @@ open import commutative-algebra.commutative-rings
 open import elementary-number-theory.additive-group-of-rational-numbers
 open import elementary-number-theory.multiplication-rational-numbers
 open import elementary-number-theory.multiplicative-monoid-of-rational-numbers
+open import elementary-number-theory.rational-numbers
+open import elementary-number-theory.ring-of-integers
 
 open import foundation.coproduct-types
 open import foundation.dependent-pair-types
+open import foundation.homotopies
+open import foundation.identity-types
 open import foundation.unital-binary-operations
 open import foundation.universe-levels
 
 open import group-theory.semigroups
 
+open import ring-theory.homomorphisms-rings
 open import ring-theory.rings
 ```
 
@@ -63,4 +68,22 @@ pr2 ring-ℚ = has-mul-abelian-group-add-ℚ
 commutative-ring-ℚ : Commutative-Ring lzero
 pr1 commutative-ring-ℚ = ring-ℚ
 pr2 commutative-ring-ℚ = commutative-mul-ℚ
+```
+
+### The inclusion of integers in the rationals is the initial ring homomorphism
+
+```agda
+hom-ring-rational-ℤ : hom-Ring ℤ-Ring ring-ℚ
+hom-ring-rational-ℤ =
+  ( hom-add-rational-ℤ) ,
+  ( λ {x y} → inv (mul-rational-ℤ x y)) ,
+  ( refl)
+
+htpy-map-initial-hom-ring-rational-ℤ : map-initial-hom-Ring ring-ℚ ~ rational-ℤ
+htpy-map-initial-hom-ring-rational-ℤ =
+  htpy-initial-hom-Ring ring-ℚ hom-ring-rational-ℤ
+
+eq-initial-hom-ring-rational-ℤ : initial-hom-Ring ring-ℚ ＝ hom-ring-rational-ℤ
+eq-initial-hom-ring-rational-ℤ =
+  contraction-initial-hom-Ring ring-ℚ hom-ring-rational-ℤ
 ```
--- a/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
@@ -27,6 +27,7 @@ open import foundation.action-on-identifications-functions
 open import foundation.binary-transport
 open import foundation.dependent-pair-types
 open import foundation.functoriality-dependent-pair-types
+open import foundation.identity-types
 
 open import group-theory.groups
 ```
@@ -117,4 +118,44 @@ smaller-reciprocal-ℚ⁺ q⁺@(q , _) =
           ( rational-ℚ⁺ (positive-rational-ℕ⁺ n⁺ *ℚ⁺ q⁺))
           ( 1<nq)))
     ( bound-archimedean-property-ℚ⁺ q⁺ one-ℚ⁺)
+```
+
+### The reciprocal of `n` is a multiplicative inverse of `n`
+
+```agda
+module _
+  (n : ℕ⁺)
+  where
+
+  left-inverse-law-positive-reciprocal-rational-ℕ⁺ :
+    mul-ℚ⁺
+      ( positive-reciprocal-rational-ℕ⁺ n)
+      ( positive-rational-ℕ⁺ n) ＝
+    one-ℚ⁺
+  left-inverse-law-positive-reciprocal-rational-ℕ⁺ =
+    left-inverse-law-mul-ℚ⁺ (positive-rational-ℕ⁺ n)
+
+  left-inverse-law-reciprocal-rational-ℕ⁺ :
+    mul-ℚ
+      ( reciprocal-rational-ℕ⁺ n)
+      ( rational-ℚ⁺ (positive-rational-ℕ⁺ n)) ＝
+    one-ℚ
+  left-inverse-law-reciprocal-rational-ℕ⁺ =
+    ap rational-ℚ⁺ left-inverse-law-positive-reciprocal-rational-ℕ⁺
+
+  right-inverse-law-positive-reciprocal-rational-ℕ⁺ :
+    mul-ℚ⁺
+      ( positive-rational-ℕ⁺ n)
+      ( positive-reciprocal-rational-ℕ⁺ n) ＝
+    one-ℚ⁺
+  right-inverse-law-positive-reciprocal-rational-ℕ⁺ =
+    right-inverse-law-mul-ℚ⁺ (positive-rational-ℕ⁺ n)
+
+  right-inverse-law-reciprocal-rational-ℕ⁺ :
+    mul-ℚ
+      ( rational-ℚ⁺ (positive-rational-ℕ⁺ n))
+      (reciprocal-rational-ℕ⁺ n) ＝
+    one-ℚ
+  right-inverse-law-reciprocal-rational-ℕ⁺ =
+    ap rational-ℚ⁺ right-inverse-law-positive-reciprocal-rational-ℕ⁺
 ```
--- a/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
@@ -14,8 +14,8 @@ open import elementary-number-theory.inequality-integers
 open import elementary-number-theory.inequality-rational-numbers
 open import elementary-number-theory.integer-fractions
 open import elementary-number-theory.integers
-open import elementary-number-theory.multiplication-integers
 open import elementary-number-theory.multiplication-integer-fractions
+open import elementary-number-theory.multiplication-integers
 open import elementary-number-theory.multiplication-rational-numbers
 open import elementary-number-theory.multiplicative-group-of-positive-rational-numbers
 open import elementary-number-theory.natural-numbers
--- a/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
@@ -12,12 +12,15 @@ module elementary-number-theory.unit-fractions-rational-numbers where
 open import elementary-number-theory.archimedean-property-positive-rational-numbers
 open import elementary-number-theory.inequality-integers
 open import elementary-number-theory.inequality-rational-numbers
+open import elementary-number-theory.integer-fractions
 open import elementary-number-theory.integers
 open import elementary-number-theory.multiplication-integers
+open import elementary-number-theory.multiplication-integer-fractions
 open import elementary-number-theory.multiplication-rational-numbers
 open import elementary-number-theory.multiplicative-group-of-positive-rational-numbers
 open import elementary-number-theory.natural-numbers
 open import elementary-number-theory.nonzero-natural-numbers
+open import elementary-number-theory.positive-integers
 open import elementary-number-theory.positive-rational-numbers
 open import elementary-number-theory.rational-numbers
 open import elementary-number-theory.strict-inequality-integers
@@ -61,6 +64,20 @@ reciprocal-rational-succ-ℕ n =
   reciprocal-rational-ℕ⁺ (succ-nonzero-ℕ' n)
 ```
 
+### Reciprocals of positive integers
+
+```agda
+positive-reciprocal-rational-ℤ⁺ : ℤ⁺ → ℚ⁺
+positive-reciprocal-rational-ℤ⁺ k =
+  positive-reciprocal-rational-ℕ⁺ (positive-nat-ℤ⁺ k)
+
+reciprocal-rational-ℤ⁺ : ℤ⁺ → ℚ
+reciprocal-rational-ℤ⁺ k =
+  reciprocal-rational-ℕ⁺ (positive-nat-ℤ⁺ k)
+```
+
+## Properties
+
 ### If `m ≤ n`, the reciprocal of `n` is less than or equal to the reciprocal of `n`
 
 ```agda
@@ -92,8 +109,6 @@ abstract
       ( left-unit-law-mul-ℤ (int-ℕ n))
       ( le-natural-le-ℤ m n m<n)
 ```
-
-## Properties
 
 ### For every positive rational number, there is a smaller unit fraction
 
@@ -154,7 +169,7 @@ module _
   right-inverse-law-reciprocal-rational-ℕ⁺ :
     mul-ℚ
       ( rational-ℚ⁺ (positive-rational-ℕ⁺ n))
-      (reciprocal-rational-ℕ⁺ n) ＝
+      ( reciprocal-rational-ℕ⁺ n) ＝
     one-ℚ
   right-inverse-law-reciprocal-rational-ℕ⁺ =
     ap rational-ℚ⁺ right-inverse-law-positive-reciprocal-rational-ℕ⁺
--- a/src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/additive-group-of-rational-numbers.lagda.md
@@ -11,6 +11,7 @@ module elementary-number-theory.additive-group-of-rational-numbers where
 ```agda
 open import elementary-number-theory.addition-rational-numbers
 open import elementary-number-theory.difference-rational-numbers
+open import elementary-number-theory.group-of-integers
 open import elementary-number-theory.rational-numbers
 
 open import foundation.dependent-pair-types
@@ -21,6 +22,7 @@ open import foundation.universe-levels
 open import group-theory.abelian-groups
 open import group-theory.commutative-monoids
 open import group-theory.groups
+open import group-theory.homomorphisms-abelian-groups
 open import group-theory.monoids
 open import group-theory.semigroups
 ```
@@ -98,4 +100,11 @@ abstract
 
   left-swap-add-ℚ : (p q r : ℚ) → p +ℚ (q +ℚ r) ＝ q +ℚ (p +ℚ r)
   left-swap-add-ℚ = left-swap-add-Ab abelian-group-add-ℚ
+```
+
+### The inclusion of integers in the rationals is an additive homomorphism
+
+```agda
+hom-add-rational-ℤ : hom-Ab ℤ-Ab abelian-group-add-ℚ
+hom-add-rational-ℤ = (rational-ℤ , λ {x y} → inv (add-rational-ℤ x y))
 ```
--- a/src/elementary-number-theory/positive-integers.lagda.md
+++ b/src/elementary-number-theory/positive-integers.lagda.md
@@ -98,6 +98,9 @@ module _
 ```agda
 one-positive-ℤ : positive-ℤ
 one-positive-ℤ = (one-ℤ , star)
+
+one-ℤ⁺ : ℤ⁺
+one-ℤ⁺ = one-positive-ℤ
 ```
 
 ## Properties
--- a/src/elementary-number-theory/positive-integers.lagda.md
+++ b/src/elementary-number-theory/positive-integers.lagda.md
@@ -156,6 +156,26 @@ is-positive-int-is-nonzero-ℕ (succ-ℕ x) H = star
 
 positive-int-ℕ⁺ : ℕ⁺ → positive-ℤ
 positive-int-ℕ⁺ (n , n≠0) = int-ℕ n , is-positive-int-is-nonzero-ℕ n n≠0
+
+positive-nat-ℤ⁺ : positive-ℤ → ℕ⁺
+positive-nat-ℤ⁺ (inr (inr x) , k>0) = succ-nonzero-ℕ' x
+
+is-section-positive-nat-ℤ⁺ :
+  (k : ℤ⁺) → positive-int-ℕ⁺ (positive-nat-ℤ⁺ k) ＝ k
+is-section-positive-nat-ℤ⁺ (inr (inr k) , k>0) =
+  eq-type-subtype subtype-positive-ℤ refl
+
+is-retraction-positive-nat-ℤ⁺ :
+  (n : ℕ⁺) → positive-nat-ℤ⁺ (positive-int-ℕ⁺ n) ＝ n
+is-retraction-positive-nat-ℤ⁺ (zero-ℕ , n≠0) = ex-falso (n≠0 refl)
+is-retraction-positive-nat-ℤ⁺ (succ-ℕ n , n≠0) = eq-nonzero-ℕ refl
+
+is-equiv-positive-int-ℕ⁺ : is-equiv positive-int-ℕ⁺
+is-equiv-positive-int-ℕ⁺ =
+  is-equiv-is-invertible
+    ( positive-nat-ℤ⁺)
+    ( is-section-positive-nat-ℤ⁺)
+    ( is-retraction-positive-nat-ℤ⁺)
 ```
 
 ### The canonical equivalence between natural numbers and positive integers


malarbol:
I still have a few more ideas that I think could be useful for #1451. If @djspacewhale wants to join in / take anything interesting, they are welcome.

malarbol:
This should fill most of the blanks of #1451. 
Independently, I think I introduces some interesting results (unity of ring homomorphisms `ℚ → R`, unicity of left rational module structure, etc.) which could also be useful for #1435 (rational rings are good candidates to define the exponential series, maybe even without convergence nor commutativity).

malarbol:
This overlaps a bit with #1451 so if this gets merged, @djspacewhale could be credited in this PR.
I keep getting new ideas but this is already quite big so I think I'll stop here.

src/elementary-number-theory/unit-fractions-rational-numbers.lagda.md
```diff
@@ -60,6 +66,54 @@ reciprocal-rational-succ-ℕ n =
...
+  positive-reciprocal-rational-ℕ⁺ (positive-nat-ℤ⁺ k)
+
+reciprocal-rational-ℤ⁺ : ℤ⁺ → ℚ
+reciprocal-rational-ℤ⁺ k =
+  reciprocal-rational-ℕ⁺ (positive-nat-ℤ⁺ k)
+```
+
+## Properties
+
+### The numerator of a unit fracion is one
```
fredrik-bakke:
```suggestion
### The numerator of a unit fraction is one
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+open import ring-theory.rational-rings
+open import ring-theory.rings
+```
+
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
+[Abelian group](group-theory.abelian-groups.md) whose
```
fredrik-bakke: By convention we do not capitalize "Abelian" even though it is derived from a name 
```suggestion
[abelian group](group-theory.abelian-groups.md) whose
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+[ring of endomorphisms](group-theory.endomorphism-rings-abelian-groups.md) is
+[rational](ring-theory.rational-rings.md). The type of **rational modules** is
```
fredrik-bakke: Please add an explanation for what a rational ring is so the reader does not have to go on a goose chase to get to the definition.

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+
+  rational-ring-endomorphism-Rational-Module : Rational-Ring l
+  rational-ring-endomorphism-Rational-Module =
+    ( endomorphism-ring-Ab ab-Rational-Module ,
+      is-rational-endomorphism-ring-ab-Rational-Module)
+```
+
+## Properties
+
+### An Abelian group is a rational module if and only if for any `k : ℤ⁺` the map `x ↦ k x` is an isomorphism
```
fredrik-bakke:
```suggestion
### An abelian group `M` is a rational module if and only if for any positive integer `k` the action of `k` on `M` is an automorphism
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
+[Abelian group](group-theory.abelian-groups.md) whose
+[ring of endomorphisms](group-theory.endomorphism-rings-abelian-groups.md) is
+[rational](ring-theory.rational-rings.md). The type of **rational modules** is
+equivalent to the type of [left modules](linear-algebra.left-modules-rings.md)
+and [right modules](linear-algebra.right-modules-rings.md) over the
```
fredrik-bakke:
```suggestion
equivalent to the type of [left](linear-algebra.left-modules-rings.md)
or [right modules](linear-algebra.right-modules-rings.md) over the
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+open import ring-theory.opposite-rings
+open import ring-theory.rational-rings
+open import ring-theory.rings
+```
+
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
```
fredrik-bakke, line 46: Please add a citation for the term "rational module". If this term is used on wikipedia or nlab a link to either in an "External links" section will suffice.

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,199 @@
...
+        ( λ M →
+          equiv-iff-is-prop
+            ( is-prop-has-rational-hom-Ring (op-Ring (endomorphism-ring-Ab M)))
+            ( is-prop-is-rational-module-Ab M)
+            ( is-rational-is-rational-op-Ring (endomorphism-ring-Ab M) ∘
+              is-rational-has-rational-hom-Ring
+                ( op-Ring (endomorphism-ring-Ab M)))
+            ( initial-hom-Rational-Ring ∘
+              op-Rational-Ring ∘
+              pair (endomorphism-ring-Ab M)))
```
fredrik-bakke: since you're proving two propositions are logically equivalent here as part of your proof, it is useful to factor that out into a lemma. I also believe it is better if you mark only the logical equivalence as abstract, and then unabstract this definition so it computes on the carrier sets.

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+open import ring-theory.localizations-rings
+open import ring-theory.opposite-rings
+open import ring-theory.rings
+```
+
+</details>
+
+## Idea
+
+A {{#concept "rational ring" Agda=Rational-Ring}} is a
```
fredrik-bakke: Add a citation for the term "rational ring". If the term is used on wikipedia or nlab then an external link will suffice.

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  integer-initial-hom-Rational-Ring : hom-Ring ℤ-Ring (ring-Rational-Ring R)
+  integer-initial-hom-Rational-Ring =
+    initial-hom-Ring (ring-Rational-Ring R)
+
+  map-integer-initial-hom-Rational-Ring : ℤ → type-Rational-Ring R
+  map-integer-initial-hom-Rational-Ring =
+    map-initial-hom-Ring (ring-Rational-Ring R)
+```
+
+### Fractional extension of the initial ring homomorphism `ι : ℤ → R` into a rational ring: `φ : (p/q : fraction-ℤ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)`
```
fredrik-bakke: Headers shouldn't be this long unless they have to, because they make the contents list hard to navigate, and moreover equations/code does not display well in headers.

```suggestion
### Fractional extensions of the initial ring homomorphism into a rational ring

**YOUR EXPLANATION USING  `ι : ℤ → R` AND `φ : (p/q : fraction-ℤ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)`**
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+      ( associative-mul-Ring (ring-Rational-Ring R) pr qr qr') ∙
+      ( ap
+        ( mul-Rational-Ring R pr)
+        ( right-inverse-law-positive-integer-Rational-Ring R q)) ∙
+      ( right-unit-law-mul-Ring
+        ( ring-Rational-Ring R)
+        ( pr))
+```
+
+### Rational extension of the initial ring homomorphism `ι : ℤ → R` into a rational ring: `γ : (p/q : ℚ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p) `
```
fredrik-bakke: See comment above

```suggestion
### Rational extensions of the initial ring homomorphism into a rational ring

**YOUR EXPLANATION USING  `ι : ℤ → R` AND `γ : (p/q : ℚ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)`**
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  ( inv-tr
+    ( inverts-subset-hom-Ring
+      ( ℤ-Ring)
+      ( ring-ℚ)
+      ( subtype-positive-ℤ))
+      ( eq-initial-hom-ring-rational-ℤ)
+      ( inverts-positive-integers-rational-ℤ))
+```
+
+### The inverse of `one-ℤ` in a rational ring is `one`
```
fredrik-bakke:
```suggestion
### The inverse of the image of one in a rational ring is the multiplicative unit
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+      ( inv-positive-integer-Rational-Ring R one-ℤ⁺)
+      ( ( ap
+          ( mul-Ring'
+            ( ring-Rational-Ring R)
+            ( inv-positive-integer-Rational-Ring R one-ℤ⁺))
+          ( inv (preserves-one-initial-hom-Ring (ring-Rational-Ring R)))) ∙
+        ( right-inverse-law-positive-integer-Rational-Ring R one-ℤ⁺))
+```
+
+### The inverse of the product of positive integers in a rational ring is the product of their inverses
```
fredrik-bakke:
```suggestion
### The inverse of a product of positive integers in a rational ring is the product of their inverses
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+                ( ring-Rational-Ring R)
+                ( map-integer-initial-hom-Rational-Ring R (int-positive-ℤ q)))
+              ( left-inverse-law-positive-integer-Rational-Ring R p)) ∙
+            ( left-unit-law-mul-Ring
+              ( ring-Rational-Ring R)
+              ( map-integer-initial-hom-Rational-Ring R (int-positive-ℤ q))))) ∙
+      ( right-inverse-law-positive-integer-Rational-Ring R q)
+```
+
+### The fractional initial map `φ : fraction-ℤ → R` into a rational ring extends the initial ring homomorphism `ι : ℤ → R`
```
fredrik-bakke: See previous comments on use of equations/code in headers

```suggestion
### The fractional initial map into a rational ring extends the initial ring homomorphism

**YOUR EXPLANATION USING `φ : fraction-ℤ → R` AND `ι : ℤ → R`
```

fredrik-bakke: also in this case it seems evident enough that I'd just omit any explanation
```suggestion
### The fractional initial map into a rational ring extends the initial ring homomorphism
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  htpy-map-integer-fraction-map-initial-Rational-Ring k =
+    ( ap
+      ( mul-Rational-Ring R (map-integer-initial-hom-Rational-Ring R k))
+      ( inv (eq-one-inv-positive-one-Rational-Ring R))) ∙
+    ( right-unit-law-mul-Ring
+      ( ring-Rational-Ring R)
+      ( map-integer-initial-hom-Rational-Ring R k))
+```
+
+### The fractional initial map `φ : fraction-ℤ → R` maps similar fractions to identical elements
```
fredrik-bakke:
```suggestion
### The fractional initial map maps similar fractions to identical elements
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+                ( λ z → (rny *R z) *R rdy')
+                ( right-inverse-law-positive-integer-Rational-Ring R dx)
+          ＝ map-fraction-initial-hom-Rational-Ring R y
+            by
+              ap
+                ( mul-Ring' (ring-Rational-Ring R) rdy')
+                ( right-unit-law-mul-Ring (ring-Rational-Ring R) rny)
+```
+
+### The fractional initial map `φ : fraction-ℤ → R` preserves addition
```
fredrik-bakke:
```suggestion
### The fractional initial map preserves addition
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+          ( rdx'))) ∙
+      ( ap
+        ( λ z → (rny *R z) *R rdy')
+        ( right-inverse-law-positive-integer-Rational-Ring R dx)) ∙
+      ( ap
+        ( mul-Ring' (ring-Rational-Ring R) (rdy'))
+        ( right-unit-law-mul-Ring (ring-Rational-Ring R) rny))
+```
+
+### The fractional initial map `φ : fraction-ℤ → R` preserves multiplication
```
fredrik-bakke:
```suggestion
### The fractional initial map preserves multiplication
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+    rnx rdx rdx' rny rdy rdy' : type-Rational-Ring R
+    rnx = map-integer-initial-hom-Rational-Ring R nx
+    rdx = map-integer-initial-hom-Rational-Ring R (int-positive-ℤ dx)
+    rdx' = inv-positive-integer-Rational-Ring R dx
+    rny = map-integer-initial-hom-Rational-Ring R ny
+    rdy = map-integer-initial-hom-Rational-Ring R (int-positive-ℤ dy)
+    rdy' = inv-positive-integer-Rational-Ring R dy
+```
+
+### The rational initial ring map `γ : ℚ → R` extends the initial ring homomorphism `ι : ℤ → R`
```
fredrik-bakke: Isn't "initial rational ring map" a better name for "rational initial ring map"?

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  htpy-integer-map-initial-hom-Rational-Ring k =
+    ( ap
+      ( mul-Rational-Ring R (map-integer-initial-hom-Rational-Ring R k))
+      ( inv (eq-one-inv-positive-one-Rational-Ring R))) ∙
+    ( right-unit-law-mul-Ring
+      ( ring-Rational-Ring R)
+      ( map-integer-initial-hom-Rational-Ring R k))
+```
+
+### The rational initial ring map `γ : ℚ → R` preserves one
```
fredrik-bakke:
```suggestion
### The initial rational ring map preserves the multiplicative unit
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  preserves-one-rational-initial-hom-Rational-Ring :
+    map-rational-initial-hom-Rational-Ring R one-ℚ ＝
+    one-Ring (ring-Rational-Ring R)
+  preserves-one-rational-initial-hom-Rational-Ring =
+    is-right-inverse-inv-is-invertible-element-Ring
+      ( ring-Rational-Ring R)
+      ( is-invertible-positive-integer-Rational-Ring R one-ℤ⁺)
+```
+
+### The initial ring map `ℚ → R` is a ring homomorphism
```
fredrik-bakke:
```suggestion
### The initial rational ring map is a ring homomorphism
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  rational-ring-has-rational-hom-Ring f =
+    ( R , is-rational-has-rational-hom-Ring f)
+
+  inv-positive-integer-has-rational-hom-Ring :
+    hom-Ring ring-ℚ R → (k : ℤ⁺) → type-Ring R
+  inv-positive-integer-has-rational-hom-Ring =
+    inv-positive-integer-Rational-Ring ∘ rational-ring-has-rational-hom-Ring
+```
+
+### Any ring homomorphism `f : ℚ → R` is an extension of the initial ring homomorphism `ι : ℤ → R`
```
fredrik-bakke:
```suggestion
### Any ring homomorphism `ℚ → R` is an extension of the initial ring homomorphism `ℤ → R`
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  htpy-map-reciprocal-rational-hom-Ring :
+    map-hom-Ring ring-ℚ R f ∘ reciprocal-rational-ℤ⁺ ~
+    inv-positive-integer-has-rational-hom-Ring R f
+  htpy-map-reciprocal-rational-hom-Ring =
+    htpy-map-reciprocal-hom-Rational-Ring
+      ( rational-ring-has-rational-hom-Ring R f)
+      ( f)
+```
+
+### Any homomorphism `ℚ → R` is homotopic to the rational initial ring homomorphism `γ : ℚ → R`
```
fredrik-bakke:
```suggestion
### Any homomorphism `ℚ → R` is homotopic to the initial rational ring homomorphism
```

src/ring-theory/rational-rings.lagda.md
```diff
@@ -0,0 +1,1312 @@
...
+  has-rational-hom-is-rational-Ring : is-rational-Ring R → hom-Ring ring-ℚ R
+  has-rational-hom-is-rational-Ring =
+    initial-hom-Rational-Ring ∘ pair R
+
+  iff-is-rational-has-rational-hom-Ring : hom-Ring ring-ℚ R ↔ is-rational-Ring R
+  iff-is-rational-has-rational-hom-Ring =
+    ( is-rational-has-rational-hom-Ring R , has-rational-hom-is-rational-Ring)
+```
+
+### A ring `R` is rational if and only if the type ring homomorphism `ℚ → R` is contractible
```
fredrik-bakke:
```suggestion
### A ring `R` is rational if and only if the type of ring homomorphisms `ℚ → R` is contractible
```

src/ring-theory/rational-rings.lagda.md
fredrik-bakke, line 1: A lot of this file is about initiality of the rational ring of rationals. Could this be factored into a new file `elementary-number-theory.rational-ring-of-rational-numbers`? Please also add a "see also" link from `ring-of-rational-numbers` to `rational-ring-of-rational-numbers`

fredrik-bakke:
> This overlaps a bit with #1451 so if this gets merged, @djspacewhale could be credited in this PR. I keep getting new ideas but this is already quite big so I think I'll stop here.

See here for how to add a co-author: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors
You can add the relevant line to your main pull request message.

src/ring-theory/rational-rings.lagda.md
fredrik-bakke: or
```suggestion
### The inverse of the image of one in a rational ring is one
```

src/ring-theory/rational-rings.lagda.md
fredrik-bakke: You have not explained the terminology "_fractional_ map" anywhere either.

malarbol: That's actually kind of intensional, I didn't want to talk about "fractional maps" in general (It's a bit unclear to me what the definition should be; it would probably involve the notion of "fraction rings" or something like this). And for the same reason, neither I defined "rational maps" in general.

I'm using "*the* fractional initial map" and "*the* rational initial map" as names for the map(s) `p/q ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)`.

src/ring-theory/rational-rings.lagda.md
malarbol: Is about "the order of adjectives in the English language" or about the the intrinsic meaning of the name?
I'm afraid "initial rational ring map" (and I guess I "initial fractional ring map" instead of "fractional initial ring map") could suggest there exists some concept of "rational ring map" which is not the case (yet?).

Moreover, the idea is that the important map is the "initial ring map" `p/q ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)`, and whether it's defined on `fraction-ℤ` or `ℚ` is kind of an implementation detail and the fractional version is mainly a helper to prove that the rational version _is_ a ring homorphism. Hence, we consider some kind of "initial ring map" that can either be "fractional" or "rational".

That being said, I'm not trying to argue, just to explain my thoughts. If you're 100% sure about this I'll happily change the names.

src/linear-algebra/rational-modules.lagda.md
malarbol: see https://github.com/UniMath/agda-unimath/pull/1452/commits/cfa72664ab4c8d9ccdd9e4cfe3d65076a3899202

src/linear-algebra/rational-modules.lagda.md
malarbol: The closest reference I could find is this: https://ncatlab.org/nlab/show/rational+vector+space 
Would that be ok?

fredrik-bakke: Ah, right, this is a rational vector space. Then it should be called that

src/linear-algebra/rational-modules.lagda.md
malarbol: I think it's more akin to an `eq-type-subtype` situation. I'll refactor this bits to make it more readable.

src/ring-theory/rational-rings.lagda.md
fredrik-bakke: My thinking was that "rational map" is a stepping stone towards "rational homomorphism". In general a rational homomorphism should just be a homomorphism but between rational vector spaces. I suppose the disambiguation is important here because you consider the initial homomorphism, so maybe your terminology is better

src/linear-algebra/rational-modules.lagda.md
malarbol: We don't really have a `Vector-Space` type yet. Would it be ok to add a note explaining that it's equivalent to the concept of rational vector spaces and refactor this later, when we have a better vector spaces infrastructure?

src/ring-theory/rational-rings.lagda.md
malarbol: I can't find any reference for this. Since it ends up being equivalent to rings containing `ℚ`, the name seemed sensible enough. Maybe I should prove the fact that the rational initial map is an embedding so a rational ring is a "ring extension" of `ℚ`? 
Do you prefer the term "rational extension ring" to convey the fact that these are the rings that extend the ring of rational numbers?

src/linear-algebra/rational-modules.lagda.md
malarbol: > Ah, right, this is a rational vector space. Then it should be called that

And when me have vector spaces, we might actually want a separate module `rational-vector-spaces` where the distinctions between a "rational module" and a "rational vector space" is essentially the same as the difference with `ring-ℚ` and `discrete-field-ℚ`. 

I'll try to rephrase the idea section, mentioning "rational vector spaces", but I think "rational module" is ok.

src/linear-algebra/rational-modules.lagda.md
malarbol: See https://github.com/UniMath/agda-unimath/pull/1452/commits/c756d2dcc25b84734e1455ef1a9a71204202fa51

src/linear-algebra/rational-modules.lagda.md
malarbol: See https://github.com/UniMath/agda-unimath/pull/1452/commits/c756d2dcc25b84734e1455ef1a9a71204202fa51

src/ring-theory/rational-rings.lagda.md
malarbol: see ~https://github.com/UniMath/agda-unimath/pull/1452/commits/6a4ccd0298508ab5c9e479f64bb9b15af53d4555~ current version

src/ring-theory/rational-rings.lagda.md
malarbol: I introduced the type
```agda
  hom-Rational-Ring : UU (l1 ⊔ l2)
  hom-Rational-Ring =
    hom-Ring
      ( ring-Rational-Ring A)
      ( ring-Rational-Ring B)
```
and made a few name change:
- ~`map-integer-initial-hom-Rational-Ring`~ -> `map-initial-hom-integer-Rational-Ring`:
   I don't think it should live in the `hom-Rational-Space` namespace (`ℤ` is not a rational ring);
- ~`map-fraction-initial-hom-Rational-Ring`~ -> `map-initial-hom-fraction-Rational-Ring`:
   I don't think it should live in the `hom-Rational-Space` namespace (`fraction-ℤ ` is not a ring);
- ~`map-rational-initial-hom-Rational-Ring`~ -> `map-initial-hom-Rational-Ring`:
     it **is**  equal to `map-hom-Rational-Ring (initial-hom-Rational-Ring)` (defined in the `rational-ring-of-rational-numbers` module).

Is that ok? 

> In general a rational homomorphism should just be a homomorphism but between rational vector spaces. I suppose the disambiguation is important here because you consider the initial homomorphism [...].

Yes, the focus here is more on the **initial** map and the fractional/rational flavor is more of a detail. I suggest we stay with "rational initial ring map" until we develop the `vector-spaces` module, and then maybe we'll find a better name.

malarbol:
> > This overlaps a bit with #1451 so if this gets merged, @djspacewhale could be credited in this PR. I keep getting new ideas but this is already quite big so I think I'll stop here.
> 
> See here for how to add a co-author: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors You can add the relevant line to your main pull request message.

Do you mean like I did in the main message of the PR? You'll be the one eventually merging this and I'm not sure how the commit message will be handled.

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+open import ring-theory.rings
+```
+
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
+[abelian group](group-theory.abelian-groups.md) for which any of the following
+[logically](foundation.logical-equivalences.md) equivalent
```
fredrik-bakke:
```suggestion
[logically equivalent](foundation.logical-equivalences.md)
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+[propositions](foundation.propositions.md) holds:
+
+- its [ring of endomorphisms](group-theory.endomorphism-rings-abelian-groups.md)
+  is [rational](ring-theory.rational-rings.md), i.e., the
+  [initial ring homomorphism](elementary-number-theory.ring-of-integers.md) in
+  its ring of endomorphisms [inverts](ring-theory.localizations-rings.md) the
+  [positive integers](elementary-number-theory.positive-integers.md);
```
fredrik-bakke: It doesn't "invert" them, it "makes them invertible", right?

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+    is-rational-Ring (endomorphism-ring-Ab ab-Rational-Module)
+  is-rational-endomorphism-ring-ab-Rational-Module = pr2 M
+
+  rational-ring-endomorphism-Rational-Module : Rational-Ring l
+  rational-ring-endomorphism-Rational-Module =
+    ( endomorphism-ring-Ab ab-Rational-Module ,
+      is-rational-endomorphism-ring-ab-Rational-Module)
+```
+
+### The property of being a left module on the rationals
```
fredrik-bakke:
```suggestion
### The predicate on abelian groups of being a left module on the rationals
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+  is-prop-is-rational-left-module-Ab =
+    is-prop-has-rational-hom-Ring (endomorphism-ring-Ab A)
+
+  subtype-is-rational-left-module : Prop l
+  subtype-is-rational-left-module =
+    ( is-rational-left-module-Ab ,
+      is-prop-is-rational-left-module-Ab)
+```
+
+### The property of being a right module on the rationals
```
fredrik-bakke:
```suggestion
### The predicate on abelian groups of being a right module on the rationals
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+    inv-tr
+      ( is-invertible-element-Ring (endomorphism-ring-Ab M))
+      ( htpy-initial-hom-integer-multiple-endomorphism-ring-Ab M k)
+      ( ev-pair H k k>0)
+```
+
+## External links
+
+- [rational vector space](https://ncatlab.org/nlab/show/rational+vector+space)
+  at $n$lab
```
fredrik-bakke:
```suggestion
  at $n$Lab
```

src/linear-algebra/rational-modules.lagda.md
fredrik-bakke: hmm, I'm a bit unsure about wheter this disambiguation is worthwhile, but if you prefer it then okay

src/elementary-number-theory/rational-ring-of-rational-numbers.lagda.md
```diff
@@ -0,0 +1,724 @@
...
+</details>
+
+## Idea
+
+The [reciprocal](elementary-number-theory.unit-fractions-rational-numbers.md) of
+a [positive integer](elementary-number-theory.positive-integers.md) is its
+[inverse](ring-theory.invertible-elements-rings.md) in the
+[ring of rational numbers](elementary-number-theory.ring-of-rational-numbers.md)
+so `ℚ` is a [rational ring](ring-theory.rational-rings.md); this is the
+{{#concept "rational ring of rationa numbers" Agda=rational-ring-ℚ}}.
```
fredrik-bakke:
```suggestion
{{#concept "rational ring of rational numbers" Agda=rational-ring-ℚ}}.
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,285 @@
...
+  [ring of rational numbers](elementary-number-theory.ring-of-rational-numbers.md).
+
+**Note:** Because `ℚ` is a
+[discrete field](commutative-algebra.discrete-fields.md), rational modules are
+the vector spaces on the
+[field of rational numbers](elementary-number-theory.field-of-rational-numbers.md).
+
+## Definitions
+
+### The property of being a rational module
```
fredrik-bakke:
```suggestion
### The predicate on abelian groups of being a rational module
```

src/ring-theory/rational-rings.lagda.md
fredrik-bakke: Fair enough. Could you add a note on terminology that this is not standard terminology, and that more standard terminology would be a ring extension of the rationals?

fredrik-bakke: > I can't find any reference for this. Since it ends up being equivalent to rings containing `ℚ`, the name seemed sensible enough. Maybe I should prove the fact that the rational initial map is an embedding so a rational ring is a "ring extension" of `ℚ`? Do you prefer the term "rational extension ring" to convey the fact that these are the rings that extend the ring of rational numbers?

"Rational ring extensions", or "ring extensions of ℚ" would probably be my preferred terminology, from the limited googling I just did on the topic

fredrik-bakke: "extension ring of ℚ" is also good

fredrik-bakke: although I'm not a number theorist, given the fact that the term "rational ring" is not used online I believe we should avoid it and instead use the more verbose, but also actually recognizable name of "ring extension/extension ring (of the rationals)". This name also contextualizes the structure better.

src/linear-algebra/rational-modules.lagda.md
fredrik-bakke: Ah, I see, nice

src/linear-algebra/rational-modules.lagda.md
malarbol: That's the terminology used in the [localizations-rings](https://unimath.github.io/agda-unimath/ring-theory.localizations-rings.html) module but I can use "make them invertible" if you prefer.

src/ring-theory/rational-rings.lagda.md
malarbol: ok. I'll rephrase / rename this module to focus more on "ring extension of the rationals"

malarbol: See https://github.com/UniMath/agda-unimath/pull/1452/commits/dbb9d4a857fdd142dc7a5cac699c8028814c3686

I had to re-include the results of `rational-ring-of-rational-number` to construct the ring homomorphism in the same module.

malarbol: Note that, if I'm not mistaken, these are also **rational unital algebras** (unital algebras over the ring of rational numbers) so `rational-unital-algebras` could also be a good candidate for the name of this module. However, the type of [unital algebras](https://unimath.github.io/agda-unimath/ring-theory.algebras-rings.html) is not defined yet so maybe it's precipitated.

src/linear-algebra/rational-modules.lagda.md
fredrik-bakke: Ah, I see, then it's probably fine.

src/ring-theory/rational-rings.lagda.md
fredrik-bakke: > See [dbb9d4a](https://github.com/UniMath/agda-unimath/commit/dbb9d4a857fdd142dc7a5cac699c8028814c3686)
> 
> I had to re-include the results of `rational-ring-of-rational-number` to construct the ring homomorphism in the same module.

Wait, why can't you keep a module on the initial rational extension ring?

malarbol: > Wait, why can't you keep a module on the initial rational extension ring?

In the old settings, I proved that "ring homorphisms from ℚ to a rational ring are unique" in the `rational-rings` module and actually construct the ring homomorphism in `rational-ring-of-rational-number` (proving that ℚ is the initial rational ring).

With the new setting, we can't really name it "ring extension of ℚ" without defining actual ring extension, `ℚ -> R` so it needs to be in the same module. Maybe there's a better way to phrase/refactor this. We can think about it.

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
```diff
@@ -64,3 +72,61 @@ commutative-ring-ℚ : Commutative-Ring lzero
 pr1 commutative-ring-ℚ = ring-ℚ
 pr2 commutative-ring-ℚ = commutative-mul-ℚ
 ```
+
+### The inclusion of integers in the rationals is the initial ring homomorphism
+
+```agda
+hom-ring-rational-ℤ : hom-Ring ℤ-Ring ring-ℚ
```
fredrik-bakke, line 79: The discrepancy between `ℤ-Ring` and `ring-ℚ` should be fixed...

fredrik-bakke: It seems like it is more conventional to write things like `XY-Set` as opposed to `set-XY`, although I wouldn't be opposed to the latter. Either way, I'm declaring this out of scope for this PR.

src/ring-theory/rational-extensions-rings.lagda.md
fredrik-bakke, line 1: this file name reads like "rational extensions of rings", but it should read like "ring extensions of rationals", so I suggest the file name `ring-extensions-rational-numbers`

fredrik-bakke, line 1: Where did the file about the rational numbers as a ring extension of the rational numbers go?

fredrik-bakke: I still think such a file is very worthwhile. A good name for it might be `ring-extension-rational-numbers-of-rational-numbers`

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
malarbol: IIRC we talked about this when defining `ring-ℚ` and there was some name scheme transition where `ℤ-Ring` was the _old_ scheme while `ring-ℚ`fitted the new scheme better.

src/ring-theory/rational-extensions-rings.lagda.md
malarbol: See https://github.com/UniMath/agda-unimath/pull/1452/commits/e7178658f3f893af9bd300b8ad92660f5decd6c0

(I also fixed a few headers where I had left a few "rational rings" instead of "ring extension of `ℚ`").

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1451 @@
...
+    hom-Rational-Extension-Ring →
+    type-Rational-Extension-Ring A →
+    type-Rational-Extension-Ring B
+  map-hom-Rational-Extension-Ring =
+    map-hom-Ring
+      ( ring-Rational-Extension-Ring A)
+      ( ring-Rational-Extension-Ring B)
+```
+
+### The initial ring homomorphism into a ring extension of `ℚ`
```
fredrik-bakke, line 188: I thought the stuff about initial ring (extension) homomorphisms would fit best in the file about the initial ring extension: `ring-extension-rational-numbers-of-rational-numbers`. Is there a constraint that makes this impossible?

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1451 @@
...
+{{#concept "ring extension of ℚ" Agda=Rational-Extension-Ring}}.
+
+The unique ring homomorphism from `ℚ` to a ring extension of `ℚ` is given by
+`(p/q : ℚ) ↦ (ι p)(ι q)⁻¹` where `ι : ℤ → R` is the
+[initial ring homomorphism](elementary-number-theory.ring-of-integers.md) in
+`R`.
+
+## Definitions
+
+### The property of being a rational ring
```
fredrik-bakke:
```suggestion
### The predicate on a ring of being a ring extension of the rational numbers
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
fredrik-bakke, line 1: This file could still use with a bit of partitioning. Some files you should consider factoring out are
- products of ring extensions of rational numbers
- morphisms of ring extensions of rational numbers

src/ring-theory/ring-extensions-rational-numbers.lagda.md
fredrik-bakke: That would mean most of the remainder of this file should go there

malarbol: I don't know. It feels weird to call them "ring extensions of ℚ" here, but _actually_ proving that they are ring extensions of ℚ (i.e., constructing the unique ring homomorphism `ℚ → R`) in another module.

Or, put in other words, the main idea of this module is 
```text
For a [ring](ring-theory.rings.md) `R`, the following conditions are equivalent
propositions:
- the images of
  [positive integers](elementary-number-theory.positive-integers.md) by the
  [initial ring homomorphism](elementary-number-theory.ring-of-integers.md)) are
  [invertible elements](ring-theory.invertible-elements-rings.md), i.e., the
  initial ring homomorphism [inverts](ring-theory.localizations-rings.md) the
  subset of positive integers;
- there exist a [ring homomorphism](ring-theory.homomorphisms-rings.md) `ℚ → R`
  from the
  [ring of rational numbers](elementary-number-theory.ring-of-rational-numbers.md)
  to `R`;
- the type of ring homomorphisms `ℚ → R` is
  [contractible](foundation-core.contractible-types.md).
```

If I move the construction of the initial rational ring homomorphism in `ring-extension-rational-numbers-of-rational-numbers` we'd have one implication _here_, and the other implication _there_ and I think it's better to have everything together in one module.

malarbol: I guess we could split the modules and add some links / `See also` to help the reader find useful results.
I could factor out some sub-modules like 
- `fractional-extension-initial-hom-ring-extensions-rational-numbers`;
- `rational-extension-initial-hom-ring-extensions-rational-numbers`;

I understand this module is quite big, but the main idea _"the following conditions are equivalent
propositions: [...]"_ actually carries a lot (each of these are **propositions** _and_ they are **logically equivalent**)
so maybe you are right and it's better to split each result in their own modules.

malarbol: I factored out 
- `fractional-extension-initial-hom-ring-extensions-rational-numbers`;
- `rational-extension-initial-hom-ring-extensions-rational-numbers`;
- `products of ring extensions of rational numbers`;
- `morphisms of ring extensions of rational numbers`;
- `opposite ring extensions of rational numbers`;
so the results are now in smaller independent modules. Is it better?

src/ring-theory/ring-extensions-rational-numbers.lagda.md
malarbol: see https://github.com/UniMath/agda-unimath/pull/1452/files#r2265366788

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,363 @@
...
+For a [ring](ring-theory.rings.md) `R`, the following conditions are equivalent
+propositions:
+
+- the images of
+  [positive integers](elementary-number-theory.positive-integers.md) by the
+  [initial ring homomorphism](elementary-number-theory.ring-of-integers.md)) are
+  [invertible elements](ring-theory.invertible-elements-rings.md), i.e., the
+  initial ring homomorphism [inverts](ring-theory.localizations-rings.md) the
+  subset of positive integers;
+- there exist a [ring homomorphism](ring-theory.homomorphisms-rings.md) `ℚ → R`
```
fredrik-bakke:
```suggestion
- there exists a [ring homomorphism](ring-theory.homomorphisms-rings.md) `ℚ → R`
```

src/ring-theory/homomorphisms-ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,52 @@
...
+
+open import ring-theory.homomorphisms-rings
+open import ring-theory.ring-extensions-rational-numbers
+```
+
+</details>
+
+## Idea
+
+{{#concept "Homomorphisms" Disambiguation="of rational extensions of ℚ" Agda=hom-Rational-Extension-Ring}}
```
fredrik-bakke:
```suggestion
{{#concept "Homomorphisms" Disambiguation="of ring extensions of ℚ" Agda=hom-Rational-Extension-Ring}}
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,363 @@
...
+  htpy-map-reciprocal-rational-hom-Ring =
+    htpy-map-reciprocal-rational-hom-Rational-Extension-Ring
+      ( rational-ring-has-rational-hom-Ring R f)
+      ( f)
+```
+
+## See also
+
+- [`rational-extension-initial-hom-ring-extensions-rational-numbers`](ring-theory.rational-extension-initial-hom-ring-extensions-rational-numbers.md):
+  the initial ring homomorphism from `ℚ` into a rational extension of `ℚ`.
```
fredrik-bakke:
```suggestion
  the initial ring homomorphism from `ℚ` into a ring extension of `ℚ`.
```

src/ring-theory/rational-extension-initial-hom-ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,369 @@
...
+
+</details>
+
+## Idea
+
+Given a [ring extension of `ℚ`](ring-theory.ring-extensions-rational-numbers.md)
+`R`, the
+{{#concept "initial rational ring homomorphism" Agda=initial-hom-Rational-Extension-Ring}}
+into `R`, is the unique [ring homomorphism](ring-theory.homomorphisms-rings.md)
+from [`ℚ`](elementary-number-theory.ring-of-rational-numbers.md) to `R` ; it is
```
fredrik-bakke:
```suggestion
from [`ℚ`](elementary-number-theory.ring-of-rational-numbers.md) to `R`; it is
```

src/ring-theory/rational-extension-initial-hom-ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,369 @@
...
+given by:
+
+```text
+(p/q : ℚ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)
+```
+
+where `ι : ℤ → R` is the
+[initial ring homomorphism](elementary-number-theory.ring-of-integers.md) in
```
fredrik-bakke:
```suggestion
[initial ring homomorphism](elementary-number-theory.ring-of-integers.md) into
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
fredrik-bakke: > I don't know. It feels weird to call them "ring extensions of ℚ" here, but _actually_ proving that they are ring extensions of ℚ (i.e., constructing the unique ring homomorphism `ℚ → R`) in another module.

Fair enough, lets keep all of the stuff about equivalent characterizations of ring extensions of ℚ in the same module for now. Can dependent products and opposites still be their own files?

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,363 @@
...
+  subset of positive integers;
+- there exists a [ring homomorphism](ring-theory.homomorphisms-rings.md) `ℚ → R`
+  from the
+  [ring of rational numbers](elementary-number-theory.ring-of-rational-numbers.md)
+  to `R`;
+- the type of ring homomorphisms `ℚ → R` is
+  [contractible](foundation-core.contractible-types.md).
+
+In that case, `R` is called a a
+{{#concept "ring extension of ℚ" Agda=Rational-Extension-Ring}}.
```
fredrik-bakke: Please be explicit about what you take as the default definition

fredrik-bakke (approved):

src/ring-theory/ring-extensions-rational-numbers.lagda.md
malarbol: Maybe I went too fast but I already factored out 
- fractional-extension-initial-hom-ring-extensions-rational-numbers;
- rational-extension-initial-hom-ring-extensions-rational-numbers;

Is this ok?

fredrik-bakke: My issue with those is that they're not well-defined concepts, if I'm not mistaken. The current organisation also makes the story a tad fragmented, as you noted

malarbol: ok. no problem.I'll put these back and keep dependent products. opposites. and homomorphisms of ring extensions of Q in separate modules.

malarbol: What about now?

src/ring-theory/ring-extensions-rational-numbers.lagda.md
malarbol: Is this better now?

src/elementary-number-theory/ring-of-rational-numbers.lagda.md
```diff
@@ -64,3 +72,69 @@ commutative-ring-ℚ : Commutative-Ring lzero
...
+      preserves-invertible-elements-hom-Ring
+        ( ring-ℚ)
+        ( R)
+        ( f)
+        ( inverts-positive-integers-rational-ℤ k k>0)
+```
+
+## See also
+
+- [ring-extension-rational-numbers-of-rational-numbers](elementary-number-theory.ring-extension-rational-numbers-of-rational-numbers.md):
```
fredrik-bakke:
```suggestion
- [`ring-extension-rational-numbers-of-rational-numbers`](elementary-number-theory.ring-extension-rational-numbers-of-rational-numbers.md):
```

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,291 @@
...
+```
+
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
+[abelian group](group-theory.abelian-groups.md) for which any of the following
+[logically equivalent](foundation.logical-equivalences.md)
+[propositions](foundation.propositions.md) holds:
```
fredrik-bakke:
```suggestion
[propositions](foundation.propositions.md) hold:
```

fredrik-bakke: please note that you take the first condition to be the default one

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1319 @@
...
+        ( right-inverse-law-positive-integer-Rational-Extension-Ring R q)) ∙
+      ( right-unit-law-mul-Ring
+        ( ring-Rational-Extension-Ring R)
+        ( pr))
+```
+
+### Rational extension of the initial ring homomorphism into a ring extension of `ℚ`
+
+It is defined as `γ : (p/q : ℚ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)` where `ι : ℤ → R`
+is the initial ring homomorphism.
```
fredrik-bakke:
```suggestion
### Rational extension of the initial ring homomorphism into a ring extension of `ℚ`

The _rational extension of the initial ring homomorphism_ into a ring extension
of `ℚ` is defined as `γ : (p/q : ℚ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)` where
`ι : ℤ → R` is the initial ring homomorphism.
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1319 @@
...
+  map-initial-hom-integer-Rational-Extension-Ring :
+    ℤ → type-Rational-Extension-Ring R
+  map-initial-hom-integer-Rational-Extension-Ring =
+    map-initial-hom-Ring (ring-Rational-Extension-Ring R)
+```
+
+### Fractional extension of the initial ring homomorphism into a ring extension of `ℚ`
+
+It is defined as `φ : (p/q : fraction-ℤ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)` where
+`ι : ℤ → R` is the initial ring homomorphism.
```
fredrik-bakke:
```suggestion
### Fractional extension of the initial ring homomorphism into a ring extension of `ℚ`

The _fractional extension of the initial ring homomorphism_ into a ring
extension of `ℚ` is defined as
`φ : (p/q : fraction-ℤ) ↦ (ι p)(ι q)⁻¹ = (ι q)⁻¹(ι p)` where `ι : ℤ → R` is the
initial ring homomorphism.
```

fredrik-bakke:
Would a better title for this pull request be ``Ring extensions of `ℚ` and rational modules``?

src/ring-theory/ring-extensions-rational-numbers.lagda.md
fredrik-bakke: Yep, thanks!

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,291 @@
...
+
+### The predicate on abelian groups of being a rational module
+
+```agda
+module _
+  {l : Level} (A : Ab l)
+  where
+
+  is-rational-extension-module-prop-Ab : Prop l
+  is-rational-extension-module-prop-Ab =
```
malarbol: Should this just be `is-rational-module-prop-Ab`? (I think I messed-up some renaming at some point).

fredrik-bakke: yes

src/linear-algebra/rational-modules.lagda.md
```diff
@@ -0,0 +1,292 @@
...
+
+</details>
+
+## Idea
+
+A {{#concept "rational module" Agda=Rational-Module}} is an
+[abelian group](group-theory.abelian-groups.md) whose
+[ring of endomorphisms](group-theory.endomorphism-rings-abelian-groups.md) is a
+[ring extension of `ℚ`](ring-theory.ring-extensions-rational-numbers.md), i.e.,
+the [initial ring homomorphism](elementary-number-theory.ring-of-integers.md) in
```
fredrik-bakke:
```suggestion
[ring extension of `ℚ`](ring-theory.ring-extensions-rational-numbers.md). I.e.,
the [initial ring homomorphism](elementary-number-theory.ring-of-integers.md) in
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1322 @@
...
+
+## Idea
+
+A {{#concept "ring extension of ℚ" Agda=Rational-Extension-Ring}} is a
+[ring](ring-theory.rings.md) where the images of
+[positive integers](elementary-number-theory.positive-integers.md) by the
+[initial ring homomorphism](elementary-number-theory.ring-of-integers.md)) are
+[invertible elements](ring-theory.invertible-elements-rings.md), i.e., where the
+initial ring homomorphism [inverts](ring-theory.localizations-rings.md) the
+subset of positive integers;
```
fredrik-bakke:
```suggestion
subset of positive integers.
```

src/ring-theory/ring-extensions-rational-numbers.lagda.md
```diff
@@ -0,0 +1,1322 @@
...
+
+This is equivalent to the following conditions:
```
fredrik-bakke:
```suggestion
The following conditions are equivalent to being a ring extension of the
rational numbers:
```

PR #1457, opened 2025-07-24
Make rational arithmetic opaque

lowasser:
This PR makes most rational arithmetic operations `opaque`, which improves compilation performance (avoiding expansion) and improves goal readability (since expanded rational operations tend to be extremely convoluted), so most goals will be phrased e.g. in terms of `add-Q` etc.

(This is not least based on struggles I've had with basing real arithmetic on rational numbers.)


fredrik-bakke (approved):
Thank you for implementing this! I'm excited to see the performance benefits of this change.

fredrik-bakke:
I'll give @VojtechStep some time to look this over in case he has any input. Otherwise, I'm happy to merge this in its current state.

VojtechStep:
Looks good to me, thanks for going through the definitions!

PR #1459, opened 2025-07-27
Free magmas

djspacewhale:
Introducing the free magmas on one generator and on a set, using the infrastructure of full binary trees. To prove their universal property, this also characterizes the identity types of, say, `hom-Magma M N`.


djspacewhale:
Currently the universal properties for freeness of `full-binary-tree-Magma, labeled-full-binary-tree-Magma X` rely on all target magmas `M` being sets. I suspect this is due to multiplication-preservation of a magma map `f` being structure otherwise, but don't have a concrete counterexample of a higher magma where the universal property fails.

src/trees/combinator-full-binary-trees.lagda.md
```diff
@@ -0,0 +1,59 @@
+# The combinator of full binary trees
```
VojtechStep, line 1: Is the name "combinator" for this operation standard? I don't think I've ever seen it used this way

src/trees/combinator-full-binary-trees.lagda.md
```diff
@@ -0,0 +1,59 @@
...
+open import trees.full-binary-trees
+open import trees.labeled-full-binary-trees
+```
+
+</details>
+
+## Idea
+
+Two [full binary trees](trees.full-binary-trees.md) `L, R` may be combined into
+a new full binary tree in the natural way. By abstract nonsense, this extends to
```
VojtechStep: What do you mean by "abstract nonsense" here?

src/trees/labeled-full-binary-trees.lagda.md
```diff
@@ -0,0 +1,39 @@
...
+open import foundation.universe-levels
+
+open import trees.full-binary-trees
+```
+
+</details>
+
+## Idea
+
+For a type `X`, an {{concept "`X`-labeling" Agda=labeling-full-binary-tree}} of
```
VojtechStep:
```suggestion
For a type `X`, an {{#concept "`X`-labeling" Disambiguation="of full binary trees" Agda=labeling-full-binary-tree}} of
```

src/trees/labeled-full-binary-trees.lagda.md
```diff
@@ -0,0 +1,39 @@
...
+a [full binary tree](trees.full-binary-trees.md) `T` is a map from the nodes of
+`T` to `X`. A
+{{concept "labeled full binary tree" Agda=labeled-full-binary-tree}} is a pair
```
VojtechStep:
```suggestion
{{#concept "labeled full binary tree" Agda=labeled-full-binary-tree}} is a pair
```

VojtechStep (commented):
I noticed you changed the PR to WIP, so I'll not finish the review.

The way free structures are usually formalized in the library is by having an evaluation function `ev_x : hom X Y -> Y` for an arbitrary element `x : X`. Then `(X, x)` is a free structure on a generator `x` if `ev_x` is an equivalence. See e.g. `is-free-ring-with-one-generator`, `is-identity-system` or `is-identity-system-descent-data-pushout` (which are dependent version of this principle). I tried to write down an intuition for this on pages 3 and 4 of [my lecture notes](https://vojtechstep.eu/notes/2024/05-seminar-pushouts.pdf).

You have all the parts here, so I wrote this mostly to suggest naming changes:
- `image-of-leaf` -> `ev-leaf-full-binary-tree-hom-Magma` (or maybe `ev-leaf-hom-Magma`, if nobody's looking)
- `extension-of-point-hom-full-binary-tree-Magma` -> `map-inv-ev-leaf-full-binary-tree-hom-Magma`
- `is-set-is-equiv-image-of-leaf-full-binary-tree-Magma` -> `is-free-magma-on-one-generator-full-binary-tree-Magma` whose type is `is-free-magma-on-one-generator full-binary-tree-Magma leaf-full-binary-tree`, to be defined
- ...

And I'd also like to recall that the naming convention is `result-assumptions`

src/trees/combinator-full-binary-trees.lagda.md
djspacewhale: I don't know if it's standard but the library uses `combinator`s for other tree operations, cf. `trees.combinator-directed-trees`.

src/trees/combinator-full-binary-trees.lagda.md
djspacewhale: I mean to handwave details in a cute way, but can remove the quip.

djspacewhale:
> * `is-set-is-equiv-image-of-leaf-full-binary-tree-Magma` -> `is-free-magma-on-one-generator-full-binary-tree-Magma` whose type is `is-free-magma-on-one-generator full-binary-tree-Magma leaf-full-binary-tree`, to be defined

One problem that will come up, as things are now, is that `full-binary-tree-Magma` does *not* satisfy the universal property for free magmas on one generator - at least, I only have an equivalence constructed when all target magmas are sets. 

I noticed this is a similar issue Egbert and Fredrik ran into in `lists.universal-property-lists-wild-monoids`. I suspect there may be counterexamples with higher magmas/wild monoids, but admittedly I haven't thought as much about the identity types of the relevant spaces.

src/trees/combinator-full-binary-trees.lagda.md
VojtechStep: Oh, fair enough, resolving

VojtechStep:
> One problem that will come up, as things are now, is that full-binary-tree-Magma does not satisfy the universal property for free magmas on one generator

That's true, though then I'd consider just formalizing something like `is-free-magma-on-one-generator-Set`, and only look at set-level magmas. The implementation can still be defined without the `is-set M` assumption as you have it now

djspacewhale:
I can do that but first want to defend my current factorization. It's not obvious that `is-set M` is necessary for the proof, and in fact it's not necessary to prove `ev-leaf-hom-Magma` has a section nor to construct a homotopy between the relevant magma maps. The assumption only appears in proving that the homotopies witnessing multiplication-preservation coincide. Thus, I hoped to make where the set assumption appears explicit, hoping further that someone may be able to build the dependent identification without the set assumption down the line.

VojtechStep:
I got intrigued and look into it; yes, you can get the dependent identification without the `is-set` assumption: https://gist.github.com/VojtechStep/2d3cf415b06aa47a2b3dfa2ddf275d8e

djspacewhale:
Oh amazing! I really appreciate you taking a swing at it. I'm on my phone currently but will sit down and digest your write-up later today - mind if I parrot your construction for the wild monoid case while we're at it?

VojtechStep:
> mind if I parrot your construction for the wild monoid case while we're at it?

Sure, go for it! Though note that my edits aren't merge-ready, hence the `find-better-locations` module name 😅. I also think my `ap-binary-htpy` may be decomposed further, but I'm not going to put more effort into it right now

djspacewhale:
I imagine this will get abstracted into the appropriate foundation modules, I'll take care of that (and note that you contributed that material? I'm not sure how the collaborator end looks merging PRs and attributing credit therein, not sure what I need to do for you to get that...)

djspacewhale:
Here's the current state of things. I've created two new modules in `foundation` without any description, can add that in but currently just using your lemmas for the proofs. The free magma on one generator is indeed complete!! The free magma on a type is not quite complete - it would be if the labeled combinator judgementally commuted with the join, but right now there's an annoying `ap _` obstructing the last path. I'm going to use your construction for the free wild monoid as well, though there are quite a few more `eq-pair-Σ` layers so it will take time.

VojtechStep:
> I'm not sure how the collaborator end looks merging PRs and attributing credit therein, not sure what I need to do for you to get that

You can credit me as a coauthor of a commit by adding
```
Co-authored-by: Vojtěch Štěpančík <vojtechstepancik@outlook.com>
```
on a line by itself at the end of the commit message or the PR description. Because we squash PRs though, I would get co-author credit for all the changes in the PR, which you reasonably might not want. In that case I'm okay with being uncredited in this case

fredrik-bakke:
> Introducing free magmas, using the infrastructure of full binary trees. Assuming all magmas in sight are sets, we may also prove their universal property.

This is only mildly relevant, but I introduced the terminology [_set-magmoid_](https://unimath.github.io/agda-unimath/category-theory.set-magmoids.html) for a magmoid whose homs are sets. In case you were looking for such a concept.

djspacewhale:
As there's that annoying identification that wasn't judgementally refl, I'm characterizing identifications of labeled full binary trees to give a path doing the trick. Unlabeled full binary trees are characterized and labeled ones are mostly so; a few goals to complete the structure identity principle application remain.

This PR likely won't end up including the free wild monoid characterization as promised; this feels a large enough scope as is.

Issue #1458, opened 2025-07-25
Define the completion of an arbitrary metric space


lowasser:
It is not actually obvious to me right this minute that this is _possible_, and I don't think equivalence classes of Cauchy approximations will cut it.

When defining the limit of a Cauchy approximation of equivalence classes Cauchy approximations -- that is, showing that the metric space of equivalence classes of Cauchy approximations is complete -- essentially what you want is to construct a new Cauchy approximation, mapping each epsilon to the value of an inhabitant of the equivalence class at that epsilon.  It requires countable choice to pick out the inhabitant (because the domain Q+ is countable).

HoTT, now that I actually look at it for how _it_ defines Cauchy reals, uses a higher inductive-inductive type family to simultaneously construct the reals and the neighborhood relation between them.  Every place I've seen this sort of higher inductive-inductive construction done before uses Cubical Agda or some other system that does fancy shenanigans with defining equality in terms of paths.

I haven't yet figured out if you can do it with just truncation the way agda-unimath seems to.

VojtechStep:
> I haven't yet figured out if you can do it with just truncation the way agda-unimath seems to.

I'm not an expert, but while I'd expect just truncations to not be enough, I believe you should be able to construct the type with just point constructors, then define the relation on that, then take the coequaliser to add the paths, and then set truncate. It sounds like a real PITA to work with, though related formalization will probably just use the universal property and not the concrete construction

lowasser:
Some things worth noting from my research:

[nLab](https://ncatlab.org/nlab/show/Cauchy+real+number): 
> On the other hand, even the (modulated) Cauchy real numbers are not necessarily Cauchy complete, i.e. a Cauchy sequence (even a modulated one) of Cauchy real numbers need not converge to another Cauchy real number (though it always does converge to a Dedekind real number, since the Dedekind real numbers are always Cauchy complete). The problem is that without enough countable choice, we cannot lift a (modulated) Cauchy sequence of (modulated Cauchy) real numbers to a Cauchy sequence of Cauchy sequences in order to “diagonalize” it; a countermodel is constructed by [Lubarsky](https://ncatlab.org/nlab/show/Cauchy+real+number#Lubarsky).

(This uses a direct definition of the Cauchy reals as equivalence classes of Cauchy approximations, not the higher inductive-inductive definition used in HoTT definition 13.3.2.)

[...]

PR #1462, opened 2025-07-30
The axiom of countable choice

lowasser:
Decidable equality is necessary to apply the well-ordering principle, to pick a specific natural number for each element of the countable set.

This comes up in #1458, at least as far as using the axiom of countable choice (as defined in unimath) to construct the completion of a metric space in the classical way.

(Choice-free alternatives are exhibited in HoTT, but depend on higher inductive-inductive types that I'm still struggling to implement in unimath or even Agda as a whole.  Even if we do figure that out, it'd be nice to show the classical construction.)


lowasser:
Whoops, I hadn't realized I hadn't even merged countable choice in the first place.

lowasser:
Closing to fold into a bigger change.

PR #1466, opened 2025-08-04
Necessary and sufficient conditions for lower and upper reals to correspond to reals

lowasser:
This addresses #1464 .

PR #1450, opened 2025-07-09
Refactor metric spaces

malarbol:
Refactoring the metric space module (#1421)

- remove the notion of `Premetric` / `Premetric-Space`:
  
  as we discussed, the term "premetric" has too many different meanings to be reliably used in this library. What was called a `Premetric` is now a `Rational-Neighborhood-Relation` and we only focus on (pseudo)metric spaces.

- all (pseudo)metric spaces are saturated:

  including saturation in the definition of pseudometric space is coherent with the definition of [Richman premetric spaces](https://ncatlab.org/nlab/show/Richman+premetric+space) and makes a lot of things _easier_ (e.g.: the limit map is short, the metric space of Cauchy approximations in a complete metric space is complete, etc.).

- simplify hierarchy in concepts (define things for for metric spaces first, instead of the splitting premetric>pseudometric>metric):

  although a lot of concepts only depend on the underlying neighborhood relation, they are principally useful for (pseudo)metric spaces. The main concepts (isometries, Cauchy approximations and their limits) are defined for pseudometric and metric spaces; other concepts (short maps, continuity) are only defined for metric spaces but we can easily lower them to pseudometric spaces if we ever need them.

PR #1467, opened 2025-08-05
Located metric spaces

lowasser:
Combines #1450 and #1466 to define located metric spaces, a sufficient (and necessary) condition to assign a real number to the distance between two points in a metric space.

(We don't yet prove that this is necessary, but perhaps we should.)


lowasser:
Folded into #1474 .

Issue #1436, opened 2025-06-03
Consider moving sequences into `lists` package


lowasser:
Since `finite-sequences` are there now, it might make sense, though I suppose `lists` currently only contains finite structures.

fredrik-bakke:
This seems like a sensible suggestion to me. Certainly, it would make sense to put streams (#1368) in `lists`.

PR #1476, opened 2025-08-10
Move sequences to lists package

lowasser:
Addresses #1436.


fredrik-bakke (approved):

fredrik-bakke:
I also moved `partial-sequences` to lists. It seems to fit better there as well, at least on the surface-level

fredrik-bakke:
Checks pass locally 🧮✅

PR #1477, opened 2025-08-11
Bump Agda to 2.8.0

fredrik-bakke:
Bumps Agda to version 2.8.0.

### Status
- [x] ~Github Action setup-agda~
- [x] NixOS https://search.nixos.org/packages?channel=unstable&show=agda

### DIFF
--- a/src/univalent-combinatorics/complements-decidable-subtypes.lagda.md
+++ b/src/univalent-combinatorics/complements-decidable-subtypes.lagda.md
@@ -22,7 +22,7 @@ open import univalent-combinatorics.finite-types
 ## Idea
 
 The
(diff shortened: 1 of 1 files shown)


.github/actions/setup-agda/action.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Setup Agda'
+description: 'Setup a specific version of Agda'
+inputs:
+  agda-version:
+    description:
+      'Agda version. Must be one with compiled bineries provided by upstream.'
```
fredrik-bakke:
```suggestion
      'Agda version. Must be a version that is provided as a standalone binary by the Agda project.'
```

VojtechStep: I replaced your "standalone" with "prebuilt", since we don't care if it's standalone

Issue #1332, opened 2025-02-15
Syntax modules


fredrik-bakke:
I'm curious about opinions and input on this one @EgbertRijke @VojtechStep @lowasser. This is the kind of thing I had in mind when I mentioned syntax modules for wild categories the other day.

              If you add the below module, then you can "`open`" the syntax elsewhere using `open operator-syntax-Quasigroup Q`, to access these operators for your chosen quasigroup `Q`, instead of using private definitions as you do currently.

````suggestion
```

### Quasigroup operator syntax

```agda
module
  operator-syntax-Quasigroup
  {l : Level} (Q : Quasigroup l)
  where

  _*_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _*_ = mul-Quasigroup Q

  _l/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _l/_ = left-div-Quasigroup Q

  _r/_ : type-Quasigroup Q → type-Quasigroup Q → type-Quasigroup Q
  _r/_ = right-div-Quasigroup Q
```

````

_Originally posted by @fredrik-bakke in https://github.com/UniMath/agda-unimath/pull/1330#discussion_r1957111694_

VojtechStep:
Related to #768. I remember discussing this with Egbert in September '23, where we identified two issues:

[...]

PR #1481, opened 2025-08-13
Prototype: shorthand for algebraic structures

lowasser:
It has been very painful to do lots of algebra with algebraic structures due to the long function names that must be retyped (and formatted, and indented increasingly deeply) with every operation.

This is a prototype example of a possible solution that attempts to _standardize_ a shorthand.  The idea is to write something like

```
let open shorthand-Monoid Some-Monoid in
...(x *Mon y)...
```

...sticking to what seem to be the "standard" mathematical abbreviations for the category of the described algebraic type: `Grp`, `Sgrp`, `Mon`, `CMon`, and so on.  I am not currently abbreviating any of the property names like `associative`, `interchange`, or `commutative`.

The hope is that this being standardized and easy to use locally -- e.g. not reinventing which names to use every time shorthand is required -- this might be more acceptable than more extensive use of `let` local variables.

(I remain deeply convinced that one should use `let` any time an expression could be given a shorter self-explanatory name than its definition, regardless of the number of times the expression is used, and that this codebase would be more maintainable in every respect with that philosophy, but that's a separate issue.)

### DIFF
--- a/src/group-theory/commutative-monoids.lagda.md
+++ b/src/group-theory/commutative-monoids.lagda.md
@@ -185,3 +185,56 @@ module _
   is-unit-prop-Commutative-Monoid x =
     Id-Prop (set-Commutative-Monoid M) x unit-Commutative-Monoid
 ```
+
+## Shorthand
+
+When extensively referring to a commutative monoid `M` and its properties, it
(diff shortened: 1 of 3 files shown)


fredrik-bakke:
See also #1332

lowasser:
Happy to fold the discussion over there.

PR #1482, opened 2025-08-16
Bounded distance decompositions of metric spaces

malarbol:
Given a metric space `A`, this PR introduces the bounded distance decomposition of a metric space `A`: the metric space
```agda
  bounded-distance-decompsition-Metric-Space :
    Metric-Space (l1 ⊔ l2) (l1 ⊔ l2)
  bounded-distance-decompsition-Metric-Space =
    indexed-sum-Metric-Space
      ( set-bounded-distance-component-Metric-Space A)
      ( subspace-bounded-distance-component-Metric-Space A)
```
where 
- `set-bounded-distance-component-Metric-Space A` is the quotient set of `A` by the equivalence relation of being at bounded distance;
- `subspace-bounded-distance-component-Metric-Space A` is the subspace of elements in a given quotient class.
  In each subspace `subspace-bounded-distance-component-Metric-Space A X`, all elements are at bounded distance.

Any metric space is isometrically equivalent to its bounded distance decomposition. The underlying equivalence is provided by the following new results for quotient sets and equivalence classes:

```agda
module _
  {l1 l2 : Level} {A : UU l1} (R : equivalence-relation l2 A)
  where

  equiv-total-equivalence-class :
    Σ (equivalence-class R) (type-subtype ∘ is-in-equivalence-class-Prop R) ≃ A
  equiv-total-equivalence-class = [...]

  equiv-total-set-quotient :
    Σ ( set-quotient R)
       ( type-subtype ∘ is-in-equivalence-class-set-quotient-Prop R) ≃
    ( A)
 equiv-total-set-quotient = [...]
```

PR #1484, opened 2025-08-17
Induced metric space of pseudometric spaces

malarbol:
This PR introduces the *induced metric space* of a pseudometric space `M`: the metric space whose points are
quotient classes of `M` by the similarity relation and neighborhoods given by neighborhoods of inhabitants of the quotient classes: two quotient classes `X`, `Y` are in a `d`-neighborhood if for all `x ∈ X` and `y ∈ Y`, `x` and `y` are `d`-neighbors in the pseudometric space.

Co-authored-by: Louis Wasserman <wasserman.louis@gmail.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macOS-13, ubuntu-latest]
+        os: [ubuntu-latest]
         agda: ['2.7.0']
     steps:
       - name: Checkout our repository


fredrik-bakke:
This PR should be marked as a draft until #1482 is merged.

src/metric-spaces/decomposition-metric-spaces-elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,364 @@
...
+```
+
+### The sum of metric subspaces of elements at bounded distance indexed by the set of components
+
+```agda
+module _
+  {l1 l2 : Level} (A : Metric-Space l1 l2)
+  where
+
+  Σ-components-at-bounded-distance-Metric-Space :
```
fredrik-bakke, line 109: I think `sum` is more appropriate and conventional to use than `Σ` here

src/metric-spaces/decomposition-metric-spaces-elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -0,0 +1,364 @@
+# Decomposition of metric spaces as indexed sums of metric spaces with elements at bounded distance
```
fredrik-bakke, line 1:
```suggestion
# The bounded distance decomposition of metric spaces
```

fredrik-bakke:
oops, I started my review in the wrong PR, my bad!

fredrik-bakke:
I'm wondering about the name of this concept. Should it not be called the quotient metric space associated to a pseudometric space or the (universal) metric space generated by a pseudometric space?

malarbol:
> I'm wondering about the name of this concept. Should it not be called the quotient metric space associated to a pseudometric space or the (universal) metric space generated by a pseudometric space?

Sure. I kept the original name by @lowasser in https://github.com/lowasser/agda-unimath/blob/induced-metric-space-pseudometric/src/metric-spaces/induced-metric-space-of-pseudometric-spaces.lagda.md but I guess we can change it. 
I quite like "quotient metric space associated to a pseudometric space". Do you like `metric-quotient-Pseudometric-Space` for the name of the construction?

fredrik-bakke:
_Completion_ (and maybe even _saturation_) are used in some other contexts too, I know. How do you feel about `metric-completion-Pseudometric-Space`?

malarbol:
> _Completion_ (and maybe even _saturation_) are used in some other contexts too, I know. How do you feel about `metric-completion-Pseudometric-Space`?

mmh. I don't really like this 😅. I think it could bring confusion with the concept of Cauchy completion of (pseudo)metric spaces (which we are also trying to define). Moreover, a "completion" construction is more about "adding new points to fill the voids". Here we're doing the opposite, shrinking the space identifying similar elements. So I don't think "completion" is appropriate. "Saturation" is also already used in another context so, for the same reason, it's not really an option. 

What don't you like about `metric-quotient-Pseudometric-Space`?

fredrik-bakke:
Ah, yes, fair point. I don't have anything against "quotient", I was just spitting ideas

malarbol:
> Ah, yes, fair point. I don't have anything against "quotient", I was just spitting ideas

It seems like the name comes from the [wikipedia page](https://en.wikipedia.org/wiki/Pseudometric_space#Metric_identification): "[...] called the metric space induced by the pseudometric space [...]" but I must confess I still prefer `metric-quotient-Pseudometric-Space`.

BTW, is it ok to link to specific sections of a wikipedia page like [this](https://en.wikipedia.org/wiki/Pseudometric_space#Metric_identification) in a `## See also` section?

fredrik-bakke:
External links, yes

malarbol:
Restarting this in #1507.

PR #1493, opened 2025-08-20
Only run `typecheck` on pull requests that are ready for review

fredrik-bakke:
#1514


fredrik-bakke:
We can also consider running the less costly pre-commit job (and IDK about link-check?) less restrictively.

fredrik-bakke:
Actually, I made that change just now

VojtechStep:
Nice, let's merge this one without CI as well?

fredrik-bakke:
I haven't been able to test if I wrote the condition correctly, so let's maybe hold off until the CI is back up and running, in case we experience other issues at the same time?

VojtechStep:
Okay

PR #1498, opened 2025-08-26
Fix typos in subfinite-indexing

fredrik-bakke:


fredrik-bakke:
I've run the checks locally and I'm merging this one.

PR #1499, opened 2025-08-26
Add runner parameter to CI

fredrik-bakke:
For the potential to run self-hosted runners.


fredrik-bakke:
I've added my macbook as a self-hosted runner for this repo and want to make it possible to run specific jobs on it until the GitHub hosted ones are enabled again.

fredrik-bakke:
It is unclear if this actually achieves _anything_, since it seems it is "GitHub Actions" and not just GitHub's hosted runners that are disabled.

PR #1502, opened 2025-08-28
Some cleanup of species

fredrik-bakke:
- Add concept macro invocations and cross references
- Tidy constructions of equivalences
- Fix indentation
- Improve universe polymorphism
- Fix some grammatical errors and typos
- Make `private` definitions into local ones

### DIFF
--- a/src/foundation/type-arithmetic-cartesian-product-types.lagda.md
+++ b/src/foundation/type-arithmetic-cartesian-product-types.lagda.md
@@ -146,6 +146,16 @@ pr2 (equiv-add-redundant-prop is-prop-B f) =
     ( pr1)
     ( λ p → eq-pair refl (eq-is-prop is-prop-B))
     ( refl-htpy)
+
+equiv-remove-redundant-prop :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  is-prop B → (f : A → B) → A × B ≃ A
+pr1 (equiv-remove-redundant-prop is-prop-B f) = pr1
+pr2 (equiv-remove-redundant-prop is-prop-B f) =
+  is-equiv-is-invertible
+    ( λ a → (a , f a))
+    ( refl-htpy)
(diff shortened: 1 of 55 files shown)


src/foundation/type-arithmetic-cartesian-product-types.lagda.md
```diff
@@ -146,6 +146,16 @@ pr2 (equiv-add-redundant-prop is-prop-B f) =
...
+
+equiv-remove-redundant-prop :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  is-prop B → (f : A → B) → A × B ≃ A
+pr1 (equiv-remove-redundant-prop is-prop-B f) = pr1
+pr2 (equiv-remove-redundant-prop is-prop-B f) =
+  is-equiv-is-invertible
+    ( λ a → (a , f a))
+    ( refl-htpy)
+    ( λ p → eq-pair refl (eq-is-prop is-prop-B))
```
VojtechStep, line 158: `eq-pair-eq-fiber`

fredrik-bakke:
@VojtechStep This PR consists almost entirely of tedious edits and contributes almost no new content, so I don't think it is particularly productive to review. So if you'd like, we can merge this without a review.

fredrik-bakke:
Thanks!

src/foundation/type-arithmetic-cartesian-product-types.lagda.md
fredrik-bakke: Good catch! Unfortunately, since this is equality in cartesian products and we dont import equality in dependent pair types to this file I'm going to leave this as is for now. It should be noted that `eq-pair` is not defined in terms of `eq-pair-Σ`, so they do not unify afaict. For that we would need to define both in terms of something like a binary `apd`. Moreover, there is no `eq-pair-eq-right`, which would be the correct lemma here.

fredrik-bakke:
Checks pass locally 🧹

PR #1494, opened 2025-08-22
Commutative semigroups

lowasser:
I probably need to add sums over finite types, not just the standard finite type, too.

I would love to get away without this, but soon we will need maxima and minima over finite families, and max and min are not monoids in general (and certainly not for the rationals and reals where we need them).

PR #1506, opened 2025-08-29
Joins and meets of finite families of elements

lowasser:
Depends on #1494 .

PR #1510, opened 2025-08-31
Suprema and infima of totally bounded subsets of the reals

lowasser:
This was a lot, and I still need to do infima, but this feels pretty exciting.  Depends on #1494 .


lowasser:
(I would be happy to break this into smaller components, but it's not obvious to me if you'd prefer I do so.)

lowasser:
(...,and further depends on #1506.)

src/foundation/surjective-maps.lagda.md
```diff
@@ -237,6 +238,16 @@ surjection-retract R =
   ( map-retraction-retract R , is-surjective-has-section (section-retract R))
 ```
 
+### If an empty type has a surjection into another type, that type is empty
+
+```agda
+abstract
+  is-empty-surjection-is-empty :
+    {l1 l2 : Level} {A : UU l1} → is-empty A → {B : UU l2} → A ↠ B → is-empty B
```
fredrik-bakke: The following order is more natural
```suggestion
    {l1 l2 : Level} {A : UU l1} {B : UU l2} → A ↠ B → is-empty A → is-empty B
```

src/foundation/surjective-maps.lagda.md
```diff
@@ -237,6 +238,16 @@ surjection-retract R =
   ( map-retraction-retract R , is-surjective-has-section (section-retract R))
 ```
 
+### If an empty type has a surjection into another type, that type is empty
+
+```agda
+abstract
+  is-empty-surjection-is-empty :
```
fredrik-bakke: and  I think you can get away with the name `is-empty-surjection`

fredrik-bakke: which matches `is-inhabited-surjection`

lowasser:
Now also dependent on #1513 .

src/univalent-combinatorics/finitely-enumerable-types.lagda.md
```diff
@@ -118,6 +118,12 @@ finite-enumeration-equiv (n , Fin-n↠X) X≃Y =
   ( n ,
     map-equiv X≃Y ∘ map-surjection Fin-n↠X ,
     is-surjective-left-comp-equiv X≃Y (is-surjective-map-surjection Fin-n↠X))
+
+is-finitely-enumerable-equiv :
+  {l1 : Level} {X : UU l1} → is-finitely-enumerable X →
+  {l2 : Level} {Y : UU l2} → X ≃ Y → is-finitely-enumerable Y
```
fredrik-bakke:
```suggestion
  {l1 l2 : Level} {X : UU l1} {Y : UU l2} → 
  X ≃ Y →
  is-finitely-enumerable X →
  is-finitely-enumerable Y
```

fredrik-bakke:
> (I would be happy to break this into smaller components, but it's not obvious to me if you'd prefer I do so.)

Yes, this PR is a bit too large for me at the moment. Would you be able to do that for me?

lowasser:
Marking as a draft until I think it's small enough (and the broken-up pieces are submitted).

lowasser:
Broken into #1521 , #1522, and #1523, and should be a pretty small PR (+644 lines) after those three are merged.

fredrik-bakke:
Thank you so much for doing that! I'll review them when I have the time

src/metric-spaces/nets-metric-spaces.lagda.md
```diff
@@ -86,6 +87,24 @@ module _
       is-approximation-subset-net-Metric-Space)
 ```
 
+### If a metric space is inhabited, so is any net
```
fredrik-bakke:
```suggestion
### If a metric space is inhabited, so is any net on it
```

src/real-numbers/totally-bounded-subsets-real-numbers.lagda.md
```diff
@@ -0,0 +1,431 @@
...
+  is-totally-bounded-totally-bounded-subset-ℝ = pr2 S
+
+  is-inhabited-totally-bounded-subset-ℝ : UU (l1 ⊔ lsuc l2)
+  is-inhabited-totally-bounded-subset-ℝ =
+    is-inhabited-subset-ℝ subset-totally-bounded-subset-ℝ
+```
+
+## Properties
+
+### The negation of a totally bounded subset of ℝ is totally bounded
```
fredrik-bakke:
```suggestion
### The elementwise negation of a totally bounded subset of ℝ is totally bounded
```

src/real-numbers/totally-bounded-subsets-real-numbers.lagda.md
```diff
@@ -0,0 +1,431 @@
...
+            ( max-net ε' -ℝ real-ℚ⁺ ε')
+            ( x)
+            ( max-net-ε'-ε'<x)
+            ( concatenate-le-leq-ℝ
+              ( sup -ℝ real-ℚ⁺ ε)
+              ( sup -ℝ real-ℚ⁺ (ε' +ℚ⁺ ε'))
+              ( max-net ε' -ℝ real-ℚ⁺ ε')
+              ( reverses-le-diff-ℝ sup _ _ (preserves-le-real-ℚ _ _ ε'+ε'<ε))
+              ( tr (λ y → leq-ℝ y (max-net ε' -ℝ real-ℚ⁺ ε'))
+                ( associative-add-ℝ _ _ _ ∙
```
fredrik-bakke:
```suggestion
              ( tr
                ( λ y → leq-ℝ y (max-net ε' -ℝ real-ℚ⁺ ε'))
                ( associative-add-ℝ _ _ _ ∙
```

src/real-numbers/totally-bounded-subsets-real-numbers.lagda.md
```diff
@@ -0,0 +1,431 @@
...
+        ( is-inhabited-net-inhabited-Metric-Space
+          ( metric-space-subset-ℝ S)
+          ( |S|)
+          ( δ)
+          ( M δ))
+
+    max-net : ℚ⁺ → ℝ l2
+    max-net δ =
+      max-inhabited-finitely-enumerable-subset-ℝ
+        ( net δ)
```
fredrik-bakke, line 198: as @VojtechStep has pointed out a couple of times, `private` definitions in Agda suffer from certain deficiencies making it inadvisable to use them, so these should be local `where` or `let` bindings instead.

lowasser: That's depressing, given that they need to get reused between several different functions -- or I would've used `let` originally.

lowasser: FWIW -- the only deficiency I remember seeing a comment about is that they show up strangely in goals, but these are `abstract` definitions so they should never be visible in goals anyway..?

lowasser: (adding a few skipped words: the only complaint I've seen about `private` definitions in Agda is that references to those private definitions show up strangely in goals, but because every function that uses one of these `private` definitions is `abstract`, the bodies of those functions (including those private references) should never show up anyway in goals, and as a result it's not obvious to me what the downside is)

fredrik-bakke: Oh yeah, that might be true. Do recall that you can do let bindings in the module declaration though.

fredrik-bakke: If thats the only deficiency then I suppose there is no harm in using them here, although I always worry that people will observe coding patterns in the library and then misapply them

lowasser: That's valid, though it might be reasonably addressed by documentation in the practices.  I can write something about that if you like.

I am inclined to leave this as it is.

fredrik-bakke: Yeah, fair! If you'd like to do that I'd be very glad. Btw, I had partially forgotten about the do-syntax doc PR. I'll try to get to that at some point soon in return.

src/real-numbers/totally-bounded-subsets-real-numbers.lagda.md
```diff
@@ -0,0 +1,432 @@
...
+                  ( sup-modulated-totally-bounded-subset-ℝ -ℝ real-ℚ⁺ ε)
+                  ( s)))
+      in do
+        (ε' , ε'+ε'<ε) ← double-le-ℚ⁺ ε
+        ((x , x∈net-ε') , max-net-ε'-ε'<x) ←
+          is-approximated-below-max-inhabited-finitely-enumerable-subset-ℝ
+            ( net ε')
+            ( is-inhabited-net ε') ε'
+        intro-exists
+          (x , net⊆S ε' x x∈net-ε')
```
fredrik-bakke:
```suggestion
          ( x , net⊆S ε' x x∈net-ε')
```

fredrik-bakke:
Checks pass 🎖️

PR #1520, opened 2025-09-02
Metric spaces induced by metric functions

lowasser:
I do find myself wondering if it'd be better to define the metric space of real numbers in terms of the distance function, rather than in terms of the neighborhood shenanigans, though they are of course equivalent.  (I haven't bothered proving that yet.)  

This of course gives locatedness and bounded distance for free.

(I also haven't yet proven that the distance function available to all located metric spaces with all elements at bounded distance is the same as the original distance function.)

### DIFF
--- a/src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
+++ b/src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
@@ -260,4 +260,17 @@ module _
             ( q≤0)
             ( le-zero-is-positive-ℚ ε is-pos-ε))
           ( ε<q)
+
+    leq-zero-upper-real-dist-Metric-Space :
+      leq-upper-ℝ (upper-real-ℚ zero-ℚ) upper-real-dist-Metric-Space
+    leq-zero-upper-real-dist-Metric-Space q =
+      rec-trunc-Prop
+        ( le-ℚ-Prop zero-ℚ q)
+        ( λ ((ε , Nεxy) , ε<q) →
+          transitive-le-ℚ
+            ( zero-ℚ)
+            ( rational-ℚ⁺ ε)
+            ( q)
+            ( ε<q)
+            ( le-zero-is-positive-ℚ (pr1 ε) (pr2 ε)))
 ```
--- a/src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
+++ b/src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
@@ -262,7 +262,7 @@ module _
           ( ε<q)
 
     leq-zero-upper-real-dist-Metric-Space :
-      leq-upper-ℝ (upper-real-ℚ zero-ℚ) upper-real-dist-Metric-Space
+      leq-upper-ℝ zero-upper-ℝ upper-real-dist-Metric-Space
     leq-zero-upper-real-dist-Metric-Space q =
       rec-trunc-Prop
         ( le-ℚ-Prop zero-ℚ q)
--- a/src/metric-spaces/metrics.lagda.md
+++ b/src/metric-spaces/metrics.lagda.md
@@ -154,7 +154,7 @@ module _
     leq-prop-ℝ⁰⁺ (dist-Metric X μ x y) (nonnegative-real-ℚ⁺ ε)
 ```
 
-#### The rational neighborhood relation induced by a metric is reflexive
+### The rational neighborhood relation induced by a metric is reflexive
 
 ```agda
 module _
--- a/src/metric-spaces/metrics.lagda.md
+++ b/src/metric-spaces/metrics.lagda.md
@@ -1,0 +1,358 @@
+# Metrics
+
+```agda
+module metric-spaces.metrics where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import elementary-number-theory.nonnegative-rational-numbers
+open import elementary-number-theory.positive-rational-numbers
+
+open import foundation.conjunction
+open import foundation.dependent-pair-types
+open import foundation.disjunction
+open import foundation.identity-types
+open import foundation.negation
+open import foundation.propositions
+open import foundation.sets
+open import foundation.subtypes
+open import foundation.transport-along-identifications
+open import foundation.universe-levels
+
+open import logic.functoriality-existential-quantification
+
+open import metric-spaces.elements-at-bounded-distance-metric-spaces
+open import metric-spaces.extensionality-pseudometric-spaces
+open import metric-spaces.located-metric-spaces
+open import metric-spaces.metric-spaces
+open import metric-spaces.pseudometric-spaces
+open import metric-spaces.rational-neighborhood-relations
+open import metric-spaces.reflexive-rational-neighborhood-relations
+open import metric-spaces.saturated-rational-neighborhood-relations
+open import metric-spaces.symmetric-rational-neighborhood-relations
+open import metric-spaces.triangular-rational-neighborhood-relations
+
+open import real-numbers.dedekind-real-numbers
+open import real-numbers.nonnegative-real-numbers
+open import real-numbers.rational-real-numbers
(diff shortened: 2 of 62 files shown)


lowasser:
(Notably, one thing this is shooting is for is that there is a distance function between uniformly continuous functions on totally bounded subsets of R; that metric space is in fact located, complete, and every element is at bounded distance.)

src/metric-spaces/elements-at-bounded-distance-metric-spaces.lagda.md
```diff
@@ -260,4 +260,17 @@ module _
             ( q≤0)
             ( le-zero-is-positive-ℚ ε is-pos-ε))
           ( ε<q)
+
+    leq-zero-upper-real-dist-Metric-Space :
+      leq-upper-ℝ (upper-real-ℚ zero-ℚ) upper-real-dist-Metric-Space
```
fredrik-bakke: `upper-real-ℚ zero-ℚ` should be defined as `zero-upper-ℝ`

src/metric-spaces/metrics.lagda.md
```diff
@@ -0,0 +1,358 @@
...
+            ( not-leq-le-ℝ
+              ( real-ℚ⁺ δ)
+              ( real-ℝ⁰⁺ (dist-Metric X μ x y))
+              ( δ<d)))
+        ( λ d<ε →
+          inr-disjunction
+            ( leq-le-ℝ
+              ( real-ℝ⁰⁺ (dist-Metric X μ x y))
+              ( real-ℚ⁺ ε)
+              ( d<ε)))
```
fredrik-bakke: Looks like you're actually applying the ~binary~ functorial action of disjunctions here. It would be nice to have that

fredrik-bakke: (Not a requirement for merging)

fredrik-bakke: I.e., with the proper definition, this could ~perhaps~ be written as

```suggestion
      map-disjunction
        ( not-leq-le-ℝ (real-ℚ⁺ δ) (real-ℝ⁰⁺ (dist-Metric X μ x y)))
        ( leq-le-ℝ (real-ℝ⁰⁺ (dist-Metric X μ x y)) (real-ℚ⁺ ε))
```
which looks quite a bit more appealing to me (for this you will have to use the conflation between `disjuction` and `disjunction-type`)

fredrik-bakke: Actually, I'm promoting this to a requirement for merging, since it simplifies the instantiation so much, and is more broadly a useful addition.

src/metric-spaces/metrics.lagda.md
fredrik-bakke, line 1: Very nice addition!

src/real-numbers/nonnegative-real-numbers.lagda.md
```diff
@@ -37,4 +53,347 @@ is-nonnegative-prop-ℝ = leq-ℝ-Prop zero-ℝ
...
+```
+
+### Strict inequality on nonnegative real numbers
+
+```agda
+module _
+  {l1 l2 : Level} (x : ℝ⁰⁺ l1) (y : ℝ⁰⁺ l2)
+  where
+
+  le-ℝ⁰⁺-Prop : Prop (l1 ⊔ l2)
```
fredrik-bakke: should be `le-prop-ℝ⁰⁺`

src/real-numbers/similarity-real-numbers.lagda.md
```diff
@@ -44,8 +44,8 @@ opaque
   sim-prop-ℝ : {l1 l2 : Level} → ℝ l1 → ℝ l2 → Prop (l1 ⊔ l2)
   sim-prop-ℝ x y = sim-prop-subtype (lower-cut-ℝ x) (lower-cut-ℝ y)
 
-  sim-ℝ : {l1 l2 : Level} → ℝ l1 → ℝ l2 → UU (l1 ⊔ l2)
-  sim-ℝ x y = type-Prop (sim-prop-ℝ x y)
+sim-ℝ : {l1 l2 : Level} → ℝ l1 → ℝ l2 → UU (l1 ⊔ l2)
+sim-ℝ x y = type-Prop (sim-prop-ℝ x y)
```
fredrik-bakke: I find this choice a bit strange. Sure, you want `sim-ℝ` to definitionally be the first projection of `sim-prop-ℝ`, but the better way to arrange that is to first define `sim-ℝ` opaquely, and then define `sim-prop-ℝ` non-opaquely as the pair `sim-ℝ , is-prop-sim-ℝ`. This way the `unfolding` statements written elsewhere are more relevant to what you actually want to unfold, namely `sim-ℝ`.

fredrik-bakke: Perhaps this "anti-pattern" is also present in #1521. If so, I recommend you do the same for other definitions in that PR.

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -244,6 +244,38 @@ module _
...
 
+### The real numbers are located
+
+```agda
+module _
+  {l : Level} (x : ℝ l) (p q : ℚ) (p<q : le-ℚ p q)
+  where
+
+  is-located-le-ℝ :
+    type-disjunction-Prop (le-ℝ-Prop (real-ℚ p) x) (le-ℝ-Prop x (real-ℚ q))
```
fredrik-bakke: you could also, if you wanted, write this as
```suggestion
    disjunction-type (le-ℝ (real-ℚ p) x) (le-ℝ x (real-ℚ q))
```

src/real-numbers/strict-inequality-real-numbers.lagda.md
```diff
@@ -647,6 +691,53 @@ module _
...
+          dense-rational-le-ℝ zero-ℝ (y -ℝ x)
+            ( preserves-le-left-sim-ℝ
+              ( y -ℝ x)
+              ( x -ℝ x)
+              ( zero-ℝ)
+              ( right-inverse-law-add-ℝ x)
+              ( preserves-le-right-add-ℝ (neg-ℝ x) x y x<y))
+        intro-exists
+          ( q , is-positive-le-zero-ℚ q (reflects-le-real-ℚ _ _ 0<q))
+          ( le-transpose-right-diff-ℝ' _ _ _ q<y-x)
```
fredrik-bakke: looks like a `map-exists`

src/metric-spaces/metrics.lagda.md
lowasser: The functoriality of disjunction [already exists](https://unimath.github.io/agda-unimath/foundation.functoriality-disjunction.html), but turns out to require so many explicit type arguments that it's much more painful to use than this approach.

lowasser: I can have a go at improving that state of affairs, I suppose?

fredrik-bakke: Oh, yes please! Using the `disjunction-type ↜?↝ disjunction` conflation, I hope most of that file can use implicit type parameters rather than explicit Props

src/real-numbers/strict-inequality-real-numbers.lagda.md
lowasser: I don't think `map-exists` will actually quite do it?  In `exists A B -> exists C D`, you can only use `A` in computing `C`, not `B`, and I don't think that's good enough here.

It's not clear to me it'd be clearer or more readable either way.

fredrik-bakke: You're right, my bad.

fredrik-bakke:
This PR does not pass the checks after updating the branch. Please fix the issues, so it can be merged. ❌

lowasser:
On it.

lowasser:
A complete build now succeeds for me.

fredrik-bakke:
Alright, good to go 👍

PR #1521, opened 2025-09-02
Opacify real numbers

lowasser:
Marks more things in `real-numbers` opaque, and does a few other cleanups here and there.

In general, it's strongly preferable to see things like `neg-R` in goals rather than...the mess you get if it's not opaque...and the same for most things about the real numbers.

PR #1522, opened 2025-09-02
Maximum and minimum of finite families of real numbers

lowasser:
Pulled out of #1510; depends on #1521.


lowasser:
Marking as a draft until #1521 is submitted.

fredrik-bakke:
Checks are good now 🐗

PR #1524, opened 2025-09-02
Simplify the minimum of real numbers

lowasser:
It's simpler to reuse the logic for maximum than to reinvent it for the minimum of real numbers.


lowasser:
Arguably we could even delete the minimum of lower and upper Dedekind real numbers, too, since that's very unlikely to be independently useful, but eh.

fredrik-bakke:
checks pass ☑️

PR #1527, opened 2025-09-03
Finite sums in groups and abelian groups

lowasser:
Just filling these out.


src/linear-algebra/finite-sequences-in-abelian-groups.lagda.md
```diff
@@ -0,0 +1,158 @@
+# Finite sequences in abelian-groups
```
fredrik-bakke:
```suggestion
# Finite sequences in abelian groups
```

fredrik-bakke:
Checks also pass 🏔️

PR #1528, opened 2025-09-03
Series in metric abelian groups

lowasser:
I think this is a reasonable place to start -- both for series in general, and for an `analysis` package.  Depends on #1527.

### DIFF
--- a/src/analysis/series-metric-abelian-groups.lagda.md
+++ b/src/analysis/series-metric-abelian-groups.lagda.md
@@ -1,0 +1,208 @@
+# Series in metric abelian groups
+
+```agda
+module analysis.series-metric-abelian-groups where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import analysis.metric-abelian-groups
+
+open import elementary-number-theory.addition-natural-numbers
+open import elementary-number-theory.natural-numbers
+
+open import foundation.action-on-identifications-functions
+open import foundation.dependent-pair-types
+open import foundation.equivalences
+open import foundation.function-extensionality
+open import foundation.function-types
+open import foundation.homotopies
+open import foundation.identity-types
+open import foundation.universe-levels
+
+open import group-theory.abelian-groups
+open import group-theory.sums-of-finite-sequences-of-elements-abelian-groups
+
+open import lists.sequences
+
+open import univalent-combinatorics.coproduct-types
+open import univalent-combinatorics.standard-finite-types
+```
+
+</details>
+
+## Idea
+
+A
+{{#concept "series" WD="series" WDID=Q170198 disambiguation="in an abelian group" Agda=series-Metric-Ab}}
+in a [metric abelian group](analysis.metric-abelian-groups.md) `G` is a sum over
+an infinite [sequence](lists.sequences.md) of elements of `G`.
+
+## Definition
+
+```agda
+record series-Metric-Ab {l1 l2 : Level} (G : Metric-Ab l1 l2) : UU l1 where
+  constructor series-terms-Metric-Ab
+  field
+    term-series-Metric-Ab : sequence (type-Metric-Ab G)
+
+open series-Metric-Ab
+```
+
+## Properties
+
+### Equality of series
+
+```agda
+module _
+  {l1 l2 : Level} (G : Metric-Ab l1 l2)
+  where
+
+  equiv-series-sequence-Metric-Ab :
+    series-Metric-Ab G ≃ sequence (type-Metric-Ab G)
+  equiv-series-sequence-Metric-Ab =
+    ( term-series-Metric-Ab ,
+      is-equiv-is-invertible series-terms-Metric-Ab (λ x → refl) (λ x → refl))
+
+  eq-series-Metric-Ab :
+    {f g : series-Metric-Ab G} →
+    term-series-Metric-Ab f ＝ term-series-Metric-Ab g → f ＝ g
+  eq-series-Metric-Ab = ap series-terms-Metric-Ab
+
+  eq-htpy-term-series-Metric-Ab :
+    {f g : series-Metric-Ab G} →
+    term-series-Metric-Ab f ~ term-series-Metric-Ab g → f ＝ g
+  eq-htpy-term-series-Metric-Ab f~g = eq-series-Metric-Ab (eq-htpy f~g)
+```
+
+### Partial sums of a series
+
+```agda
+module _
+  {l1 l2 : Level} (G : Metric-Ab l1 l2)
+  where
+
+  partial-sum-series-Metric-Ab : series-Metric-Ab G → ℕ → type-Metric-Ab G
+  partial-sum-series-Metric-Ab (series-terms-Metric-Ab f) n =
+    sum-fin-sequence-type-Ab (ab-Metric-Ab G) n (λ i → f (nat-Fin n i))
+```
+
+### The `n`th term of a series is the difference of the `succ-ℕ n`th partial sum and the `n`th
+
+```agda
+module _
+  {l1 l2 : Level} (G : Metric-Ab l1 l2)
+  where
+
+  eq-term-diff-partial-sum-series-Metric-Ab :
+    (f : series-Metric-Ab G) (n : ℕ) →
+    add-Metric-Ab G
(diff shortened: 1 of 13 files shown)


src/analysis/series-metric-abelian-groups.lagda.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+</details>
+
+## Idea
+
+A
+{{#concept "series" WD="series" WDID=Q170198 disambiguation="in an abelian group" Agda=series-Metric-Ab}}
+in a [metric abelian group](analysis.metric-abelian-groups.md) `G` is a sum over
+an infinite [sequence](lists.sequences.md) of elements of `G`.
```
fredrik-bakke: Seems like you actually mean 
```suggestion
a countably infinite [sequence](lists.sequences.md) of elements of `G`.
```
but I could be wrong. At least that is what your formalization says. Note that infinite sets are a bit strange in constructive math. Although I don't know much about them, there is at least a definition in the library.

src/analysis/series-metric-abelian-groups.lagda.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Definition
+
+```agda
+record series-Metric-Ab {l1 l2 : Level} (G : Metric-Ab l1 l2) : UU l1 where
+  constructor series-terms-Metric-Ab
+  field
+    term-series-Metric-Ab : sequence (type-Metric-Ab G)
+
+open series-Metric-Ab
```
fredrik-bakke: Why do you choose to formalize series like this rather than just `series-Metric-Ab = sequence (type-Metric-Ab G)`? This way you don't have to transport along the trivial equivalence `equiv-series-sequence-Metric-Ab`. I suppose, if you wanted, you could define it as `opaque` also.

lowasser: I want to be careful about mixing up a series and its terms.  Marking it opaque might be good enough, but I don't actually want these things to be substitutable for one another.

src/analysis/series-metric-abelian-groups.lagda.md
lowasser: I have to admit I assume that "infinite sequence" would _always_ refer to a sequence indexed by the natural numbers.  An "infinite set" might not be, but _sequence_ implies natural indexing to me.

src/analysis/series-metric-abelian-groups.lagda.md
lowasser: But the downside of making it opaque is that the terms of a series of a sequence of terms is not judgmentally equal to the original sequence.  So this seems like the only viable strategy that ensures you can't accidentally mix up a series and its terms.

src/analysis/series-metric-abelian-groups.lagda.md
fredrik-bakke: I get your sentiment, but it can also be understood as a map from any ordinal. I was initially confused because "infinite" is a scary word to me and I didn't know what you meant by it, and some other people may be in the same position as me. Since "countably infinite" leaves no room for confusion, why not use it?

src/analysis/series-metric-abelian-groups.lagda.md
fredrik-bakke: Gotcha, that makes a lot of sense! Would you mind adding this explanation, that we define it as a record because it is intensionally different from the terms of the underlying sequence? I imagine others might also get confused by this trivial definition.

src/analysis/series-metric-abelian-groups.lagda.md
```diff
@@ -0,0 +1,211 @@
...
+
+A
+{{#concept "series" WD="series" WDID=Q170198 disambiguation="in an abelian group" Agda=series-Metric-Ab}}
+in a [metric abelian group](analysis.metric-abelian-groups.md) `G` is a sum over
+a countably infinite [sequence](lists.sequences.md) of elements of `G`.
+
+Series are defined with a record to make them intensionally distinct from the
+sequence of their terms.
+
+## Definition
```
fredrik-bakke: This remark is more relevant to the definition section than the idea section
```suggestion
## Definition

Series are defined with a record to make them intensionally distinct from the
sequence of their terms.
```

src/analysis/series-metric-abelian-groups.lagda.md
```diff
@@ -0,0 +1,211 @@
...
+Series are defined with a record to make them intensionally distinct from the
+sequence of their terms.
+
+```agda
+record series-Metric-Ab {l1 l2 : Level} (G : Metric-Ab l1 l2) : UU l1 where
+  constructor series-terms-Metric-Ab
+  field
+    term-series-Metric-Ab : sequence (type-Metric-Ab G)
+
+open series-Metric-Ab
```
fredrik-bakke: Don't you mean to open this record publicly?

fredrik-bakke:
Alright, checks pass, this is ready to go 👨🏻‍✈️

VojtechStep:
As a consequence of making `series-Metric-Ab` a record, you also get better type inference, which means you can make most of the `G : Metric-Ab` arguments implicit, since record types are injective --- when you have `f : {G : Metric-Ab} -> series-Metric-Ab G -> ...` and you call `f foo` with `foo : series-Metric-Ab my-cool-ab`, Agda knows that unifying `series-Metric-Ab ?G = series-Metric-Ab my-cool-ab` is only possible when `?G = my-cool-ab`.

This breaks a little when you try to call `f` with a constructor literal, like `f (series-terms-Metric-Ab s)`, since the type of `series-terms-Metric-Ab s` can't infer the precise type `series-Metric-Ab ?G` (that's why arguments like `G` need to be explicit elsewhere in the first place). You do that in the proof of `partial-sum-drop-series-Metric-Ab`, but that can be fixed by using `s@(series-terms-Metric-Ab σ)` instead of `(series-terms-Metric-Ab σ)` in your argument binding, and replacing `(series-terms-Metric-Ab σ)` with `s`. This gets rid of all but two places where you specify `G`, and even those are unnecessary if you replace ` series-terms-Metric-Ab (s ∘ add-ℕ n)` with `drop-series-Metric-Ab n σ` (I'm not sure if the unfolding was done for pedagogical reasons). The resulting reasoning would then look like
```agda
    partial-sum-drop-series-Metric-Ab n σ@(series-terms-Metric-Ab s) i =
      inv
        ( equational-reasoning
          diff-Metric-Ab G
            ( partial-sum-series-Metric-Ab σ
              ( n +ℕ i))
            ( partial-sum-series-Metric-Ab σ n)
          ＝
            diff-Metric-Ab G
              ( add-Metric-Ab G
                ( sum-fin-sequence-type-Ab (ab-Metric-Ab G) n
                  ( s ∘ nat-Fin (n +ℕ i) ∘ inl-coproduct-Fin n i))
                ( sum-fin-sequence-type-Ab (ab-Metric-Ab G) i
                  ( s ∘ nat-Fin (n +ℕ i) ∘ inr-coproduct-Fin n i)))
              ( partial-sum-series-Metric-Ab σ n)
            by
              ap-diff-Metric-Ab G
                ( split-sum-fin-sequence-type-Ab
                  ( ab-Metric-Ab G)
                  ( n)
                  ( i)
                  ( s ∘ nat-Fin (n +ℕ i)))
                ( refl)
          ＝
            diff-Metric-Ab G
              ( add-Metric-Ab G
                ( partial-sum-series-Metric-Ab σ n)
                ( partial-sum-series-Metric-Ab (drop-series-Metric-Ab n σ) i))
              ( partial-sum-series-Metric-Ab σ n)
            by
              ap-diff-Metric-Ab G
                ( ap-add-Metric-Ab G
                  ( htpy-sum-fin-sequence-type-Ab (ab-Metric-Ab G) n
                    ( λ j → ap s (nat-inl-coproduct-Fin n i j)))
                  ( htpy-sum-fin-sequence-type-Ab (ab-Metric-Ab G) i
                    ( λ j → ap s (nat-inr-coproduct-Fin n i j))))
                ( refl)
          ＝
            partial-sum-series-Metric-Ab
              ( drop-series-Metric-Ab n σ)
              ( i)
            by is-identity-left-conjugation-Ab (ab-Metric-Ab G) _ _)
```

PR #1537, opened 2025-09-08
Add linear combinations, linear spans and tuple concatenation

SimonBrandner:
This PR adds linear combinations, linear spans and tuple concatenation. It contains a proof that a linear span has the structure of a left submodule.

PR #1540, opened 2025-09-09
Add left modules with bases

SimonBrandner:
Builds on top of https://github.com/UniMath/agda-unimath/pull/1537


lowasser:
FWIW; I'd prefer to see support for "a basis can be infinite and elements of the module must be a finite linear combination of elements of that basis", to make polynomials a module and monomials the basis.

Issue #946, opened 2023-11-26
Idea: use coherently invertible maps as default notion of equivalence


fredrik-bakke:
This is useful because it can give definitional computation laws for the inverse. It will take a large effort to refactor the greater library, but perhaps it is possible to start with a smaller refactoring to at least instate this as the default for future contributions. On the other hand, having two competing defaults may be detrimental to the quality and usability of the library, so we should discuss this rigorously before attempting to implement it.

fredrik-bakke:
it would be very useful to implement #1030 before we undertake this refactoring. This will let us see if the refactoring leads to a performance improvement.

PR #1543, opened 2025-09-11
Interchanging multiplication and evaluation on polynomials

lowasser:


src/commutative-algebra/polynomials-commutative-semirings.lagda.md
```diff
@@ -749,6 +757,268 @@ module _
       is-polynomial-mul-polynomial-Commutative-Semiring pp qq)
 ```
 
+#### The product of two polynomials, evaluated at `x`, is equal to the sum of each polynomial evaluated at `x`
```
fredrik-bakke:
```suggestion
#### The product of two polynomials, evaluated at `x`, is equal to the scalar multiplication of each polynomial evaluated at `x`
```

src/commutative-algebra/polynomials-commutative-semirings.lagda.md
```diff
@@ -749,6 +757,268 @@ module _
...
+              ( λ (i , i<Np) → (cp i *R power-R i x)) *R
+            ev-polynomial-Commutative-Semiring qq x
+            by
+              inv
+                ( right-distributive-mul-sum-finite-Commutative-Semiring R
+                  ( _)
+                  ( _)
+                  ( _))
+          ＝
+            ev-polynomial-Commutative-Semiring pp x *R
```
fredrik-bakke, line 1005: I suppose the extra line break actually helps a bit with readability since each step is so large

src/foundation-core/equivalences.lagda.md
```diff
@@ -420,11 +420,10 @@ module _
   {l1 l2 l3 : Level} {A : UU l1} {B : UU l2} {X : UU l3}
   where
 
-  abstract
-    is-equiv-comp :
-      (g : B → X) (h : A → B) → is-equiv h → is-equiv g → is-equiv (g ∘ h)
-    pr1 (is-equiv-comp g h (sh , rh) (sg , rg)) = section-comp g h sh sg
-    pr2 (is-equiv-comp g h (sh , rh) (sg , rg)) = retraction-comp g h rg rh
+  is-equiv-comp :
```
fredrik-bakke: So, `is-equiv` proofs are very intentionally made `abstract`, since we do not want our proofs to depend on the underlying choice of representation of equivalences.

lowasser: Is the intent specifically that I can depend on `map-equiv (f ∘e g)` to be `map-equiv f  ∘ map-equiv g`, but *not* that map-inv-equiv (f ∘e g) is `map-inv-equiv g ∘ map-inv-equiv f`?

That's what I'm working around here.  I'm happy to leave the proofs that these are sections and retractions abstract, but I do absolutely expect that if I compose explicit equivalences with nonabstract maps and inverses, that the map and inverse maps are preserved in the expected way.

lowasser: I have changed the equivalence to be composed in the opposite direction, but I have to admit I do still find it nonsensical that the forward direction of the equivalence is judgmentally just the appropriate composition, and the backward direction is not.

fredrik-bakke: I suppose, kinda; It's been too long since I thought about this, but #946 was written as a potential solution to the computational issues with `is-equiv`. Part of the issue is that we don't want definitions in the library to depend on whether `map-inv-equiv` returns the section definitionally or the retraction, so in general we avoid using definitional properties of `map-inv-equiv`. I can try and dig for what Egbert said about it if you want, and he may also have talked about it on the Discord server in the past.

fredrik-bakke: This is why I think coherently invertibility maps are better, because, disregarding the 2-coherence, the inverse operation is a definitional involution, — and we can just mark the 2-coherences as `abstract` in general to save on Agda thinking time.

src/logic/complements-decidable-subtypes.lagda.md
```diff
@@ -120,18 +120,15 @@ module _
...
-        by
-        left-distributive-Σ-coproduct A
-          ( is-in-decidable-subtype P)
-          ( is-in-complement-decidable-subtype P)
+              ( is-decidable-decidable-subtype P x)) ∘e
+        inv-equiv
+          ( left-distributive-Σ-coproduct
+            ( A)
+            ( is-in-decidable-subtype P)
+            ( is-in-complement-decidable-subtype P)))
```
fredrik-bakke: This is not an improvement

lowasser: I agree.  But it works around is-equiv-comp being abstract, and it's the least invasive way I can find to do it.

lowasser: Like, fundamentally, I'm 100% behind the use of `abstract` when it's hiding details that can't be relevant to anything using the function (e.g. the function is proving a proposition), or when it's being used to hide implementation details that shouldn't be depended on -- but these _are_ relevant details for this case.

I could write the entire explicit equivalence here, of course, but that gives up on the whole idea of building proofs out of lemmas; we have two useful equivalences already that we can compose together to prove the result.

The use case here is `sum-finite-Commutative-Monoid M A f ＝sum-finite-Commutative-Monoid M B (f ∘ map-inv-equiv H)`, as applied to `A ≃ P + A\P`.  We need a _concrete_ mapping from `P + A\P` to `A` that is an equivalence; we cannot in fact get away without one, because we have to eliminate it from the final equations anyway.

I do think dropping `abstract` on `is-equiv-comp` is by a significant margin the clearest and even most principled way to accomplish this, even if it adds complexity to a future migration of equivalences to be more conveniently structured.  But if we can't do that, I don't have a better solution than either this, or giving up on composing equivalences.

fredrik-bakke: I'm not quite following what your use case is. Does the map `f` and the equivalence `H` have any specific relation? Which line is it on? And is it not possible to use something like uniqueness of inverses, uniqueness of `is-equiv` proofs, or transport of `is-equiv` along homotopies? I.e. `htpy-map-inv-is-equiv`, `is-property-is-equiv`, and `is-equiv-htpy` respectively.

I'd expect that un`abstract`ing `is-equiv-comp` would lead to a noticable typechecking overhead, since this is one of the most used definitions in the library. Not that I'm seriously considering this option at the moment, but I am not a fan of the possibility that some proofs may have to invoke funext because of the `abstract` design choice. Another alternative, if push comes to shove, would be to `opaque` it; That would at least maintain some restrictions on the abuse of the definitional properties of `is-equiv`.

fredrik-bakke: This is not really about mitigating problems in the event of a potential future port to a different notion of equivalence. At this point I don't seriously think that move will happen, at least not in the current situation of the library. The design choices had already been made by the point we started discussing alternative definitions, and they were made by Egbert to, as far as I can recall a quote, stop contributors from using bad practices when constructing and using equivalences. The idea was to force people to explicitly reason through code about how the constructed equivalences act instead. And to be clear I support his idea, it has forced me to rethink how I formalized certain things in the past, and each time that has led me to find neater ways of formalizing my arguments.

lowasser: The place where it's relevant is decompose-sum-decidable-subset-finite-Commutative-Monoid.

What that use site needs is for `f ∘ map-inv-equiv (equiv-coproduct-decomposition-subset-Finite-Type A P)` to reduce to the obvious trivial map `f ∘ inclusion-decidable-subtype P`.

lowasser: We know nothing at all about `f`.  The equivalence is not relevant to the result of `decompose-sum-decidable-subset-finite-Commutative-Monoid` and certainly needs to get eliminated for its end users `decompose-sum-decidable-subset-finite-Commutative-Monoid`.

lowasser: Er, technically I think it's that `f (map-inv-equiv (equiv-coproduct-decomposition-subset-Finite-Type A P)) (inl (a , a-in-A)) = f a`.

lowasser: We need to end up with a function `λ (a , a∈A) → f a`.

lowasser: Just pinging on this to say I've made `is-equiv-comp` opaque and unfolded it at this use site.  I'm sympathetic to the goals described, but I continue to actively believe there is no better alternative for this use case; `sum-equiv-finite-Commutative-Monoid` seems a) to be the best way to express the claim it makes, and b) to really be only useful in equational reasoning when you can inline `map-inv-equiv`.

fredrik-bakke:
@lowasser Please forgive the slow response time on this one. Hopefully I'll find the time to look at the final issues over the weekend

fredrik-bakke (approved):

PR #1552, opened 2025-09-24
Hide `typecheck` workflow on master

fredrik-bakke:
The previous change to the `typecheck` workflow #1493 contained a mistake that made it skip on commits to the master branch, which was unintended. However, I believe that we in fact do not need to run it on master, since it is already tested in the pull request branch, and besides, the benchmarking workflow also verifies that master type checks. Hence, this PR hides the workflow from the workflow agenda on master. It does so by creating a new ci file.

I've still added an inclusion rule so that if we in the future wish to run it on commits to a certain branch we only need to add an "on push" entry and not edit the if clause for `typecheck`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,49 +27,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  typecheck:
-    if:
-      ${{ github.event_name == 'workflow_dispatch' || (github.event_name ==
-      'pull_request' && github.event.pull_request.draft == false) }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        agda: ['2.7.0']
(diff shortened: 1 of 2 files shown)


VojtechStep:
Since we require PRs to be up-to-date before merging them, does it even make sense to run precommit and link-check on master? We could have incremental typechecking and style checking in PRs, and then master does a clean build (which typechecks) and deployment (which linkchecks). WDYT?

fredrik-bakke:
Yeah that sounds good! In the past I was thinking of these redundancies as double checks that the workflows were working properly, but in the current state of affairs I agree we can do without them. The problems don't have to be detected on master anyways; if we find an issue later, _we've found the issue_.

fredrik-bakke:
@VojtechStep updated now

VojtechStep:
LGTM! (for some reason GitHub decided to delay the ping notification for two hours)

PR #1556, opened 2025-09-29
fix: store cache from `clean-build`

fredrik-bakke:


.github/workflows/clean-build.yaml
```diff
@@ -33,6 +33,12 @@ jobs:
           mkdir -p temp
           make check-profile 2> temp/memory-results.txt | tee temp/benchmark-results.txt
 
+      - name: Save Agda build cache
+        uses: actions/cache/save@v4
+        with:
+          path: repo/_build
+          key: '${{ runner.os }}-check-${{ github.ref }}-${{ matrix.agda }}-'
```
VojtechStep:
```suggestion
          key: '${{ runner.os }}-check-${{ github.ref }}-${{ matrix.agda }}-${{ hashFiles('repo/src/**') }}'
```
This is necessary; cache keys need to be unique, because caches are immutable, so the action (silently?) fails if you ask it to overwrite an existing cache

VojtechStep (commented):
I looked over our current usage of caches, and I don't think we can do much better (except caching Graphviz installation in website deployment, but that's for another PR). Technically we don't need to have `github.ref` in the key, because caches have unique scope per branch (plus they inherit caches from `master`), but eh

Issue #1545, opened 2025-09-14
Split out operations and properties of signed numbers


lowasser:
I am increasingly dissatisfied with all the arithmetic and logical operations on e.g. positive rational numbers all being stuffed into `elementary-number-theory.positive-rational-numbers`.

I'm convinced by the theory of splitting things out further, and ought to do it.

fredrik-bakke:
Full support for this 👍

lowasser:
Calling this done enough.

PR #1562, opened 2025-10-03
Split out operations on positive rational numbers

lowasser:
Continued work on #1545 .


src/elementary-number-theory/multiplication-rational-numbers.lagda.md
```diff
@@ -36,9 +36,9 @@ open import foundation.interchange-law
 
 ## Idea
 
-**Multiplication** on the
-[rational numbers](elementary-number-theory.rational-numbers.md) is defined by
-extending
+{{#concept "Multiplication" WDID=Q40276 WD="multiplication" Disambiguation="of rational numbers" Agda=mul-ℚ}}
```
lowasser, line 39: Not sure if this deserves the Wikidata link, but as far as "this is the real concept of multiplication goes," the main number systems seem likely to qualify.

fredrik-bakke: I agree with your judgment, even though the page can also be read as if it is about the abstract concept of a "multiplication" binary operation.

fredrik-bakke (approved):
No objections

PR #1384, opened 2025-03-26
Multiplication of real numbers

lowasser:
Work-in-progress, not ready for review, but at least putting this here to show how the parts fit together.

Will complete #1353 .

PR #1563, opened 2025-10-03
Poset of closed intervals

lowasser:
Closed intervals in a poset (or total order) themselves form a poset.


lowasser:
Maybe in this PR, maybe in a follow-up: all the operations we've really used for the poset of closed intervals on total orders are about lattices, not just total orders -- and I don't actually see a proof anywhere that total orders are lattices.

The main reason this matters is that real numbers (at least at a given universe level) form a lattice but not a total order, and I think we're about to need intervals of reals.  Notably, once #1384 goes in, we can define differentiability of real functions on an interval, and that'll get us _properly_ started on real analysis...

lowasser:
Went ahead and did that as a draft in #1565.

lowasser:
(To be explicit, this is ready to review as-is; the lattice follow-up will follow.)

src/elementary-number-theory/intersections-closed-intervals-rational-numbers.lagda.md
```diff
@@ -0,0 +1,121 @@
...
+## Idea
+
+Two
+[closed intervals](elementary-number-theory.closed-intervals-rational-numbers.md)
+[a, b] and [c, d] of
+[rational numbers](elementary-number-theory.rational-numbers.md)
+{{#concept "intersect" Disambiguation="closed intervals of rational numbers"}}
+if `a ≤ d` and `c ≤ b`.
+
+If [a, b] and [c, d] intersect, their intersection is itself a closed interval.
```
fredrik-bakke, line 34: Just a comment, but this is one reason why one might want to include the empty interval. Then you would be able to define the intersection of any two closed intervals, and define the meet-semilattice of closed intervals.

src/order-theory/spans-closed-intervals-total-orders.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+## Idea
+
+The
+{{#concept "span" Disambiguation="of closed intervals in a total order" Agda=span-closed-interval-Total-Order}}
+of two [closed intervals](order-theory.closed-intervals-total-orders.md) in a
+[total order](order-theory.total-orders.md) is the smallest interval which
+[contains](order-theory.poset-closed-intervals-total-orders.md) both intervals,
+or in other words, their
+[least upper bound](order-theory.least-upper-bounds-posets.md) in the
+[poset](order-theory.posets.md) of the containment relation.
```
fredrik-bakke, line 31: What's the source of the terminology "span"? Another recognizable name might be the _join_ of two closed intervals.

lowasser:
[Wikipedia](https://en.wikipedia.org/wiki/Interval_(mathematics)):

> For any set X of real numbers, the interval enclosure or interval span of X is the unique interval that contains X, and does not properly contain any other interval that also contains X.

(This is also the name we used for this operation in one of my [professional open source projects](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Range.html#span(com.google.common.collect.Range)).)

src/elementary-number-theory/intersections-closed-intervals-rational-numbers.lagda.md
lowasser: That may be, but the concept I actually need for the real numbers is this one, whether or not it's a meet-semilattice; every interval needs a lower and an upper bound.

fredrik-bakke:
Great, thank you for the explanation!

PR #1570, opened 2025-10-07
Square roots of nonnegative real numbers

lowasser:
I've defined the square root and shown that its square is the original number, but not any other interesting properties.

### DIFF
--- a/src/category-theory/codensity-monads-on-precategories.lagda.md
+++ b/src/category-theory/codensity-monads-on-precategories.lagda.md
@@ -67,7 +67,7 @@ module _
   unit-codensity-monad-Precategory =
     map-inv-is-equiv
       ( KR (id-functor-Precategory D))
-      ( pr2 (id-right-extension-Precategory C D F))
+      ( id-natural-transformation-Precategory C D F)
 
   abstract
     compute-unit-codensity-monad-Precategory :
@@ -80,7 +80,9 @@ module _
           ( F)) ＝
       id-natural-transformation-Precategory C D F
     compute-unit-codensity-monad-Precategory =
-      is-section-map-inv-is-equiv (KR _) _
+      is-section-map-inv-is-equiv
+        ( KR (id-functor-Precategory D))
+        ( id-natural-transformation-Precategory C D F)
 
   mul-codensity-monad-Precategory :
     natural-transformation-Precategory D D
@@ -89,9 +91,8 @@ module _
   mul-codensity-monad-Precategory =
     map-inv-is-equiv
       ( KR (comp-functor-Precategory D D D R R))
-      ( pr2
-        ( square-right-extension-Precategory C D F
-          ( right-extension-right-kan-extension-Precategory C D D F F Rk)))
+      ( natural-transformation-square-right-extension-Precategory C D F
+        ( right-extension-right-kan-extension-Precategory C D D F F Rk))
 
   abstract
     compute-mul-codensity-monad-Precategory :
@@ -110,7 +111,9 @@ module _
           ( R)
           ( α))
     compute-mul-codensity-monad-Precategory =
-      is-section-map-inv-is-equiv (KR _) _
+      is-section-map-inv-is-equiv (KR RR)
+        ( natural-transformation-square-right-extension-Precategory C D F
+            ( right-extension-right-kan-extension-Precategory C D D F F Rk))
 ```
 
 ## Monad laws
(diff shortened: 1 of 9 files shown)


lowasser:
Thought of some simplifications in my sleep, but just to communicate;

My plan is to do square roots, because those are only defined for reals on nonnegative reals, then define all odd roots generally (because their domain is the full set of reals), then factor out powers of two to define arbitrary roots on nonnegative reals.

src/elementary-number-theory/minimum-rational-numbers.lagda.md
```diff
@@ -120,6 +121,14 @@ abstract
       ( z<y)
 ```
 
+### If `z` is less or equal to than both `x` and `y`, it is less than or equal to their minimum
```
fredrik-bakke:
```suggestion
### If `z` is less than or equal to both `x` and `y`, it is less than or equal to their minimum
```

src/elementary-number-theory/positive-and-negative-rational-numbers.lagda.md
```diff
@@ -112,6 +129,39 @@ abstract
       ( le-zero-is-positive-ℚ q pos-q)
 ```
 
+#### A negative rational number is less than a positive rational number
```
fredrik-bakke:
```suggestion
#### Negative rational numbers are less than all positive rational numbers
```

src/elementary-number-theory/positive-and-negative-rational-numbers.lagda.md
```diff
@@ -112,6 +129,39 @@ abstract
...
+    transitive-le-ℚ p zero-ℚ q
+      ( le-zero-is-positive-ℚ q pos-q)
+      ( le-zero-is-negative-ℚ p neg-p)
+
+  leq-negative-positive-ℚ :
+    (p : ℚ⁻) (q : ℚ⁺) → leq-ℚ (rational-ℚ⁻ p) (rational-ℚ⁺ q)
+  leq-negative-positive-ℚ p q = leq-le-ℚ (le-negative-positive-ℚ p q)
+```
+
+### If `p ≤ q` for positive `p`, then `q` is positive
```
fredrik-bakke:
```suggestion
### If `p ≤ q` and `p` is positive, then `q` is positive
```

src/real-numbers/square-roots-nonnegative-real-numbers.lagda.md
```diff
@@ -0,0 +1,603 @@
...
+      ( inv
+        ( eq-sim-ℝ
+          ( unique-sqrt-ℝ⁰⁺
+            ( nonnegative-real-ℚ⁰⁺ q⁰⁺)
+            ( nonnegative-real-ℚ⁰⁺ p⁰⁺)
+            ( sim-eq-ℝ
+              ( mul-real-ℚ p p ∙ ap real-ℚ (ap rational-ℚ⁰⁺ p²=q))))))
+```
+
+### The square root operation preserves similarity
```
fredrik-bakke, line 588: Every operation preserves every similarity relation.

src/real-numbers/square-roots-nonnegative-real-numbers.lagda.md
```diff
@@ -0,0 +1,603 @@
...
+                  ( leq-negative-positive-ℚ
+                    ( mul-negative-positive-ℚ p⁻ q⁺)
+                    ( q⁺ *ℚ⁺ q⁺)))
+                ( leq-max-leq-both-ℚ _ _ _
+                  ( leq-negative-positive-ℚ
+                    ( mul-positive-negative-ℚ q⁺ p⁻)
+                    ( q⁺ *ℚ⁺ q⁺))
+                  ( refl-leq-ℚ _))))))
+
+  unique-sqrt-ℝ⁰⁺ : {l1 l2 : Level} → (x : ℝ⁰⁺ l1) (y : ℝ⁰⁺ l2) →
```
fredrik-bakke, line 562: It should follow from this that `sqrt-ℝ⁰⁺` is a (nontrivial) autoequivalence and a left inverse to squaring, which is another useful fact to record

lowasser: First I have to show that squares are nonnegative, and other related properties to the effects of ordering on multiplication that are turning out to be less trivial than I'd like.  When I can do that, I'll come back and show that.

fredrik-bakke: Ah okay no pressure

PR #1500, opened 2025-08-27
Organize repository files

fredrik-bakke:
Tidies and organizes the repository files a bit better. Also includes some fixes I discovered along the way

Fixes
- Header in Contributors file is now top-level
- Disable git metadata content for `STATEMENT-OF-INCLUSION.md`
- Remove the "generate main index" pre-commit hook. It should never fail on the user-side, and the description was wrong.

PR #1576, opened 2025-10-08
fix `check-profile`

fredrik-bakke:
#1500 broke the profiling CI, this fixes it I think

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,8 @@ MDBOOK_SRC := $(MDBOOK_DIR)/src
 # Base directory where Agda interface files are stored
 AGDA_BUILD := _build
 # Agda profiling directory
-AGDA_PROFILING_TEMP := temp
-AGDA_PROFILING_OUTPUT := $(AGDA_PROFILING_TEMP)/typecheck_output.txt
+TEMP_DIR := temp
+AGDA_PROFILING_OUTPUT := $(TEMP_DIR)/typecheck_output.txt
 # Docs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -100,10 +100,9 @@ check: ./$(SOURCE_DIR)/everything.lagda.md
 	${TIME} ${AGDA} $?
 
 .PHONY: check-profile
-# `clean` is specified second so that the $< variable stores the everything file.
-# We don't mind, because the `clean` target busts the typechecking and website cache,
-# but doesn't touch the everything file.
-check-profile: ./$(SOURCE_DIR)/everything.lagda.md clean
+check-profile: $(SOURCE_DIR)/everything.lagda.md
+	@# Remove cached data
+	@rm -Rf ./$(AGDA_BUILD)/ ./$(AGDA_PROFILING_TEMP)/
```
VojtechStep: The CI creates the `temp` directory, runs `make check-profile` with its output redirected to `temp/memory-results.txt` and `temp/benchmark-results.txt`, and the first thing `make` does is delete `temp`, so this doesn't fix the problem.

I think the proper fix is for CI to use a different directory than `temp`, or rather that the Makefile should not use `temp` as a de facto reserved name

fredrik-bakke: right, the profiling step doesn't itself actually store anything in the temp directory. I'll just remove the deletion step all together

fredrik-bakke: except for the deletion of agda cache of course

PR #1580, opened 2025-10-09
Retractive types and pointed type duality

fredrik-bakke:
Defines retractive types and proves pointed type duality: families of pointed types over `X` are equivalent to retractive types under `X`.


fredrik-bakke:
This PR has now been waiting for 3 weeks, so I'm invoking the self-review policy.

PR #1586, opened 2025-10-11
Ring of real numbers

lowasser:
Just at universe level 0 for now; still working up to large rings.


fredrik-bakke:
FYI. I've noticed that you usually write proofs using the additive unit element at universe level zero. The large structure of a ring should probably have an additive unit element at each universe level so it is still possible to extract the ring at universe level `l`. I don't know if there are some tricks you can employ to avoid having to repeat proofs with respect to zeros at each universe level.

lowasser:
My best attempt to deal with this is in #1587 ; we can postpone this one until afterwards.

lowasser:
Marking this as a draft; I think we'll probably go ahead and build all the way to large fields here, with fields needing an additional apartness relation to describe nonzero/invertible elements.

PR #1572, opened 2025-10-07
Large similarity relations

lowasser:
Working towards definitions of large groups and rings, since they're more relevant now...

PR #1587, opened 2025-10-11
Large monoids

lowasser:
It's not needed for the _internal_ laws, but the large similarity relation and raising of universe levels seems necessary both to match the models of the prototypical large structures (the real numbers, but I suppose unions and intersections of subtypes are both monoids as well), to have useful operations available like natural number powers, and to get small monoids from large ones.

Related to #1334 , as that started to grapple with what large algebraic structures would look like.

### DIFF
--- a/src/foundation/intersections-subtypes.lagda.md
+++ b/src/foundation/intersections-subtypes.lagda.md
@@ -26,9 +26,6 @@ open import foundation.universe-levels
 open import foundation-core.decidable-propositions
 open import foundation-core.propositions
 
-open import group-theory.large-monoids
-open import group-theory.large-semigroups
-
 open import order-theory.greatest-lower-bounds-large-posets
 ```
 
@@ -183,32 +180,4 @@ abstract
   preserves-sim-intersection-subtype _ _ (S⊆T , T⊆S) _ _ (U⊆V , V⊆U) =
     ( ( λ x → map-product (S⊆T x) (U⊆V x)) ,
       ( λ x → map-product (T⊆S x) (V⊆U x)))
-```
-
-### The large monoid of subtypes under intersection
-
-```agda
-module _
-  {l : Level} (X : UU l)
-  where
-
-  large-semigroup-intersection-subtype : Large-Semigroup (λ l2 → l ⊔ lsuc l2)
-  large-semigroup-intersection-subtype =
-    make-Large-Semigroup
-      ( powerset-Set X)
-      ( intersection-subtype)
-      ( associative-intersection-subtype)
-
-  large-monoid-intersection-subtype :
-    Large-Monoid (λ l2 → l ⊔ lsuc l2) (λ l1 l2 → l ⊔ l1 ⊔ l2)
-  large-monoid-intersection-subtype =
-    make-Large-Monoid
-      ( large-semigroup-intersection-subtype)
-      ( large-similarity-relation-sim-subtype X)
-      ( λ l → raise-subtype l)
-      ( sim-raise-subtype)
-      ( preserves-sim-intersection-subtype)
-      ( full-subtype lzero X)
-      ( left-unit-law-intersection-subtype)
-      ( right-unit-law-intersection-subtype)
 ```
--- a/src/foundation/unions-subtypes.lagda.md
+++ b/src/foundation/unions-subtypes.lagda.md
@@ -21,9 +21,6 @@ open import foundation.raising-universe-levels
 open import foundation.subtypes
 open import foundation.universe-levels
 
-open import group-theory.large-monoids
-open import group-theory.large-semigroups
-
 open import logic.de-morgan-propositions
 open import logic.de-morgan-subtypes
 open import logic.double-negation-stable-subtypes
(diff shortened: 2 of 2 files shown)


lowasser:
Only a draft because of the dependency on #1572 .

src/group-theory/large-monoids.lagda.md
```diff
@@ -0,0 +1,242 @@
...
+
+  eq-sim-Large-Monoid :
+    {l : Level} (x y : type-Large-Monoid l) → sim-Large-Monoid x y → x ＝ y
+  eq-sim-Large-Monoid =
+    eq-sim-Large-Similarity-Relation large-similarity-relation-Large-Monoid
+
+  field
+    raise-Large-Monoid :
+      {l1 : Level} (l2 : Level) → type-Large-Monoid l1 →
+      type-Large-Monoid (l1 ⊔ l2)
```
fredrik-bakke, line 70: The raise operation should also be expected to be a monoid monomorphism. However, I'm not convinced that having a raise operation is the way to go. An alternative, seemingly simpler way would be to ask for a _family_ of unit elements at each universe level that are all similar, and then omit the raise operation. What's your opinion on that?

fredrik-bakke: oh, I suppose that is exactly what you get with the `sim-raise-Large-Monoid` field, hmm...

fredrik-bakke: is it provable that `raise-Large-Monoid` is a monoid homomorphism from `monoid-Large-Monoid l1` to  `monoid-Large-Monoid (l1 ⊔ l2)`?

fredrik-bakke: Also, your solution does resolve the issue I raised earlier where raising from `(l1 ⊔ l2)` to `l2` should be the identity function, so good work on that.

fredrik-bakke: > is it provable that `raise-Large-Monoid` is a monoid homomorphism from `monoid-Large-Monoid l1` to `monoid-Large-Monoid (l1 ⊔ l2)`?

Yep, I've convinced myself that this works, so I'm all for this definition. Would you mind adding that the raise operation is a monoid monomorphism? Otherwise this PR is impeccable.

lowasser:
I'm actually going to make a change: the similarity relation can be characterized in terms of the raise operation, so we don't need both.

fredrik-bakke:
It's not in general a bad thing to have redundant information in a definition. You should consider whether the similarity relation that is generated by the raise operation is the one you want to use in applications. For instance, if you define the large monoid of real numbers, you probably don't want to use the similarity operation that follows from the existence of your raise operation, but instead you want to use `sim-R`, so it is useful if you can insert that as part of your definition, such as it does currently.

lowasser:
All right, I'm convinced.

fredrik-bakke:
Oh, I just remembered, here's another important property that similarity relations should satisfy, and it generalizes antisymmetry: It should be true for every `x : X l1` and every `l2` that `is-prop (Σ (y : X l2) (x ~ y))`. If this is not provable with the current definition, then it should be added. Maybe it is also possible to go the other way around and recover a raise operation from this property? See `foundation-core.small-types`, in particular `is-prop-is-small` for something similar.

fredrik-bakke:
I guess `raise + raise-sim` precisely states that we have `Σ (y : X (l1 ⊔ l2)) (x ~ y)` for all `l2`, but it doesn't say anything about uniqueness for general `l2`.

lowasser:
It's provable, and I've proved it, but the presence of the raise operation is required to show that every universe level is inhabited at all; you can't extract the raise operation from similarity.

src/group-theory/large-monoids.lagda.md
```diff
@@ -152,6 +121,99 @@ open Large-Monoid public
...
+    eq-sim-Large-Similarity-Relation (large-similarity-relation-Large-Monoid M)
+
+  sim-raise-Large-Monoid' :
+    {l1 : Level} (l2 : Level) (x : type-Large-Monoid M l1) →
+    sim-Large-Monoid M (raise-Large-Monoid M l2 x) x
+  sim-raise-Large-Monoid' l2 x =
+    symmetric-sim-Large-Monoid _ _ (sim-raise-Large-Monoid M l2 x)
+```
+
+### Raising an element of a large monoid by two universe levels is equivalent to raising it by the lub of the universe levels
```
fredrik-bakke:
```suggestion
### Raising an element of a large monoid by two universe levels is equivalent to raising it by the least upper bound of the universe levels
```

lowasser:
I had one more thing to add, so it might need re-approval.

src/foundation/intersections-subtypes.lagda.md
```diff
@@ -106,3 +116,99 @@ module _
...
+    make-Large-Monoid
+      ( large-semigroup-intersection-subtype)
+      ( large-similarity-relation-sim-subtype X)
+      ( λ l → raise-subtype l)
+      ( sim-raise-subtype)
+      ( preserves-sim-intersection-subtype)
+      ( full-subtype lzero X)
+      ( left-unit-law-intersection-subtype)
+      ( right-unit-law-intersection-subtype)
+```
```
fredrik-bakke: These fit better in `foundation.large-locale-of-subtypes`. Potentially you could show that every large meet semilattice has an underlying large semigroup

src/foundation/unions-subtypes.lagda.md
```diff
@@ -101,3 +108,105 @@ module _
...
+    make-Large-Monoid
+      ( large-semigroup-union-subtype)
+      ( large-similarity-relation-sim-subtype X)
+      ( λ l2 → raise-subtype l2)
+      ( sim-raise-subtype)
+      ( preserves-sim-union-subtype)
+      ( empty-subtype lzero X)
+      ( left-unit-law-union-subtype)
+      ( right-unit-law-union-subtype)
+```
```
fredrik-bakke: also fits better in `large-locale-of-subtypes`

src/foundation/intersections-subtypes.lagda.md
lowasser: I'll save that for another PR, then.

PR #1588, opened 2025-10-12
Large groups

lowasser:


src/foundation/disjunction.lagda.md
```diff
@@ -349,6 +349,32 @@ module _
   swap-disjunction = map-trunc-Prop (map-commutative-coproduct A B)
 ```
 
+### Associativity of the disjunction
```
fredrik-bakke, line 352: also please record `associative-disjunction : disjunction-type A (disjunction-type B C) ↔ disjunction-type (disjunction-type A B) C`

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,743 @@
...
+```agda
+module _
+  {α : Level → Level} {β : Level → Level → Level} (G : Large-Group α β)
+  where
+
+  open similarity-reasoning-Large-Group G
+
+  private
+    _*_ = mul-Large-Group G
+    neg = inv-Large-Group G
```
fredrik-bakke: If I understood correctly, let bindings are safer than `private` declarations

```suggestion
  {α : Level → Level} {β : Level → Level → Level} (G : Large-Group α β)
  (let _*_ = mul-Large-Group G) (let neg = inv-Large-Group G)
  where

  open similarity-reasoning-Large-Group G
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,743 @@
...
+  where
+
+  abstract
+    is-prop-map-left-mul-Large-Group :
+      is-prop-map (mul-Large-Group G {l2 = l2} x)
+    is-prop-map-left-mul-Large-Group y =
+      let
+        open similarity-reasoning-Large-Group G
+        _*_ = mul-Large-Group G
+        neg = inv-Large-Group G
```
fredrik-bakke: the multiplicative inverse is not the "negative", so your choice of short hand is not consistent

src/group-theory/integer-powers-of-elements-large-groups.lagda.md
```diff
@@ -0,0 +1,562 @@
...
+module _
+  {α : Level → Level} {β : Level → Level → Level} (G : Large-Group α β)
+  where
+
+  private
+    _^_ : {l : Level} → type-Large-Group G l → ℤ → type-Large-Group G l
+    x ^ k = int-power-Large-Group G k x
+
+    _*_ = mul-Large-Group G
+    mul-inv = inv-Large-Group G
```
fredrik-bakke: again, we try our best not to use `private` declarations. Feel free to define `int-power-Large-Group' G x k := int-power-Large-Group G k x`

lowasser: FWIW, I am being carefully consistent in using them exclusively in private declarations.  (And the corresponding version in small groups is [even worse](https://unimath.github.io/agda-unimath/group-theory.integer-powers-of-elements-groups.html#x%E1%B5%8F%C2%B9--x%E1%B5%8Fx%C2%B9-and-x%E1%B5%8F%C2%B9--x%C2%B9x%E1%B5%8F), using them in public signatures.

If I have to repeat them in `let`s for each declaration in this module, I will.

lowasser: Excuse me, abstract declarations.

fredrik-bakke: it's possible to do it once for the entire module, as I showed in a previous suggestion

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,740 @@
...
+      left-inverse-law-mul-Large-Group G ,
+      right-inverse-law-mul-Large-Group G)
+```
+
+### Cancellations in a large group
+
+```agda
+module _
+  {α : Level → Level} {β : Level → Level → Level} (G : Large-Group α β)
+  (let _*_ = mul-Large-Group G) (let neg = inv-Large-Group G)
```
fredrik-bakke:
```suggestion
  (let _*_ = mul-Large-Group G) (let mul-inv = inv-Large-Group G)
```
keep in mind rippling effects

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,740 @@
...
+    distributive-inv-mul-Large-Group :
+      {l1 l2 : Level} (x : type-Large-Group G l1) (y : type-Large-Group G l2) →
+      inv-Large-Group G (mul-Large-Group G x y) ＝
+      mul-Large-Group G (inv-Large-Group G y) (inv-Large-Group G x)
+    distributive-inv-mul-Large-Group {l1} {l2} x y =
+      inv
+        ( let
+            open similarity-reasoning-Large-Group G
+            _*_ = mul-Large-Group G
+            neg = inv-Large-Group G
```
fredrik-bakke: this one is still not fixed. apologies if you're just still working on it

src/group-theory/integer-powers-of-elements-large-groups.lagda.md
lowasser: Done, though I definitely don't see any advantage on any metric from doing it that way.

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,740 @@
...
+    {l1 : Level} (l2 : Level) → type-Large-Group l1 → type-Large-Group (l1 ⊔ l2)
+  raise-Large-Group = raise-Large-Monoid large-monoid-Large-Group
+
+  field
+    inv-Large-Group : {l : Level} → type-Large-Group l → type-Large-Group l
+
+    preserves-sim-inv-Large-Group :
+      {l1 l2 : Level} (x : type-Large-Group l1) (y : type-Large-Group l2) →
+      sim-Large-Group x y →
+      sim-Large-Group (inv-Large-Group x) (inv-Large-Group y)
```
fredrik-bakke, line 130: I suppose for propositionally valued relations it follows that the following square commutes,

```text
                 sim-eq
      (x = y) -----------> (x ~ y)
         |                    |
ap (_⁻¹) |                    | preserves-sim-inv-Large-Group
         v                    v
   (x ⁻¹ = y ⁻¹) -----> (x ⁻¹ ~ y ⁻¹)
                 sim-eq
```

 but this won't follow in general... Just something for the next guy to think about.

src/group-theory/integer-powers-of-elements-large-groups.lagda.md
fredrik-bakke: Yeah, no, me neither; except for with regards to my learned aversion to `private`.

src/group-theory/large-groups.lagda.md
fredrik-bakke: could you record this square, with a comment that it follows from similarity being propositionally valued? I think that would be worthwhile, in order to point out limitations of the definition.

fredrik-bakke: or efficiency of the definition, depending on your point of view

lowasser: Uh, I guess?  What do I call this, or label its section in the page?

It doesn't seem very interesting, since of course any element of the output type is equal, because, as you point out, propositions.  This is presumably just `eq-type-Prop (sim-prop-Large-Group G (inv-Large-Group G x) (inv-Large-Group G y)`.

lowasser: Also, does similarity being propositionally valued follow from [the characterization of similarity in terms of the raise operation and identity in the set of values](https://unimath.github.io/agda-unimath/group-theory.large-monoids.html#8931)?  I don't follow why similarity being propositionally valued is interesting.

lowasser: (If similarity being propositionally valued _is_ interesting, I would be more tempted to make it uninteresting by defining it in terms of the raise operation instead of letting it be its own thing.)

lowasser: I think I've convinced myself that the characterization in terms of the raise operation can't suffice to prove that similarity is a proposition, since I can imagine similarity being defined as "that plus some other random integer just for fun" and all the laws still hold.  (I was fooled by how we often prove identity is a proposition by showing equality iff some other proposition, but I guess that's specific to identity and is why you bother calling it a "fundamental theorem of identity.")

I'm not a fan of it being interesting that similarity is a proposition, but if you have suggestions for how to name the proof that that diagram commutes, I'll let it be for this PR.

fredrik-bakke: It's not that similarity being propositionally valued is interesting, but the opposite. My comment highlights what is needed to generalize to the case where similarity is instead set-valued. But, as you say, it is not interesting for your use, so I'm happy to leave this be. A simple name for this entry would've been `coh-preserves-sim-inv-Large-Group`.

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,740 @@
...
+
+    left-inverse-law-mul-Large-Group :
+      {l : Level} (x : type-Large-Group l) →
+      mul-Large-Group (inv-Large-Group x) x ＝ raise-unit-Large-Group l
+
+    right-inverse-law-mul-Large-Group :
+      {l : Level} (x : type-Large-Group l) →
+      mul-Large-Group x (inv-Large-Group x) ＝ raise-unit-Large-Group l
```
fredrik-bakke, line 138: I believe you usually prefer the formluation

```suggestion
    left-inverse-law-mul-Large-Group :
      {l : Level} (x : type-Large-Group l) →
      sim-Large-Group (mul-Large-Group (inv-Large-Group x) x) (unit-Large-Group)

    right-inverse-law-mul-Large-Group :
      {l : Level} (x : type-Large-Group l) →
      sim-Large-Group (mul-Large-Group x (inv-Large-Group x)) (unit-Large-Group)
```
IIRC this is for instance how it is formulated for real numbers. And, of course, the two laws are equivalent so I'd be inclined to prefer this more practical choice that avoids reasoning about raising universe levels.

In essence, since we are specifying a particular similarity relation in this structure, we should prefer it over identity.

fredrik-bakke: But we have to make a choice on naming convention.
Either we should write `left-inverse-law-mul-Large-Group` for the law using similarity, and then the corresponding law using equality can be called `eq-left-inverse-law-mul-Large-Group`. Or, conversely, we write `sim-left-inverse-law-mul-Large-Group` for the law phrased in terms of similarity and `left-inverse-law-mul-Large-Group` for the law using equality. If we want to prefer similarity over equality, which I think we do, then I'd propose to use the former.

fredrik-bakke: The way I see it, while the raise operation is necessary (in the sense that it is equivalent to having units at each universe level) and ensures some sort of cohesion of structures between universe levels, it is not a practical construct for reasoning about structures. It clutters arguments with needless trivial intermediate steps. Instead, we can use similarity to avoid it and phrase relations directly between elements at different universe levels.

This has an additional benefit that arguments that don't need to transfer from similarity to equality no longer depend on the function extensionality axiom or the propositional extensionality axiom (univalence for propositions) either, meaning that they in particular compute. Of course, computation doesn't matter since similarity is propositional, but it is practical for other purposes and I believe this might just be a more elegant way to do things.

fredrik-bakke: Notice that this is akin to how we use equivalence to reason about types rather than equality, even though they are equivalent notions by univalence. This allows us to compute things more easily, and avoids use of `raise`.

lowasser: I have found exactly the opposite in practice.  Equational reasoning with raise is much easier than the alternative, and I begin to regret introducing similarity to the picture at all.

lowasser: I previously had things the way you describe, and found it much more cumbersome.  Writing things this way is the product of that experience.

To elaborate if only a little, I don't want to reinvent every wheel that exists for identity.  It doesn't add value.  Transferring back and forth is painful, and we often have to start with equality.  When we can transfer results from small types, which we want to do as much as possible, they are formulated in terms of equality, and must also be formulated in terms of raise.

I think I will probably attempt to rip out similarity as a separate relation the next opportunity I have to work on this.

lowasser: Elaborating further: in addition to needing to transfer from equality to use small results, we also need to constantly transfer to equality to prove results about embeddings, equivalences, homomorphisms, and all those other tools that are not worth reinventing again.  Having a potentially unique notion of similarity for each large monoid makes it still more difficult to generalize.

Using identity everywhere is just much simpler even when we need to track universe levels in the elements being identified.

lowasser: I think I'm going to pursue that in a follow up PR, leave the naming (and primary definition) as is for now, but add the similarity version in addition.

fredrik-bakke: I see. I'm surprised by the rebuttal, but I see where you are coming from. This is why some people prefer to work with [DURGs](https://philpapers.org/rec/SCHHGV-2), meaning you equip every type with a similarity relation up front and then define embeddings and all those things in terms of the specified similarity relation. Of course, that would be a huge uprooting so I'm not suggesting you do that, I'm just giving some context.

Then the question really boils down to _when_ do you want to translate from similarity to equality. I still find the reasons I laid out compelling, but I'll not push further.

Looking forward to seeing what you do next!

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    sim-Large-Group G x (raise-Large-Group G l2 x)
+  sim-raise-Large-Group = sim-raise-Large-Monoid (large-monoid-Large-Group G)
+
+  sim-raise-Large-Group' :
+    {l1 : Level} (l2 : Level) (x : type-Large-Group G l1) →
+    sim-Large-Group G (raise-Large-Group G l2 x) x
+  sim-raise-Large-Group' = sim-raise-Large-Monoid' (large-monoid-Large-Group G)
+
+  raise-raise-Large-Group :
+    {l1 l2 l3 : Level} → (x : type-Large-Group G l1) →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 : Level} (x : type-Large-Group G l1) →
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    raise-Large-Group G l2 (raise-Large-Group G l3 x) ＝
+    raise-Large-Group G (l2 ⊔ l3) x
+  raise-raise-Large-Group =
+    raise-raise-Large-Monoid (large-monoid-Large-Group G)
+
+  raise-left-mul-Large-Group :
+    {l1 l2 l3 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    (x : type-Large-Group G l1) (y : type-Large-Group G l2) →
+    mul-Large-Group G (raise-Large-Group G l3 x) y ＝
+    raise-Large-Group G l3 (mul-Large-Group G x y)
+  raise-left-mul-Large-Group =
+    raise-left-mul-Large-Monoid (large-monoid-Large-Group G)
+
+  raise-right-mul-Large-Group :
+    {l1 l2 l3 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    (x : type-Large-Group G l1) (y : type-Large-Group G l2) →
+    mul-Large-Group G x (raise-Large-Group G l3 y) ＝
+    raise-Large-Group G l3 (mul-Large-Group G x y)
+  raise-right-mul-Large-Group =
+    raise-right-mul-Large-Monoid (large-monoid-Large-Group G)
+
+  raise-mul-Large-Group :
+    {l1 l2 l3 l4 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 l4 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+
+### The raise operation characterizes the similarity relation
+
+```agda
+module _
+  {α : Level → Level} {β : Level → Level → Level} (G : Large-Group α β)
+  where
+
+  sim-iff-eq-raise-Large-Group :
+    {l1 l2 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    (x : type-Large-Group G l1) (y : type-Large-Group G l2) →
+    ( sim-Large-Group G x y) ↔
+    ( raise-Large-Group G l2 x ＝ raise-Large-Group G l1 y)
+  sim-iff-eq-raise-Large-Group =
+    sim-iff-eq-raise-Large-Monoid (large-monoid-Large-Group G)
+
+  sim-eq-raise-Large-Group :
+    {l1 l2 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    (x : type-Large-Group G l1) (y : type-Large-Group G l2) →
+    (raise-Large-Group G l2 x ＝ raise-Large-Group G l1 y) →
+    sim-Large-Group G x y
+  sim-eq-raise-Large-Group x y =
+    backward-implication (sim-iff-eq-raise-Large-Group x y)
+
+  eq-raise-sim-Large-Group :
+    {l1 l2 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    type-Large-Group (l1 ⊔ l2)
+  mul-Large-Group = mul-Large-Monoid large-monoid-Large-Group
+
+  mul-Large-Group' :
+    {l1 l2 : Level} → type-Large-Group l1 → type-Large-Group l2 →
+    type-Large-Group (l1 ⊔ l2)
+  mul-Large-Group' x y = mul-Large-Group y x
+
+  ap-mul-Large-Group :
+    {l1 l2 : Level} →
```
fredrik-bakke:
```suggestion
    {l1 l2 : Level}
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+  eq-sim-Large-Group = eq-sim-Large-Monoid large-monoid-Large-Group
+
+  symmetric-sim-Large-Group :
+    {l1 l2 : Level} (x : type-Large-Group l1) (y : type-Large-Group l2) →
+    sim-Large-Group x y → sim-Large-Group y x
+  symmetric-sim-Large-Group =
+    symmetric-sim-Large-Monoid large-monoid-Large-Group
+
+  preserves-sim-left-mul-Large-Group :
+    {l1 l2 l3 : Level} (y : type-Large-Group l1) →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 : Level} (y : type-Large-Group l1)
```

src/group-theory/large-groups.lagda.md
```diff
@@ -0,0 +1,777 @@
...
+    (x : type-Large-Group l2) (x' : type-Large-Group l3) →
+    sim-Large-Group x x' →
+    sim-Large-Group (mul-Large-Group x y) (mul-Large-Group x' y)
+  preserves-sim-left-mul-Large-Group =
+    preserves-sim-left-mul-Large-Monoid large-monoid-Large-Group
+
+  preserves-sim-right-mul-Large-Group :
+    {l1 l2 l3 : Level} (x : type-Large-Group l1) →
```
fredrik-bakke:
```suggestion
    {l1 l2 l3 : Level} (x : type-Large-Group l1)
```

PR #1591, opened 2025-10-14
Complemented inequality of cardinalities

fredrik-bakke:
Introduces another inequality relation on cardinalities that is antisymmetric assuming WLPO, and coincides with the usual inequality assuming LEM. Also does some cleanup for the cardinality formalizations.

~The main results of this PR are currently blocked on #1206 which is in turn _still_ blocked on #1264. However, it can still be merged, as those results are not type checked in the present version.~


fredrik-bakke:
Since #1206 is now merged I can ~formalize~ *typecheck the main result of this PR, hence I'll mark it as a draft until that is done.

fredrik-bakke:
Okay done :)

src/set-theory/complemented-inequality-cardinals.lagda.md
```diff
@@ -0,0 +1,267 @@
...
+open import set-theory.cardinals
+open import set-theory.equality-cardinals
+```
+
+</details>
+
+## Idea
+
+We may say a [cardinal](set-theory.cardinals.md) `X` is
+{{#concept "(complemented) less than or equal to" Agda=leq-complemented-Cardinal}}
```
lowasser, line 37: What about this definition makes the word "complemented" relevant or appropriate?  Or is this a standard term in literature?

fredrik-bakke: The term "complemented" is used in the topological sense. In topological models of type theory, a subtype $X \subseteq Y$ is complemented if and only if it is a closed subspace such that there exists another closed subspace $\overline{X} \subseteq Y$ with  $X \cap \overline{X} = \emptyset$ and $X \cup \overline{X} = Y$.

fredrik-bakke: Of course, in boolean models every subtype satisfies this property since there every object has trivial topology, but in general a type that is a set need not have a trivial topology in a model.

fredrik-bakke: The term is used in the TypeTopology library and also in David Wärn's "path spaces of pushouts" paper. I'd venture to bet it also appears elsewhere in the literature, but I can't think of more examples at the moment.

It is appropriate to apply this term in the setting at hand. The inequality that is defined in this file states that a cardinal $\alpha$ is less than a cardinal $\beta$ if every set in $\alpha$'s isomorphism class can be embedded as a complemented subset into any set of $\beta$'s isomorphism class.

Issue #1611, opened 2025-10-18
Distributivity of absolute value over multiplication in the Dedekind reals


lowasser:
Just putting this out into space because I'm having an incredibly hard time with it.

This would be easy if inequality on the reals were linear, but it isn't in general; that implies one or another omniscience principle.

I am not sure, but I'm not convinced, that arbitrary order preserving maps distribute over the binary maximum or minimum on the reals, as they do in total orders.  I just spent an hour trying to prove that unsuccessfully.  I know it's not true in general for meet/join semilattices, but it might be true for the reals even so.

The main result I want to get out of this is that multiplication is uniformly continuous in each variable on the reals.

fredrik-bakke:
This is a shot in the dark, but can you use that $\left\lvert {x y}\right\rvert = \sqrt{{(xy)}^2}=\sqrt{x^2y^2}=\sqrt{x^2}\sqrt{y^2} = \left\lvert {x}\right\rvert\left\lvert {y}\right\rvert$?

fredrik-bakke:
Oh, I suppose this suffers from the same issue.

lowasser:
That sounds plausible.  I'll try it.

lowasser:
(That is, we still have to prove that the square root of x squares is the absolute value of x, but that should at least be easier to prove directly than the general result for multiplication.)

fredrik-bakke:
Indeed. Otherwise, if this is impossible, it might be reasonable to name it as an alternative definition of the absolute value function.

PR #1616, opened 2025-10-21
The absolute value distributes over multiplication in the reals

lowasser:
As suggested in #1611 , going through the square root of the square was easier.  I'm pretty sure this could be proved directly, but I have no complaints about this approach.

### DIFF
--- a/scripts/preprocessor_concepts.py
+++ b/scripts/preprocessor_concepts.py
@@ -56,9 +56,10 @@ def make_definition_regex(definition):
 
 
 def make_loose_definition_regex(definition):
+    escaped = re.escape(definition)
     return re.compile(
-        definition + r'\s+:(?=[\s({])|' +
-        r'(?:data|record|infix[lr]?(?:\s+\d+)?)\s+' + definition
+        escaped + r'\s+:(?=[\s({])|' +
+        r'(?:data|record|infix[lr]?(?:\s+\d+)?)\s+' + escaped
     )
 
 
--- a/src/complex-numbers/eisenstein-integers.lagda.md
+++ b/src/complex-numbers/eisenstein-integers.lagda.md
@@ -32,9 +32,10 @@ open import ring-theory.rings
 
 ## Idea
 
-The {{#concept "Eisenstein integers" WDID=Q262370 WD="Eisenstein integer"}} are
-the [complex numbers](complex-numbers.complex-numbers.md) of the form `a + bω`,
-where `ω = -½ + ½√3i`, and where `a` and `b` are
+The
+{{#concept "Eisenstein integers" WDID=Q262370 WD="Eisenstein integer" Agda=ℤ[ω]}}
+are the [complex numbers](complex-numbers.complex-numbers.md) of the form
+`a + bω`, where `ω = -½ + ½√3i`, and where `a` and `b` are
 [integers](elementary-number-theory.integers.md). Note that `ω` is a solution to
 the equation `ω² + ω + 1 = 0`.
 
--- a/src/complex-numbers/gaussian-integers.lagda.md
+++ b/src/complex-numbers/gaussian-integers.lagda.md
@@ -33,9 +33,11 @@ open import ring-theory.rings
 
 ## Idea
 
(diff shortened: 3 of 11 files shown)


src/real-numbers/strict-inequality-nonnegative-real-numbers.lagda.md
```diff
@@ -0,0 +1,39 @@
...
+[standard strict ordering](real-numbers.strict-inequality-real-numbers.md) on
+[real numbers](real-numbers.dedekind-real-numbers.md).
+
+## Definition
+
+```agda
+le-prop-ℝ⁰⁺ : {l1 l2 : Level} → ℝ⁰⁺ l1 → ℝ⁰⁺ l2 → Prop (l1 ⊔ l2)
+le-prop-ℝ⁰⁺ x y = le-prop-ℝ (real-ℝ⁰⁺ x) (real-ℝ⁰⁺ y)
+
+le-ℝ⁰⁺ : {l1 l2 : Level} → ℝ⁰⁺ l1 → ℝ⁰⁺ l2 → UU (l1 ⊔ l2)
```
fredrik-bakke, line 45: Don't forget to add `is-prop-le`

fredrik-bakke (approved):
Nice, I'm glad it worked :)

src/real-numbers/absolute-value-real-numbers.lagda.md
```diff
@@ -244,3 +252,69 @@ module _
...
+        ( leq-abs-sqrt-square-ℝ)
+        ( leq-leq'-ℝ
+          ( real-sqrt-ℝ⁰⁺ (nonnegative-square-ℝ x))
+          ( abs-ℝ x)
+          ( leq-abs-sqrt-square-ℝ'))
+```
+
+### Absolute values distribute over multiplication
+
+```agda
```
fredrik-bakke, line 302: It would be nice if you added a comment which states that the proof goes through the characterization as the square root of the square, but that a direct proof should also be possible. It's up to you if you wish to add this or not, though.

src/real-numbers/absolute-value-real-numbers.lagda.md
```diff
@@ -244,3 +252,69 @@ module _
...
 ```
+
+### The absolute value of `x` is the square root of `x²`
+
+```agda
+module _
+  {l : Level} (x : ℝ l)
+  where
+
+  opaque
```
VojtechStep: Nit: these should all be rather `abstract opaque`, right?

fredrik-bakke: Off-topic: it's kinda silly that one has to specify both, though

lowasser: I have not considered it worth bothering.  `opaque` is like "abstract, but you can break the glass if you need to," and if nobody ever breaks the glass, there's no need to worry about the difference.

fredrik-bakke: I think the signaling is clearer with `abstract opaque`. Then it irrefutably states that the glass is not to be broken.

VojtechStep: I asked on the Agda Zulip and apparently there are places where `abstract` unfolds and `opaque` doesn't — namely `abstract` definitions are unfolded in sibling/children `abstract` blocks, `opaque` definitions aren't, so in the following code
```agda
open import Agda.Builtin.Equality
data Foo : Set where mkFoo : Foo

module X where
  abstract
    foo : Foo
    foo = mkFoo
  abstract
    is-foo : foo ≡ mkFoo
    is-foo = refl

module Y where
  abstract opaque
    foo : Foo
    foo = mkFoo
  abstract opaque
    is-foo : foo ≡ mkFoo
    is-foo = {!refl!}
```
Agda doesn't accept `refl` for `Y.is-foo`

fredrik-bakke: petition for a new keyword `impenetrable`? (*jk)

lowasser: I had not previously realized that you can put `abstract opaque` on the same line, which makes me much more willing to use it.

PR #1507, opened 2025-08-30
Metric pseudocompletion of pseudometric spaces

malarbol:
(WIP)

- pseudometric completions of a pseudometric space`M`: the pseudometric space of Cauchy approximations in `M` where two Cauchy approximations `x` and `y` are in a `d`-neighborhood of one another if for all `δ ε : ℚ⁺`, `x δ` and `y ε` are in a (`δ + ε + d`)-neighborhood of one another in `M`;

- metric quotients of a pseudometric space `M`: the metric space whose points are quotient classes of `M` by the similarity relation and neighborhoods given by neighborhoods of inhabitants of the quotient classes: two quotient classes `X`, `Y` are in a `d`-neighborhood if for all `x ∈ X` and `y ∈ Y`, `x` and `y` are `d`-neighbors in the pseudometric space.

- [...]

Co-authored-by: Louis Wasserman [wasserman.louis@gmail.com](mailto:wasserman.louis@gmail.com)

PR #1619, opened 2025-10-21
Cauchy pseudocompletions of (pseudo)metric spaces

malarbol:

- the **Cauchy pseudocompletion**  of a pseudometric space `M`: the pseudometric space of Cauchy approximations in `M` where two Cauchy approximations `x` and `y` are in a `d`-neighborhood of one another if for all `δ ε : ℚ⁺`, `x δ` and `y ε` are in a `δ + ε + d`-neighborhood of one another in `M`;

- the **Cauchy pseudocompletion**  of a metric space: the Cauchy pseudocompletion of its underlying pseudometric space.

Cauchy approximations in a Cauchy pseudocompletion have a limit; any complete metric space is a retract of its Cauchy pseudocompletion.

Co-authored-by: Louis Wasserman [wasserman.louis@gmail.com](mailto:wasserman.louis@gmail.com)

### DIFF
--- a/src/foundation-core.lagda.md
+++ b/src/foundation-core.lagda.md
@@ -21,6 +21,7 @@ open import foundation-core.decidable-propositions public
 open import foundation-core.dependent-identifications public
 open import foundation-core.diagonal-maps-cartesian-products-of-types public
 open import foundation-core.discrete-types public
+open import foundation-core.double-negation-stable-equality public
 open import foundation-core.embeddings public
 open import foundation-core.empty-types public
 open import foundation-core.endomorphisms public
--- a/src/foundation-core/double-negation-stable-equality.lagda.md
+++ b/src/foundation-core/double-negation-stable-equality.lagda.md
@@ -1,0 +1,308 @@
+# Double negation stable equality
+
+```agda
+module foundation-core.double-negation-stable-equality where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.action-on-identifications-dependent-functions
+open import foundation.action-on-identifications-functions
+open import foundation.coproduct-types
+open import foundation.dependent-pair-types
+open import foundation.double-negation
+open import foundation.equality-cartesian-product-types
+open import foundation.equality-dependent-pair-types
+open import foundation.equivalences
+open import foundation.injective-maps
+open import foundation.irrefutable-equality
+open import foundation.negation
+open import foundation.reflecting-maps-equivalence-relations
+open import foundation.sets
+open import foundation.unit-type
+open import foundation.universe-levels
+
+open import foundation-core.cartesian-product-types
+open import foundation-core.empty-types
+open import foundation-core.identity-types
+open import foundation-core.propositions
+open import foundation-core.retractions
+open import foundation-core.retracts-of-types
+
+open import logic.double-negation-elimination
+```
+
+</details>
+
+## Idea
+
+A type `A` is said to have
+{{#concept "double negation stable equality" Disambiguation="type" Agda=has-double-negation-stable-equality}}
+if `x ＝ y` has
+[double negation elimination](logic.double-negation-elimination.md) for every
+`x y : A`. By the
+[fundamental theorem of identity types](foundation.fundamental-theorem-of-identity-types.md),
+types with double negation stable equality are [sets](foundation-core.sets.md).
+
+## Definitions
+
+```agda
+has-based-double-negation-stable-equality : {l : Level} (A : UU l) → A → UU l
+has-based-double-negation-stable-equality A x =
+  (y : A) → has-double-negation-elim (x ＝ y)
+
+has-based-double-negation-stable-equality' : {l : Level} (A : UU l) → A → UU l
+has-based-double-negation-stable-equality' A x =
+  (y : A) → has-double-negation-elim (y ＝ x)
+
+has-double-negation-stable-equality : {l : Level} → UU l → UU l
+has-double-negation-stable-equality A =
+  (x : A) → has-based-double-negation-stable-equality A x
+```
+
+## Examples
+
+### Propositions have double negation stable equality
+
+```agda
+abstract
+  has-double-negation-stable-equality-is-prop :
+    {l1 : Level} {A : UU l1} → is-prop A → has-double-negation-stable-equality A
+  has-double-negation-stable-equality-is-prop H x y =
+    double-negation-elim-is-contr (H x y)
+```
+
+### The empty type has double negation stable equality
+
+```agda
+has-double-negation-stable-equality-empty :
+  has-double-negation-stable-equality empty
+has-double-negation-stable-equality-empty ()
+```
+
+### The unit type has double negation stable equality
+
+```agda
+has-double-negation-stable-equality-unit :
+  has-double-negation-stable-equality unit
+has-double-negation-stable-equality-unit _ _ _ = refl
+```
+
+## Properties
+
+### Types with double negation stable equality are sets
+
+```agda
+module _
+  {l : Level} {A : UU l}
+  where
+
+  is-prop-based-Id-has-based-double-negation-stable-equality :
+    {x : A} →
+    has-based-double-negation-stable-equality A x → (y : A) → is-prop (x ＝ y)
+  is-prop-based-Id-has-based-double-negation-stable-equality {x} =
+    is-prop-based-Id-prop-in-based-id x
+      ( λ y → ¬¬ (x ＝ y))
+      ( λ y → is-prop-neg)
+      ( intro-double-negation refl)
+
+  is-set-has-double-negation-stable-equality :
+    has-double-negation-stable-equality A → is-set A
+  is-set-has-double-negation-stable-equality H x =
+    is-prop-based-Id-has-based-double-negation-stable-equality (H x)
+```
+
+### Types with double negation stable equality are closed under injections
+
+```agda
+abstract
+  has-double-negation-stable-equality-injection :
+    {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+    injection A B →
+    has-double-negation-stable-equality B →
+    has-double-negation-stable-equality A
+  has-double-negation-stable-equality-injection (f , H) d x y =
+    has-double-negation-elim-iff (ap f , H) (d (f x) (f y))
+```
+
+### Types with double negation stable equality are closed under retracts
+
+```agda
+abstract
+  has-double-negation-stable-equality-retract-of :
+    {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+    A retract-of B →
+    has-double-negation-stable-equality B →
+    has-double-negation-stable-equality A
+  has-double-negation-stable-equality-retract-of (i , r , R) =
+    has-double-negation-stable-equality-injection
+      ( i , is-injective-has-retraction i r R)
+```
+
+### Types with double negation stable equality are closed under equivalences
+
+```agda
+abstract
+  has-double-negation-stable-equality-equiv :
+    {l1 l2 : Level} {A : UU l1} {B : UU l2} (e : A ≃ B) →
+    has-double-negation-stable-equality B →
+    has-double-negation-stable-equality A
+  has-double-negation-stable-equality-equiv e =
+    has-double-negation-stable-equality-retract-of (retract-equiv e)
+
+abstract
+  has-double-negation-stable-equality-equiv' :
+    {l1 l2 : Level} {A : UU l1} {B : UU l2} (e : A ≃ B) →
+    has-double-negation-stable-equality A →
+    has-double-negation-stable-equality B
+  has-double-negation-stable-equality-equiv' e =
+    has-double-negation-stable-equality-retract-of (retract-inv-equiv e)
+```
+
+### Having double negation stable equality is a property
+
+```agda
+abstract
+  is-prop-has-based-double-negation-stable-equality :
+    {l1 : Level} {X : UU l1} (x : X) →
+    is-prop (has-based-double-negation-stable-equality X x)
+  is-prop-has-based-double-negation-stable-equality x =
+    is-prop-has-element
+      ( λ d →
+        is-prop-Π
+        ( λ y →
+          is-prop-function-type
+            ( is-prop-based-Id-has-based-double-negation-stable-equality d y)))
+
+abstract
+  is-prop-has-double-negation-stable-equality :
+    {l1 : Level} {X : UU l1} → is-prop (has-double-negation-stable-equality X)
+  is-prop-has-double-negation-stable-equality =
+    is-prop-Π is-prop-has-based-double-negation-stable-equality
+
+has-double-negation-stable-equality-Prop : {l1 : Level} → UU l1 → Prop l1
+has-double-negation-stable-equality-Prop X =
+  ( has-double-negation-stable-equality X ,
+    is-prop-has-double-negation-stable-equality)
+```
+
+### Any map into a type with double negation stable equality reflects irrefutable equality
+
+```agda
+module _
+  {l1 l2 : Level} {A : UU l1} {X : UU l2}
+  (H : has-double-negation-stable-equality X) (f : A → X)
+  where
+
+  reflects-irrefutable-eq :
+    reflects-equivalence-relation (irrefutable-eq-equivalence-relation A) f
+  reflects-irrefutable-eq {x} {y} r =
+    H (f x) (f y) (map-double-negation (ap f) r)
+
+  reflecting-map-irrefutable-eq :
+    reflecting-map-equivalence-relation
+      ( irrefutable-eq-equivalence-relation A)
+      ( X)
+  reflecting-map-irrefutable-eq = (f , reflects-irrefutable-eq)
+```
+
+### A product of types with double negation stable equality has double negation stable equality
+
+```agda
+has-double-negation-stable-equality-product :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  has-double-negation-stable-equality A →
+  has-double-negation-stable-equality B →
+  has-double-negation-stable-equality (A × B)
+has-double-negation-stable-equality-product d e x y p =
+  eq-pair
+    ( d (pr1 x) (pr1 y) (map-double-negation (ap pr1) p))
+    ( e (pr2 x) (pr2 y) (map-double-negation (ap pr2) p))
+```
+
+### Double negation stability of equality of the factors of a cartesian product
+
+If `A × B` has double negation stable equality and `B` has an element, then `A`
+has double negation stable equality; and vice versa.
+
+```agda
+has-double-negation-stable-equality-left-factor :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  has-double-negation-stable-equality (A × B) →
+  B →
+  has-double-negation-stable-equality A
+has-double-negation-stable-equality-left-factor d b x y p =
+  ap pr1 (d (x , b) (y , b) (map-double-negation (λ q → eq-pair q refl) p))
+
+has-double-negation-stable-equality-right-factor :
+  {l1 l2 : Level} {A : UU l1} {B : UU l2} →
+  has-double-negation-stable-equality (A × B) →
+  A → has-double-negation-stable-equality B
+has-double-negation-stable-equality-right-factor d a x y p =
+  ap pr2 (d (a , x) (a , y) (map-double-negation (eq-pair refl) p))
+```
+
+### If the total space has double negation stable equality, and `B` has a section, then the base type has double negation stable equality
+
+```agda
+abstract
+  has-double-negation-stable-equality-base-has-double-negation-stable-equality-Σ :
+    {l1 l2 : Level} {A : UU l1} {B : A → UU l2} (b : (x : A) → B x) →
+    has-double-negation-stable-equality (Σ A B) →
+    has-double-negation-stable-equality A
+  has-double-negation-stable-equality-base-has-double-negation-stable-equality-Σ
+    b dΣ x y nnp =
+    ap
+      ( pr1)
+      ( dΣ
+        ( x , b x)
+        ( y , b y)
+        ( map-double-negation (λ p → eq-pair-Σ p (apd b p)) nnp))
+```
+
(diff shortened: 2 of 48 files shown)


fredrik-bakke:
To me the term "pseudometric completion" seems tautological when applied to pseudometrics. "Pseudometric completion" reads like "the universal way to turn something into a pseudometric", which it evidently isn't. Wouldn't a better term for this concept be simply "The completion of pseudometric spaces", or "Cauchy completion of pseudometric spaces"?

Is it not the case that the "pseudometric completion" of a metric space is a metric space? If so, maybe a better term is "(Cauchy) pseudocompletion"?

malarbol:
Unfortunately I have no reference to give for the term "pseudometric completion".

Nonetheless it's some kind of canonical pseudometric space, isometric to the original one, where all Cauchy approximations have a limit; so it seemed appropriate.

On the other hand, we don't really have the concept of **complete pseudometric space** (limits aren't unique so this would not be a subtype of pseudometric spaces) so talking about "completion" of pseudometric spaces might be misleading. 

We could also consider the names `cauchy-neighborhood-relation-Pseudometric-Space` and `cauchy-pseudometric-space-Pseudometric-Space`.

fredrik-bakke:
Ah, I see, it freely adjoins a limit to every cauchy sequence? This is a nice construction!

I don't think "pseudometric completion" is good. The "pseudocompletion" or "Cauchy pseudocompletion" should work though, right? Unless if the construction already produces a metric space when applied to metric spaces. If so, you might as well call it the "completion" with no extra qualifier, or "Cauchy completion" if you prefer.

malarbol:
> To me the term "pseudometric completion" seems tautological when applied to pseudometrics. "Pseudometric completion" reads like "the universal way to turn something into a pseudometric", which it evidently isn't. Wouldn't a better term for this concept be simply "The completion of pseudometric spaces", or "Cauchy completion of pseudometric spaces"?
> 

Yes. It's basically about constructing a natural pseudometric space, isometric to the original one, where all Cauchy approximations have a limit. I guess "Cauchy completion of pseudometric spaces" would be ok but there's no `complete-pseudometric-spaces` module so maybe it would be weird. 

> Is it not the case that the "pseudometric completion" of a metric space is a metric space? If so, maybe a better term is "(Cauchy) pseudocompletion"?

The term "(Cauchy) pseudocompletion" will appear later, namely as the metric quotient of these pseudometric Cauchy completions. There're the closest we could get to Cauchy completion but we can't prove they're complete, hence the name "pseudocompletion". 

It's all in #1507 if you want to have a look.

fredrik-bakke:
> The term "(Cauchy) pseudocompletion" will appear later, namely as the metric quotient of these pseudometric Cauchy completions. There're the closest we could get to Cauchy completion but we can't prove they're complete, hence the name "pseudocompletion".

I would advise against using the qualifier "pseudo-" for that, since you are already using that extensively for your notion of a nonextensional metric space.

fredrik-bakke:
Okay, so let's think. This PR contributes the universal construction of adjoining a limit point to every Cauchy sequence in a pseudometric space. I think "the (Cauchy) pseudocompletion" is a perfectly reasonable name for this construction, since it is the (Cauchy) completion up to similarity, which is exactly the failure of a **pseudo**metric space, contra a metric space.

malarbol:
> > The term "(Cauchy) pseudocompletion" will appear later, namely as the metric quotient of these pseudometric Cauchy completions. There're the closest we could get to Cauchy completion but we can't prove they're complete, hence the name "pseudocompletion".
> 
> I would advise against using the qualifier "pseudo-" for that, since you are already using that extensively for your notion of a nonextensional metric space.

ok. We'll have to think of a better name.

fredrik-bakke:
Now, surely we can find a more suitable name for the quotient you describe that is coming in a later contribution. Though, why does it deserve a special name if it fails to satisfy the property you are after?

malarbol:
> Though, why does it deserve a special name if it fails to satisfy the property you are after?

It satisfies almost all properties you'd expect from a Cauchy completion but not exactly (without ACC for all we know). Maybe that's not the right concept.

In the meantime this construction still seems intrinsically interesting. Should it be called `cauchy-pseudocompletion-Pseudometric-Space`?

fredrik-bakke:
> In the meantime this construction still seems intrinsically interesting. Should it be called `cauchy-pseudocompletion-Pseudometric-Space`?

The concept you are contributing in this PR? Yes, I think that is the best name.

fredrik-bakke:
> It satisfies almost all properties you'd expect from a Cauchy completion but not exactly (with our ACC for all we know).

That's a bummer. And sorry I don't have a solution for you. maybe a reasonable name is the _Cauchy quotient_? Because it reminds everyone of the defining property/construction as well as the (potential) deficiency of it.

malarbol:
> > Though, why does it deserve a special name if it fails to satisfy the property you are after?
> 
> It satisfies almost all properties you'd expect from a Cauchy completion but not exactly (with our ACC for all we know). Maybe that's not the right concept.
> 
> In the meantime this construction still seems intrinsically interesting. Should it be called `cauchy-pseudocompletion-Pseudometric-Space`?



> > It satisfies almost all properties you'd expect from a Cauchy completion but not exactly (with our ACC for all we know).
> 
> That's a bummer. And sorry I don't have a solution for you. maybe a reasonable name is the _Cauchy quotient_? Because it reminds everyone of the defining property/construction as well as the (potential) deficiency of it.

I'm not sure. The problem is not about the _quotient_ (the **metric quotient** of a pseudometric space is indeed a well-defined metric space) but more about _completeness_ itself. (we cannot show that the metric quotient of a pseudocomplete pseudometric space is complete).

fredrik-bakke:
Right, hence why I suggested the juxtaposed _Cauchy quotient_. Anyways, it's just a suggestion.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+{{#concept "Cauchy pseudocompletion" Disambiguation="of a pseudometric space" Agda=cauchy-pseudocompletion-Pseudometric-Space}}
+of a [pseudometric space](metric-spaces.pseudometric-spaces.md) `M` is the
+pseudometric space of
+[Cauchy approximations](metric-spaces.cauchy-approximations-pseudometric-spaces.md)
+in `M` where two Cauchy approximations `x` and `y` are in a
+[`d`-neighborhood](metric-spaces.rational-neighborhood-relations.md) of one
+another if for all `δ ε : ℚ⁺`, `x δ` and `y ε` are in a
+`(δ + ε + d)`-neighborhood of one another in `M`.
+
+Any Cauchy approximation in the pseudometric completion has a
```
fredrik-bakke:
```suggestion
Any Cauchy approximation in the Cauchy pseudocompletion has a
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+        in
+          saturated-neighborhood-Pseudometric-Space M
+            ( δ +ℚ⁺ ε +ℚ⁺ d)
+            ( xδ)
+            ( yε)
+            ( λ θ →
+              tr
+                ( λ η → neighborhood-Pseudometric-Space M η xδ yε)
+                ( inv (associative-add-ℚ⁺ _ _ _))
+                ( H θ δ ε))
```
fredrik-bakke:
```suggestion
      let
        xδ = map-cauchy-approximation-Pseudometric-Space M x δ
        yε = map-cauchy-approximation-Pseudometric-Space M y ε
      in
        saturated-neighborhood-Pseudometric-Space M
          ( δ +ℚ⁺ ε +ℚ⁺ d)
          ( xδ)
          ( yε)
          ( λ θ →
            tr
              ( λ η → neighborhood-Pseudometric-Space M η xδ yε)
              ( inv (associative-add-ℚ⁺ _ _ _))
              ( H θ δ ε))
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+
+### The map from a Cauchy approximation in the pseudometric completion to its limit is an isometry
+
+```agda
+module _
+  { l1 l2 : Level} (M : Pseudometric-Space l1 l2)
+  ( d : ℚ⁺)
+  ( u v :
+    cauchy-approximation-Pseudometric-Space
+    ( cauchy-pseudocompletion-Pseudometric-Space M))
```
fredrik-bakke:
```suggestion
    cauchy-approximation-Pseudometric-Space
      ( cauchy-pseudocompletion-Pseudometric-Space M))
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+              ( cauchy-pseudocompletion-Pseudometric-Space M)
+              ( v)
+              ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+                ( M)
+                ( v))
+              ( is-limit-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+                ( M)
+                ( v)))
+            ( d)
+            ( Nuv)
```
fredrik-bakke: This can be defined as a separate entry

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+        ( v) →
+      neighborhood-Pseudometric-Space
+        ( cauchy-pseudocompletion-Pseudometric-Space M)
+        ( d)
+        ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+          ( M)
+          ( u))
+        ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+          ( M)
+          ( v))
```
fredrik-bakke: There should be a type `preserves-neighborhoods-map-Pseudometric-Space` that captures this signature. You should define that if it isn't already and instantiate it here.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,766 @@
...
+          ( u))
+        ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+          ( M)
+          ( v)) →
+      neighborhood-Pseudometric-Space
+        ( cauchy-pseudocompletion-Pseudometric-Space
+          ( cauchy-pseudocompletion-Pseudometric-Space M))
+        ( d)
+        ( u)
+        ( v)
```
fredrik-bakke, line 702: There should be a type `reflects-neighborhoods-map-Pseudometric-Space` that captures this signature. You should define that if it isn't already and instantiate it here.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: This is captured by [short functions](https://unimath.github.io/agda-unimath/metric-spaces.short-functions-pseudometric-spaces.html) between premetric spaces. 
Would you prefer `is-short-function-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space` with type signature [`is-short-function-Pseudometric-Space [...]`](https://unimath.github.io/agda-unimath/metric-spaces.short-functions-pseudometric-spaces.html#2382)?

fredrik-bakke: yes

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: I don't think we have anything for this one.
I guess we could have "neighborhoods-preserving maps", "neighborhood-reflecting maps" and relate them with the definitions of short functions (same has neighborhood-preserving) and isometries (neighborhood-preserving AND reflecting). 
I'm not sure if this PR is the best place to do this or better have it in another.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: see https://github.com/UniMath/agda-unimath/pull/1619/commits/6b9a0a441757f8e07b94b90d2841103ead7b93ef

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: see https://github.com/UniMath/agda-unimath/pull/1619/commits/6b9a0a441757f8e07b94b90d2841103ead7b93ef

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+module _
+  { l1 l2 : Level} (M : Pseudometric-Space l1 l2)
+  where
+
+  abstract
+    reflects-neighborhood-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space :
+      ( d : ℚ⁺) →
+      ( u v :
+        cauchy-approximation-Pseudometric-Space
+      ( cauchy-pseudocompletion-Pseudometric-Space M)) →
```
fredrik-bakke:
```suggestion
          ( cauchy-pseudocompletion-Pseudometric-Space M)) →
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+      ( is-limit-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+        ( M)
+        ( u))
+```
+
+### The map from a Cauchy approximation in the pseudometric completion to its limit is short
+
+```agda
+module _
+  { l1 l2 : Level} (M : Pseudometric-Space l1 l2)
```
fredrik-bakke:
```suggestion
  {l1 l2 : Level} (M : Pseudometric-Space l1 l2)
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+  short-lim-cauchy-approximation-pseudocompletion-Pseudometric-Space =
+    ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space M ,
+      is-short-function-lim-cauchy-approximation-pseudocompletion-Pseudometric-Space)
+```
+
+### The map from a Cauchy approximation in the pseudometric completion to its limit is an isometry
+
+```agda
+module _
+  { l1 l2 : Level} (M : Pseudometric-Space l1 l2)
```
fredrik-bakke:
```suggestion
  {l1 l2 : Level} (M : Pseudometric-Space l1 l2)
```

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: I tried to make this all part more readable. 
There might be something deeper going on:
Basically we have a diagram `C (C M) → C M → C (C M)` where the left map is the limit and the right map is the constant Cauchy approximation.
The composition `f :  C (C M) → C (C M)` satisfies `∀ u : f u ~ u` so because the right map is an isometry, so is the limit map. 
I'm not sure what to make of this but this seems interesting somehow.

fredrik-bakke: Yes, let's postpone this change. Clearly we should keep "short maps". I'd expect there'd be an established name for the dual in the literature.

malarbol: > Yes, let's postpone this change. Clearly we should keep "short maps". I'd expect there'd be an established name for the dual in the literature.

I think I've seen the terms "expansive mappings" or "expanding mappings" somewhere. Wikipedia https://en.wikipedia.org/wiki/Metric_map mentions "non-expansive" and "non-expanding" as synonyms of short maps so this is coherent.

fredrik-bakke: Great, those seem to be good names!

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+      in
+        monotonic-neighborhood-Pseudometric-Space M xδ xε
+          ( δ +ℚ⁺ ε)
+          ( δ +ℚ⁺ ε +ℚ⁺ d)
+          ( le-right-add-rational-ℚ⁺ _ d)
+          ( is-cauchy-approximation-map-cauchy-approximation-Pseudometric-Space
+            ( M)
+            ( x)
+            ( δ)
+            ( ε))
```
fredrik-bakke: theres no reason to have `let` bindings here

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+      lemma-δ+ε =
+        ( interchange-law-add-add-ℚ⁺
+          ( δ₁)
+          ( ε₁)
+          ( δ₂)
+          ( ε₂)) ∙
+        ( ap-binary
+          ( add-ℚ⁺)
+          ( δ₁+δ₂=δ)
+          ( ε₁+ε₂=ε))
```
fredrik-bakke: the linebreaking is excessive here, this should be a two-line expression.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+      ( M)
+      ( map-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space)
+  is-cauchy-map-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+    δ ε =
+    let
+      (δ₁ , δ₂ , δ₁+δ₂=δ) = split-ℚ⁺ δ
+      (ε₁ , ε₂ , ε₁+ε₂=ε) = split-ℚ⁺ ε
+
+      lemma-δ+ε :
+        (δ₁ +ℚ⁺ ε₁ +ℚ⁺ (δ₂ +ℚ⁺ ε₂)) ＝ δ +ℚ⁺ ε
```
fredrik-bakke: Is this
```suggestion
        (δ₁ +ℚ⁺ (ε₁ +ℚ⁺ (δ₂ +ℚ⁺ ε₂))) ＝ δ +ℚ⁺ ε
```
or
```suggestion
        ((δ₁ +ℚ⁺ ε₁) +ℚ⁺ (δ₂ +ℚ⁺ ε₂)) ＝ δ +ℚ⁺ ε
```
?
please add the missing parenthesis

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,773 @@
...
+      ( cauchy-pseudocompletion-Pseudometric-Space M)
+      ( U)
+      ( lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space)
+  is-limit-lim-cauchy-approximation-cauchy-pseudocompletion-Pseudometric-Space
+    δ ε η θ =
+    let
+      (θ₁ , θ₂ , θ₁+θ₂=θ) = split-ℚ⁺ θ
+
+      lemma-η+θ+δ :
+        ( η +ℚ⁺ θ₁ +ℚ⁺ (δ +ℚ⁺ θ₂)) ＝ η +ℚ⁺ θ +ℚ⁺ δ
```
fredrik-bakke: please also add the missing parenthesis on the left hand side here. Having a sum expression with some parenthization but not all is just confusing and odd.

src/metric-spaces/cauchy-pseudocompletion-of-pseudometric-spaces.lagda.md
malarbol: Fixed in https://github.com/UniMath/agda-unimath/pull/1619/commits/e1e95b61376d084c2fb729b366941cb3f32b1b63 (it was `((δ₁ +ℚ⁺ ε₁) +ℚ⁺ (δ₂ +ℚ⁺ ε₂))`)

src/metric-spaces/cauchy-pseudocompletion-of-metric-spaces.lagda.md
```diff
@@ -0,0 +1,437 @@
...
+          { y}
+          { const-cauchy-approximation-Metric-Space
+            ( M)
+            ( map-lim-cauchy-pseudocompletion-is-complete-Metric-Space y)}
+          ( sim-const-map-lim-cauchy-pseudocompletion-is-complete-Metric-Space
+            ( x))
+          ( sim-const-map-lim-cauchy-pseudocompletion-is-complete-Metric-Space
+            ( y))
+          ( d)
+          ( Nxy)
```
fredrik-bakke: This lemma shouldn't be hidden away in a `where` block. Make it global and call it `is-short-const-lim-cauchy-pseudocompletion-Metric-Space` or something.

fredrik-bakke (approved):

PR #1624, opened 2025-10-23
Multiplication on the reals is uniformly continuous on each argument

lowasser:
It's not uniformly continuous on the product metric space, which will require some care, but it is uniformly continuous across the products of totally bounded subsets of R.

PR #1630, opened 2025-10-24
Subrings

lowasser:
With #1624 , we have the infrastructure to prove that the uniformly continuous functions on a totally bounded subset of the reals form a ring, but that'll presumably be a subring of all functions on that subset, etc.

### DIFF
--- a/src/ring-theory/subrings.lagda.md
+++ b/src/ring-theory/subrings.lagda.md
@@ -1,0 +1,174 @@
+# Subrings
+
+```agda
+module ring-theory.subrings where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.conjunction
+open import foundation.dependent-pair-types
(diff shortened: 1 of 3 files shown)


src/ring-theory/subrings.lagda.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+</details>
+
+## Idea
+
+A
+{{#concept "subring" Disambiguation="of a ring" WD="subring" WDID=Q929536 Agda=Subring}}
+of a [ring](ring-theory.rings.md) `R` is a
+[subset](ring-theory.subsets-rings.md) of `R` that contains the zero and one of
```
fredrik-bakke, line 33:
```suggestion
[subset](ring-theory.subsets-rings.md) of `R` that contains the zero and one element of
```

PR #1633, opened 2025-10-25
Compactness of ℕ∞

fredrik-bakke:


src/set-theory/increasing-binary-sequences.lagda.md
```diff
@@ -298,10 +305,8 @@ retract-cantor-space-ℕ∞↗ = (sequence-ℕ∞↗ , retraction-sequence-ℕ
 abstract
   preserves-order-ℕ∞↗ :
     {x : ℕ∞↗} → preserves-order-Poset ℕ-Poset bool-Poset (sequence-ℕ∞↗ x)
-  preserves-order-ℕ∞↗ {x} =
-    preserves-order-ind-ℕ-Poset bool-Poset
-      ( sequence-ℕ∞↗ x)
-      ( is-increasing-sequence-ℕ∞↗ x)
+  preserves-order-ℕ∞↗ {x , p} =
```
VojtechStep: This feels like it's unnecessarily breaking the abstraction of defining the type `ℕ∞↗` in the first place, since pattern matching here essentially just binds the projections

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
```diff
@@ -0,0 +1,182 @@
...
+```agda
+abstract
+  has-decidable-Σ-pointed-bool-ℕ∞↗' :
+    has-decidable-Σ-pointed-bool' ℕ∞↗
+  has-decidable-Σ-pointed-bool-ℕ∞↗' p = (a , lemma)
+    where
+    a : ℕ∞↗
+    a = force-ℕ∞↗ (p ∘ increasing-binary-sequence-ℕ)
+
+    dagger₀ :
```
VojtechStep: What does `dagger` mean? And those look quite generalizable

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
```diff
@@ -0,0 +1,182 @@
+# Decidability of dependents sums over increasing binary sequences
```
VojtechStep: The page title disagrees with the file name on whether it's "dependent sum" or "dependent pair types". While other files sometimes use "dependent sum", the title and file name always agree; and there is a file on "decidability of dependent pair types", so I'd say that this should stick to "pair types" rather than "sums"

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
```diff
@@ -0,0 +1,182 @@
...
+open import foundation.tight-apartness-relations
+open import foundation.transport-along-identifications
+open import foundation.types-with-decidable-dependent-pair-types
+open import foundation.types-with-decidable-dependent-product-types
+open import foundation.unit-type
+open import foundation.universe-levels
+
+open import foundation-core.identity-types
+
+open import order-theory.order-preserving-maps-posets
```
VojtechStep: These imports look excessive. I haven't minimized them, but at least this one seems irrelevant

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
fredrik-bakke: Indeed, I believe this proof can be vastly generalized. I'd prefer to keep that for a separate PR, though, so as not to *waste my time polishing dirt, as one says.

fredrik-bakke: I don't know what `dagger` means. It is, as stated above, taken verbatim from Martín's proof.

VojtechStep: Don't take this the wrong way please, but then why do we want a verbatim copy of a very TypeTopology-style proof in the library? Do you have an immediate use for it, or can this be a draft until it's assimilated?

fredrik-bakke: Yes, I want this in the library without delay.

fredrik-bakke: In general, it is a lot better to have a suboptimal formalization than to have no formalization, and it is much easier to make iterative improvements than produce one-and-done masterpieces.

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
fredrik-bakke: you were absolutely right

src/set-theory/decidable-dependent-pair-types-over-increasing-binary-sequences.lagda.md
fredrik-bakke: Moreover, the generalization I have in mind requires large additions, so it will have to be iteratively merged regardless.

fredrik-bakke: This is to say nothing of the effects on my sanity.

fredrik-bakke:
I finished refactoring the formalization, and am much happier with the implementation now. Let me know if you are happy with this, @VojtechStep. Otherwise I'll merge in a couple of days.

fredrik-bakke:
There seems to be licensing issues with this PR #1635, and since I already have generalizations in mind that I would not consider derivative, hence circumventing the licensing issue, I am going to go ahead and close this one.

PR #1637, opened 2025-10-27
Well-founded material sets

fredrik-bakke:
Defines elementhood structures, material sets and well-founded material sets, as introduced in https://arxiv.org/pdf/2312.13024.

### DIFF
--- a/src/set-theory/elementhood-structures.lagda.md
+++ b/src/set-theory/elementhood-structures.lagda.md
@@ -41,7 +41,7 @@ is an [equivalence](foundation-core.equivalences.md).
 
 An extensional elementhood relation on `A` endows the type `A` with the
 [structure](foundation.structure.md) of
-{{#concept "elementhood" Disambiguation="on a type" Agda=Elementhood-Structure}}.
+{{#concept "elementhood" Disambiguation="on a type" Agda=Elementhood-Structure WD="is an element of" WDID=Q655288}}.
 
 ## Definitions
 


docs/PROJECTS.md
```diff
@@ -4,6 +4,7 @@ Here is a list of projects that use the agda-unimath library:
 
 - <https://git.app.uib.no/hott/hott-set-theory>
 - <https://git.app.uib.no/hott/containers>
+- <https://elisabeth.stenholm.one/univalent-material-set-theory/>
```
elisabethstenholm: This can be removed, as it is just a copy of (some of) the code at `https://git.app.uib.no/hott/hott-set-theory`

references.bib
```diff
@@ -520,6 +520,15 @@ @article{GK12
...
 
+@misc{GS24,
+  title         = {Univalent Material Set Theory},
+  author        = {Håkon Robbestad Gylterud and Elisabeth Stenholm},
+  year          = 2024,
+  eprint        = {2312.13024},
+  archiveprefix = {arXiv},
+  primaryclass  = {math.LO},
+}
+
```
elisabethstenholm: The paper has been published: https://www.sciencedirect.com/science/article/pii/S0168007225001113 🥳

src/set-theory/elementhood-structures.lagda.md
```diff
@@ -0,0 +1,217 @@
...
+        ( uniqueness-comprehension-Elementhood-Structure' ϕ)
+```
+
+## References
+
+{{#bibliography}}
+
+## External links
+
+- <https://elisabeth.stenholm.one/univalent-material-set-theory/e-structure.core.html>
```
elisabethstenholm:
```suggestion
- <https://git.app.uib.no/hott/hott-set-theory/-/blob/master/src/e-structure/core.agda>
```

elisabethstenholm: Actually, unsure what is the best link here now that I think about it... The code at my webpage is a copy of the actual repo at `git.app.uib.no`. However, that code may change, and the link may break in the future. Whereas the link to my webpage is to a snapshot of the code, that will not change.

docs/PROJECTS.md
elisabethstenholm: Hm, or is it nice to have a link to a snapshot of the code in nice clickable html? 🤔

src/set-theory/well-founded-material-sets.lagda.md
```diff
@@ -0,0 +1,251 @@
...
+      ( well-founded-relation-Well-Founded-Material-Set A)
+```
+
+## References
+
+{{#bibliography}}
+
+## External links
+
+- <https://elisabeth.stenholm.one/univalent-material-set-theory/e-structure.property.foundation.html>
```
elisabethstenholm, line 240:
```suggestion
- <https://git.app.uib.no/hott/hott-set-theory/-/blob/master/src/e-structure/property/foundation.agda>
```

elisabethstenholm: Same here. Which link is preferable?

docs/PROJECTS.md
fredrik-bakke: Oh, I didn't notice the main version does not have a clickable html page. That is definitely very handy.

fredrik-bakke:
@elisabethstenholm Does it make sense to call a type with an elementhood structure a "material set" at all? Or is it  the elements of such a type that are "material sets"? Or neither?

elisabethstenholm:
> @elisabethstenholm Does it make sense to call a type with an elementhood structure a "material set" at all? Or is it the elements of such a type that are "material sets"? Or neither?

I would say that it is the elements of the type with the associated elementhood relation, that are "material sets".

fredrik-bakke:
> I would say that it is the elements of the type with the associated elementhood relation, that are "material sets".

Hm, then what do you call a type with an elementhood structure? A "type with elementhood"?

elisabethstenholm:
> > I would say that it is the elements of the type with the associated elementhood relation, that are "material sets".
> 
> Hm, then what do you call a type with an elementhood structure? A "type with elementhood"?

A type together with an extensional elementhood relation, is what we have called an "elementhood structure" in our paper.

https://git.app.uib.no/hott/hott-set-theory/-/blob/master/src/e-structure/core.agda?ref_type=heads#L54-L57

fredrik-bakke:
Turns out there is considerable overlap with formalizations in `trees`, and in addition the names in this PR need to be rethought, so I'll mark this as a draft for now.

PR #1644, opened 2025-10-29
Show recent changes sitewide on homepage

fredrik-bakke:


fredrik-bakke:
Here's how the front page looks with this change (here with some garbage since it is on a local branch):
image

scripts/preprocessor_git_metadata.py
```diff
@@ -145,6 +146,55 @@ def get_author_element_for_file(filename, include_contributors, contributors, co
...
+
+        formatted_authors = format_multiple_authors_attribution([
+            contributors[idx]['displayName'] for idx in author_indices
+        ])
+        recent_changes += f'- {date}. {formatted_authors}. <i><a target="_blank" href={github_page_for_commit(sha)}>{message}.</a></i>\n'
+
+    if skipped_authors:
+        print_skipping_contributors_warning(skipped_authors, contributors_file)
+
+    return recent_changes
```
VojtechStep: Please extract this part to a function accepting an optional path and returning `(formatted_string, skipped_authors)`, and use it in `get_author_element_for_file` as well

fredrik-bakke: done

scripts/preprocessor_git_metadata.py
```diff
@@ -104,37 +148,8 @@ def get_author_element_for_file(filename, include_contributors, contributors, co
...
-                continue
-            author_indices.append(author_index)
-        if len(author_indices) == 0:
-            continue
-        formatted_authors = format_multiple_authors_attribution([
-            contributors[author_index]['displayName'] for author_index in author_indices
-        ])
-        recent_changes += f'- {date}. {formatted_authors}. <i><a target="_blank" href={github_page_for_commit(sha)}>{message}.</a></i>\n'
+    recent_changes, skipped_authors = get_recent_changes(
+        contributors, '--', filename)
```
VojtechStep: This reassigns `skipped_authors`, which could already have some contents by this point. You should do

```suggestion
    recent_changes, recent_skipped_authors = get_recent_changes(
        contributors, '--', filename)
    skipped_authors.update(recent_skipped_authors)
```
exactly as it's done 20 lines above

VojtechStep:
LGTM, thanks!

fredrik-bakke:
yay!

PR #1646, opened 2025-10-30
chore: Concepts in graphs

fredrik-bakke:


fredrik-bakke:
I am amused by the fact that "alcohols" has wikidata identifier number [156](https://www.wikidata.org/wiki/Q156)

fredrik-bakke:
Gotta love that according to wikipedia/wikidata's definitions, an Eulerian cycle is not a cycle.

PR #1656, opened 2025-10-30
chore: Concepts in `finite-group-theory`

fredrik-bakke:


fredrik-bakke:
TIL we have an almost 3000 lines long file about orbits of permutations that doesn't even define orbits 🙃

PR #522, opened 2023-03-19
Universal algebra first commit

shiranaiyo:
First commit on universal algebra. Credits to @plt-amy for coming up on how to do the `is-algebra` thing.

I'm not convinced on the names for the stuff under the sections `Abstract equations` and `Theories`, but otherwise I think it's fine.
I think that doing
```
group-Algebra-Group (((A , is-set-A) , models-A) , satisfies-A) =
```
is not common in this library but it really helps to reduce a lot of the boilerplate. Not sure if creating types equivalent to the standard finite types is fine. It certainly helps readability in my opinion.

PR #1657, opened 2025-10-30
Refactor `universal-algebra`

fredrik-bakke:


src/universal-algebra/algebraic-theory-of-groups.lagda.md
```diff
@@ -31,8 +31,11 @@ open import universal-algebra.terms-over-signatures
 
 ## Idea
 
-There is an algebraic theory of groups. They type of all such algebras is
-equivalent to the type of groups.
+There is an
+{{#concept "algebraic theory of groups" Disambiguation=theory-Group}}. The type
```
VojtechStep: This call is wrong

src/universal-algebra/algebraic-theory-of-groups.lagda.md
```diff
@@ -84,43 +87,43 @@ pr2 group-Theory =
...
 ## Properties
 
 ### The algebra of groups is equivalent to the type of groups
 
 ```agda
-group-Algebra-Group :
+group-algebra-Group :
   {l : Level} →
-  Algebra group-signature group-Theory l →
+  Algebra group-signature theory-Group l →
```
VojtechStep: Better yet, changing this to `algebra-Group l` makes it more apparent that this definition should have the name it has. Here and later in the file

VojtechStep (commented):
I pretty much only looked at the concept calls, before I realized this is a refactoring PR

fredrik-bakke:
@VojtechStep workingonit

fredrik-bakke:
This is in preparation of some work Mark might want to contribute.

src/universal-algebra/algebraic-theory-of-groups.lagda.md
fredrik-bakke: was already going to do that, thanks

fredrik-bakke:
Sorry for the switch-up @VojtechStep

src/universal-algebra/models-of-signatures.lagda.md
```diff
@@ -36,45 +34,46 @@ open import universal-algebra.signatures
...
-    ( tuple X (arity-operation-signature σ f) → X)
+```agda
+is-model : {l1 l2 : Level} → signature l1 → Set l2 → UU (l1 ⊔ l2)
+is-model σ X = is-model-type σ (type-Set X)
+```
 
-  is-model-signature : {l2 : Level} → (Set l2) → UU (l1 ⊔ l2)
-  is-model-signature X = is-model (type-Set X)
+```agda
+Model : {l1 : Level} (l2 : Level) → signature l1 → UU (l1 ⊔ lsuc l2)
```
VojtechStep: This used to be called `Model` in the first PR on universal algebra, but Egbert [explicitly requested](https://github.com/UniMath/agda-unimath/pull/522#discussion_r1142274384) the change to `Model-Signature`, and I agree with him that this definition doesn't cover many of the concepts one would reasonably want to call models (e.g. models of GATs or SOGATs would not be `Model`s)

fredrik-bakke: I agree, but this is not consistent with the rest of the namespace. If we want to be specific here, we should qualify everything as algebraic, single-sorted, and finitary, as I also pointed out in that first PR. Either everything should be qualified or nothing should be qualified. Anything in between is confusing.

fredrik-bakke: If you insist that we disambiguate, I can do that, but in a subsequent PR. Note that this will entail naming changes like

- `signature` ->  `single-sorted-finitary-algebraic-signature`
- `Model` ->  `Model-Of-Single-Sorted-Finitary-Algebraic-Signature`
- `abstract-equation` -> `abstract-equation-over-single-sorted-finitary-algebraic-signature`

VojtechStep: My comment was about not changing the name in this PR, since IMO no disambigutation is worse than incomplete disambiguation. If we wanted to be more precise, I'd rather suggest `Algebraic-Theory` as the "namespace" (the "AT" part of "GAT"), i.e.
- `signature` -> `Signature-Algebraic-Theory`
- `Theory` -> `Algebraic-Theory`
- `Model` -> `Structure-Algebraic-Theory`
- `Algebra` -> `Model-Algebraic-Theory`
- `abstract-equation` -> `Equation-Algebraic-Theory`

The above also raises the point that our definition of "model" isn't what I've seen people in general to call a "model": in Crole (which Fernando said he was following) an interpretation of a signature is a "structure" and a "model" is a structure satisfying the imposed equalities. This is also the same convention that Caramello uses, and the one I'm familiar with.

fredrik-bakke: This seems like a reasonable proposal.

fredrik-bakke: Opened an issue, and resolving this conversation for now

fredrik-bakke:
Merging this one soon

PR #1670, opened 2025-11-06
chore: Display more changes on front page

fredrik-bakke:


fredrik-bakke:
here's what it looks like with 10 items
image

fredrik-bakke:
here's what it looks like with 8 items
image

PR #1683, opened 2025-11-09
Dual construction of coherent unit laws

fredrik-bakke:


src/foundation/unital-binary-operations.lagda.md
```diff
@@ -71,29 +78,84 @@ is-coherently-unital {A = A} μ = Σ A (coherent-unit-laws μ)
 
 ### The unit laws for an operation `μ` with unit `e` can be upgraded to coherent unit laws
 
+#### Preserving the underlying left unit law
+
+Given a pair of unit laws `λ` and `ρ` we construct a new coherent pair of unit
+laws where the left unit law is the same, `λ`, and the right unit law is
+replaced by `ρ'(x) := ap (μ x (-)) (ρ(e)⁻¹ ∙ λ(e)) ∙ ρ(x)`.
```
VojtechStep: This is not the definition in the code though. I think it's important that if the prose poses something as a definitional equality with `:=`, then the code should reflect that, not just be propositionally equal

fredrik-bakke: fair fair

PR #1693, opened 2025-11-11
Infinitely differentiable real functions defined on proper closed intervals

lowasser:
Still getting a handle on how we like to work with coinductive records, though the conatural numbers are an example I'm taking notes from.

### DIFF
--- a/src/complex-numbers.lagda.md
+++ b/src/complex-numbers.lagda.md
@@ -4,12 +4,17 @@
 module complex-numbers where
 
 open import complex-numbers.addition-complex-numbers public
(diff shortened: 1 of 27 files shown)


fredrik-bakke:
> Still getting a handle on how we like to work with coinductive records, though the conatural numbers are an example I'm taking notes from.

Could an infinitely differentiable function be defined as a sequence of functions such that each is a derivative of the next? Then you would not need to be bothered with coinduction. For reference, a closer example would perhaps be [Unbounded π-finite types](https://unimath.github.io/agda-unimath/univalent-combinatorics.unbounded-pi-finite-types.html).

PR #1703, opened 2025-11-11
Double negation stability of equality on Dedekind real numbers

fredrik-bakke:


src/real-numbers/equality-real-numbers.lagda.md
```diff
@@ -0,0 +1,69 @@
...
+open import real-numbers.similarity-real-numbers
+```
+
+</details>
+
+## Idea
+
+We record that [equality](foundation-core.identity-types.md) and
+[similarity](real-numbers.similarity-real-numbers.md) of
+[Dedekind real numbers](real-numbers.dedekind-real-numbers.md) is
```
lowasser: "are" vs "is"?

PR #1706, opened 2025-11-12
chore: Rename `extension` to `extension-map` and `lift` to `lift-map`

fredrik-bakke:


src/orthogonal-factorization-systems/extensions-maps.lagda.md
```diff
@@ -93,29 +95,29 @@ module _
...
 
-  is-extension-along-self : is-extension f f id
-  is-extension-along-self = refl-htpy
+  is-extension-of-map-along-self : is-extension-of-map f f id
+  is-extension-of-map-along-self = refl-htpy
 
-  extension-along-self : extension f f
-  extension-along-self = (id , is-extension-along-self)
+  self-extension-map : extension-map f f
+  self-extension-map = (id , is-extension-of-map-along-self)
```
VojtechStep, line 120: Perhaps `is-extension-map-id` and `id-extension-map`?

fredrik-bakke: is-extension-map-id has two different interpretations, so I'd prefer just keeping this wording

PR #1709, opened 2025-11-13
Register two Wikipedia theorems

fredrik-bakke:


fredrik-bakke:
I'll go ahead and merge this, it should not be controversial.

VojtechStep:
For sure, I was just waiting for the pre-commit fix

PR #1713, opened 2025-11-14
Hilbert spaces

lowasser:
Uses the Cauchy-Schwarz inequality to prove the triangle inequality for inner product spaces, and that inner product spaces are normed.


src/linear-algebra/real-hilbert-spaces.lagda.md
fredrik-bakke, line 1: I just realized you've been putting these files in `linear-algebra`. But actually the study of hilbert spaces, inner product spaces, banach spaces, and even normed vector spaces, are usually categorized in the theory of functional analysis. Maybe it is time to open yet another namespace?

fredrik-bakke: Either way, I suggest such a change is performed in an isolated PR

PR #1717, opened 2025-11-16
The real number `e`

lowasser:
Or maybe I should get the power series for `exp(x)` working everywhere and define `e = exp(1)`?  I think I slightly prefer this way.


malarbol:
> I think I slightly prefer this way.

Not sure which way you prefer but I'm voting for

> [...]  get the power series for `exp(x)` working everywhere and define `e = exp(1)`?

PR #1719, opened 2025-11-17
Large function commutative rings

lowasser:
Starting to build this out so we can do analysis on these too.


lowasser:
I swear I'm going to try to slow down, I know I've built up a large review queue, but I think it's worth it, that it's all significant content...

fredrik-bakke:
Your contributions are most valued and we're very excited to see your speedy progress! That said, I will likely be more busy going forward and until christmas, so reviews might slow down on my end :)

PR #1604, opened 2025-10-16
Dependent sums and products of cardinals

fredrik-bakke:
Depends on #1591

PR #1720, opened 2025-11-17
Kőnig's theorem (indexed)

fredrik-bakke:
Formalizes Kőnig's theorem (for indexed inequality): given two families of cardinals `A` and `B` over projective `I` such that `A i < B i` for all `i`, then `ΣᵢAᵢ < ΠᵢBᵢ`.

Depends on #1604 and #1591, and #1846 and #1853.


fredrik-bakke:
@lowasser it turns out cardinality-projectivity is a redundant notion, since all projective types are cardinality-projective. I'll therefore refactor to reflect that before marking it ready for review.

fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #1730, opened 2025-11-27
Cleanup of descent properties

fredrik-bakke:
Cleans up certain files on descent properties, and compiles such properties into a table.


fredrik-bakke:
pre-commit now fails in the cloud but not locally due to the character counting issue with certain symbols such as `↪`, in files that are not touched by this PR!

PR #1735, opened 2025-11-28
Operations on cospans

fredrik-bakke:
Just a blueprint copy of the duals of the operations for spans.


fredrik-bakke:
@VojtechStep @EgbertRijke I'm going ahead and invoking the self-review policy on this PR and its dependencies.

PR #1736, opened 2025-11-28
Equivalences of cospan diagrams

fredrik-bakke:
Builds on #1735


fredrik-bakke:
@VojtechStep @EgbertRijke I'm going ahead and invoking the self-review policy on this PR and its dependencies.

PR #1742, opened 2025-12-02
Extensions of (pseudo)metric spaces

malarbol:
- `extension-Pseudometric-Space P`: a pseudometric space `Q` with an isometry `i : P → Q`;
- `extension-Metric-Space M`: a metric space `N` with an isometry `i : M → N`;
- actions on Cauchy approximations of extensions of (pseudo)metric spaces;
- isometries between extensions of a metric space.

### DIFF
--- a/src/metric-spaces/action-on-cauchy-approximations-extensions-pseudometric-spaces.lagda.md
+++ b/src/metric-spaces/action-on-cauchy-approximations-extensions-pseudometric-spaces.lagda.md
@@ -17,6 +17,7 @@ open import metric-spaces.cauchy-approximations-pseudometric-spaces
 open import metric-spaces.cauchy-pseudocompletion-of-pseudometric-spaces
 open import metric-spaces.extensions-pseudometric-spaces
 open import metric-spaces.isometries-pseudometric-spaces
+open import metric-spaces.maps-pseudometric-spaces
 open import metric-spaces.pseudometric-spaces
 ```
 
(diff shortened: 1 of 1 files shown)


fredrik-bakke:
What you call a `Metric-Extension` should really be called something of the form `is-metric-extension`. **A** metric extension _of_ `P` would be a choice of a metric space `M` together with an isometry. I.e., the `M` is bundled with the isometry. I am still entirely unconvinced about the terminology choice of `Metric-Extension`, though, but I haven't come up with a better name which is why you haven't heard from me.

fredrik-bakke:
I believe* it should be called `is-extension-metric-space-Pseudometric-Space`, and presumably you would want the rest of your formalizations to be in terms of the bundled version instead, whatever that should be called. Potentially `extension-metric-space-Pseudometric-Space`.

fredrik-bakke:
another candidate is `extensional-extension-Pseudometric-Space`, although I like that less

fredrik-bakke:
it would fit into a larger picture though, together with extensional extensions of preorders and strictly* preordered sets, and such things.

fredrik-bakke:
@malarbol this pull request is very large. Is there any possibility of splitting it into smaller chunks, preferrably of less than 1000 lines of code each?

malarbol:
> @malarbol this pull request is very large. Is there any possibility of splitting it into smaller chunks, preferrably of less than 1000 lines of code each?

Should be better now.

malarbol:
I'm re-drafting this to start over, again. The end game is still #1458 and _Cauchy precompletions_ (metric quotients of Cauchy pseudocompletions). I found a way to formalize some of what I wanted, i.e. "any isometry from a metric space into a complete metric space uniquely extends to the Cauchy precompletion" or "any Cauchy-dense complete extension is isometrically equivalent to the Cauchy precompletion" (see #1776 for those interested) but maybe I don't actually _need_ the concepts of **extension of metric** spaces to state them. I'll try another road; at least I know a little bit better where I go. And I'll do my best to keep my PRs under 1k LOC 😉.

PR #1716, opened 2025-11-16
The ratio test for convergence in real Banach spaces

lowasser:
Builds on #1714 .

PR #1747, opened 2025-12-06
The natural exponential function on real numbers

lowasser:
Depends on #1716.


src/analysis/exponential-power-series-real-numbers.lagda.md
```diff
@@ -0,0 +1,336 @@
...
+
+This page describes properties of the power series.
+
+## Definition
+
+```agda
+exp-power-series-at-zero-ℝ : (l : Level) → power-series-at-zero-ℝ l
+exp-power-series-at-zero-ℝ l =
+  power-series-at-zero-coefficients-ℝ
+    ( λ n → raise-real-ℚ l (reciprocal-rational-ℕ⁺ (nonzero-factorial-ℕ n)))
```
malarbol, line 78: Couldn't we define `exp-power-series-XXX` in the settings of [ring extension of ℚ](https://unimath.github.io/agda-unimath/ring-theory.ring-extensions-rational-numbers.html#concept-Rational-Extension-Ring). For the definition you only need to invert `factorial-ℕ n` in a ring structure so this should be enough to define the power series. Then apply this to Banach algebras to define exponentials of real numbers, complex numbers, matrices, etc.

lowasser: We can, but it's not obvious to me we'd ever find a use for it in something that isn't also a ring extension of the real numbers?

lowasser: We can, but it's not obvious to me we'd ever find a use for it in something that isn't also a ring extension of the real numbers?

malarbol: Sure. But now you're defining the exponential series living in `formal-power-series-Commutative-Ring (commutative-ring-ℝ l)` not some ring extension of the real numbers. I think it should be in some `formal-power-series-Commutative-Ring (???)` for a larger class of rings, be it rational extensions, or real extensions if you prefer, but not restricted to real numbers themselves.

lowasser: Fair.  And there's another reason that occurs to me: computing rates of convergence ought be aided by having explicit rationals instead of reals, preventing the need for any flavor of choice.

malarbol: FWIW, I finally implemented my idea of `coefficient-exponential-series-Rational-Extension-Ring`, the sequence `n ↦ 1/n!` in ring extensions of the rationals. The basic definitions and properties are in #1996 if you want to have a look.

malarbol: #1996 has now been merged 🥳.
I hope we can use it to define the natural exponential of real numbers, complex numbers, real matrices, etc.

PR #1749, opened 2025-12-08
Smallness of the monomorphism predicate

fredrik-bakke:
A bit random. It optimizes some universe level bounds here and there. To get it working without cyclic dependencies I had to split up a couple of files though. Also renames retracts of maps to retracts of arrows. This is both a less ambiguous name, and fits better with our other naming scheme.


fredrik-bakke:
I'm going ahead and merging this one.

PR #1752, opened 2025-12-16
Standardize an "is zero" predicate on the real numbers

lowasser:
We use similarity because it is at a lower universe level, which should probably be its own thing.

### DIFF
--- a/src/complex-numbers/similarity-complex-numbers.lagda.md
+++ b/src/complex-numbers/similarity-complex-numbers.lagda.md
@@ -14,6 +14,7 @@ open import foundation.dependent-pair-types
 open import foundation.identity-types
 open import foundation.large-equivalence-relations
 open import foundation.large-similarity-relations
+open import foundation.locally-small-types
 open import foundation.propositions
 open import foundation.universe-levels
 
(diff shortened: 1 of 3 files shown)


src/real-numbers/field-of-real-numbers.lagda.md
```diff
@@ -45,11 +46,11 @@ The [real numbers](real-numbers.dedekind-real-numbers.md) form a
 
 ```agda
 abstract
-  is-zero-is-noninvertible-commutative-ring-ℝ :
+  eq-zero-is-noninvertible-commutative-ring-ℝ :
```
fredrik-bakke: This should actually be named `eq-zero-is-not-invertible-ℝ`*

src/real-numbers/field-of-real-numbers.lagda.md
```diff
@@ -63,12 +64,19 @@ abstract
                 ( sim-raise-ℝ' l zero-ℝ)
                 ( x#raise-l-zero)))))
 
+  is-zero-is-noninvertible-commutative-ring-ℝ :
```
fredrik-bakke:
```suggestion
  is-zero-is-not-invertible-ℝ :
```

src/real-numbers/zero-real-numbers.lagda.md
```diff
@@ -0,0 +1,81 @@
...
+open import real-numbers.dedekind-real-numbers
+open import real-numbers.rational-real-numbers
+open import real-numbers.similarity-real-numbers
+```
+
+</details>
+
+## Idea
+
+A [real number](real-numbers.dedekind-real-numbers.md) is zero if it is
```
fredrik-bakke: "zero" is a new concept here

fredrik-bakke: so please use `#concept`

fredrik-bakke (approved):

PR #1739, opened 2025-12-02
Odd roots of real numbers

lowasser:
We do something more general, actually, which is show that any strictly increasing, pointwise continuous, unbounded function from ℝ to ℝ is an equivalence.

PR #1760, opened 2025-12-23
Rewrite the implementation of the multiplicative inverse of positive reals

lowasser:
Depends on #1739, using how that PR shows that any strictly increasing, pointwise continuous, unbounded function from the reals to the reals is invertible; left multiplication by a positive real number satisfies those conditions, so the multiplicative inverse of x is the application of the inverse function of left multiplication by x to 1.  Saves ~450 lines and makes the reasoning much easier to follow.

### DIFF
--- a/src/analysis.lagda.md
+++ b/src/analysis.lagda.md
@@ -3,6 +3,8 @@
 ```agda
 module analysis where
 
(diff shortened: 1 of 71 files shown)


lowasser:
Redoing.

PR #1761, opened 2025-12-23
Change to multiplicative terminology for combining finite sequences in semigroups, monoids, and groups

lowasser:
I'm no longer convinced the additive terminology avoids any confusion.


src/group-theory/products-of-finite-sequences-of-elements-monoids.lagda.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Products are equal to the zero-th term plus the rest
+
+```agda
+module _
+  {l : Level} (M : Monoid l)
+  where
+
+  abstract
+    cons-product-fin-sequence-type-Monoid :
```
lowasser, line 106: GitHub doesn't make the difference easy to see, so I will call it out: the original versions of these functions expected you to exhibit the value at the head of the sequence and an identity proving it is the head of the sequence, but I think that should really be an `ap` after the invocation of this function, not combined in.  The main use case that had to change is `binomial-theorem-semirings`.

PR #1766, opened 2025-12-24
Split out functional analysis as its own module

lowasser:
As suggested earlier, stuff about Banach and Hilbert vector spaces falls into this area.


src/linear-algebra/normed-real-vector-spaces.lagda.md
fredrik-bakke, line 1: I believe this file too should go in `functional-analysis`. Do you disagree?

lowasser: I will take your word for it.

fredrik-bakke: After thinking a little more, I take this back. The most sensible splitting point seems to be at the completeness condition, seeing as it is undisputable that orthogonality is a concept belonging to linear algebra. My apologies!

PR #1770, opened 2025-12-26
Correct the naming of monotonic neighborhoods in metric spaces

lowasser:
I remain convinced that "monotonic" is the correct term to describe a property P such that if `x ≤ y` and `P x`, then `P y`.  However, "strictly monotonic" is _probably_ a better term for properties such that if `x < y` and `P x`, then `P y`.

### DIFF
--- a/src/metric-spaces/located-metric-spaces.lagda.md
+++ b/src/metric-spaces/located-metric-spaces.lagda.md
@@ -139,7 +139,8 @@ module _
     neighborhood-Located-Metric-Space d₁ x y →
     neighborhood-Located-Metric-Space d₂ x y
   strictly-monotonic-neighborhood-Located-Metric-Space =
-    strictly-monotonic-neighborhood-Metric-Space metric-space-Located-Metric-Space
+    strictly-monotonic-neighborhood-Metric-Space
+      ( metric-space-Located-Metric-Space)
 
   is-located-Located-Metric-Space :
     is-located-Metric-Space metric-space-Located-Metric-Space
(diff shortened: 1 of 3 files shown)


fredrik-bakke:
> I remain convinced that "monotonic" is the correct term to describe a property P such that if `x ≤ y` and `P x`, then `P y`. However, "strictly monotonic" is _probably_ a better term for properties such that if `x < y` and `P x`, then `P y`.

"Strictly monotonic" would be the statement that  if `x < y` then `P x` implies`P y` _and `P y` does not imply `P x`_.  I.e., it would be a strictly order-preserving map into the powerset.

src/metric-spaces/located-metric-spaces.lagda.md
```diff
@@ -133,13 +133,14 @@ module _
   triangular-neighborhood-Located-Metric-Space =
     triangular-neighborhood-Metric-Space metric-space-Located-Metric-Space
 
-  monotonic-neighborhood-Located-Metric-Space :
+  strictly-monotonic-neighborhood-Located-Metric-Space :
     (x y : type-Located-Metric-Space) (d₁ d₂ : ℚ⁺) →
     le-ℚ⁺ d₁ d₂ →
     neighborhood-Located-Metric-Space d₁ x y →
     neighborhood-Located-Metric-Space d₂ x y
```
fredrik-bakke, line 140: This condition might be better described as weak monotonicity, because it is a weaker condition than monotonicity: it only works when we know d1 is strictly less than d2.

fredrik-bakke: however, "weak monotonicity" is not a very descriptive name. I'd be open to other suggestions too

lowasser: Maybe the correct approach is just to delete this condition entirely and replace it for every metric space with the nonstrict inequality version...

fredrik-bakke: If these things are always monotonic that seems reasonable.

PR #1782, opened 2026-01-01
Centers of rings

lowasser:


src/group-theory/centers-groups.lagda.md
```diff
@@ -23,7 +25,13 @@ open import group-theory.subgroups
 
 ## Idea
 
-The **center** of a group consists of those elements that are central.
+The
+{{#concept "center" Disambiguation="of a group" WDID=Q1195852 WD="center" Agda=center-Group}}
+of a [group](group-theory.groups.md) `G` is the
+[abelian](group-theory.abelian-groups.md)
+[normal](group-theory.normal-subgroups.md) [subgroup](group-theory.subgroups.md)
```
fredrik-bakke: Links directly next to eachother should be avoided since they do not appear as distinct links to the reader. Instead use
```suggestion
[normal subgroup](group-theory.normal-subgroups.md)
```

fredrik-bakke: optionally add that this normal subgroup is abelian as a separate statement

src/group-theory/centers-monoids.lagda.md
```diff
@@ -22,8 +24,11 @@ open import group-theory.submonoids
 
 ## Idea
 
-The **center** of a [monoid](group-theory.monoids.md) consists of those elements
-that are central.
+The {{#concept "center" Disambiguation="of a monoid" Agda=center-Monoid}} of a
+[monoid](group-theory.monoids.md) `M` is the
+[commutative](group-theory.commutative-monoids.md)
+[submonoid](group-theory.submonoids.md) consisting of those elements of `M`
```
fredrik-bakke:
```suggestion
[submonoid](group-theory.submonoids.md) consisting of those elements of `M`
```

fredrik-bakke: again optionally add that this submonoid is commutative.

src/group-theory/centers-semigroups.lagda.md
```diff
@@ -22,7 +24,11 @@ open import group-theory.subsemigroups
 
 ## Idea
 
-The center of a semigroup consists of those elements that are central.
+The {{#concept "center" Disambiguation="of a semigroup" Agda=center-Semigroup}}
+of a [semigroup](group-theory.semigroups.md) `G` is the
+[commutative](group-theory.commutative-semigroups.md)
+[subsemigroup](group-theory.subsemigroups.md) consisting of those elements of
```
fredrik-bakke:
```suggestion
[subsemigroup](group-theory.subsemigroups.md) consisting of those elements of
```
and add that it is commutative in a separate statement.

fredrik-bakke (approved):

src/group-theory/centers-groups.lagda.md
lowasser: I would very much like if we could fix this as a matter of the format on the site.

fredrik-bakke: I proposed a solution to this in the past, but it was not deemed satisfactory. See the history here: #1413

PR #1788, opened 2026-01-06
Functoriality of metric quotients and short maps

malarbol:
This PR introduces the **functorial action of metric quotients on short maps**: any short map `f : P  → Q` between pseudometric spaces induces a short map of metric spaces `g : [P]  → [Q]` between the metric quotients. This action preserves the identity and composition of short maps.

PR #1795, opened 2026-01-13
Rename unit map metric quotient

malarbol:
This PR introduces a renaming scheme for the unit map of metric quotients.


malarbol:
@fredrik-bakke I know you said you didn't deem necessary to make separated modules for the `unit-map` but it was easier to refactor this way. If you ~really want~ prefer it all in the same module, I can try to put it back together.

fredrik-bakke:
I said I didn't yet see why it should have a separate module, not that I thought it shouldn't.

malarbol:
The name scheme tries to follow this pattern:

- we have concepts `P` and `M` (`Pseudometric-Space` and `Metric-Space`);
- each of them have som concepts of _morphisms_ `H` (`map`, `short-map`, `isometry`, `ucont-map`, etc.);
- the names of the space of _morphisms_ follow the template `{H}-{X}` (`short-map-Pseudometric-Space`, `isometry-Metric-Space`, etc.);

I want to consider the construction `metric-quotient-Pseudometric-Space`, and and combine it with the different types of morphisms so I need some names do describe the spaces of morphisms in, out and between metric quotients, i.e. morphisms 
- `P → [Q]`;
- `[P] → Q`;
- `[P] → [Q]`.

(where `[X]` denote the metric quotient of `X`).

So I'll need names `{H}-??-metric-quotient-??-Pseudometric-Space` to describe these. Then the unit map ought to be called `unit-{H}-??-metric-quotient-??-Pseudometric-Space` for different flavors of _morphisms_ `H`.

The current suggestion uses  `{H}-metric-quotient-Pseudometric-Space` but it may be wrong.

The name scheme we chose should also apply to **Cauchy pseudocompletions** (pseudometric space of Cauchy approximations with the neighborhoods that makes it _pseudocomplete_) to consider morphisms 
`{H}-??-cauchy-pseudocompletion-??-{X}` (`X` either `Metric-Space` or `Pseudometric-Space`): 
- `X → C Y`;
- `C X → Y`;
- `C X → C Y`.

For example, the case `C X → C Y` is what we call `action-on-cauchy-approximations-{H}-{X}` but I think the name is wrong. The fact that Cauchy approximations are the habitants of the Cauchy pseudocompletion is incidental and the names should be something like `??-cauchy-pseudocompletion-??-{H}-??-{X}`

malarbol:
> I said I didn't yet see why it should have a separate module, not that I thought it shouldn't.

And I didn't say you _thought it shouldn't_, I said you didn't deem it as necessary, so I guess we agree 😅? 

Does this PR make you see the appeal of having them in their own module?

Either way, I'll need names for morphisms 
- `P → [Q]`;
- `[P] → Q`;
- `[P] → [Q]`;
- `X → C Y`;
- etc.

So we should discuss this. 

I guess the names should also apply to the construction `pseudometric-Metric-Space` (forgetting extensionality) so the names  `{H}-??-pseudometric-??-Metric-Space` would describe morphisms in/out/between forgetting extensionality. they are all the same concept in this case, so it's probably not necessary to define name, but _if we were to define them_, the name scheme should work too.

I'll also need it for the **Cauchy precompletions** (metric quotients of Cauchy pseudocompletions) of (pseudo)metric spaces.

malarbol:
Ultimately, I'll also need names for the actions of these constructions between the different types of morphisms, e.g.
- `(P → [Q]) → ([P] → [Q])` (for the universal property of metric quotients);
- `(P → Q) → ([P] → [Q])` (some kind of _functorial_ action. what I called `hom-metric-quotient-{H}-Pseudometric-Space` in #1788 but I guess the name was wrong and should be something like `??-{H}-??-metric-quotient-Pseudometric-Space`?);
- `(C X → Y) → (X → Y)`;
- `(X → Y) → (C X → C Y)` (_action on Cauchy approximations of morphisms_, should be renamed);
- `(X → Y) → (C X → Y)` (the ability to extend a morphism to the Cauchy pseudocompletion, i.e., all Cauchy approximations in `X` converge in `Y`).
- etc.

malarbol:
> I'll also need it for the Cauchy precompletions (metric quotients of Cauchy pseudocompletions) of (pseudo)metric spaces.

In that case, the `cauchy-precompletion-Metric-Space` construction will also define [pointed endofunctors](https://unimath.github.io/agda-unimath/category-theory.pointed-endofunctors-categories.html) on metric spaces (for the `short-map` and `isometry` morphisms) and  `unit-{H}-??-cauchy-precompletion-??-Metric-Space` will be the pointing family `η`. So ideally the naming scheme should also be coherent with this.

malarbol:
Maybe we could use the placeholders `in` / `out` / `hom` for the situations
- `X → C Y`;
- `C X → Y`;
- `C X → C Y`.

and a template `{H}-{pos}-{Y←}-{X}` with names like
- `short-map-in-metric-quotient-Pseudometric-Space` (short map `X → [Y]`);
- `isometry-out-cauchy-pseudocompletion-Metric-Space` (isometry `C X → Y`);
- `isometry-hom-cauchy-precompletion-Metric-Space` (isometry `[C X] → [C Y]`).

fredrik-bakke:
* `P → [Q]`: This just looks like the postcomposition map of the unit map applied to something. I.e. it is `postcomp-map-unit-metric-quotient-... f`

* `[P] → Q`: `map-extension-metric-quotient-...`

* `[P] → [Q]`: `map-metric-quotient-...`

- `X → C Y`: `postcomp-map-unit-caucy-precomplection-... f`

- `C X → Y`: `map-extension-cauchy-precompletion-...`

- `C X → C Y`: `map-cauchy-precompletion-...`

The noun `map` should be replaced by whichever concept one is considering, be it `isometry`, `short-map`, `ucont-map` and so on.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -111,7 +112,7 @@ module _
     short-map-cauchy-approximation-short-map-Pseudometric-Space
       ( M)
       ( pseudometric-metric-quotient-Pseudometric-Space M)
-      ( short-map-metric-quotient-Pseudometric-Space M)
+      ( unit-short-map-metric-quotient-Pseudometric-Space M)
```
fredrik-bakke:
```suggestion
      ( short-map-unit-metric-quotient-Pseudometric-Space M)
```

think of it as the "underlying short map of the unit"

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -211,7 +212,7 @@ module _
     is-limit-cauchy-approximation-Metric-Space
       ( metric-quotient-Pseudometric-Space M)
       ( map-cauchy-approximation-metric-quotient-Pseudometric-Space M u)
-      ( map-metric-quotient-Pseudometric-Space M lim)
+      ( unit-map-metric-quotient-Pseudometric-Space M lim)
```
fredrik-bakke:
```suggestion
      ( map-unit-metric-quotient-Pseudometric-Space M lim)
```

src/metric-spaces/metric-quotients-of-metric-spaces.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+
+### The unit map from a metric space into its metric quotient is an equivalence
+
+```agda
+module _
+  {l1 l2 : Level}
+  (M : Metric-Space l1 l2)
+  where abstract
+
+  is-contr-map-unit-map-metric-quotient-Metric-Space :
```
fredrik-bakke:
```suggestion
  is-contr-map-map-unit-metric-quotient-Metric-Space :
```

src/metric-spaces/metric-quotients-of-metric-spaces.lagda.md
```diff
@@ -0,0 +1,187 @@
...
+## Applications
+
+### The isometric equivalence between a metric space and its metric quotient
+
+```agda
+module _
+  {l1 l2 : Level} (M : Metric-Space l1 l2)
+  where
+
+  isometric-equiv-metric-quotient-Metric-Space' :
```
fredrik-bakke:
```suggestion
  isometric-equiv-unit-metric-quotient-Metric-Space' :
```
to avoid confusion with the functorial action on isometric equivalences

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
+# Unit map of metric quotients of pseudometric spaces
```
fredrik-bakke:
```suggestion
# The unit map of metric quotients of pseudometric spaces
```

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+#### Maps
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  (P : Pseudometric-Space l1 l2)
+  (Q : Pseudometric-Space l3 l4)
+  where
+
+  map-metric-quotient-Pseudometric-Space : UU (l1 ⊔ l3 ⊔ l4)
```
fredrik-bakke: This should be the name of the functorial action of metric quotients. It's a bit non-standard to name the types you've defined in this file.

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+
+### The unit map from a pseudometric space to its quotient metric space
+
+```agda
+module _
+  {l1 l2 : Level}
+  (P : Pseudometric-Space l1 l2)
+  where
+
+  unit-map-metric-quotient-Pseudometric-Space :
```
fredrik-bakke:
```suggestion
  map-unit-metric-quotient-Pseudometric-Space :
```

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+    map-metric-quotient-Pseudometric-Space P P
+  unit-map-metric-quotient-Pseudometric-Space =
+    quotient-map (equivalence-relation-sim-Pseudometric-Space P)
+
+  is-in-class-unit-map-metric-quotient-Pseudometric-Space :
```
fredrik-bakke:
```suggestion
  is-in-class-map-unit-metric-quotient-Pseudometric-Space :
```

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+  compute-unit-map-metric-quotient-Pseudometric-Space X {x} x∈X =
+    eq-set-quotient-equivalence-class-set-quotient
+      ( equivalence-relation-sim-Pseudometric-Space P)
+      ( X)
+      ( x∈X)
+```
+
+## Properties
+
+### The unit map from a pseudometric space its quotient metric space is an isometry
```
fredrik-bakke:
```suggestion
### The unit map is an isometry
```

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+### The unit isometry from a pseudometric space to its quotient metric space
+
+```agda
+module _
+  {l1 l2 : Level} (P : Pseudometric-Space l1 l2)
+  where
+
+  unit-isometry-metric-quotient-Pseudometric-Space :
```
fredrik-bakke:
```suggestion
  isometry-unit-metric-quotient-Pseudometric-Space :
```

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+### The unit short map from a pseudometric space to its quotient metric space
+
+```agda
+module _
+  {l1 l2 : Level} (P : Pseudometric-Space l1 l2)
+  where
+
+  unit-short-map-metric-quotient-Pseudometric-Space :
```
fredrik-bakke:
```suggestion
  short-map-unit-metric-quotient-Pseudometric-Space :
```

fredrik-bakke:
> Does this PR make you see the appeal of having them in their own module?

Sure, yeah, it seems reasonable.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: What do you mean by "the unit". The way I see it, what we define is a lot like the identity in some https://en.wikipedia.org/wiki/Kleisli_category (except we are jumping from `Pseudometric-Space` to `Metric-Space` and pseudometric spaces don't form a category, but the analogy remains). 

This way, the scheme `unit-short-map-{??}-metric-quotient-Pseudometric-Space` or `unit-isometry-{??}-metric-quotient-Pseudometric-Space` follow the names of `id-short-map-Pseudometric-Space`, `id-isometry-Metric-Space`, etc. 
If we find a good name for `{??}` this will also follow the more generic naming scheme `{identifier}-{Type}` (i.e. `zero-ℕ`,  `one-ℚ`, etc.): what we are defining are `unit` in the types `short-map-{??}-metric-quotient-Pseudometric-Space` or `isometry-{??}-metric-quotient-Pseudometric-Space` the very same way `id-isometry-Metric-Space` is the `id` in the type `isometry-Metric-Space`.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: Same here: my idea is that `unit-map-{??}-metric-quotient-Pseudometric-Space` follows the same pattern as `id-map-Pseudometric-Space`: it's the `unit` of the type `map-{??}-metric-quotient-Pseudometric-Space` the very same way `id-map-Pseudometric-Space` is the `id` in the type `map-Pseudometric-Space`.

fredrik-bakke: Yes, there is a discrepancy here, which stems from two different intended readings:

- `unit-map-{??}-metric-quotient-Pseudometric-Space`, reads as "unit as a map relating to metric quotient ..."
- `id-map-Pseudometric-Space` reads as "the identity as a map of pseudometric spaces".

However, there is no underlying idea of a "unit" at play here, we are dealing with the underlying unit **of** the metric quotient, which is qualified as `unit-metric-quotient-...`. **Then** we want to take the underlying _map_ of that so we prepend `map-`. On the other hand, there is no underlying concept of a "`id-Pseudometric-Space`" to take the underlying map of, so we don't write `map-id-Pseudometric-Space`.

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: > This should be the name of the functorial action of metric quotients.

I agree the name is wrong (at the moment I'd vote for `map-in-metric-quotient-Pseudometric-Space`) but I'm not sure _this_ should be the name of the functorial action(s): the name will probably involve `isometry` or `short-map` somewhere in their names.

> It's a bit non-standard to name the types you've defined in this file.

Yet I'll need them further to study the connection between metric quotients and Cauchy pseudocompletions so it would it would help to have good names for these. Could we try to think about it? 

It's seems related to [Kleisli](https://ncatlab.org/nlab/show/Kleisli+category) or [co-Kleisli](https://ncatlab.org/nlab/show/Kleisli+category+of+a+comonad) categories (but it's not so we shouldn't name them that). Haskell seem to call these the [UpStar](https://hackage.haskell.org/package/profunctors-4.3.2/docs/Data-Profunctor.html#t:UpStar) and [DownStar](https://hackage.haskell.org/package/profunctors-4.3.2/docs/Data-Profunctor.html#t:DownStar) of a functor.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: > However, there is no underlying idea of a "unit" at play here,

But there kind of _is_: we'd have three types

- `map-{??}-metric-quotient-Pseudometric-Space`;
- `short-map-{??}-metric-quotient-Pseudometric-Space`;
- `isometry-{??}-metric-quotient-Pseudometric-Space`;

corresponding to maps/short maps/isometries `P -> [Q]` and there are composition operators

```
(Q -> [R]) -> (P -> [Q]) -> P -> [R]
```

for which `unit-{hom}-{??}-metric-quotient-Pseudometric-Space` (for `hom` ∈  {`short-map`, `isometry`}) are "units". (it doesn't quite work for `map` but the naming scheme extends quite nicely).
 
> we are dealing with the underlying unit **of** the metric quotient,

I'm not sure what you mean by that.

> which is qualified as `unit-metric-quotient-...`. **Then** we want to take the underlying _map_ of that so we prepend `map-`. 

well, that's kind of the point. I don't think there's such thing as `unit-metric-quotient-...` but rather `unit-short-map-{??}-metric-quotient-...` and `unit-isometry-{??}-metric-quotient-...`. Then, what _you_ want to call `map-unit-metric-quotient-...` would/could be either `map-unit-short-map-{??}-metric-quotient-...` or `map-unit-isometry-{??}-metric-quotient-...`, the "underlying map of the unit short map (resp. isometry)",  both of type `map-{??}-metric-quotient-...`.

They happen to be **the same map**, which _could_ be called `map-unit-map-{??}-metric-quotient-...` but it's obviously redundant since "the underlying map of the unit map" is just the map itself: hence `unit-map-{??}-metric-quotient-...`.

> On the other hand, there is no underlying concept of a "`id-Pseudometric-Space`" to take the underlying map of,

I believe the situation is similar here: there's no concept of `unit-metric-quotient-...` to take the underlying map of. There _are_ concepts of `{hom}-{??}-metric-quotient-Pseudometric-Space` for which we define units.

> so we don't write `map-id-Pseudometric-Space`.

Indeed. We _could_ write `map-id-map-Pseudometric-Space`, "the underlying map of the identity map of a pseudometric space", but it would be redundant too.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: Oh. I think I understand. Do you mean there **is** a concept of `unit-metric-quotient-Pseudometric-Space` from which we can take either the underlying `map`, `short-map`, or `isometry`?

fredrik-bakke: Yes.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
fredrik-bakke: > They happen to be **the same map**, which _could_ be called `map-unit-map-{??}-metric-quotient-...` but it's obviously redundant since "the underlying map of the unit map" is just the map itself: hence `unit-map-{??}-metric-quotient-...`.

Exactly, so the naming system is designed to avoid this redundancy in the names too.

fredrik-bakke: > I believe the situation is similar here: there's no concept of `unit-metric-quotient-...` to take the underlying map of. There _are_ concepts of `{hom}-{??}-metric-quotient-Pseudometric-Space` for which we define units.

I believe there is. I suppose `unit-metric-quotient-...` would have the type of being a *unit for an adjunction. But you don't have to construct that to name these concepts.

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
fredrik-bakke: > Yet I'll need them further to study the connection between metric quotients and Cauchy pseudocompletions so it would it would help to have good names for these. Could we try to think about it?

It's a very specific type. Is it so important you cannot write out

```agda
    map-Pseudometric-Space
      ( P)
      ( pseudometric-metric-quotient-Pseudometric-Space Q)
```

every time you need it?

I worry that this sort of definition will cause too much bloat if we start doing this by convention. And it seems to me it will make reading type signatures more opaque unless we'd make it very clear from the name that we only apply the metric quotient to `Q`.

fredrik-bakke: In short, I think this would set a bad precedence.

fredrik-bakke: > It's seems related to [Kleisli](https://ncatlab.org/nlab/show/Kleisli+category) or [co-Kleisli](https://ncatlab.org/nlab/show/Kleisli+category+of+a+comonad) categories (but it's not so we shouldn't name them that). Haskell seem to call these the [UpStar](https://hackage.haskell.org/package/profunctors-4.3.2/docs/Data-Profunctor.html#t:UpStar) and [DownStar](https://hackage.haskell.org/package/profunctors-4.3.2/docs/Data-Profunctor.html#t:DownStar) of a functor.

Right, because Kleisli categories describe derivative structures from a monad, and monads arise from adjunctions, and the metric quotient is the *left adjoint to the "underlying pseudometric space of a metric space" construction.

fredrik-bakke: You have an idempotent moand on pseudometric spaces, which creates a kleisli precategory (up to truncatedness *or extensionality issues) and so on.

src/metric-spaces/cauchy-approximations-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: > Yes.

Ok. Can't say I really understand the logic at play here but I don't want to make you lose more time. I'll do my best to follow your recommendations.

src/metric-spaces/unit-map-metric-quotients-of-pseudometric-spaces.lagda.md
malarbol: yes. One problem with (pre)categories and pseudometric spaces is that the Hom-~sets~ are not sets. But if we consider isometries / short maps `P -> [Q]`, the codomain is a set so these morphisms are the Hom-sets of some  precategories. 
So there are (large?) precategories where objects are pseudometric spaces and hom-sets are short-maps / isometries `P -> [Q]`. I wonder if they are categories; I'd have to think a bit more about this. If I were to formalize this, what would be a good name for it?

malarbol: And I guess we could also consider the precategories where objects are pseudometric spaces and hom-sets are isometries / short maps `[P] -> [Q]`.

malarbol:
> it seems to be that it's best to delete the definitions `(short-)map-metric-quotient-Pseudometric-Space` and so on, as they conflict with the functorial actions, and they don't seem particularly useful.

I guess I spent too much time thinking about #1458 and I'm getting lost between what's useful and what's not. I'll try to refocus on the important topics. 

Thanks a lot for your patience and explanations.

PR #1794, opened 2026-01-11
Unbounded above and below, strictly increasing, pointwise epsilon-delta continuous endomaps on the real numbers are automorphisms

lowasser:
Broken off and redone from #1739.

PR #1796, opened 2026-01-14
Rewrite multiplicative inverses of positive real numbers

lowasser:
The definition becomes pretty simple with #1794, and all the useful properties follow.

### DIFF
--- a/src/real-numbers/multiplicative-inverses-positive-real-numbers.lagda.md
+++ b/src/real-numbers/multiplicative-inverses-positive-real-numbers.lagda.md
@@ -216,7 +216,7 @@ abstract
               by ap-mul-ℝ⁺ (eq-ℝ⁺ _ _ (inv (eq-raise-sim-ℝ x~y))) refl)))
 ```
 
-### The embedding of positive rational numbers in the positive real numbers preserves multiplicative inverses
+### The inclusion of positive rational numbers in the positive real numbers preserves multiplicative inverses
 
 ```agda
 abstract


src/elementary-number-theory/multiplicative-group-of-positive-rational-numbers.lagda.md
```diff
@@ -253,3 +266,24 @@ abstract
   inv-one-ℚ⁺ : inv-ℚ⁺ one-ℚ⁺ ＝ one-ℚ⁺
   inv-one-ℚ⁺ = inv-unit-Group group-mul-ℚ⁺
 ```
+
+### Multiplication by a positive rational number is unbounded above and below
+
+```agda
+abstract
+  is-unbounded-above-left-mul-rational-ℚ⁺ :
```
fredrik-bakke, line 274: I'm noting that the incorrect terminology "unbounded above" and "unbounded below" is used in this PR as well, but let's fix it in another common PR.

lowasser: I do claim that it is at least applicable in ucnbounded posets like Q and R.

fredrik-bakke: That may be true, but it is best that we name things by what they encode rather than what they're equivalent to. It should be a theorem rather than a definition that cofinality is equivalent to unboundedness from above.

PR #1806, opened 2026-01-31
Cauchy pseudocompletions of indexed sums of pseudometric spaces

malarbol:
The Cauchy pseudocompletion of an indexed sum is the indexed sum of the Cauchy pseudocompletions.

This PR introduces the concept of **indexed sums of pseudometric spaces** (same as _indexed sums of metric spaces_ without extensionality).

Then, we define:
- **Cauchy pseudocompletions of indexed sums**;
- **indexed sums of Cauchy pseudocompletions**:

```text
module _
  {la lp lp' : Level}
  (A : Set la)
  (P : type-Set A → Pseudometric-Space lp lp')
  where

  cauchy-pseudocompletion-indexed-sum-Pseudometric-Space :
    Pseudometric-Space (la ⊔ lp ⊔ lp') (la ⊔ lp')
  cauchy-pseudocompletion-indexed-sum-Pseudometric-Space =
    cauchy-pseudocompletion-Pseudometric-Space
      ( indexed-sum-Pseudometric-Space A P)

[...]

  indexed-sum-cauchy-pseudocompletion-Pseudometric-Space :
    Pseudometric-Space (la ⊔ lp ⊔ lp') (la ⊔ lp')
  indexed-sum-cauchy-pseudocompletion-Pseudometric-Space =
    indexed-sum-Pseudometric-Space A
      ( cauchy-pseudocompletion-Pseudometric-Space ∘ P)
```

Finally, we construct an isometric equivalence between them to conclude these are _equal_ pseudometric spaces.


malarbol:
To put this PR in context, this is also related to #1458 :

If `F` denotes the _metric quotient of the Cauchy pseudocompletion_, `F X = [C X]`,
for any pseudometric space `X` we have a sequence of isometries of metric spaces:

```
[X] → F[X] → F²[X] → F³[X] → ... → Fⁿ[X] → ...

```
and for any metric space `M`, if `M` is complete, this sequence naturally extends to `M`:
```
[X] → F[X] → F²[X] → F³[X] → ... → Fⁿ[X] → ... → M

```
and a short map (not an isometry)
```
Σ (n : ℕ) (Fⁿ[X]) → M
```

So the metric space `Σ (n : ℕ) (Fⁿ[X])` seems relevant somehow in the construction of an hypothetical Cauchy completion of `X` and I'm trying to understand the construction
```
ΣF[X] = Σ (n : ℕ) (Fⁿ[X])
```
better.


For example, this PR implies that
```
C (ΣF[X]) ≃ Σ (n : ℕ) (C Fⁿ[X])

```
and since each `C Fⁿ[X]` can be mapped into `Fⁿ⁺¹[X]`, this will ultimately give us a map
```
C (ΣF[X]) → ΣF[X].
```

Unfortunately, this is not a _limit map_: composing with the unit map:

```
ΣF[X] → C (ΣF[X]) → ΣF[X]
```
is not the identity, but the _shifting map_  `ΣF[X] → ΣF[X]`. But this still seems interesting.

So, I'm not 100% sure of where I'm going, but there seem to be something worth exploring there.
This PR is part of that exploration.

PR #1808, opened 2026-02-02
Quotients of modules and vector spaces

lowasser:
I don't go through the universal property yet, but the construction in terms of quotient groups is perfectly sensible.


fredrik-bakke:
You might be interested in the following paper, on how a formalisation of quotient (localisation) rings was a total failure because they didn't formalise them using universal properties:
- https://arxiv.org/abs/2101.02602

lowasser:
Planning to redo this with the universal-algebra strategy, since that has quotients and even their universal properties well handled.

Issue #1699, opened 2025-11-11
Non-denumerability of the continuum


lowasser:
Theorem 22 of the 100 Theorems.  This seems like a pain, but only a pain.

fredrik-bakke:
As far as I know this theorem requires some classicality axiom by the way: https://arxiv.org/abs/2404.01256

lowasser:
I'm uncertain.  I am mostly planning to go from Bishop's proof in Foundations of Constructive Analysis, and I don't see any choice principles in that proof yet; he accepts countable choice in general, so it might be hiding in there and I haven't spotted it yet.

fredrik-bakke:
This seems likely. In particular, the counter model in the cited article refutes countable choice.

lowasser:
I see now that it uses dependent choice, actually.

fredrik-bakke:
The cited article demonstrates that ~the Cauchy reals and~ the MacNeille reals are constructively uncountable. I think formalizing ~either of these two facts~ *this fact should be satisfactory for this issue. ~If you want to be extra rigorous, it can additionally be demonstrated that the Dedekind reals coincide with the Cauchy reals assuming countable choice.~

fredrik-bakke:
Correction: for the Cauchy reals it is only demonstrated that they are uncountable in a particular model.

PR #1816, opened 2026-02-06
Conditional order completeness of the MacNeille reals

fredrik-bakke:
Shows that every inhabited upper bounded family of MacNeille reals has a least upper bound. Follows Sketches of an Elephant Lemma D4.7.7. This is work towards #1699.

Builds on #1830.


fredrik-bakke:
I factored out the basic definition to #1830, so that has to be merged before this PR now.

fredrik-bakke:
I got around to writing exposition for the order completeness now.

PR #1817, opened 2026-02-06
Located MacNeille reals

fredrik-bakke:
Shows equivalences between the MacNeille and Dedekind reals.

Builds on #1830 #1816. Builds towards #1699.


lowasser:
I find myself wondering if we should simply refactor Dedekind reals to _be_ located MacNeille reals.

fredrik-bakke:
> I find myself wondering if we should simply refactor Dedekind reals to _be_ located MacNeille reals.

That sounds like a lot of work!

lowasser:
I think it's a pretty doable amount of work, though I can take it on.  I think almost nothing deconstructs the Dedekind reals through tuples, except the accessors built in.  Several places construct reals through the pair constructor, but I'm pretty certain I can count them on my fingers, and there it would just be shuffling parentheses.

Off the top of my head, places we construct reals directly include:

- Cauchy completeness
- embedding of rationals
- the inverse of strictly increasing functions done recently
- the square root function, which could be made into a slightly more general version of inverting strictly increasing functions
- addition, multiplication, negation
- binary maximum 

...and I can't think of any others.

PR #1819, opened 2026-02-07
Suspensions of projective types

fredrik-bakke:
Shows that suspensions of 𝑘-projective 𝑘-truncated types are (𝑘+1)-projective and that suspensions of projective types are 1-projective.


fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #1821, opened 2026-02-07
Fix definition of cotransitivity

### LINKED ISSUE #1803: Definition of cotransitivty is wrong
Our current definition of cotransitivity states that if `R a b` then `R a c` or `R b c` for all `c`, but this condition is only correct for symmetric orders.

fredrik-bakke:
Resolves #1803.


fredrik-bakke:
Unless I get any objections I'll go ahead and merge this in a couple of days.

PR #1826, opened 2026-02-08
Sequence-avoiding sets

fredrik-bakke:
Builds towards #1699.


src/set-theory/sequence-avoiding-sets.lagda.md
```diff
@@ -0,0 +1,145 @@
...
+```agda
+module set-theory.sequence-avoiding-sets where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import elementary-number-theory.natural-numbers
+
+open import foundation.dependent-pair-types
```
ElifUskuplu, line 12: `Π-Prop` moved to `foundation.dependent-products-propositions` in #1373, after this file was written, so it is no longer in `foundation.propositions`.

```suggestion
open import foundation.dependent-pair-types
open import foundation.dependent-products-propositions
```

src/set-theory/sequence-avoiding-sets.lagda.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+### Weakly sequence-avoiding sets
+
+```agda
+weakly-avoids-sequence :
+  {l : Level} (X : Set l) → (ℕ → type-Set X) → UU l
+weakly-avoids-sequence X a =
+  is-surjective a →
+  ¬¬ type-Prop (∃ (type-Set X) (avoids-sequence-Prop X a))
```
ElifUskuplu, line 70: I didn't get the point in this definition. If `a` is surjective, then no element can avoid it: any `x` is `a n` for some `n`, which contradicts `x ≠ a n`. So under the hypothesis `is-surjective a`, the proposition `∃ x, avoids-sequence x a` is false, and its double negation is false as well. That makes

```text
weakly-avoids-sequence X a  ↔  ¬ is-surjective a
```

and hence

```text
is-weakly-sequence-avoiding X  ↔  ¬ is-directly-countable X
```

The proof of `is-not-directly-countable-is-weakly-sequence-avoiding` is essentially this unfolding, and the converse direction holds for the same reason.

Is there a reason to have this as a separately named notion, rather than stating the results directly in terms of `¬ is-directly-countable`? I may be missing an intended use. As far as I can tell, Bauer–Hanson only define the strong notion, and #1840 only uses `is-sequence-avoiding`.

PR #1831, opened 2026-02-09
Variants of the classical König's theorem

fredrik-bakke:
Shows the classical formulation of König's theorem assuming WLPO.

Builds on #1834 #1720 #1604 #1591 #1846 #1853 #2001 #2002 #2003 #2004. 

It remains to refactor some of the work still, but the main results are already formalised.


src/foundation/projective-types.lagda.md
```diff
@@ -11,6 +11,7 @@ open import elementary-number-theory.addition-natural-numbers
 open import elementary-number-theory.natural-numbers
 
 open import foundation.connected-maps
+open import foundation.connected-types
```
fredrik-bakke:
```suggestion
```

PR #1834, opened 2026-02-10
Types with decidable existential quantification

fredrik-bakke:
Develops the infrastructure for types with decidable existential quantification, and fills a few missing holes regarding the untruncated variant. Notably that the subuniverse of propositions has decidable sums.


fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #1836, opened 2026-02-11
Sums and differences of natural numbers

lowasser:
Trying to make both sums and differences for natural numbers more closely resemble the infrastructure for other number types.

### DIFF
--- a/src/foundation.lagda.md
+++ b/src/foundation.lagda.md
@@ -106,6 +106,7 @@ open import foundation.coproducts-pullbacks public
 open import foundation.coslice public
 open import foundation.cospan-diagrams public
 open import foundation.cospans public
+open import foundation.cumulative-large-sets public
 open import foundation.decidable-dependent-function-types public
 open import foundation.decidable-dependent-pair-types public
 open import foundation.decidable-embeddings public
@@ -417,6 +418,10 @@ open import foundation.sets public
 open import foundation.sigma-closed-subuniverses public
 open import foundation.sigma-decomposition-subuniverse public
 open import foundation.sigma-decompositions public
+open import foundation.similarity-preserving-binary-maps-cumulative-large-sets public
+open import foundation.similarity-preserving-binary-maps-large-similarity-relations public
+open import foundation.similarity-preserving-maps-cumulative-large-sets public
+open import foundation.similarity-preserving-maps-large-similarity-relations public
 open import foundation.similarity-subtypes public
 open import foundation.singleton-induction public
 open import foundation.singleton-subtypes public
--- a/src/foundation/cumulative-large-sets.lagda.md
+++ b/src/foundation/cumulative-large-sets.lagda.md
@@ -1,0 +1,442 @@
+# Cumulative large sets
+
+```agda
+module foundation.cumulative-large-sets where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.action-on-identifications-functions
+open import foundation.dependent-pair-types
+open import foundation.embeddings
+open import foundation.identity-types
+open import foundation.injective-maps
+open import foundation.large-binary-relations
+open import foundation.large-equivalence-relations
+open import foundation.large-similarity-relations
+open import foundation.logical-equivalences
+open import foundation.sets
+open import foundation.universe-levels
+```
+
+</details>
+
+## Idea
+
+Many structures in mathematics are universe polymorphic in nature, which calls
+for some way to compare elements at different universe levels.
+[Identity types](foundation-core.identity-types.md) fall short with this respect
+because elements at different universe levels don't inhabit the same type. To
+solve this issue we have two standard mechanisms: raising universe levels of
+elements, and
+[large similarity relations](foundation.large-similarity-relations.md).
+
+A {{#concept "cumulative large set" Agda=Cumulative-Large-Set}} captures such a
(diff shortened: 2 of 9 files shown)


src/elementary-number-theory/difference-natural-numbers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+    is-prop-all-elements-equal (all-elements-equal-type-subtraction-ℕ n m)
+
+subtraction-prop-ℕ : Relation-Prop lzero ℕ
+subtraction-prop-ℕ n m =
+  ( type-subtraction-ℕ n m , is-prop-type-subtraction-ℕ n m)
+```
+
+## Properties
+
+### We have `n ≤ m` if and only if there is a number `l` such that `l+n=m`
```
fredrik-bakke:
```suggestion
### We have `n ≤ m` if and only if there is a number `l` such that `l + n = m`
```

src/elementary-number-theory/difference-natural-numbers.lagda.md
```diff
@@ -0,0 +1,139 @@
...
+          by
+            left-add-diff-leq-ℕ
+              ( k +ℕ m)
+              ( k +ℕ n)
+              ( preserves-leq-right-add-ℕ k n m n≤m)
+        ＝ k +ℕ (diff-leq-ℕ m n n≤m +ℕ n)
+          by ap-add-ℕ refl (inv (left-add-diff-leq-ℕ m n n≤m))
+        ＝ diff-leq-ℕ m n n≤m +ℕ (k +ℕ n)
+          by left-swap-add-ℕ k (diff-leq-ℕ m n n≤m) n)
+```
```
fredrik-bakke, line 145: See also distance of natural numbers

fredrik-bakke: and please show that diff and dist agree

fredrik-bakke: maybe also mention in the idea section that that in the agda-unimath library we prefer the distance function because it is a total binary function

fredrik-bakke (approved):

src/elementary-number-theory/difference-natural-numbers.lagda.md
```diff
@@ -0,0 +1,155 @@
...
+Given two [natural numbers](elementary-number-theory.natural-numbers.md) `a` and
+`b`, there is a difference `a - b` such that the
+[sum](elementary-number-theory.addition-natural-numbers.md) of `a - b` and `b`
+is `a` if and only if `b` is
+[less than or equal to](elementary-number-theory.inequality-natural-numbers.md)
+`a`.
+
+The difference of `a` and `b` is equivalent to the
+[distance](elementary-number-theory.distance-natural-numbers.md) between `a` and
+`b`, but is only defined when `b ≤ a`.
```
fredrik-bakke:
```suggestion
**Note.** The difference of `a` and `b` is equivalent to the
[distance](elementary-number-theory.distance-natural-numbers.md) between `a` and
`b`, but is only defined when `b ≤ a`. For this reason, it is often preferred to use
the distance function over the difference function on natural numbers.
```

src/elementary-number-theory/difference-natural-numbers.lagda.md
```diff
@@ -0,0 +1,155 @@
...
+### Where defined, the distance and difference of natural numbers agree
+
+```agda
+abstract
+  eq-diff-dist-leq-ℕ :
+    (m n : ℕ) (n≤m : leq-ℕ n m) → diff-leq-ℕ m n n≤m ＝ dist-ℕ m n
+  eq-diff-dist-leq-ℕ 0 0 star = refl
+  eq-diff-dist-leq-ℕ (succ-ℕ m) 0 star = refl
+  eq-diff-dist-leq-ℕ (succ-ℕ m) (succ-ℕ n) m≤n = eq-diff-dist-leq-ℕ m n m≤n
+```
```
fredrik-bakke, line 156:
````suggestion
```

## See also

- [The distance between natural numbers](elementary-number-theory.distance-natural-numbers.md), which is a total function substitute for the difference function considered on this page
````

lowasser: I don't tend to explicitly add a see also when it's already linked from the Idea section, which it is now.

fredrik-bakke: I think it is reasonable to have in case the reader misses it in the idea section, since we generally don't want them to use the difference function.

fredrik-bakke:
I think we should also add something to the sentiment of that the difference file is only there for completeness and that it is not meant to be built on

fredrik-bakke:
Or maybe that's too strict

lowasser:
I'm not at all inclined to say that.  It's not universally defined, but that doesn't make it unuseful, and I have specific use cases in mind for it.

fredrik-bakke:
Okay okay I'm being a bit overzealous. I edited the notice suggestion. I still think we should err on the side of overabundance that the distance function is an alternative, since wrong implementations of the difference function on natural numbers are widespread. E.g. the difference function in mathlib and the builtin difference function in Agda, just to name the ones that immediately come to mind.

fredrik-bakke:
Also this is like the 4th time I'm having the discussion on differences of natural numbers (you give the first good implementation of it).

PR #1837, opened 2026-02-11
Prove Mather's second cube theorem

fredrik-bakke:
This might be my least proud work yet, but hey, it typechecks!


fredrik-bakke:
Closing this as I want to insepct a second proof route first.

PR #1840, opened 2026-02-13
Uncountability of the MacNeille reals

### LINKED ISSUE #1699: Non-denumerability of the continuum

fredrik-bakke:
Proves that the MacNeille reals form an uncountable set following a predicative and constructive version of Levy's "unusual proof of uncountability of the reals". As far as I can tell a predicative proof is novel.

There's still some cleanup to do, but the main argument is already here and typechecks.

Resolves #1699.

Builds on #1830 #1816 #1817 #1820 #1822 #1824 #1825 #1826 #1827 #1839.


fredrik-bakke:
This development has made me think that "addition of MacNeille reals" should be renamed to "translation of MacNeille reals by located MacNeille reals", since that is all it achieves.

PR #1843, opened 2026-02-15
Unital and associative algebras over a commutative ring

lowasser:


fredrik-bakke:
Thank you for these additions!

PR #1846, opened 2026-02-16
𝑘-projective 𝑘-types are (𝑘 + 𝑛)-projective for all 𝑛

fredrik-bakke:
Also refactors some of the basic infrastructure surrounding projective types.

PR #1853, opened 2026-02-19
Distributivity of truncation over projective types

fredrik-bakke:
Shows that set-truncation distributes over projective types, and derives the finite case as a direct corollary. Also gives a weaker result for distributivity of truncation over truncation-projective types.

Builds on #1846.


fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #1898, opened 2026-03-18
The construction of freely generated algebras satisfies the universal property

lowasser:
Defines the universal property of freely generated algebras and shows that the construction in #1896 satisfies it.


lowasser:
Depends on #1900 and #1896.

lowasser:
Folding into #1896 .

PR #1818, opened 2026-02-06
Universal property of Cauchy pseudocompletions and short maps of pseudometric spaces 

malarbol:
This PR introduces the concept of **precomplete** short maps `f : P → M` from a pseudometric space in a metric space: short maps such that for all Cauchy approximation `u : C P`, `Cf u` is convergent in `M`. Precomplete short maps extend along the unit map of Cauchy pseudocompletions `κ : P → C P`: if `f` is precomplete, there exists some `g : C P → M` such that `f ~ g ∘ κ`.

Then, we prove the universal property of Cauchy pseudocompletions of pseudometric spaces w.r.t. short maps to metric spaces:

Given a metric space `M` and a pseudometric space `P`, precomposition with the unit map of Cauchy pseudocompletions `κ : P → C P` maps short maps `g : C P → M` to short maps `g ∘ κ : P → M`. For any Cauchy approximation `u : C P`, its pointwise image `C(g ∘ κ) u : C M` converges  to `g u` so `g` is determined by its restriction to `P`, `g ∘ κ`, which is precomplete. Therefore, extensions of short maps `f : P → M` along the unit map of Cauchy pseudocompletions `κ : P → C P` are unique and exist iff `f` is **precomplete**.

In particular, a metric space `M` is _complete_ if and only if the unit map `κ : M → C M` has a short retraction, i.e. a short map `lim : C M → M` such that `lim ∘ κ ~ id`.

PR #1902, opened 2026-03-18
Precomplete isometries from pseudometric spaces to metric spaces

malarbol:
Follow up on #1818 

This PR defines the notion of **precomplete isometries** from a pseudometric space to a metric space, isometries for which the underlying short map is _precomplete_.


malarbol:
This will be followed by the corresponding universal property, as in #1818, precomplete isometries **are exactly** the isometries that extend as isometries from the Cauchy pseudocompletions. 

I'll do it in another PR, to keep them smallish.

PR #1906, opened 2026-03-19
The external direct sum of abelian groups

lowasser:
Demonstrating what the free generation and quotients in `universal-algebra` can do, really, but also useful in its own right, and it feels like the right construction.


lowasser:
Closing in favor of an even more general direct sum construction of algebras.

PR #1910, opened 2026-03-21
The algebraic theory of monoids

lowasser:
Continues the work from #1899 , but these files are big enough they might as well be separate PRs.


lowasser:
Folding into #1912.

PR #1911, opened 2026-03-21
Refactor the algebraic theory of groups

lowasser:
Modernize the naming schema and build it as an extension from the monoid theory in #1910 .


lowasser:
Folding into #1912.

PR #1933, opened 2026-04-01
Refactor tuples

lowasser:
Organize things into separate files, and pull out definitions that were in random other files -- `fold-tuple` lived in `arrays`?.


lowasser:
We do add one new semantically interesting result, that `reverse-tuple` is an involution.

PR #1935, opened 2026-04-17
Matrix overhaul

lowasser:
DO NOT SUBMIT: to be broken up into several smaller PRs; this PR exists to prototype the entire whole and to show the long-term goal of the sub-PRs.

This PR:

- renames the existing tuple-based `matrix` concept to "grid"
- defines a "matrix" concept around finite sequences instead of tuples
- gets addition and scalar multiplication on matrices "for free" from the dependent product of modules (one of several reasons to use finite sequences)
- defines square matrices, diagonal matrices, and identity matrices
- defines matrix multiplication and shows its properties (which never happened for the original tuple-based matrix concept)
- shows that square matrices on commutative rings form a unital associative algebra

That's probably enough to start with; matrix similarity, diagonalizability, determinants, traces, and more can come next...


src/finite-group-theory/counting-permutations-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,273 @@
...
+
+### `Permutation n` is finite
+
+```agda
+finite-type-Permutation : (n : ℕ) → Finite-Type lzero
+finite-type-Permutation n =
+  ( Permutation n , is-finite-count (count-Permutation n))
+```
+
+### The number of elements of `Permutation n` is n
```
malarbol, line 94: The `!` is missing in `n!`.

PR #1936, opened 2026-04-18
The type of permutations on n elements is finite with cardinality n!

lowasser:
This implies that it can be summed over, which will allow us to define determinants.


src/finite-group-theory/counting-permutations-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,297 @@
...
+the [standard finite type](univalent-combinatorics.standard-finite-types.md)
+with `n` elements are a [finite type](univalent-combinatorics.finite-types.md)
+with `n` [factorial](elementary-number-theory.factorials.md) elements.
+
+## Proof
+
+### `Permutation (n + 1) ≃ Fin (n + 1) × Permutation n`
+
+```agda
+ap-lemma :
```
lowasser: Surely this is a lemma elsewhere, but I couldn't find it...

src/finite-group-theory/transpositions.lagda.md
lowasser, line 1: The only diffs here are making things abstract to improve compilation.

src/finite-group-theory/counting-permutations-standard-finite-types.lagda.md
```diff
@@ -0,0 +1,273 @@
...
+      pr1 x
+    pr1-is-section-map-inv-equiv-succ-Permutation (k , σ) =
+      map-swap-left-permutation-Fin (succ-ℕ n) (inr star) k
+
+    pr2-is-section-map-inv-equiv-succ-Permutation' :
+      (k : Fin (succ-ℕ n)) (σ : Permutation n) →
+      htpy-equiv
+        ( left-cancel-head-Permutation (map-inv-equiv-succ-Permutation (k , σ)))
+        ( σ)
+    pr2-is-section-map-inv-equiv-succ-Permutation' k σ i =
```
malarbol: `ap-lemma` is a composition of `ap`, `ind-Σ`, and `eq-pair-Σ` so this could be refactored as
```
      ap
        ( ind-Σ left-is-left)
        ( eq-pair-Σ
          ( ( ap
              ( λ j →
                map-swap-permutation-Fin
                  ( succ-ℕ n)
                  ( inr star)
                  ( j)
                  ( map-swap-permutation-Fin
                    ( succ-ℕ n)
                    ( inr star)
                    ( k)
                    ( inl (map-equiv σ i))))
                ( map-swap-left-permutation-Fin (succ-ℕ n) (inr star) k)) ∙
            ( is-involution-map-swap-permutation-Fin
              ( succ-ℕ n)
              ( inr star)
              ( k)
              ( inl (map-equiv σ i))))
          ( eq-is-prop (is-prop-is-left {Y = unit} (inl (map-equiv σ i)))))
```

EgbertRijke:
I made some minor modifications in this PR, while reviewing it, just a heads-up. Still working on it.

EgbertRijke:
I noticed that there should be a much cleaner way of handling the count of automorphisms of finite types, which should be parametric (using arbitrary types as arguments), rather than trying to rely entirely on the construction for the standard finite types. By doing so, we can probably improve a lot on the old code as well.

I'm trying to implement it over the next couple of days.

EgbertRijke:
I have finished the equivalence `Aut A ≃ Aut C × isolated-element A`. The next step is to lift this equivalence to arbitrary types with decidable equality. After that, the code originally proposed in this pull request can be significantly simplified.

PR #1938, opened 2026-04-22
Increasing arrays in posets

lowasser:
Finally attempting to get Riemann integrals up and running, and this is the first step.


src/order-theory/increasing-finite-sequences-posets.lagda.md
```diff
@@ -0,0 +1,173 @@
...
+
+The
+[standard ordering](elementary-number-theory.inequality-standard-finite-types.md)
+on the [standard finite types](univalent-combinatorics.standard-finite-types.md)
+defines `neg-one-Fin n` as the greatest element of `Fin (succ-ℕ n)`, but finite
+sequences use `neg-one-Fin n` as the head, which is conventionally the first
+element of the sequence.
+
+We choose to adopt the convention that the head of an increasing sequence is its
+least element, requiring us to reverse the order of `Fin n`.
```
malarbol: Maybe we should re-factor the `finite-sequences` module to use `zero-Fin n` as head instead of `neg-one-Fin n`? So we don't have to switch endianness back and forth like this?

malarbol: Or maybe define `is-increasing-prop-fin-sequence-type-Poset` as an increasing finite sequence?

If I understand correctly you aim to define refinements of intervals `a = a₀ ≦ a₁ ≦ ... ≦ aₙ = b` and I don't think the double-reversing of orderings is necessary.

lowasser: > Maybe we should re-factor the `finite-sequences` module to use `zero-Fin n` as head instead of `neg-one-Fin n`? So we don't have to switch endianness back and forth like this?

I have to say I don't love this idea, because the ability to define `cons` in terms of `rec-coproduct` is nice.  Honestly, I think _right_ thing to do would be to flip the built-in ordering and the names, e.g. `zero-Fin n = inr star`.  But that would require pulling up a lot more stuff throughout the library than I feel up to right now.

But I don't think we're double-reversing orderings; we're just single-reversing orderings.

malarbol: > > Maybe we should re-factor the `finite-sequences` module to use `zero-Fin n` as head instead of `neg-one-Fin n`? So we don't have to switch endianness back and forth like this?
> 
> I have to say I don't love this idea, because the ability to define `cons` in terms of `rec-coproduct` is nice. Honestly, I think _right_ thing to do would be to flip the built-in ordering and the names, e.g. `zero-Fin n = inr star`. But that would require pulling up a lot more stuff throughout the library than I feel up to right now.
> 

ok. I agree changing to `zero-Fin n = inr star` would be too much work and I don't think that's worth it.

> But I don't think we're double-reversing orderings; we're just single-reversing orderings.

maybe not. But why reversing in the first place?

The convention `head v = v (neg-one-Fin n)` just means that `head` is the right-most element, the element with larger index; so the `head` of `(a₀ , a₁ , ... , aₙ)` is `aₙ`.  Why do you also want the `head` to be the smallest element of a partition of an interval?

What you are describing in this file are actually `decreasing-fin-sequence-type-Poset`.

src/lists/finite-sequences.lagda.md
```diff
@@ -54,12 +54,19 @@ module _
   empty-fin-sequence ()
 
   head-fin-sequence : (n : ℕ) → fin-sequence A (succ-ℕ n) → A
-  head-fin-sequence n v = v (inr star)
+  head-fin-sequence n v = v (neg-one-Fin n)
+
+  last-fin-sequence : (n : ℕ) → fin-sequence A (succ-ℕ n) → A
+  last-fin-sequence n v = v (zero-Fin n)
```
malarbol: Or maybe that's where the problem is. the `head` **is** the last element of a `fin-sequence`.

src/order-theory/increasing-finite-sequences-posets.lagda.md
```diff
@@ -0,0 +1,173 @@
...
+## Definition
+
+```agda
+module _
+  {l1 l2 : Level}
+  (P : Poset l1 l2)
+  where
+
+  is-increasing-prop-fin-sequence-type-Poset :
+    (n : ℕ) → subtype l2 (fin-sequence (type-Poset P) n)
```
malarbol, line 51: Why not
```
is-increasing-prop-fin-sequence-type-Poset n = 
  preserves-order-prop-Poset
    ( Fin-Poset n)
    ( P)
```
and have the `head` be the largest element. So you can prove things like "an array is increasing iff its tail is increasing and its head it the top-element" etc.

src/elementary-number-theory/inequality-standard-finite-types.lagda.md
```diff
@@ -145,6 +146,18 @@ abstract
...
+### The reverse embedding of the standard finite types in the natural numbers reverses inequality
+
+```agda
+abstract
+  reverses-leq-nat-Fin-reverse :
+    (k : ℕ) (x y : Fin k) → leq-Fin k x y →
+    leq-ℕ (nat-Fin-reverse k y) (nat-Fin-reverse k x)
+  reverses-leq-nat-Fin-reverse (succ-ℕ k) x (inr star) x≤y = star
+  reverses-leq-nat-Fin-reverse (succ-ℕ k) (inl x) (inl y) x≤y =
+    reverses-leq-nat-Fin-reverse k x y x≤y
```
malarbol: > But I don't think we're double-reversing orderings; we're just single-reversing orderings.

Sorry. _This_ was what I was referring to.

src/order-theory/increasing-finite-sequences-posets.lagda.md
EgbertRijke: I think this is a good suggestion. It is better to reuse the existing notion of order-preserving map rather than to reinvent the wheel

src/order-theory/increasing-finite-sequences-posets.lagda.md
lowasser: Refactored to this approach.

src/elementary-number-theory/inequality-standard-finite-types.lagda.md
```diff
@@ -145,6 +146,18 @@ abstract
   reflects-leq-nat-Fin (succ-ℕ k) {inr star} {inr star} H = star
 ```
 
+### The reverse embedding of the standard finite types in the natural numbers reverses inequality
+
+```agda
+abstract
+  reverses-leq-nat-Fin-reverse :
```
malarbol: this could also be called `is-decreasing-nat-Fin-reverse`

src/elementary-number-theory/inequality-standard-finite-types.lagda.md
```diff
@@ -168,3 +181,26 @@ linear-leq-Fin (succ-ℕ k) (inl x) (inl y) = linear-leq-Fin k x y
 linear-leq-Fin (succ-ℕ k) (inl x) (inr y) = inl star
 linear-leq-Fin (succ-ℕ k) (inr x) y = inr star
 ```
+
+### `skip-zero-Fin` preserves inequality
+
+```agda
+abstract
+  preserves-order-inr-Fin :
```
malarbol, line 190: It feels like this name should refer `skip-zero-Fin` somehow. Maybe `preserves-order-skip-zero-Fin`, or `is-increasing-skip-zero-Fin` ?

malarbol: or `leq-skip-zero-Fin`?

lowasser: `inr-Fin` and `skip-zero-Fin` are synonyms; maybe we should standardize on one?

lowasser: In the meantime, fixed the docs.

PR #1939, opened 2026-04-22
Partitions of real intervals

lowasser:
Step 2 on the way to Riemann integrals, building on #1938.


malarbol:
> Step 2 on the way to Riemann integrals

What kind of Riemann integrals do you have in mind? one of 
- https://en.wikipedia.org/wiki/Riemann_integral;
- https://en.wikipedia.org/wiki/Riemann–Stieltjes_integral;
- https://en.wikipedia.org/wiki/McShane_integral ?

lowasser:
I'm not yet certain about Riemann vs. Riemann-Stieltjes, but one of those for sure.

lowasser:
Ah, I understand the distinction now.  I meant Riemann.

PR #1940, opened 2026-04-23
Sums in normed real vector spaces

lowasser:
Comes with a lot of other pieces all over `linear-algebra`, but they're mostly very formulaic (copying properties from the next layer of abstraction down).


lowasser:
This can probably be broken up further if it'd help.

PR #1941, opened 2026-04-23
Break out real-analysis from analysis

lowasser:
This kept feeling a little weird.


EgbertRijke:
These are very sensible changes. Thanks!

PR #1946, opened 2026-04-30
Linear maps between normed vector spaces are bounded iff they are continuous

lowasser:


lowasser:
Builds atop #1940 .

lowasser:
Removed the dependency on #1940.

PR #1949, opened 2026-05-06
Clamping into closed intervals of ℝ is idempotent

malarbol:
This PR  refactors the [clamping function](https://unimath.github.io/agda-unimath/real-numbers.closed-intervals-real-numbers.html#5792) into its own module and introduces a few more properties:

- if `x ∈ [a, b]` then clamping `x` into `[a, b]` gives `x`;
- the clamping function is idempotent.


malarbol:
Do you think the clamping function deserves its own module?

lowasser:
Certainly plausible.

EgbertRijke:
Yes, it deserves its own module. The design philosophy in agda-unimath is that pretty much anything that has its own name should have its own module. 

This would provide a natural space to prove properties about the clamping function, explain its idea, write a user interface for clamping function, and so on. All of these help the maintainability of the library.

malarbol:
> Certainly plausible.

Would you rather have it _here_, or in the `real-analysis` module? 

Do you think we should also move more modules from `real-numbers` to `real-analysis`? Like, the power / root functions, or the absolute value function.
E.g., keep the _primitive_ operations like add, neg, diff, mul , max, min, etc. in `real-numbers` but move _composite_ functions  in the `real-analysis` module?

EgbertRijke:
Wherever closed intervals are defined.

src/real-numbers/closed-intervals-real-numbers.lagda.md
```diff
@@ -279,6 +303,41 @@ abstract
...
+```
+
+### The clamping function is idempotent
+
+```agda
+abstract
+  is-idempotent-map-clamp-closed-interval-ℝ :
+    {l1 l2 l3 : Level} ([a,b] : closed-interval-ℝ l1 l2) (x : ℝ l3) →
+    map-clamp-closed-interval-ℝ [a,b] (map-clamp-closed-interval-ℝ [a,b] x) ＝
+    map-clamp-closed-interval-ℝ [a,b] x
```
EgbertRijke: Please use the concept of idempotent map from `foundation.idempotent-maps`

EgbertRijke:
> > Certainly plausible.
> 
> Would you rather have it _here_, or in the `real-analysis` module?
> 
> Do you think we should also move more modules from `real-numbers` to `real-analysis`? Like, the power / root functions, or the absolute value function. E.g., keep the _primitive_ operations like add, neg, diff, mul , max, min, etc. in `real-numbers` but move _composite_ functions in the `real-analysis` module?

I think most elementary functions on the reals should just be in `real-numbers`, and the `real-analysis` folder should be for the study of continuity, differentiability, integrability, and so on, for functions in one variable.

malarbol:
> I think most elementary functions on the reals should just be in `real-numbers`, and the `real-analysis` folder should be for the study of continuity, differentiability, integrability, and so on, for functions in one variable.

I agree. The question is "where do you draw the line at _elementary_?". Certainly, the algebraic operations or min/max are primitive enough to be in `real-numbers`; but maybe the _clamping function_, as a **composition** of primitive operators `max` and `min`, could be considered **non**-elementary enough to go into the `real-analysis` module. And one could also argue that the definition of the odd-root of a real number is more analytical than _elementary_, etc.

But that's just some thoughts. I'll start working on a `real-numbers.clamp-function-[xxx]` and we'll see where it gets us.

EgbertRijke:
Elementary function is a technical term:

https://en.wikipedia.org/wiki/Elementary_function

I'd include all the standard functions in `real-numbers`: polynomials, exponentials, log, trig functions, absolute values, powers with real exponents, and so on, and possibly also those constructed by power series.

Something like a Taylor series, however, would fit more naturally in `real-analysis`, since it uses differentiation in its definition. 

This suggests the following cut-off: If you use differentiation/integration in the definition of a real function, then it goes into `real-analysis`.

src/real-numbers/closed-intervals-real-numbers.lagda.md
malarbol: > Please use the concept of idempotent map from `foundation.idempotent-maps`

I don't think the `foundation.idempotent-maps` module is level-parametric enough to describe the situation here: the `map-clamp-closed-interval-ℝ [a,b]` has type  `ℝ l3 → ℝ (l1 ⊔ l2 ⊔ l3)` so I think we'd need some notion of `large-idempotent-maps` somehow.

EgbertRijke: Ah, that's a very good observation. Let's not make `large-idempotent-maps` for now, but we can consider making it if the issue comes up again

malarbol: FYI, this also comes up in the `metric-quotient` construction: the metric quotient of a metric space is isometrically equivalent to itself so the map 
```
metric-quotient-Metric-Space :
  Metric-Space l1 l2 → Metric-Space (l1 ⊔ l2) (l1 ⊔ l2)
```
is idempotent but the first application shifts the levels. I guess we should have that in mind when we think about `large-idempotent-maps`.

src/real-numbers/clamp-function-closed-interval-real-numbers.lagda.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+For any `x ∈ ℝ`,
+
+- if `x ≤ a` then `clamp-[a,b] x ＝ a`;
+- if `b ≤ x` then `clamp-[a,b] x ＝ a`;
+- if `a ≤ x ≤ b` then `clamp-[a,b] x ＝ x`.
+
+In particular, the **clamp function** is
+[idempotent](foundation.idempotent-maps.md).
```
EgbertRijke, line 52: Beautiful!

PR #1950, opened 2026-05-07
Editing coding style

EgbertRijke:
It seems that a couple of years ago the maintainers had a bit of fun with ChatGPT to help them write the guidelines of the library. Some of these gpt-isms are quite embarrassing. With this pull request, I hope to correct some past mistakes.

### DIFF
--- a/docs/CODINGSTYLE.md
+++ b/docs/CODINGSTYLE.md
@@ -15,7 +15,7 @@ clarity and readability. We hope that with this style guide, we are able to help
 new contributors write code that is both functional and understandable, as well
 as easily maintainable.
 
-Please reach out to us if you have any questions or remarks about the
+Please reach out to us on [Discord](https://discord.gg/Zp2e8hYsuX) if you have any questions or remarks about the
 agda-unimath style, or need any help getting started with your formalization
 project.
 
@@ -39,21 +39,21 @@ become elements of the recorded knowledge base of agda-unimath.
 Other obvious benefits of refactoring your code into small parts is that these
 parts thereby become reusable across the library. Work that is properly
 refactored is more easily understood and used by others, which is ultimately
-what it's all about.
+what it's all about!
 
 ### Representation independence
 
 Most of our concepts are introduced along with a "user interface", which
-consists of entries that specify and determine the basic use cases of a concept.
+consists of entries that specify and determine the basic use cases of the concept.
 This way, our code itself serves as a practical guide on how it can be used
 later in new definitions. Furthermore, later code that correctly uses the
-interface functions for a concept don't depend on the specific implementations
-of its dependants, which makes them conceptually a lot cleaner, and more robust
+interface of a concept don't depend on the specific implementations
+of its dependencies, which makes them conceptually a lot cleaner, and more robust
 to change. John C. Reynolds explained the benefits of representation
 independence more fully in {{#cite Reynolds1974}}.
 
 In essence, our code structuring conventions are designed and practiced with the
-goal of making the library as conceptually clear and as understandable as
+goal of making the library as conceptually clear and understandable as
 possible. Finally, we hope that with the coding practices outlined here, you'll
 find our codebase welcoming and pleasant to work with.
 
--- a/docs/CODINGSTYLE.md
+++ b/docs/CODINGSTYLE.md
@@ -1,16 +1,23 @@
 # The agda-unimath library coding style guide
 
-The agda-unimath library is an ever-expanding library of formalized
-mathematics from a univalent point of view.
-Agda-unimath stands as the largest library of formalized mathematics in the Agda proof assistant, covering a broad range of mathematical subjects. 
-The library's corresponding website displays the formalization pages in a
-markdown book, with over 3000 separate pages as of May 2026.
+The agda-unimath library is an ever-expanding library of formalized mathematics
+from a univalent point of view. Agda-unimath stands as the largest library of
+formalized mathematics in the Agda proof assistant, covering a broad range of
+mathematical subjects. The library's corresponding website displays the
+formalization pages in a markdown book, with over 3000 separate pages as of
+May 2026.
 
-In this coding style guide we will outline some of the design principles of these pages. The purpose of these principles is to create a clean framework through which we present our work, and to ensure good maintainability and reusability of the code. A core value of our formalization project is conceptual clarity and readability. We hope that with this style guide, we are able to help new contributors write code that is both functional and understandable, as well as easily maintainable.
+In this coding style guide we will outline some of the design principles of
+these pages. The purpose of these principles is to create a clean framework
+through which we present our work, and to ensure good maintainability and
+reusability of the code. A core value of our formalization project is conceptual
+clarity and readability. We hope that with this style guide, we are able to help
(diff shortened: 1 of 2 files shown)


docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+these pages. The purpose of these principles is to create a clean framework
+through which we present our work, and to ensure good maintainability and
+reusability of the code. A core value of our formalization project is conceptual
+clarity and readability. We hope that with this style guide, we are able to help
+new contributors write code that is both functional and understandable, as well
+as easily maintainable.
+
+Please reach out to us if you have any questions or remarks about the
+agda-unimath style, or need any help getting started with your formalization
+project.
```
fredrik-bakke: Since you mention reaching out, please also provide means to get in touch. E.g. the discord server?

EgbertRijke: Good point, thanks!

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+is by design. By breaking down entries into small and simple entries, we make
+the library easier to maintain in several ways: Faster compilation; cleaner
+appearance; the code will break less, or less severely, if definitions are
+changed. Moreover, by factoring parts out of a proof or construction, they
+become elements of the recorded knowledge base of agda-unimath.
+
+Other obvious benefits of refactoring your code into small parts is that these
+parts thereby become reusable across the library. Work that is properly
+refactored is more easily understood and used by others, which is ultimately
+what it's all about.
```
fredrik-bakke: The final sentence here sounds a bit withheld. The easiest fix is:


```suggestion
Other obvious benefits of refactoring your code into small parts is that these
parts thereby become reusable across the library. Work that is properly
refactored is more easily understood and used by others, which is ultimately
what it's all about!
```

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+consists of entries that specify and determine the basic use cases of a concept.
+This way, our code itself serves as a practical guide on how it can be used
+later in new definitions. Furthermore, later code that correctly uses the
+interface functions for a concept don't depend on the specific implementations
+of its dependants, which makes them conceptually a lot cleaner, and more robust
+to change. John C. Reynolds explained the benefits of representation
+independence more fully in {{#cite Reynolds1974}}.
+
+In essence, our code structuring conventions are designed and practiced with the
+goal of making the library as conceptually clear and as understandable as
```
fredrik-bakke:
```suggestion
goal of making the library as conceptually clear and understandable as
```

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+
+Other obvious benefits of refactoring your code into small parts is that these
+parts thereby become reusable across the library. Work that is properly
+refactored is more easily understood and used by others, which is ultimately
+what it's all about.
+
+### Representation independence
+
+Most of our concepts are introduced along with a "user interface", which
+consists of entries that specify and determine the basic use cases of a concept.
```
fredrik-bakke:
```suggestion
consists of entries that specify and determine the basic use cases of the concept.
```

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+This way, our code itself serves as a practical guide on how it can be used
+later in new definitions. Furthermore, later code that correctly uses the
+interface functions for a concept don't depend on the specific implementations
```
fredrik-bakke:
```suggestion
interface of a concept don't depend on the specific implementations
```

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
+of its dependants, which makes them conceptually a lot cleaner, and more robust
```
fredrik-bakke: I believe you mean its dependencies here, which is the dual of dependants.

```suggestion
of its dependencies, which makes them conceptually a lot cleaner, and more robust
```

fredrik-bakke (commented):
I actually think the previously exhibited confidence of the codingstyle document was quite fitting and reasonable. We are proud of our principles and they help us a lot! 
Also, a point list highlighting why we do things the way we do is nice, although I think the previous version should be heavily reworked. We can consider including the following points

- Simplicity
- Reusability
- Maintainability/Scalability
- Reliability/navigability (in the sense that the structure of our documents are easy to navigate once you learn their structure)
- Ease of development. Rigid rules resolve otherwise tedious questions like: should I create a new file for this? What should I name this?
- Rediscoverability

EgbertRijke:
Those are good points. The reason why I removed the list with benefits was that this list appeared before we described what we do. Maybe it can be rewritten again:)

fredrik-bakke:
Ah okay. I'm happy to have this merged if you consider yourself done.

EgbertRijke:
I think for now I'm done; I'd rather work on a fresh PR if I have some fresh ideas to improve it, than to make it perfect now.

docs/CODINGSTYLE.md
```diff
@@ -1,76 +1,61 @@
...
-functional and understandable, as well as easily maintainable. Please reach out
-to us if you have any questions or remarks about the agda-unimath style, or need
-any help getting started with your formalization project. Our code, and also
-this guide, are open to refinement to best support our community and the
-project's goals.
+# The agda-unimath library coding style guide
+
+The agda-unimath library is an ever-expanding library of formalized mathematics
+from a univalent point of view. Agda-unimath stands as the largest library of
+formalized mathematics in the Agda proof assistant, covering a broad range of
```
malarbol, line 5: I'm not sure being "the largest library [...]" should be part of the _coding style guide_. We're certainly happy and proud to be "the largest [...]" but that shouldn't guide the style of the contributions.

Moreover, at any point, some _other_ "library of formalized mathematics in the Agda proof assistant" may become larger than Agda-unimath, which would turn this statement into false-advertising.

Why not just _a large library [...]_, or _one of the largest [...]_?

EgbertRijke: Fair point! I'll edit it

fredrik-bakke: > I'm not sure being "the largest library [...]" should be part of the _coding style guide_. We're certainly happy and proud to be "the largest [...]" but that shouldn't guide the style of the contributions.
> 
> Moreover, at any point, some _other_ "library of formalized mathematics in the Agda proof assistant" may become larger than Agda-unimath, which would turn this statement into false-advertising.
> 
> Why not just _a large library [...]_, or _one of the largest [...]_?

I like the wording "one of the largest" because it is less arrogant, and gives flexibility. Being "the largest" calls for us to define the metric we are measuring ourselves against other libraries on. For instance, we are not the largest library in terms of community. If the metric is "lines of code", then any John with enough LLM tokens can generate a "larger" library in a week. Lines of code is a terrible metric for quality.

EgbertRijke: I think the point I was trying to make in this text was that agda-unimath is a large library, so we are necessarily serious about coding style. I think "one of the largest" is still comparative, which is unnecessary in the context of a coding style, and just "a large library" is better imho. It's just a large chunk of work.

PR #1952, opened 2026-05-08
strictly increasing real maps on proper closed real intervals

malarbol:

- `real-map-proper-closed-interval-ℝ I`: maps `I → ℝ` from a proper closed interval `I`;
- `map-proper-closed-interval-ℝ I J`: maps `I → J` between proper closed intervals;
- `is-strictly-increasing-prop-real-map-proper-closed-interval-ℝ I`: the property of being a strictly increasing  map `I → ℝ`.

### DIFF
--- a/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
+++ b/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
@@ -31,7 +31,7 @@ of [functions](foundation.function-types.md) `I → J`, i.e., the type of
 
 ## Definition
 
-### The type of real maps on a proper closed interval of real numbers
+### Maps between proper closed intervals of real numbers
 
 ```agda
 module _
--- a/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
+++ b/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
@@ -1,0 +1,78 @@
+# Functions between proper closed intervals of real numbers
+
+```agda
+{-# OPTIONS --lossy-unification #-}
+
+module real-numbers.maps-between-proper-closed-intervals-real-numbers where
+```
+
+<details><summary>Imports</summary>
+
+```agda
+open import foundation.dependent-pair-types
+open import foundation.function-types
+open import foundation.universe-levels
+
+open import real-numbers.dedekind-real-numbers
+open import real-numbers.proper-closed-intervals-real-numbers
+```
+
+</details>
+
+## Idea
+
+The type of
+{{#concept "maps" Disambiguation="between proper closed intervals" Agda=map-proper-closed-interval-ℝ}}
+between
+[proper closed intervals](real-numbers.proper-closed-intervals-real-numbers.md)
+of [real numbers](real-numbers.dedekind-real-numbers.md) `I` and `J` is the type
+of [functions](foundation.function-types.md) `I → J`, i.e., the type of
+[real maps](real-numbers.real-maps-proper-closed-intervals-real-numbers.md) on
+`I` with values [in](foundation.subtypes.md) `J`.
+
+## Definition
+
+### The type of real maps on a proper closed interval of real numbers
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} (l l' : Level)
+  (I : proper-closed-interval-ℝ l1 l2)
+  (J : proper-closed-interval-ℝ l3 l4)
+  where
+
+  map-proper-closed-interval-ℝ :
+    UU (lsuc l ⊔ lsuc l' ⊔ l1 ⊔ l2 ⊔ l3 ⊔ l4)
+  map-proper-closed-interval-ℝ =
+    type-proper-closed-interval-ℝ l I → type-proper-closed-interval-ℝ l' J
+```
+
+## Properties
+
+### The identity map on a proper closed interval
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level} (l : Level)
+  (I : proper-closed-interval-ℝ l1 l2)
+  where
+
+  id-proper-closed-interval-ℝ : map-proper-closed-interval-ℝ l l I I
+  id-proper-closed-interval-ℝ = id
+```
+
+### Composition of maps between proper closed intervals
+
+```agda
+module _
+  {l1 l2 l3 l4 l5 l6 l l' l'' : Level}
+  {I : proper-closed-interval-ℝ l1 l2}
+  {J : proper-closed-interval-ℝ l3 l4}
+  {K : proper-closed-interval-ℝ l5 l6}
+  (g : map-proper-closed-interval-ℝ l' l'' J K)
+  (f : map-proper-closed-interval-ℝ l l' I J)
+  where
+
+  comp-map-proper-closed-interval-ℝ : map-proper-closed-interval-ℝ l l'' I K
+  comp-map-proper-closed-interval-ℝ = g ∘ f
+```
--- a/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
+++ b/src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
@@ -9,7 +9,6 @@ module real-numbers.maps-between-proper-closed-intervals-real-numbers where
 <details><summary>Imports</summary>
 
 ```agda
-open import foundation.dependent-pair-types
 open import foundation.function-types
 open import foundation.universe-levels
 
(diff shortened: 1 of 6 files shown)


malarbol:
Note that the type `real-map-proper-closed-interval-ℝ I` was already used in [real-numbers.uniformly-continuous-real-maps-proper-closed-intervals-real-numbers](https://unimath.github.io/agda-unimath/real-numbers.uniformly-continuous-real-maps-proper-closed-intervals-real-numbers.html) or [real-analysis.differentiable-real-maps-on-proper-closed-intervals-real-numbers](https://unimath.github.io/agda-unimath/real-analysis.differentiable-real-maps-on-proper-closed-intervals-real-numbers.html) although it didn't have a proper name. I can modify these modules to use this new concept if you wish; either here or in another PR. 

I'm not sure how many other modules use the type `type-proper-closed-interval-ℝ l1 [a,b] → ℝ l2`.

malarbol:
I'm trying to see if the arguments of [real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html) could be adapted to [uniformly-continuous](https://unimath.github.io/agda-unimath/real-numbers.uniformly-continuous-real-maps-proper-closed-intervals-real-numbers.html) _strictly increasing_ maps on proper closed intervals: 

hopefully, a _strictly increasing_ uniformly continuous map `f : [a,b] → ℝ` should have a strictly increasing uniformly continuous inverse `f⁻¹ :  [f(a),f(b)] → [a,b]`. I believe this could be useful in the study of the exponential map or trigonometric functions.

I still have a lot of details to figure out, but this is a first step, I think.

src/real-numbers/strictly-increasing-endomaps-real-numbers.lagda.md
```diff
@@ -100,8 +100,7 @@ module _
       leq-ℝ (f x) (f y) →
       leq-ℝ x y
     reflects-leq-is-strictly-increasing-endomap-ℝ x y fx≤fy =
-      leq-not-le-ℝ y x
-        ( λ x<y → not-le-leq-ℝ _ _ fx≤fy (H y x x<y))
+      leq-not-le-ℝ y x (not-le-leq-ℝ _ _ fx≤fy ∘ (H y x))
```
malarbol, line 103: Note that the local variable `x<y` had type `y < x` which was a bit awkward. Instead of fixing the name I just chose to get rid of it using `∘`.

EgbertRijke: Thanks for fixing!

To be honest, I think the code gets too littered with mathematical symbols with this variable naming convention. I'm not very convinced that it is such a good practice.

malarbol: > To be honest, I think the code gets too littered with mathematical symbols with this variable naming convention. I'm not very convinced that it is such a good practice.

Duly noted. 

I agree we may have gone over-the-top with mathematical symbols in variable naming.
Although I think it can be helpful to follow _some_ definitions, it forces us to be extra-careful with the naming, and, as this example illustrates, we may fail sometimes. So it's probably not a good practice. I'll try to have that in mind.

src/real-numbers/maps-between-proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -0,0 +1,77 @@
...
+between
+[proper closed intervals](real-numbers.proper-closed-intervals-real-numbers.md)
+of [real numbers](real-numbers.dedekind-real-numbers.md) `I` and `J` is the type
+of [functions](foundation.function-types.md) `I → J`, i.e., the type of
+[real maps](real-numbers.real-maps-proper-closed-intervals-real-numbers.md) on
+`I` with values [in](foundation.subtypes.md) `J`.
+
+## Definition
+
+### The type of real maps on a proper closed interval of real numbers
```
EgbertRijke: The use of singular "a proper closed interval" is a little strange here, since there are two proper closed intervals involved.


```suggestion
### Maps between proper closed intervals of real numbers
```

src/real-numbers/proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -129,6 +137,79 @@ width-proper-closed-interval-ℝ (a , b , _) = b -ℝ a
 
 ## Properties
 
+### Similarity of real numbers preserves proper closed intervals
```
EgbertRijke:
```suggestion
### Proper closed intervals are closed under similarity of real numbers
```

src/real-numbers/proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -862,3 +943,77 @@ module _
                   ( is-positive-real-ℚ⁺ ε'')))
           ( cotransitive-le-ℝ a xℝ b a<b)
 ```
+
+### Raising elements of proper closed interval
```
EgbertRijke:
```suggestion
### Raising universe levels of elements of proper closed interval
```

src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -0,0 +1,300 @@
...
+      ( map-double-negation
+        ( rec-coproduct
+          ( λ u~v →
+            leq-eq-ℝ
+              ( ap f
+                ( eq-type-subtype
+                  ( subtype-proper-closed-interval-ℝ l1 [a,b])
+                  ( eq-sim-ℝ u~v))))
+          ( λ u<v → leq-le-ℝ (H x y u<v)))
+        ( irrefutable-sim-or-le-leq-ℝ u v u≤v))
```
EgbertRijke, line 116: It is a little unexpected that this requires an involved proof such as this, but maybe I am not fully appreciating some intricacies. Isn't any strictly increasing map increasing (i.e., shouldn't this be an application of a general lemma)?

src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -0,0 +1,300 @@
...
+
+```agda
+module _
+  {l1 l2 l3 l4 : Level}
+  ([a,b] : proper-closed-interval-ℝ l3 l4)
+  (f : real-map-proper-closed-interval-ℝ l1 l2 [a,b])
+  (H : is-strictly-increasing-real-map-proper-closed-interval-ℝ [a,b] f)
+  where abstract
+
+  is-injective-is-strictly-increasing-real-map-proper-closed-interval-ℝ :
```
EgbertRijke, line 148: Here again, I would expect there to be a lemma that asserts that every strictly increasing map is injective. Isn't this in the library somewhere?

EgbertRijke: Ah, not every strictly increasing map, but every strictly increasing map between strict linear orders should be injective. We probably don't have that in the library yet.

src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
malarbol: This is directly adapted from [is-increasing-is-increasing-on-strict-inequalities-endomap-ℝ](https://unimath.github.io/agda-unimath/real-numbers.increasing-endomaps-real-numbers.html#2900)

I'm not sure how much it could be simplified but be aware there _are_ some intricacies in the definitions of `leq-ℝ` and `le-ℝ`. I'll try to have another look.

src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
malarbol: yes, you need to relate the strict preorder to the preorder somehow and I don't think we have the generic infrastructure for this.

EgbertRijke: It would be nice to have at some point, but I'll mark this comment as resolved.

src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
malarbol: Is this the same as in https://github.com/UniMath/agda-unimath/pull/1952#discussion_r3212193408 ? 
Should we just mark this as resolved too?

PR #1955, opened 2026-05-10
Some small changes

space3time1:
new:
localizations-at-families-of-maps
action-on-dependent-identifications-functions

changes:
natural-numbers
booleans
type-arithmetic-booleans


EgbertRijke:
Hi, thank you so much for your pull request!

We are not looking to reconfigure our entire library infrastructure at this time. If you wanted to make a new pull request with only contributions to the formalization, it would be most welcome.

Thank you for your understanding!

space3time1:
Actuallly I only made some edits to the formalization part. As is shown in my previous comment, that is what all I edited by hand. I really cannot understand why such additional changes appereared. I just ran the "make pre-commit" and the "make website" after editing, and can these two programs cause these unexpected changes?

space3time1:
As you can see, most of these "changes" do not actually change anything at all: it says that some contents are deleted and the exactly same things are added back again. So why are they recognized as changes? Can this be caused by "make pre-commit" and "make website"?

EgbertRijke:
I was able to revert the changes to the library infrastructure. I'm sorry that you were having issues!

src/foundation/action-on-dependent-identifications-functions.lagda.md
```diff
@@ -0,0 +1,43 @@
...
+Given a type family `B : A → UU l` and an
+[identification](foundation-core.identity-types.md) `p : x ＝ y` in `A`, a
+function `f : (x : A) → (u : B x) → C` maps a
+[dependent-identification](foundation-core.dependent-identifications.md)
+`q : dependent-identification B p u v` to an identification
+`adp : (f x u) ＝ (f y v)` in `C`.
+
+## Definition
+
+### Functorial action of functions on dependent identity types
```
EgbertRijke: Since there is no functoriality involved here, and the function `f` is actually a family of functions `B x → C` I'd suggest the following rewording:

```suggestion
### The action of families of functions on dependent identifications
```

malarbol: Hi! 
This module already exists with a slightly different name: 
https://unimath.github.io/agda-unimath/foundation.action-on-identifications-dependent-functions.html

But maybe the prose might be fixed _there_?

VojtechStep:
If I had to guess you're on Windows, and git decided to change line endings from LF to CRLF when cloning the repo. You can configure git to convert from CRLF to LF when committing with the command `git config --global core.autocrlf true`

EgbertRijke:
> If I had to guess you're on Windows, and git decided to change line endings from LF to CRLF when cloning the repo. You can configure git to convert from CRLF to LF when committing with the command `git config --global core.autocrlf true`

Would it make sense to add `eol=lf` to `.gitattributes`? If this is a common Windows issue, then surely this student isn't the only one encountering this problem.

src/foundation/action-on-dependent-identifications-functions.lagda.md
EgbertRijke: > Hi! This module already exists with a slightly different name: https://unimath.github.io/agda-unimath/foundation.action-on-identifications-dependent-functions.html
> 
> But maybe the prose might be fixed _there_?

That module does something slightly different: It defines the action on (ordinary) identifications of dependent functions. This module defines the action on dependent identifications of families of maps.

My comment about functoriality was actually inaccurate, and @space3time1 was correct in using it in this context: We use the word functoriality in all the files about the action on identifications in this same sense. I just had forgotten about it.

VojtechStep:
I honestly don't know. It does something different (with `autocrlf` you get local files with CRLF and they get converted in the background, while with `eol=lf` you get local files with LF) and `eol=lf` could break some Windows tooling by not having CRLFs, but it's been years since I've done any development on Windows. In any case Windows isn't supported as a development platform for agda-unimath (since the Makefile and pre-commit scripts assume Linux or macOS), so IMO it's more of an issue of documenting this non-support and suggesting doing everything in WSL

src/foundation/action-on-dependent-identifications-functions.lagda.md
malarbol: > That module does something slightly different: It defines the action on (ordinary) identifications of dependent functions. This module defines the action on dependent identifications of families of maps.
> 

Oh!. Indeed. I read to fast.
Sorry @space3time1, don't mind my comment.

PR #1956, opened 2026-05-12
Inverses of strictly increasing real maps on proper real closed intervals (WIP)

malarbol:
This PR proves the monotonic inversion theorem for pointwise ε-δ continuous strictly increasing maps on proper closed intervals: 

Any strictly increasing map `f : [a, b] → ℝ` has a left inverse `f⁻¹ : [f(a), f(b)] → [a, b]`. Moreover, if `f` is pointwise ε-δ continuous, `f⁻¹` is also a right inverse so `f` induces an equivalence `[a, b] ≃ [f(a), f(b)]`.


malarbol:
This needs a lot of cleaning up and refactoring but it contains some interesting results I wanted to share/discuss.

malarbol:
I've started to split this PR into smaller ones. See #1957 for a start.

In the meantime, I've also made some interesting progress. I expected to need some kind of continuity to define the inverse of a strictly increasing map on a proper closed interval, but it doesn't seem to be the case: the _strictly increasing_ condition is actually enough to define an inverse map `f⁻¹ : [f(a), f(b)] → ℝ` (or so it seems; I still have to prove that my definition is an actual inverse, maybe I'll need continuity for this?). 

Did you encounter anything similar, @lowasser, when you worked on cofinal & coinitial strictly increasing pointwise continuous real endomaps? Should I expect to need continuity sometimes further in the construction of the inverse?

src/real-numbers/inverses-strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -0,0 +1,539 @@
...
+          ( q)) ∨
+        ( upper-cut-map-inv-is-strictly-increasing-real-map-proper-closed-interval-ℝ
+          ( I)
+          ( f)
+          ( H)
+          ( y)
+          ( r)))
+      ( λ q<a →
+        inl-disjunction
+          ( is-in-lower-cut-map-inv-le-lower-bound-is-strictly-increasing-real-map-properd-closed-interval-ℝ
```
lowasser: properd?

lowasser:
> Did you encounter anything similar, @lowasser, when you worked on cofinal & coinitial strictly increasing pointwise continuous real endomaps? Should I expect to need continuity sometimes further in the construction of the inverse?

Continuity should be involved in the locatedness condition of the result, to show the lower and upper cuts get arbitrarily close.

malarbol:
> > Did you encounter anything similar, @lowasser, when you worked on cofinal & coinitial strictly increasing pointwise continuous real endomaps? Should I expect to need continuity sometimes further in the construction of the inverse?
> 
> Continuity should be involved in the locatedness condition of the result, to show the lower and upper cuts get arbitrarily close.

That's what I expected too, but as the lemma `is-located-cut-map-inv-is-strictly-increasing-real-map-proper-closed-interval-ℝ` in `real-numbers.inverses-strictly-increasing-real-maps-proper-closed-intervals-real-numbers` proves, this is not necessary; somehow the `is-strictly-increasing` condition seems enough. 

BTW, I'm having some trouble proving that my construction actually produces an inverse. Do you have any insight?

malarbol:
> Do you have any insight?

To give more context on the construction:

Given some strictly increasing map `f : [a, b] → ℝ` and `y ∈ [f(a), f(b)]`, we can define:
- `Ly = { q : ℚ | (q < b) ∧ (f q ≤ y)}`: the rational numbers lesser than `b` with image lesser than or equal to `y`;
- `Uy = {r : ℚ | (a < r) ∧ (y ≤ f r)}`: the rational numbers greater than `a` with image greater than or equal to `y`

Since we're looking for some `x` such that `y ＝ f x`, these feel somehow relevant.

These define inhabited lower/upper subtypes but they are not necessarily _rounded_; this can be fixed by **rounding** (see #1958) the lower/upper subtypes, hence the current definitions. We end up we some lower/upper reals with disjoint and located cuts, i.e. a _real number_.

But I don't know how to prove that this produces and actual inverse, nor that this inverse is strictly increasing (which should also be expected).

lowasser:
Let's picture `f` that maps `x < 0` to `x` and `x >= 0` to `x + 1`.  Clearly this is discontinuous and strictly increasing.  What is the inverse of `f` at 0.5?

Perhaps the failure here is that what you've constructed succeeds in giving a real number for every output, but fails to be an inverse?

lowasser:
Er, let me rephrase.  I think that is a counterexample showing that you can't actually successfully construct an inverse if it's discontinuous, even if you can construct something that behaves like an inverse on continuous regions?

lowasser:
I note that the continuity condition in the existing file is used in [roundedness](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#14089); I wonder if the generic rounding is losing something.

malarbol:
> Let's picture `f` that maps `x < 0` to `x` and `x >= 0` to `x + 1`. Clearly this is discontinuous and strictly increasing. What is the inverse of `f` at 0.5?
> 

I have no idea 🤯 . That's what's bugging me off!. 

> Perhaps the failure here is that what you've constructed succeeds in giving a real number for every output, but fails to be an inverse?

I guess so. As I said, I'm struggling trying to prove the inverse laws and, at this point, I don't know if and where I should be using some continuity property.

malarbol:
> Let's picture `f` that maps `x < 0` to `x` and `x >= 0` to `x + 1`. Clearly this is discontinuous and strictly increasing. What is the inverse of `f` at 0.5?

Note that you need some kind of choice to define this map. But even then, I wonder how my construction behaves on this example. I'll try to think about it.

lowasser:
> > Let's picture `f` that maps `x < 0` to `x` and `x >= 0` to `x + 1`. Clearly this is discontinuous and strictly increasing. What is the inverse of `f` at 0.5?
> 
> I have no idea 🤯 . That's what's bugging me off!.

Fundamentally, you can't define the inverse of `f` at 0.5, even in doing math "the normal way," and still have it be an inverse.  You can't define a `g` such that `f(g(x)) = 0.5`, so there is no inverse.

It took me quite a while to work out how to do the inverse proof for the all-real-numbers case, but the proof went

- [showing the inverse function is a retraction](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-a-retraction) (don't know if this definition falls apart with the rounding of the subsets?)
- [the inverse function is strictly increasing](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-strictly-increasing), which notably uses the continuity condition too
- the inverse function is injective (follows immediately from strictly increasing)
- [the inverse being a section follows from being an injective retraction](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-a-section)

malarbol:
> > > Let's picture `f` that maps `x < 0` to `x` and `x >= 0` to `x + 1`. Clearly this is discontinuous and strictly increasing. What is the inverse of `f` at 0.5?
> > 
> > 
> > I have no idea 🤯 . That's what's bugging me off!.
> 
> Fundamentally, you can't define the inverse of `f` at 0.5, even in doing math "the normal way," and still have it be an inverse. You can't define a `g` such that `f(g(x)) = 0.5`, so there is no inverse.

Sure. What I meant was: "what does **this** inverse construction evaluates at 0.5".  I think the output function would be the map `g` such that
- `g x = x` if `x < 0`;
- `g x = 0` if `0 ≤  x ≤ 1`;
- `g x = x -1` if `1 ≤ x`. 

This map is neither strictly increasing nor a right inverse of `f`, but it is a left inverse. Maybe I should start there. Or do you have a counter-example in mind that would invalidate the existence of a left inverse too?

> It took me quite a while to work out how to do the inverse proof for the all-real-numbers case, but the proof went
> 
> * [showing the inverse function is a retraction](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-a-retraction) (don't know if this definition falls apart with the rounding of the subsets?)
> * [the inverse function is strictly increasing](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-strictly-increasing), which notably uses the continuity condition too
> * the inverse function is injective (follows immediately from strictly increasing)
> * [the inverse being a section follows from being an injective retraction](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-a-section)

Yes. That's why I started to investigate the relations between `f x ≤ y`,  `x ≤ f⁻¹ y`, `y ≤ f x` and `f ⁻¹ y ≤ x` but I can't figure out how to build my proof. I'll give it some more thoughts. Thanks for your help.

malarbol:
> I note that the continuity condition in the existing file is used in [roundedness](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#14089); I wonder if the generic rounding is losing something.

I wonder too. But, it shouldn't lose anything relevant. Basically you'd want to use the lower cuts `{q : ℚ : f q < y}` and you need continuity to prove that _these_ are rounded. But you can **force** roundedness considering `{q : ℚ | (q < b) ∧ (f q ≤ y)}` instead, and the resulting lower reals shouldn't be sensibly different.

malarbol:
> > I note that the continuity condition in the existing file is used in [roundedness](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#14089); I wonder if the generic rounding is losing something.
> 
> I wonder too. But, it shouldn't lose anything relevant. Basically you'd want to use the lower cuts `{q : ℚ : f q < y}` and you need continuity to prove that _these_ are rounded. But you can **force** roundedness considering `{q : ℚ | (q < b) ∧ (f q ≤ y)}` instead, and the resulting lower reals shouldn't be sensibly different.

Note that the need for rounding also comes from the boundary conditions on `[a, b]`:  because `f : [a, b] → ℝ` is only defined on the interval `[a, b]`, `f q` **actually** means `f (clamp-[a,b] q)`.  So, for `q < a`, you have `f q = f a`; in other words, even if `q < a`, `q ∉ {q : ℚ : f q < f a}`  so `{q : ℚ : f q < f a}` is not a good candidate for a lower cut of the inverse at `f(a)`. The **rounding** was meant to fix that too.

malarbol:
> Er, let me rephrase. I think that is a counterexample showing that you can't actually successfully construct an inverse if it's discontinuous, even if you can construct something that behaves like an inverse on continuous regions?

Thanks four your help! I did manage to prove that this inverse **is** a left-inverse and increasing. After our conversation, I don't expect it to be a right inverse without some additional continuity hypothesis but I'm curious how far we can go, and where _exactly_, continuity will be needed. 

E.g., in your counter-example, we still have `f (f ⁻¹ y) ≤ y`. Could _this_ be true regardless of the continuity of `f`?

malarbol:
> * [the inverse function is strictly increasing](https://unimath.github.io/agda-unimath/real-numbers.cofinal-and-coinitial-strictly-increasing-pointwise-epsilon-delta-continuous-endomaps-real-numbers.html#the-inverse-function-is-strictly-increasing), which notably uses the continuity condition too

This seems to be the key. Once you get "`f⁻¹` is strictly increasing", it's an embedding with a right inverse, hence an equivalence, and therefore so is `f`. So the last piece would be to prove that 
```text
  is-continuous f ⇒  is-strictly-increasing f⁻¹
```
I'll see if I can wrap this up. 

Thanks again @lowasser.

malarbol:
> [...] So the last piece would be to prove that
> 
> ```
>   is-continuous f ⇒  is-strictly-increasing f⁻¹
> ```
> 
> I'll see if I can wrap this up.

This was not the way, after all. Playing with the inequalities `f x ≤ y`,  `x ≤ f⁻¹ y`, `y ≤ f x` and `f⁻¹ y ≤ x` provided the answers. 

I'm still missing a result about the continuity of `f⁻¹`. I wonder if `f⁻¹` is continuous regardless of the continuity of `f`: intuitively, `f⁻¹` maps discontinuities of `f` to locally constant values, without affecting continuity. On the other hand, a discontinuity of `f⁻¹` would result in a small interval where `f` is locally constant, which can't be if `f` is **strictly** increasing. So I expect `f⁻¹` to be continuous but I still haven't find a proof yet. Any help is welcomed.

BTW, does this seem like anything existing in the literature on constructive analysis? It could be nice to add some external  references.

PR #1957, opened 2026-05-12
misc lemmas min/max/clamp real numbers

malarbol:
This PR introduces a few lemmas on the min/max/clamp functions on real numbers:

- for any `x : ℝ`, the functions `max-ℝ x` and `min-ℝ x` are _increasing_;
- for any closed interval `I`, the clamping map `clamp-I` is _increasing_ and _strictly increasing_ on `I`;

We also prove the following location property for proper closed intervals: for any proper closed interval `[a, b]` and rational numbers `p < q`, then either:

- `p < a`;
- `q < b`;
- there exists `(s r : ℚ)` such that `a < s < r < b` and `p < s < r < q`.

### DIFF
--- a/src/real-numbers/proper-closed-intervals-real-numbers.lagda.md
+++ b/src/real-numbers/proper-closed-intervals-real-numbers.lagda.md
@@ -21,9 +21,12 @@ open import elementary-number-theory.unit-fractions-rational-numbers
 
 open import foundation.cartesian-product-types
 open import foundation.conjunction
+open import foundation.coproduct-types
 open import foundation.dependent-pair-types
+open import foundation.dependent-products-propositions
 open import foundation.disjunction
 open import foundation.existential-quantification
+open import foundation.function-types
 open import foundation.identity-types
 open import foundation.inhabited-subtypes
 open import foundation.propositional-truncations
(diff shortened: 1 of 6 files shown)


src/real-numbers/strictly-increasing-real-maps-proper-closed-intervals-real-numbers.lagda.md
malarbol, line 1: This is just an indentation change to **un**-abstract the definition of `proper-closed-interval-im-is-strictly-increasing-real-map-proper-closed-interval-ℝ`: we obviously need the explicit definitions of the lower/upper bounds of the image interval.

src/real-numbers/proper-closed-intervals-real-numbers.lagda.md
```diff
@@ -1017,3 +1026,83 @@ module _
...
+
+### Location of strictly ordered pairs of rational numbers w.r.t. a proper closed interval
+
+For any proper closed interval `[a,b]` and any pair `(p q : ℚ)` with `p < q`
+then one of the following propositions holds:
+
+- `p < a`;
+- `b < q`;
+- `∃ (r s : ℚ) | (a < r < s < b) ∧ (p < r) ∧ (s < q)`, i.e. `[r, s]` is a proper
+  sub-interval of both `[a, b] ∩ [p, q]`.
```
malarbol, line 1038: To put this lemma into context: 
a lot of properties regarding rational numbers `q` in closed intervals `[a,b]` will depend on whether `q < a` , `q ∈ [a, b]`, or `b < q`. Without trichotomy on the real numbers this cannot be decided. 

This lemma provides a _located_ version of this: for a pair `p < q`, either `p < a` , `b < q`, or there exists as smaller rational interval `[r, s]` inside `[a, b] ∩ [p, q]`.

malarbol: As an application, this allows us to prove that the lower & upper inverses of a strictly increasing map on a proper closed interval form a Dedekind cut (more precisely, the _located_ axiom; see #1956).

lowasser (approved):

PR #1958, opened 2026-05-13
Rounded lower/upper subsets of ℚ

malarbol:
This PR introduces the following modules: 
- `order-theory.upper-types-preorders`: **upper type in a preorder**, duals of `order-theory.lower-types-preorders.lagda.md`, i.e., _lower types_ in the _opposite_ preorder;
- `elementary-number-theory.rounded-lower-subsets-rational-numbers`: _lower subsets_ `L` of rational numbers such that 
```text
  ∀ (q : ℚ),  L q ⇒ ∃ (r : ℚ) | (q < r) ∧ L r;
``` 
- `elementary-number-theory.rounded-upper-subsets-rational-numbers`: _upper subsets_ `U` of rational numbers such that 
```text
  ∀ (r : ℚ),  U r ⇒ ∃ (q : ℚ) | (q < r) ∧ U q;
```

Any lower/upper subset of `ℚ` can be **rounded**. Moreover, any lower/upper set  with inhabited rounding defines a lower/upper Dedekind real number.

### DIFF
--- a/src/elementary-number-theory.lagda.md
+++ b/src/elementary-number-theory.lagda.md
@@ -207,6 +207,8 @@ open import elementary-number-theory.retracts-of-natural-numbers public
 open import elementary-number-theory.ring-extension-rational-numbers-of-rational-numbers public
 open import elementary-number-theory.ring-of-integers public
 open import elementary-number-theory.ring-of-rational-numbers public
+open import elementary-number-theory.rounded-lower-subsets-rational-numbers public
+open import elementary-number-theory.rounded-upper-subsets-rational-numbers public
 open import elementary-number-theory.semiring-of-natural-numbers public
 open import elementary-number-theory.series-rational-numbers public
 open import elementary-number-theory.sieve-of-eratosthenes public
--- a/src/elementary-number-theory/rounded-lower-subsets-rational-numbers.lagda.md
+++ b/src/elementary-number-theory/rounded-lower-subsets-rational-numbers.lagda.md
@@ -63,6 +63,34 @@ module _
     is-prop-type-Prop is-rounded-prop-lower-type-ℚ
 ```
 
+### The type of rounded lower subsets of `ℚ`
+
+```agda
+rounded-lower-type-ℚ : (l : Level) → UU (lsuc l)
+rounded-lower-type-ℚ l =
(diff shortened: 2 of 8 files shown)


malarbol:
I don't have any reference for this construction but it provides a handy way to construct lower/upper real numbers. 
As an example, this can be used to construct the inverse of a strictly increasing map (see #1956).

malarbol:
I guess it could somehow be generalized to some strict preorders (related to a preorder defining the notion of lower/upper sets) but we can start with this special case.

malarbol:
A few more thoughts:
- with this definition, the type of lower/upper reals is equivalent to the type of inhabited rounded lower/upper subsets of ℚ (modulo currying/pairing/etc.). We might consider actually **defining** lower/upper reals as _inhabited rounded lower/upper subsets of ℚ_ and maybe restructure the code around the hierarchy 
```text
  lower-type-ℚ > rounded-lower-type-ℚ > inhabited-rounded-lower-type-ℚ (= lower-ℝ)
```
- this means that `rounded-lower-type-ℚ` is basically `lower-ℝ` when you forget the _inhabited_ condition on the cut. Is it some kind of _extended lower reals_ with a `neg-infinity-lower-ℝ` corresponding to the empty lower type? Dually, is `rounded-upper-type-ℚ` some type of _extend upper reals_ with an `infinity-upper-ℝ`?
- If we drop the _inhabited_ condition in the definition of Dedekind real numbers, considering pairs of disjoint located `rounded-lower-type-ℚ` and `rounded-upper-type-ℚ`, does this produce some kind of _extended real numbers_ `[-∞ , +∞]`? Could this be interesting for analysis or integration theory?
- the existence of the idempotent rounding map implies some kind of _equivalence relation_ on the type of lower/upper subsets defined by `P ~ Q' ⇔ round P ＝ round P'` and the rounding function acts as a quotient map; I wonder if this could be interesting for the study of lower/upper reals?

malarbol:
Note also that this **rounding** construction already appears in the library; in [real numbers](https://unimath.github.io/agda-unimath/real-numbers.dedekind-real-numbers.html#10362), to prove that each cut determines the other, we actually show that each cut is _the **rounding** of the complement_ of the other.

PR #1960, opened 2026-05-16
Large substructures of large group-theoretic structures

lowasser:
Introduces everything between large subtypes and large abelian subgroups.

After this step, the goal is to reconcile the large algebraic structures of real numbers and their subsets (e.g. the positive real numbers under multiplication are a large submonoid of all real numbers under multiplication), allowing us to also reconcile the definitions of powers/exponentiation.


lowasser:
Maybe this should be broken up...

lowasser:
Now builds on #1963.

PR #1965, opened 2026-05-26
Rational powers of positive real numbers

lowasser:
Uses large subgroups etc. to keep straight the restrictions of power and root operations to positive real numbers.


lowasser:
Builds on #1964.

PR #1969, opened 2026-05-30
Insertion and removal at pairs of distinct indices

lowasser:
This required more shenanigans than I expected.

PR #1970, opened 2026-05-30
Multilinear maps

lowasser:
Still working on showing alternating multilinear maps are antisymmetric, by reducing it to the bilinear case, using #1969.


lowasser:
This is intended as infrastructure for determinants, which will build on #1967 .

PR #1984, opened 2026-06-11
Riemann integrable functions from closed intervals in the real numbers to normed real vector spaces

lowasser:
Builds on #1940 , #1982 , and #1983 .  Shows that being Riemann integrable is a proposition.


lowasser:
Bishop's Foundations of Constructive Analysis says that if the Riemann sum of the standard uniform partition has a limit as the number of partitions grows, that _is_ the integral and you can get arbitrarily close to it with any any partition with sufficiently small mesh, but the proof skips over a detail I don't yet see how to resolve: https://math.stackexchange.com/q/5140364/23995 .

malarbol:
> Bishop's Foundations of Constructive Analysis says that if the Riemann sum of the standard uniform partition has a limit as the number of partitions grows, that _is_ the integral and you can get arbitrarily close to it with any any partition with sufficiently small mesh, but the proof skips over a detail I don't yet see how to resolve: https://math.stackexchange.com/q/5140364/23995 .

~Are you talking about this proof?~
riemann-integreal-Bishop

~Do you really need to interleave arbitrary partitions or would it be sufficient to work with standard uniform partitions?
Would it be OK to have `P` and `Q` be standard uniform partitions of `[a, b]` and then have `R` be the **non-uniform** partition of `[a, b]` obtained by interleaving `P` and `Q`?~

Sorry, I had the wrong book and the wrong ideas.

malarbol:
> Bishop's Foundations of Constructive Analysis says that if the Riemann sum of the standard uniform partition has a limit as the number of partitions grows, that _is_ the integral and you can get arbitrarily close to it with any any partition with sufficiently small mesh, but the proof skips over a detail I don't yet see how to resolve: https://math.stackexchange.com/q/5140364/23995 .

I found the following note https://scispace.com/pdf/the-riemann-integral-in-constructive-mathematics-5a4du99uq1.pdf suggesting that Bishop's construction is wrong, for the reason you mention: the impossibility to define a common refinement of arbitrary partition.

Their solution is to consider only **rational** partitions of intervals for the definition of integrability and use linearity of the ordering on Q for proofs requiring common refinements of partitions.

malarbol:
> if the Riemann sum of the standard uniform partition has a limit as the number of partitions grows, that _is_ the integral and you can get arbitrarily close to it with any partition with sufficiently small mesh

BTW, is this supposed to be true without any continuity assumption on the function?

PR #1986, opened 2026-06-13
More properties of Cauchy pseudocompletions and limits

lowasser:


lowasser:
Broken out of #1985 .

malarbol:
I think this is already in https://unimath.github.io/agda-unimath/metric-spaces.functoriality-isometries-cauchy-pseudocompletions-of-pseudometric-spaces.html and others.

lowasser:
Good catch.  Modified to cover the new bits that are actually needed.

PR #1991, opened 2026-06-17
The product rule on differentiable maps from proper closed intervals in the real numbers to normed real algebras

lowasser:


lowasser:
Builds on #1989 and #1990.

PR #1997, opened 2026-07-13
Simplify swap map of two-element types

MatthiasHu:


fredrik-bakke (approved):
Hi Matthias, thank you for the simplification, it's much appreciated!

PR #2001, opened 2026-09-09
Some basic lemmas for nonsurjective maps

fredrik-bakke:
The small edits from #1831, leads up to some basic lemmas about the connections between surjectivity and non-surjectivity. I will pull out the larger additions into separate PRs subsequently.


fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #2002, opened 2026-09-09
Define some cardinal properties

fredrik-bakke:
Formalizes
- Projective cardinals
- Discrete cardinals
- Decidable cardinals
- Cardinals with decidable existential quantification

Just the definitions for now. Pulled from #1831


fredrik-bakke (commented):
Self-reviewed and merging 👍
@ElifUskuplu

PR #2003, opened 2026-09-09
Lemmas for relating different notions of inequality of cardinals

fredrik-bakke:
Main results are

- If $X$ is projective, then $X \leq^{\mathrm{i}} Y$ implies $X \leq Y$
- If $X$ is projective and $Y$ is discrete, then $X \leq^{\mathrm{i}} Y$ implies $X \leq^{\mathrm{d}} Y$

Pulled from #1831


fredrik-bakke (commented):
Another self-review completed 😮‍💨

